PR #376, opened 2013-09-21
Merge stop, restart and phased_restart  tasks

div:
While porting this cap task to mina, it took me some time to notice that cap task name is :phased_restart, while puma command is phased-restart.


evanphx:
The cap tasks need to be removed from puma, so i'm going to go ahead and just drop this .

sorentwo:
@evanphx: Awesome to hear those are getting dropped.

PR #399, opened 2013-10-28
Adding TTIN and TTOU to increment/decrement workers

softr8:
This change adds ability to increment/decrement the number of workers on the fly when running in clustered mode


sselvamani22:
If the last worker is responding to a request, then I decrement a worker by passing `TTOU` signal, will it terminate?

evanphx:
Once the worker finishes the requests it will terminate.

PR #543, opened 2014-06-01
changed pumactl restart for phased-restart in the init script

dcrec1:
Since the creation of the init script, there is a new restart method called phased restart. Isn't this option better for a default init script?


evanphx:
I'm going to reject this mostly because it would make the restart much less obvious than a strict restart.

Issue #562, opened 2014-07-27
Deferred shutdown until background threads are done executing?


johnkchow:
I have a question: is there a mechanism within Puma to defer shutdown/restarts until background worker threads are done executing? Let's say I want to run certain tasks async during a request (i.e. making analytics API calls or updating a cache). While I could use Sidekiq/Resque, I was wondering if there was a proper way of having all these async tasks run and finish before shutting down?

P.S. Great work w/ puma, first time using jruby with this bad boy and I'm pretty excited w/ this!

evanphx:
`on_worker_shutdown` gives you the ability to add your own logic to control the lifecycle.

PR #566, opened 2014-08-01
Added on_worker_shutdown mechanism

sheltond:
I implemented this because I was having the same problem as johnkchow who opened issue #562: I use a system which does operations in a background thread within a Rails process, and I need to defer the shutdown of Puma worker processes until any background operations are done.

This patch just calls an array of hooks immediately before the worker exits (after it has stopped accepting connections), and blocks until the hooks return.

The user can add an "on_worker_shutdown" section in their Puma configuration file where they can call whatever function they need to block until their background jobs are finished.


drush:
@sheltond Do you have some sample code where you utilize this for background threads that start cleanly and close gracefully?  Just looking to make sure signal hooks are used if needed, etc.

PR #587, opened 2014-09-28
Add feature to read config from STDIN

ppsshh:
Provide a dash as the argument to the `--config` flag to read config from STDIN.
Using dash as filename is a standard way to tell the program that you want to read something from STDIN.

This will break previous behavior when `--config -` meant "do not search for a config file in its default locations".
Instead, you should use something like `echo | puma -C -` to provide an empty config.


evanphx:
I haven't heard if this is good to merge.

Issue #612, opened 2014-12-01
Dependent requests can deadlock on puma if single-threaded


codehotter:
I absolutely love puma and I'm using it for all my applications. Usually, those applications are threadsafe. Sometimes they are not. When an application is running single threaded, either with threads 1, 1 in puma config or allow_concurrency=false in application config, certain kinds of requests can deadlock.

For example, let's say I have a application that generates a PDF from a page server side. The first request is for the PDF, then the PDF creator tool (e.g. phantomjs) internally does a second request to the app. If you're unlucky, and the second request is routed to the same puma worker, the second request will not be handled. It will either block in rack::lock or in puma itself until the first request is finished. However the first request will not finish until the second request is handled.
Another example is a git http server like grack, which will deadlock if you have pre-receive hooks that make calls to your application.

I realize puma may not be the best application server for a single threaded application but I really like puma and I hope I can standardize our application server on it. With the below patch, and config threads 1, 1, the problem is solved for us because puma workers that are 'busy' don't accept new requests. Therefore the request will be routed to another worker.

However I don't understand puma code that well so I'm worried I've broken some other things with this patch. What about performance for example? (Our installations are rather small.)

```
diff -ur puma-2.10.2.orig/lib/puma/server.rb puma-2.10.2/lib/puma/server.rb
--- puma-2.10.2.orig/lib/puma/server.rb 2014-12-01 08:48:32.322000000 +0100
+++ puma-2.10.2/lib/puma/server.rb      2014-12-01 08:49:48.792000000 +0100
@@ -184,7 +184,7 @@
                 break if handle_check
               else
                 begin
-                  if io = sock.accept_nonblock
+                  if !pool.busy and io = sock.accept_nonblock
                     client = Client.new io, nil
                     pool << client
                   end
@@ -293,7 +293,7 @@
                 break if handle_check
               else
                 begin
-                  if io = sock.accept_nonblock
+                  if !pool.busy and io = sock.accept_nonblock
                     client = Client.new io, @binder.env(sock)
                     pool << client
                   end
diff -ur puma-2.10.2.orig/lib/puma/thread_pool.rb puma-2.10.2/lib/puma/thread_pool.rb
--- puma-2.10.2.orig/lib/puma/thread_pool.rb    2014-12-01 08:48:32.333000000 +0100
+++ puma-2.10.2/lib/puma/thread_pool.rb 2014-12-01 09:45:59.449000000 +0100
@@ -114,6 +114,16 @@
 
[...]

PR #640, opened 2015-01-20
Add a configuration option that prevents puma from queueing requests.

codehotter:
Issue https://github.com/puma/puma/issues/612

This adds a configuration option queue_requests

With queue_requests set to true (the default), workers accept all
requests and queue them before passing them to the handlers.
With it set to false, each worker process accepts exactly as
many requests as it is configured to simultaneously handle.

In combination with threads 1, 1 this ensures that requests
are balanced across workers in a single threaded application.
This can avoid deadlocks when a single threaded app sends a
request to itself. (For example, to generate a PDF.)

### DIFF
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -222,6 +222,14 @@ module Puma
       end
     end # IS_JRUBY
 
+    def finish
+      return true if @ready
+      until try_to_finish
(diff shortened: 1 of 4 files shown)


codehotter:
https://github.com/puma/puma/issues/612

PR #657, opened 2015-02-16
Ignore multiple duplicate port declarations

schneems:
Right now if you specify a port via `-p $PORT` and via a `config/puma.rb` then puma will incorrectly try to bind to the same port twice and will fail.

This PR calls `uniq!` on the array to remove duplicate ports from the `binds` array. It also emits a warning when `binds` or `port` is called with an already existing port specified:

```
Warning :binds or :port specified twice: "test/config/app.rb:1:in `_load_from'"
```

### DIFF
--- a/lib/puma/configuration.rb
+++ b/lib/puma/configuration.rb
@@ -64,6 +64,8 @@ module Puma
       unless @options[:tag]
         @options[:tag] = infer_tag
       end
+
+      @options[:binds].uniq!
     end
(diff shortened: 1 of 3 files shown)


evanphx:
The only problem here is that legit uses of bind multiple times will now warn. Issuing the uniq! is probably enough.

schneems:
I removed the warning.

PR #662, opened 2015-03-06
Proper 'Connection' header handling compatible with HTTP/1.0 and HTTP/1.1

iaintshine:
Hi,

I hope you find this pull request useful. It includes a fix which makes puma handle the 'Connection' header properly. 

In the current state, when you make an HTTP/1.1 request with the 'Connection' header set to 'close', you expect it to always respond with the 'Connection: close'. It's not the case if the server responds with a 204 status code or any other code without a body. The response won't include the 'Connection' header at all. This behavior can cause potential issues, such as writing to a socket that's already closed if some proxy server rewrites the 'Connection' header to 'close' and a client expects the connection to last for more than one request. Even though the [HTTP/1.1 Message and Routing RFC - Persistence](http://tools.ietf.org/html/rfc7230#section-6.3) states that *connection can be closed at any time, with or without intention* it's always nice to inform a client about the server intention to close the connection. 

The pull request changes the current state to behave in the following way:

* If the received protocol is HTTP/1.1, and no 'Connection' header is present, or the header is set to 'keep-alive', the connection will persist and the header won't be present in the response
* If the received protocol is HTTP/1.1, and the 'Connection' header is present and is set to 'close', the connection will not persist and the response will include the 'Connection' header set to 'close'
* If the received protocol is HTTP/1.0, and no 'Connection' header is present, or the header is set to 'close', the connection will not persist and the header won't be present in the response
* If the received protocol is HTTP/1.0, and the 'Connection' header is present and is set to 'keep-alive', the connection will persist and the response will include the 'Connection' header set to 'keep-alive'

Hope that make sense.

Kind Regards

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -535,7 +535,7 @@ module Puma
         line_ending = LINE_END
         colon = COLON
 
-        if env[HTTP_VERSION] == HTTP_11
+        http_11 = if env[HTTP_VERSION] == HTTP_11
           allow_chunked = true
           keep_alive = env[HTTP_CONNECTION] != CLOSE
           include_keepalive_header = false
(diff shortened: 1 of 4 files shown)


evanphx:
Thanks! Great work!

PR #669, opened 2015-03-13
WIP CLI: Args over config

chulkilee:
**WIP - feedback is welcome**

- [x] add option to use args over config to keep the existing behavior.
- [ ] keep the existing behavior for `binds` option

See also #460

PR #672, opened 2015-03-15
Refactor CLI & Configuration

chulkilee:
- Made internal methods private: e.g. parse_options
- Extracted methods: e.g. setup_binds, close_binder_listeners
- Extracted file: dsl.rb for Puma::DSL (originally in configuration.rb)
- Moved Puma::cli_config from configuration.rb to cli.rb
- Reordered methods for readability
- Improved code readability
  - avoid assignment in if condition
  - avoid name conflicts between local variable and instance method
  - shorter block
  - use each_with_object
  - each over each_with_index when index is not used
  - and so on...

Only the first thing (making methods private) will *break* other apps using them - but from my understanding those methods shouldn't be called outside of puma, so I just made the change.

I started this work as part of #669 and found it would be better to merge refactoring before adding features. Please take a look and please give some feedback.


evanphx:
Thanks @chulkilee! Great work!

PR #673, opened 2015-03-15
Remove capistrano

chulkilee:
- Removed `capistrano.rb` since it is not supported anymore
- Removed capistrano3-puma configuration from README


evanphx:
I worry this will break anyone that was using it. We should probably deprecate it properly (print out a message when used) before it's removed.

kirs:
:+1: from the Capistrano team

PR #674, opened 2015-03-20
pass raw IO objects to IO.select in reactor

mmmries:
### Why this PR?
I want to make Puma work with the current maglev-head so that people can adopt maglev for web projects without relying on Webrick in production.

Maglev currently has an issue where IO.select is not coercing its arguments to IO objects so passing Puma::Client objects to IO.select was raising an error.

### The "Fix"
This commit changes the reactor loop where we call IO.select so that it passes only IO objects to IO.select and then matches them back to the Puma::Client object.

### Questions
This PR is pretty experimental at this point. I would want to do some additional benchmarking on MRI and jRuby to measure the impact before this is ready to be merged, but before I do that work I'm curious to know how open the puma project is to receiving these kinds of pull requests.

I will open an issue with the maglev project to fix the `IO.select` behavior, but I would love to get puma working before that change gets made in the maglev project.

### DIFF
--- a/lib/puma/reactor.rb
+++ b/lib/puma/reactor.rb
@@ -25,7 +25,7 @@ module Puma
 
       while true
         begin
-          ready = IO.select sockets, nil, nil, @sleep_for
+          ready = IO.select sockets.map(&:to_io), nil, nil, @sleep_for
         rescue IOError => e
           if sockets.any? { |socket| socket.closed? }
             STDERR.puts "Error in select: #{e.message} (#{e.class})"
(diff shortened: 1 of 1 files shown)


lib/puma/reactor.rb
```diff
@@ -39,7 +39,8 @@ def run_internal
 
         if ready and reads = ready[0]
           reads.each do |c|
-            if c == @ready
+            client = sockets.detect{|socket| socket.to_io == c }
```
evanphx: This line really hurts the performance of this acceptor loop because it means searching through all the sockets again on each iteration. In order to not penalize other platforms, I'd prefer to put this and the change on line 28 behind methods that are defined differently on maglev than other platforms. On other platforms, the methods would be noops that just return the value given. On maglev, the transformations would be applied.

evanphx:
I'm closing this PR because there hasn't been any updates on it.

PR #686, opened 2015-04-15
update Readme with example of custom error handler

jjb:
How's this look?

I'm still in the process of trying this in my app -- I'm not 100% sure if the lazy-loading behavior is
what I expect (nothing in the rails environment and no libraries are loaded at this point).

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,20 @@ When you use preload_app, your new code goes all in the master process, and is t
 
 Note that preload_app can’t be used with phased restart, since phased restart kills and restarts workers one-by-one, and preload_app is all about copying the code of master into the workers.
 
+### Error handler for low-level errors
+
+If puma encounters an error outside of the context of your application, it will respond with a 500 and a simple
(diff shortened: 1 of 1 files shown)


jjb:
@brianr of Rollbar recommends adding request data:

```ruby
Rollbar.scope(request: request_data_here).critical(e)
```

@evanphx What's available to grab from the Rack environment in the context of the handler?

jjb:
if anyone finds this thread-- here's a more complete solution i came up with for Rollbar

```ruby
plugin :heroku

class PumaRollbarTools
  include Rollbar::RequestDataExtractor
end

lowlevel_error_handler do |error, env|
  Rollbar.scope(request: PumaRollbarTools.new.extract_request_data_from_rack(env)).critical(error)
  [ 500, {},
    [ "An error has occurred, and our engineers have been informed. Please reload the page. If you continue to have problems, contact support@example.com\n" ]
  ]
end
```

PR #693, opened 2015-04-24
Fix example config text

rob-murray:
Update example config with correct text for `on_worker_*` blocks


rob-murray:
Shortest PR ever.

Issue #689, opened 2015-04-18
Possible regression with Rackup file using first line magic comment


nonnenmacher:
I have a service using puma, that was running fine for months, but on the last upgrade (2.11.2), it fails to start, with a error message 
     ```
     ! Unable to load application: NoMethodError: undefined method `start_with?' for :server:Symbol
     undefined method `start_with?' for :server:Symbol
    ```
I start my server using the following command 

     bundle exec pumactl -S tmp/puma/puma.state start

That read a puma config file from the `config/puma.rb` file 
 that contains a `rackup  config-puma.ru` line.

This backup file contains the magic comment line 
       #\ -s puma     (or alternate    #\ --server puma)

This allow to start either the service using command line, or using my favorite IDE, using a rack launcher.

Tracing the problem down, seems to deal with a change in puma/configuration.rb
along the lines 

      rack_app, rack_options = Rack::Builder.parse_file(rackup)
      @options.merge!(rack_options)

      config_ru_binds = rack_options.each_with_object([]) do |(k, v), b|
        b << v if k.start_with?('bind')
      end

When returning from Rack builder, the rack_options is a hash
with two **Symbol** keys   :server and :config

      { :server => "puma", :config => nil}

however the loop that try to figure out, if there is a rack options that could overrides network binding, assume the options are keyed by String and not using Symbols.

I guess a simple  caution like :

         b << v if k.start_with?('bind') if k.respond_to?(:start_with?)

will solve the issue, without too much changes.

evanphx:
This is fixed on master. I'll release it tomorrow. Sorry about that!

Issue #684, opened 2015-04-12
2.11.2 breaks config.ru compatibility


tomquas:
am using foreman and a very basic config/puma/development.rb to start my app.
my config.ru's first line is _'#\ -s puma'_.
upgrading from 2.11.1 errors out; see trace below.
removing that line from config.ru fixes the startup process

```
10:12:10 app.1        | [2514] ! Unable to load application: NoMethodError: undefined method `start_with?' for :server:Symbol
10:12:10 app.1        | /home/tom/.gem/ruby/2.2.1/gems/puma-2.11.2/lib/puma/configuration.rb:107:in `block in load_rackup': undefined method `start_with?' for :server:Symbol (NoMethodError)
10:12:10 app.1        | 	from /home/tom/.gem/ruby/2.2.1/gems/puma-2.11.2/lib/puma/configuration.rb:105:in `each'
10:12:10 app.1        | 	from /home/tom/.gem/ruby/2.2.1/gems/puma-2.11.2/lib/puma/configuration.rb:105:in `each_with_object'
10:12:10 app.1        | 	from /home/tom/.gem/ruby/2.2.1/gems/puma-2.11.2/lib/puma/configuration.rb:105:in `load_rackup'
10:12:10 app.1        | 	from /home/tom/.gem/ruby/2.2.1/gems/puma-2.11.2/lib/puma/configuration.rb:69:in `app'
10:12:10 app.1        | 	from /home/tom/.gem/ruby/2.2.1/gems/puma-2.11.2/lib/puma/runner.rb:113:in `load_and_bind'
10:12:10 app.1        | 	from /home/tom/.gem/ruby/2.2.1/gems/puma-2.11.2/lib/puma/cluster.rb:302:in `run'
10:12:10 app.1        | 	from /home/tom/.gem/ruby/2.2.1/gems/puma-2.11.2/lib/puma/cli.rb:216:in `run'
10:12:10 app.1        | 	from /home/tom/.gem/ruby/2.2.1/gems/puma-2.11.2/bin/puma:10:in `<top (required)>'
10:12:10 app.1        | 	from /home/tom/.gem/ruby/2.2.1/bin/puma:23:in `load'
10:12:10 app.1        | 	from /home/tom/.gem/ruby/2.2.1/bin/puma:23:in `<main>'
10:12:10 app.1        | exited with code 1
10:12:10 system       | sending SIGTERM to all processes
```

PR #697, opened 2015-04-29
Added tests for bind configuration on rackup file

spk:
see b8c087d967e6f018713cc4f444358754cc7dacba
see #684
see #689


spk:
Is the build error normal? It seams it just stop?

Issue #591, opened 2014-10-17
Disable SSLv3?


renier:
Can we get a configuration option to disable SSLv3?

http://googleonlinesecurity.blogspot.co.uk/2014/10/this-poodle-bites-exploiting-ssl-30.html

renier:
Hi, @evanphx. Pinging on the above.

PR #698, opened 2015-05-04
Fix C/JRuby hang on bad SSL handshake (fixes JRuby travis)

### LINKED ISSUE #603: JRuby SSL binding creates broken thread if non SSL connection targets SSL Port
Once I run an SSL server with puma:

    bundle exec puma remote-webapp/config.ru -b 'ssl://0.0.0.0:8888?keystore=remote-webapp/keystore.jks&keystore-pass=123456'

And then try to use http instead on the https, the following error occur:

    vendor/jruby/1.9/gems/puma-2.9.2-java/lib/puma/thread_pool.rb:56" java.lang.RuntimeException: javax.net.ssl.SSLException: Unrecognized SSL message, plaintext connection?

The problem is then the particular thread would become a zombie and never got recycled. Is there anyway I could prevent this from happening?

dmarcotte:
Both the C and JRuby SSL implementations would hang on a bad handshake because they were not producing the EOF expected in that case.  Update their error handling to behave correctly here (`test_ssl_v3_rejection` covers this for both impls).

This was the root cause of the recent hanging JRuby tests.  With the hang fixed, the test would fail because the SSLv3 handshake became "bad" when java updated to a version with SSLv3 disabled by default (see [here](http://www.oracle.com/technetwork/java/javase/documentation/cve-2014-3566-2342133.html)).  Fix this by removing `enable_SSLv3` support from JRuby (the C minissl implementation doesn't support it either since #591, so it clearly won't be missed).

Fixes #603, replaces #605.

Thanks again @evanphx.  Let me know if you have any questions.

### DIFF
--- a/ext/puma_http11/mini_ssl.c
+++ b/ext/puma_http11/mini_ssl.c
@@ -161,7 +161,7 @@ void raise_error(SSL* ssl, int result) {
 VALUE engine_read(VALUE self) {
   ms_conn* conn;
   char buf[512];
-  int bytes, n;
+  int bytes, n, error;
 
   Data_Get_Struct(self, ms_conn, conn);
 
@@ -173,7 +173,8 @@ VALUE engine_read(VALUE self) {
 
   if(SSL_want_read(conn->ssl)) return Qnil;
 
-  if(SSL_get_error(conn->ssl, bytes) == SSL_ERROR_ZERO_RETURN) {
+  error = SSL_get_error(conn->ssl, bytes);
+  if(error == SSL_ERROR_ZERO_RETURN || error == SSL_ERROR_SSL) {
     rb_eof_error();
   }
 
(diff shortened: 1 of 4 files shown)


dmarcotte:
@evanphx note the test failures here should be transient.  Check out a green run of these changes here: https://travis-ci.org/looker/puma/builds/60907494

evanphx:
Great find! Thanks!!

dmarcotte:
Thanks @evanphx!  Feel free to let me know when any of the JRuby stuff needs some attention in the future, and I'd be glad to lend a hand.

evanphx:
Thanks Daniel!

luxx:
Can someone release the gem please?

evanphx:
Done! Sorry for the delay.

PR #700, opened 2015-05-06
Add support for batched phased restarts

ramonsnir:
We have a large number of workers, and the one-by-one phased restart simply takes too long. Was tested in a Centos 6.6 VM.

### DIFF
--- a/lib/puma/cli.rb
+++ b/lib/puma/cli.rb
@@ -307,6 +307,7 @@ module Puma
         :debug => false,
         :binds => [],
         :workers => 0,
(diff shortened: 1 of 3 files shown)


denji:
:+1:

ramonsnir:
Don't merge yet, we suspect a bug here. Will update.

ramonsnir:
Enabling and disabling this patched correlates very well to `segmentation fault`s in `lib/puma/server.rb`. I'm closing this PR, but I think this feature still deserves a resolution.

PR #701, opened 2015-05-19
Start running tests against JRuby 9k on Travis

deepj:


deepj:
@evanphx It seems Puma will be working a bit better than now on JRuby 9k https://github.com/jruby/jruby/issues/2957 :) That was the reason why I made this PR.

Issue #691, opened 2015-04-20
Dead threads


bpaquet:
We have got a big outage on our production platform.

After lot of investigation, we found in some edge cases, puma app threads can die and are not relaunched.

I have add the following loop at https://github.com/puma/puma/blob/master/lib/puma/thread_pool.rb#L42
```rb
Thread.new do
  while true
    puts "Threads #{@workers.size} #{@workers.select{|x| x.alive?}.size}"
    sleep 5
  end
end
````

During our tests,  the second number decreases.

I do not have a precise method to reproduce the problem. What I do is
- use our rails / postgresql application
- lock a postgres table (``begin; alter table xxxx add column toto integer;``)
- launch lot of curl command to the rails app, which try to update the table (and are blocked against the lock)
- wait

We are using [rack-timeout](https://github.com/heroku/rack-timeout) and [airbrake](https://github.com/airbrake/airbrake).

We think the thread death is related to edge cases in rack-timeout, but we got no evidence. All standard uses cases are ok (eg : rack-timeout can kill a thread blocked in a postgresql transaction, and the thread is reused).

So, I have some questions
- why there is no rescue around https://github.com/puma/puma/blob/master/lib/puma/thread_pool.rb#L104 ?
- what about detecting dead threads and decreasing ``@spawned``, to allow relaunch ?
- why there is no params request_timeout in puma, like in unicorn ? multi threading issue ?
- any idea about how to kill active record connections of dead threads ?

bpaquet:
Probably related to https://github.com/heroku/rack-timeout/issues/76

evanphx:
There is no rescue around the block because that has typically been handled by puma's lowlevel handler. This should probably also do it's own management though too.

evanphx:
@bpaquet Which version of puma are you using?

bpaquet:
2.11.1

Issue #681, opened 2015-04-02
Request timeouts (H12's) when switching from Unicorn to Puma on Heroku


plindelauf:
On Heroku I switched one of our Rails 3.2 (/Ruby 2.0.0) apps from Unicorn to Puma, because we're experiencing problems with slow clients using Unicorn. Almost immediately we started seeing many 30 second timeouts (H12 errors) on at least 5% of our requests. Something we never had when running on Unicorn. 

Here's our setup:

    workers Integer(ENV['WEB_CONCURRENCY'] || 2)
    threads_count = Integer(ENV['MAX_THREADS'] || 5)
    threads threads_count, threads_count

    preload_app!

    rackup      DefaultRackup
    port        ENV['PORT']	|| 3443
    environment ENV['RACK_ENV']		|| 'development'

    on_worker_boot do
      # Valid on Rails up to 4.1 the initializer method of setting `pool` size
      ActiveSupport.on_load(:active_record) do
        config = ActiveRecord::Base.configurations[Rails.env] || Rails.application.config.database_configuration[Rails.env]
        config['pool'] = ENV['DB_POOL'] || ENV['MAX_THREADS'] || 5
        ActiveRecord::Base.establish_connection(config)
      end
    end

Since our app is likely not thread safe, we've set `WEB_CONCURRENCY` to `1` and scaled the processes by setting `WEB_CONCURRENCY` to values ranging from `4` to `8`. We generally run a number of 2x dynos, but we also tried 1x dynos. No matter how we up- or downscale these, the H12's keep coming. Memory usage stays within the normal bounds. Additionally, since we're running just 1 thread per processes, we have set the `DB_POOL` to `1`. We have also configured Rack::Timeout to timeout at 28 seconds.

I contacted Heroku support and they don't know where this is coming from either. They mentioned that they had at least one other similar issue report from someone using 1 thread and multiple processes.

Any ideas what could be causing this?

chanks:
I just diagnosed a problem with an app wherein Rack::Timeout's errors (which descend from the Exception class instead of StandardError) weren't being caught by anything and were killing Puma's workers. Then new requests that came in wouldn't be serviced and would timeout themselves.

I'm going to open an issue with rack-timeout, but in the meantime I wrote a quick middleware to catch those errors and reraise them in a way that Puma can recover from. Haven't deployed to production yet, but it seems to work alright in testing.

```
class TimeoutRecovery
  def initialize(app, options = {})
    @app, @options = app, options
  end

[...]

PR #702, opened 2015-05-20
Add thread reaping to thread pool

nesQuick:
Hi there.

# The Problem

As already discussed in #691 and #681 Puma does not recover from dead threads. Even though this is a fairly uncommon state and might only happen in edge cases. One of these edge cases seems to be using [rack-timeout](https://github.com/heroku/rack-timeout) as you can see in https://github.com/heroku/rack-timeout/issues/76 and https://github.com/rails/rails/issues/1627#issuecomment-76172622.

To reproduce this behavior use the following example rack app:

```ruby
# dummy.ru
class SimpleRacktimeout # stand-in for a more complex Rails app using all kind of different non rubycode ressouces
  def initialize(app)
    @app = app
  end

  def call(env)
    p '########### HELLO FROM APP ###########'
    begin
      app_thread     = Thread.current
      timeout_thread = Thread.start do
        app_thread.raise(Exception)
      end

      `sleep 1`
    ensure
     timeout_thread.kill
     timeout_thread.join
   end
  end
end

app = Rack::Builder.new do |b|
  b.use SimpleRacktimeout
  b.run lambda { |_env| [200, {}, ['ok']] }
end.to_app

run app
```

Start it with `puma -w 1 -b tcp://localhost -p 9292 -t 3:3 dummy.ru`. And then force Puma to concurrently serve more requests than it has (healthy) threads available by using Apache Benchmark `ab -n 8 -c 4 http://127.0.0.1:9292/`.

# Solution

This PR implements a thread reaping mechanism based on the auto trim functionality. It decreases the `@spawned` counter in the `ThreadPool` in order to allow it to spawn new threads.

# Alternative Solution

Instead you can use `th.abort_on_exception = true` in https://github.com/puma/puma/blob/master/lib/puma/thread_pool.rb#L112 which also solves the problem described above. We decided for the former solution because removing dead threads from the pool seemed a cleaner approach.

As mentioned in #691 we are not sure if and how active record connections in dead threads will be cleaned up.

> - any idea about how to kill active record connections of dead threads?

/cc @partyschaum

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -39,6 +39,7 @@ module Puma
     attr_accessor :max_threads
     attr_accessor :persistent_timeout
     attr_accessor :auto_trim_time
+    attr_accessor :reaping_time
     attr_accessor :first_data_timeout
 
     # Create a server for the rack app +app+.
@@ -60,6 +61,7 @@ module Puma
       @min_threads = 0
       @max_threads = 16
       @auto_trim_time = 1
+      @reaping_time = 1
 
       @thread = nil
       @thread_pool = nil
@@ -272,6 +274,10 @@ module Puma
       if queue_requests
         @reactor = Reactor.new self, @thread_pool
         @reactor.run_in_thread
+      end
+
+      if @reaping_time
+        @thread_pool.auto_reap!(@reaping_time)
       end
 
       if @auto_trim_time
--- a/lib/puma/thread_pool.rb
+++ b/lib/puma/thread_pool.rb
@@ -33,6 +33,7 @@ module Puma
       @workers = []
 
       @auto_trim = nil
+      @reaper = nil
 
       @mutex.synchronize do
         @min.times { spawn_thread }
@@ -155,6 +156,21 @@ module Puma
       end
     end
 
+    # If there are dead threads in the pool make them go away while decreasing
+    # spwaned counter so that new healty threads could be created again.
+    def reap
+      @mutex.synchronize do
+        dead_workers = @workers.reject(&:alive?)
+
+        dead_workers.each do |worker|
+          worker.kill
+          @spawned -= 1
+        end
+
+        @workers -= dead_workers
+      end
+    end
+
     class AutoTrim
       def initialize(pool, timeout)
         @pool = pool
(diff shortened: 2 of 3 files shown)


evanphx:
Seems like we should attack the initially problem, i.e. why threads in the pool are dying. I guess an exception is bubbling up to the top-level?

nesQuick:
I'm not completely sure what is killing the threads, from my educated guess, it's the fact that the exception is raised from another thread and thus bubbling up and killing the thread.

All the issues we've linked above and also our initial problem we've faced is that heroku recently switched their recommended web server from unicorn to puma ([ref](https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#timeout)), but due to the fact that puma does not have a request timeout `rack-timeout` is recommended. And from what I've seen in our app as well in a bunch of issues all over different ruby repos (rails/puma/rack-timeout) is that this combination is not working pretty well, I'm not sure why this happens exactly. But it seems like the ActiveRecord connection pool is having a lot of trouble with that as we as the puma thread pool. From our investigations all the following reported problems are caused by a combination of rack-timeout not playing well with puma. 

- https://github.com/puma/puma/issues/691
- https://github.com/puma/puma/issues/681
- https://github.com/heroku/rack-timeout/issues/76
- https://github.com/rails/rails/issues/1627#issuecomment-76172622

Maybe pulling in @schneems and @kch might be a good idea.

nesQuick:
@evanphx we are wondering what do you exactly mean with top level?

kch:
rack-timeout is kind of terrible and currently doesn't play well with puma, I'm working on a bunch of changes to it, which should remedy that.

But it seems to me threads could die for a number of valid reasons and puma should reap them when that happens.

nesQuick:
@kch I agree on the point that dead threads should be reaped, that is why we proposed this PR. On the other hand, could you maybe (in the rack-timeout repo) a bit more transparent on what kind a changes you are working and if you might need help? Happy to help :) Because of the issue mentioned we now switched back to unicorn.

knightq:
+1 for merge and release

sgranata82:
+1

SebastianEdwards:
+1

deepj:
+1

knightq:
We'll switch back to unicorn too until this will be resolved: it's source of dangerous raising in request queueing during high request burst periods.

schneems:
Or you could run off of 1 thread and multiple puma workers which is essentially doing the same thing as Unicorn.

acrolink:
I had encountered the issue of request time-outs when I was testing my app using `httperf` tool (conns 500, rate 50, timeout 10). Puma failed badly,out of 500 requests: 181 status 200 OK and 319 time-out.

Any new updates on this?

emerleite:
@acrolink same here. Sometimes I have only one thread working :(

acrolink:
@emerleite .. If you don't mind trying a new framework I highy recommend Phoenix (upon Elixir) .. It is super fast compared to Ruby's Rails (5 or 6 times faster). As for Puma, I have kept using Unicorn for the meantime.

emerleite:
We're currently doing that, but the Rails app must be running for now :)

evanphx:
@acrolink Can you tell me more about your setup?
@emerleite Can you tell me more about your setup?

PR #703, opened 2015-05-26
Improve Travis settings

deepj:
- JDK 7 is default on Travis
- cache dependencies
- use always the latest Ruby 2.1.x


deepj:
It seems the failure is not caused by my changes rather it comes from the last PR.

PR #711, opened 2015-06-08
Do not suppress Content-Length on partial hijack (fixes #710)

### LINKED ISSUE #710: Content-Length removed when using "rack.hijack"
I am using Puma to stream a large response. The nature of the streaming is such that I know Content-Length and Content-Range in advance. However, if I serve via Puma, I see that the both of those headers get removed from the response, and any downloading client will not be able to build a download progress report because of that.

To reproduce, make "hijack3.ru" similar to the other hijack Rackup files, with the following content:

    run lambda { |env|
      body_str = "This is a welcome message to all folks\n"
      size = body_str.bytesize
      body = lambda { |io| io.puts(body_str); io.close }
      h = {
        'Content-Length' => size.to_s,
        'rack.hijack' => body,
      }
      [200, h, []]
    }

When running it, it is very likely (hash ordering might be involved) that the Content-Length and Content-Range headers are going to be stripped from the response. AFAIK this is against the Rack spec because hijacking does not automatically mean the response is going to be chunked or an endless stream.

I wanted to put together a unit test but apparently those hijack rackup files do not get activated from the test suite (or I was not able to locate their entry points).

julik:
The partial hijack API does not mandate that the server
handler will be removing any headers from the response, so if
the app has set a Content-Length header it should be kept intact.

Closes #710

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -610,15 +610,13 @@ module Puma
           fast_write client, lines.to_s
           return keep_alive
         end
-
-        unless response_hijack
-          if content_length
-            lines.append CONTENT_LENGTH_S, content_length.to_s, line_ending
-            chunked = false
-          elsif allow_chunked
-            lines << TRANSFER_ENCODING_CHUNKED
-            chunked = true
-          end
+        
+        if content_length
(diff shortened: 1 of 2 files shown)


julik:
:heart:

PR #712, opened 2015-06-11
Make pumactl load config/puma.rb by default

chewi:
`config/puma/#{env}.rb` isn't tried because there is no environment option for pumactl, though this could be added.

### DIFF
--- a/lib/puma/control_cli.rb
+++ b/lib/puma/control_cli.rb
@@ -66,7 +66,15 @@ module Puma
       command = argv.shift
       @options[:command] = command if command
 
-      Puma::Configuration.new(@options).load if @options[:config_file]
+      unless @options[:config_file] == '-'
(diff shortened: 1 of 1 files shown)


lib/puma/control_cli.rb
```diff
@@ -66,7 +66,11 @@ def initialize(argv, stdout=STDOUT, stderr=STDERR)
       command = argv.shift
       @options[:command] = command if command
 
-      Puma::Configuration.new(@options).load if @options[:config_file]
+      unless @options[:config_file] == '-'
+        if @options[:config_file] ||= ('config/puma.rb' if File.exist?('config/puma.rb'))
```
evanphx: Can you make this more than 1 line? It's super hard to read.

chewi:
Sorry, bad habit!

PR #748, opened 2015-07-25
Fix installation from bundler's github option

deivid-rodriguez:
I'd like to be able to use latest `puma`'s master by using the `github` option in my `Gemfile`, but it's not working (see https://github.com/bundler/bundler/issues/3845).

This patch fixes it. I don't know of any alternatives that don't rely on including the generated `jar` file in source control, but I'm open to suggestions.


evanphx:
I'd rather not do this. It means balloon the git repo because I have to checkin the jar on every release then. Can bundler fix this eventually? /cc @indirect

indirect:
I have no idea how the jar gets built; if you're using the gem from git on JRuby, you have to build the jar as part of the release process, right?

deivid-rodriguez:
@evanphx You mean you have to check in a different jar file every time you release? If so, why?

evanphx:
I've thought about it and I'm not going to check the jar file in. It's just too big. If we can make it so rubygems + mkmf generate the jar, thats the best way forward.

indirect:
I would also strongly prefer that it be possible to build the jar from source if Bundler is installing from git—is it possible to make that happen via the usual gem build process?

deivid-rodriguez:
This should probably be moved to rubygems/rubygems then.

deivid-rodriguez:
Thanks for considering by the way :smiley:

PR #751, opened 2015-07-31
Ruby 1.8.7 compatibility fixes

costi:
Enumerable in 1.8.7 doesn't have each_with_object
rand() doesn't accept ranges in 1.8.7
Caught a 1.9 hash syntax in tests.


brixen:
Thank you, @evanphx!

PR #754, opened 2015-08-06
Add before_fork hooks option.

### LINKED ISSUE #303: Add `before_worker_boot` config block with clustered mode
From my understanding, preloading your application will boot the application before forking workers. In the case of Rails and ActiveRecord, this will establish a database connection pool. What happens to that pool after fork? Is it shared with the workers? 

One option is to re-establish the connection in an `on_worker_boot` hook, but does it need to call `ActiveRecord::Base.disconnect!` first? 

Unicorn provides a `before_fork` and `after_fork`. On Heroku, it's important to call `disconnect!` in the `before_fork`, or you can easily leak connections and get weird EOF messages. 

So, should users call `disconnect!` and then re-establish in the `on_worker_boot`, or just establish the connection? Or should an equivalent `before_fork` config block be added?

nathansamson:
This is usefull to close database or other remote connections
that were opened during the preload_app phase of the master process.

Fixes #303

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -379,6 +379,10 @@ module Puma
       @cli.write_state
 
       @master_read, @worker_write = read, @wakeup
+
+      hooks = @options[:before_fork]
+      hooks.each { |h| h.call }
+
       spawn_workers
 
       Signal.trap "SIGINT" do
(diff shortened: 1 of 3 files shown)


jlduran:
@rosswilson solved in 2.13.4

Issue #770, opened 2015-08-22
Phased restart ignores directory configuration when Puma is invoked from that configured directory


jamesmeador:
Our config file looks like:
```
directory "#{ENV["HOME"]}/current"
pidfile "#{ENV["HOME"]}/puma.pid"

prune_bundler
```

Where ```current``` is  a symlinked directory to the new code.

And is passed to puma as:
```
bundle exec puma -t 16:16 -w 3 -C ${HOME}/puma.rb
```

Sending SIGUSR1 to the process results in the same directory being restarted, even though the symlink was updated.

Sending SIGUSR2 works correctly.

jamesmeador:
After some digging, it appears that this is more of a problem with respecting the directory configuration.

It looks like puma "prefers" to use env['PWD'] but this is causing us issues.

We spawn Puma from a custom daemon that we use that listens to watches on a zookeeper node. We fire updates to the node, and whatever machines we have listening to that node will pick up the new configuration (a git sha and repo). 

When the daemon starts up, it clones the repo into a directory, say `root/master-gitsha1`. It changes into that directory, runs "bundle" to activate gems, as well as compile assets, etc.

When the setup is done, it's time to symlink `root/master-gitsha1` to `root/current` and launch puma for the first time. The daemon Dir.chdir's to `root/current`, and does 
```
Bundler.with_clean_env { Process.spawn('bundle exec puma -w 3 -C ${HOME}/puma.rb') }
```

Note that Dir.chdir in Ruby just changes to the target, thus anything spawned from Ruby means that ENV['PWD'] will be the target directory, not the symlink. In this case, the target directory is `root/master-gitsha1`.

Puma takes off. CLI.new sets `@cli_options[:worker_directory]` to `ENV['PWD']` because `Dir.pwd` and `ENV['PWD']` match.

CLI.run calls parse_options which then (via configuration.rb and dsl.rb) parses my config file, setting `@options[:directory]` and `@options[:worker_directory]` to `root/current`. Then, it merges cli_options in. **This is what I think is a bug:**

 **Now, `@options[:directory]` is `root/current` and `@options[:worker_directory]` is `root/master-gitsha1`**.

when `start_phased_restart` for a cluster is called, the directory will **always be changed to `root/master-gitsha1`**.

Seems unreasonable to me. Am I crazy? It seems silly to store a cli_option that the user didn't explicitly set. (and then override an attribute they **did** intentionally set).

allaire:
[...]

PR #793, opened 2015-10-06
Fix bug setting worker_directory when using a symlink directory

robdimarco:
Fix for issue #770 

The @cli_options variable is used to reflect the options passed from the command line.
The @options variable is the final set of options that will be used by the Runner instance
for configuration. When generating the restart data, we want to set the value in the @options
hash and not in the @cli_options hash.


jamesmeador:
@evanphx thoughts on this? This bug just caused some downtime for us.

robdimarco:
FWIW, we have been running this in production for the past 3 weeks with no issues.

PR #795, opened 2015-10-13
puma/cli: support specifying STD{OUT,ERR} redirections and append mode

unleashed:
Currently the only way to specify stdout/stderr redirections is through Puma's config file. This can be cumbersome in some environments where you would rather use a command line switch. Do this for both and also add a boolean option for append mode.

### DIFF
--- a/lib/puma/cli.rb
+++ b/lib/puma/cli.rb
@@ -410,6 +410,18 @@ module Puma
           @cli_options[:tag] = arg
         end
 
+        o.on "--redirect-stdout FILE", "Redirect STDOUT to a specific file" do |arg|
+          @cli_options[:redirect_stdout] = arg
+        end
+
+        o.on "--redirect-stderr FILE", "Redirect STDERR to a specific file" do |arg|
+          @cli_options[:redirect_stderr] = arg
+        end
+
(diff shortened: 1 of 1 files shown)


evanphx:
I'd rather keep it in the config file personally. Is there a reason you need the command line options?

unleashed:
This is not strictly needed but offers better ergonomics, specially when most configuration file parameters have equivalent arguments for the CLI.

I have a use case in my workplace where I have one single puma config file with common parameters used in all environments. I fill in the rest of the parameters via command line arguments.

Determining some of the remaining parameters in some environments is cumbersome to do when puma is launching. In my case I derive some things such as the number of workers just doing some simple checks with the available hardware info exposed by the OS at launch time, and I just pass in `-w n`. However, the destination storage for logs is quite dynamic based on conditions not only tied to the environment (ie. whether puma is running inside a docker container, availability of network shares, etc).

Adding the corresponding command line switches helps a lot in not maintaining extra config files and avoids hacking in some ugly text replacement before loading them with puma.

OTOH I guess the options themselves are pretty verbose. I chose to maintain their original name, but would be happier if I could convey their meaning with far fewer characters or if a shortcut was agreed upon.

PR #814, opened 2015-11-03
[ci skip] More explicit docs for worker_timeout

schneems:
Had a customer who thought setting this was preventing against long running requests (it doesn't). These docs clarify the purpose of setting this value, that it is used to detect hung processes.


schneems:
:rocket:  thanks!

PR #833, opened 2015-11-29
support TLS client auth (verify_mode) in jruby

joemiller:
Adds support for `verify_mode` to configure client authentication when running under JRuby.

Things to note:

- Assumes the CA used to verify client certs is in the same java
  keystore file that is used when setting up the HTTPS TLS listener. We
could split this out, but not sure if it's necessary.
- Friendly/helpful error messages explaining why the verification failed
  are not present in the same way they are in the CRuby/OpenSSL code
path. I'm not sure how to make them available.
- I did not include any code to create the `keystore.jks` file in the
  `examples/puma/client-certs` directory because I didn't see any
existing code to create the `examples/puma/keystore.jks` file. The
commands to create this keystore would be:

```
cd examples/puma/client-certs
  openssl pkcs12 -chain -CAfile ./ca.crt -export -password pass:blahblah -inkey server.key -in server.crt -name server -out server.p12
  keytool -importkeystore -srckeystore server.p12 -srcstoretype pkcs12 -srcstorepass blahblah -destkeystore keystore.jks -deststoretype JKS -storepass blahblah
  keytool -importcert -alias ca -noprompt -trustcacerts -file ca.crt -keystore keystore.jks -storepass blahblah
```

[Review on Reviewable](https://reviewable.io/reviews/puma/puma/833)


joemiller:
The travis tests failed only on ruby 2.2 and it looks like a timeout issue, possibly a transient issue. Can someone help re-run the test? I couldn't find an option to do so as I am not too familiar with travis-ci

joemiller:
ping. Any feedback on this PR? Thanks!

PR #856, opened 2016-01-14
Initial Seperation of CLI and Server Launcher work

schneems:
This is a WIP. This was the minimum I could do to get all tests to pass without changing any tests. Eventually I think we want all high level process controls to come from launcher, I also think we want another separate object that gets passed to Runner/Single/Cluster that will maintain a relationship with the Launcher. We could use this as the object that also gets exposed to the app like the Embeddable class we talked about earlier. 

Moving forwards i'm planning to port out the CLI tests to only test that they are parsing the correct config and launching servers. I'll port all low level unit tests over to the launcher. Making this change we could either keep all the public methods in CLI that delegate to `@launcher`, I'm guessing not many people are using the internals of CLI and we can take them out. It's your call though.

Wanted to kick this over the fence and see if you had any strong reactions or feelings about this approach.

### DIFF
--- a/lib/puma.rb
+++ b/lib/puma.rb
@@ -12,3 +12,4 @@ require 'thread'
 # Ruby Puma
 require 'puma/const'
 require 'puma/server'
+require 'puma/launcher'
--- a/lib/puma/cli.rb
+++ b/lib/puma/cli.rb
@@ -51,9 +51,6 @@ module Puma
 
       setup_options
 
-      @binder = Binder.new(@events)
-      @binder.import_from_env
-
       begin
         @parser.parse! @argv
         @cli_options[:rackup] = @argv.shift if @argv.last
@@ -61,11 +58,9 @@ module Puma
         exit 1
       end
 
-      @launcher = Puma::Launcher.new(@cli_options)
+      @launcher = Puma::Launcher.new(@cli_options, events: @events)
 
-      @launcher.events  = self.events
       @launcher.config  = self.config
-      @launcher.binder  = self.binder
       @launcher.argv    = @argv
 
       @launcher.setup(@options)
@@ -119,7 +114,9 @@ module Puma
     ## BACKWARDS COMPAT FOR TESTS
 
     # The Binder object containing the sockets bound to.
-    attr_reader :binder
+    def binder
+      @launcher.binder
+    end
 
     # The Configuration object used.
     attr_reader :config
--- a/lib/puma/cli.rb
+++ b/lib/puma/cli.rb
@@ -47,8 +47,6 @@ module Puma
       @status = nil
       @runner = nil
 
-      @config = nil
-
       setup_options
 
       begin
@@ -59,11 +57,6 @@ module Puma
       end
 
       @launcher = Puma::Launcher.new(@cli_options, events: @events, argv: @argv)
-
-      @launcher.config  = self.config
-
-      @launcher.setup(@options)
-
     end
 
     ## BACKWARDS COMPAT FOR TESTS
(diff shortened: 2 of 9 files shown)


schneems:
cc/ @pixeltrix

evanphx:
I like where this is going. Pull the code out of the CLI into the Launcher then having the CLI just use the Launcher means the code paths don't get stale.

schneems:
Okay, I got all the tests to pass here. This was WAY more work than I was thinking originally. I was successful in extracting out a general purpose puma launching class `Launcher` and re-using that in the CLI and in the rack handler. As far as I know this is 100% API backwards compatible minus one thing:

The one backwards incompatible thing: The original Puma rack handler would yield to an instance of `Puma::Server` when called with a block. Since we're not creating a server manually anymore, i'm yielding to an instance of `Launcher` instead. I'm not sure how much people use this behavior of the rack handler, or really even why it exists (though It was handy for writing tests, I added an extra one for doing an integration style test with the rack handler). The Launcher does have access to a "runner" either an instance of `Single` or `Cluster`, each of them have a server object, but neither expose it. If you want I could do the work to expose it and then keep the rack handler block API the same, but it would increase the surface area of those to classes. So either we could not care about the API that maybe not many people use or I could increase some object surface area, it's up to you.

Let me know if you've got questions. I'm available to chat about this if you want, I think we're both in the Rails contributor basecamp chatroom if you want, or ping me in rubycentral chat. I could also jump on a hangout if you need more fidelity. I'll be around tomorrow and next week after Wednesday.

evanphx:
Yay! I'm not concerned about the yield API. I doubt anyone used it, given
how strange and undocumented it was.

Given the compat layer you built in, if the tests pass I'm good with it!

I'll give it a further read tonight and ping you tomorrow with questions.

schneems:
Any questions or concerns? We could add a deprecation to the CLI methods if we want to decrease surface area in the future, but I don't think it's needed. I don't see any downside to keeping those delegate methods in there. 

The other thing I wanted to do, but it is very ambitious is to not pass the `Launcher` object to `Single` and `Cluster` and instead pass an interchange object. Ideally it would have a smaller API and if we wanted to expose some kind of intermediate control layer directly to the app I would rather we did it without going directly through the Launcher object. I spent a little time investigating that change, and it is pretty large. The classes that inherit from `Runner` directly use the `@cli` object pretty liberally, and they also directly use the `@cli.options` hash. There wasn't an immediate benefit from going down that road so I stopped.

evanphx:
Looks great! I'm going to go ahead and merge it. I'm going to probably add a few helpers as some further documentation to Launcher.

This weekend I'm going to be adding a plugin system to puma as well. After that, I'm going to go ahead and release all this work as 3.0!

schneems:
🎉🎉🎉 let me know when 3.0 goes out and I can update the Gemfile that rails generates to use a compatible version of Puma. Thanks for your help!

frodsan:
Hi,

I'm debugging https://github.com/puma/puma/issues/919. I think it's related to this change. Now that puma loads `config/puma.rb` by default, I'm having errors using my application code in hooks like `before_fork` or `on_worker_boot`. How do you make them work in Rails?

Thanks!

PD: I made a simple application that shows the error: https://github.com/frodsan/shotgun-puma-bug.
You can check `config/puma.rb` https://github.com/frodsan/shotgun-puma-bug/blob/master/config/puma.rb

frodsan:
I checked the issue again. It works if I use `puma`:

```
$ puma
[3933] Puma starting in cluster mode...
[3933] * Version 3.0.2 (ruby 2.3.0-p0), codename: Plethora of Penguin Pinatas
[3933] * Min threads: 5, max threads: 5
[3933] * Environment: development
[3933] * Process workers: 2
[3933] * Preloading application
[3933] * Listening on tcp://0.0.0.0:
```

It looks like the problem is in shotgun.

schneems:
Hey, thanks for the report can you open up a separate issue so we can track this? What do you think the problem is with shotgun? That constant should be defined when you require the puma rack handler.

PR #857, opened 2016-01-15
make phased restarts faster

osheroff:
Have the worker send the master "tPID" immediately before it shuts down; this lets the master move on to booting the next worker immediately, instead of waiting for the next iteration of the select() loop.

cc @zendesk/rules

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -136,7 +136,7 @@ module Puma
       @workers.count { |w| !w.booted? } == 0
     end
 
-    def check_workers(force)
+    def check_workers(force=false)
       return if !force && @next_check && @next_check >= Time.now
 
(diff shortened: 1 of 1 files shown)


lib/puma/cluster.rb
```diff
@@ -128,7 +136,7 @@ def all_workers_booted?
       @workers.count { |w| !w.booted? } == 0
     end
 
-    def check_workers(force=false)
+    def check_workers(force)
```
evanphx: Is there a reason you removed the default value for force?

osheroff: no, that was a mistake.  will revert.

osheroff: fixed

osheroff:
@evanphx any chance of a release with the latest?

Issue #859, opened 2016-01-15
Prune bundler causes rack to be loaded twice


boxofrad:
Howdy! :wave: 

I've been converting an app to use `prune_bundler` as recommended in [DEPLOYMENT.md](https://github.com/puma/puma/blob/master/DEPLOYMENT.md#restarting) to ensure bundled gems are properly reloaded when restarting the server with `USR1`.

I noticed a lot of warnings in my logs caused by rack constants being redefined, for example:

```
/home/vagrant/development/puma-double-rack/vendor/ruby/2.1.0/gems/rack-1.6.4/lib/rack.rb:14: warning: already initialized constant Rack::VERSION
/var/lib/gems/2.1.0/gems/rack-1.6.4/lib/rack.rb:14: warning: previous definition of VERSION was here
```

After doing some investigation, it appears that rack is being loaded twice; once from my system-wide gems, and once from my app's `vendor` folder managed by bundler.

Because `prune_bundler` removes bundled gems from the  `$LOAD_PATH`, it seems that this block (in [configuration.rb](https://github.com/puma/puma/blob/d5914573c90fe6f8b9422f397e28c0358ce4aa3b/lib/puma/configuration.rb#L130-L140)) is loading the system-wide rack instead of the bundler rack:

```ruby
# Load and use the normal Rack builder if we can, otherwise
# fallback to our minimal version.
def rack_builder
  begin
    require 'rack'
    require 'rack/builder'
  rescue LoadError
    # ok, use builtin version
    return Puma::Rack::Builder
  else
    return ::Rack::Builder
  end
end
```

Which makes sense... my understanding of this method is that it assumes that if `require 'rack'` succeeds it must be running under bundler, otherwise it uses Puma's own implementation of `Rack::Builder`. 
However in my case as the gem is installed system-wide, the `require` will succeed when not under bundler.

While the obvious solution may be to simply remove the system-wide rack, I'm not sure that's the best solution because:

1. In my particular case, it's not feasible to remove the system-wide rack.
2. This behaviour breaks the isolation / safety contract provided by bundler, if a gem is bundled it should always be preferred over the system-wide gem.

## Reproduction Steps

I've created a simple rack app that exhibits the behaviour:

https://github.com/boxofrad/puma-double-rack

Here are the conditions under which the behaviour can be observed:

- One copy of rack is installed in `GEM_HOME`
- Another copy of rack is installed by bundler
[...]

Issue #735, opened 2015-07-15
latest version of puma version 2.12 doesn't work with dashing gem -v 1.3.4


trekr5:
Hello,

I'm using the Puma gem to run a dashing dashboard. This worked fine yesterday but ever since the update today of the Puma gem, I'm now unable to get the dashboard to work as the scheduling part, of the dashing gem, doesn't work.

What do I need to do within my application to get this version of Puma to work? It is running on a Linux box with Ruby -v 1.9.3

evanphx:
What error do you get? I don't have any way to diagnose the issue without more information.

ball-hayden:
I'm also having issues running dashing after upgrading to puma 2.12.2. I see the following in `puma.err`:

```
/usr/local/rvm/rubies/ruby-2.2.2/lib/ruby/site_ruby/2.2.0/rubygems/core_ext/kernel_require.rb:54:in `require': cannot load such file -- dashing (LoadError)
	from /usr/local/rvm/rubies/ruby-2.2.2/lib/ruby/site_ruby/2.2.0/rubygems/core_ext/kernel_require.rb:54:in `require'
	from config.ru:1:in `block in <module:PumaRackCompat>'
	from /var/www/sparkseat_dash/shared/bundle/ruby/2.2.0/gems/puma-2.12.2/lib/puma/rack/builder.rb:189:in `instance_eval'
	from /var/www/sparkseat_dash/shared/bundle/ruby/2.2.0/gems/puma-2.12.2/lib/puma/rack/builder.rb:189:in `initialize'
	from config.ru:in `new'
	from config.ru:in `<module:PumaRackCompat>'
	from /var/www/sparkseat_dash/shared/bundle/ruby/2.2.0/gems/puma-2.12.2/lib/puma/rack/builder.rb:183:in `eval'
	from /var/www/sparkseat_dash/shared/bundle/ruby/2.2.0/gems/puma-2.12.2/lib/puma/rack/builder.rb:183:in `new_from_string'
	from /var/www/sparkseat_dash/shared/bundle/ruby/2.2.0/gems/puma-2.12.2/lib/puma/rack/builder.rb:174:in `parse_file'
	from /var/www/sparkseat_dash/shared/bundle/ruby/2.2.0/gems/puma-2.12.2/lib/puma/configuration.rb:104:in `load_rackup'
[...]

PR #863, opened 2016-01-18
Ensure bundler installed rack is on the $LOAD_PATH before globally installed rack

boxofrad:
This fixes the issue described in #859 where `prune_bundler` causes multiple copies of Rack to be loaded in circumstances where it is installed both globally and with bundler.

We cannot fully remove the dependency on Rack because gems such as Sinatra monkey-patch their own extensions onto Rack (see #735). In those cases we must use the real `Rack::Builder` instead of our own `Puma::Rack::Builder`.

We cannot determine the Rack path ahead of time and pass it to `puma-wild` from the `#prune_bundler` method because it would prevent Rack refreshing on restart.

Instead of shelling out to `bundle show` we could consider loading Bundler and using it to determine the gem path, this would either require loading `'bundler/setup' `which is undesirable as it would modify the `$LOAD_PATH` in unexpected ways, or manually implementing just enough of `Bundler.setup`
ourselves, which would deeply couple Puma to the internals of Bundler.

Thanks! :sparkling_heart:

### DIFF
--- a/bin/puma-wild
+++ b/bin/puma-wild
@@ -24,6 +24,28 @@ module Puma; end
 
 Puma.const_set("WILD_ARGS", ["-I", inc, gems])
 
+# Ensure the bundler installed copy of Rack is on the `$LOAD_PATH` before
+# any globally installed copy of Rack (fixes #859).
+#
+# We cannot fully remove the dependency on Rack because gems such as Sinatra
+# monkey-patch their own extensions onto Rack (see #735). In those cases we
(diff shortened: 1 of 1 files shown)


boxofrad:
Admittedly this is a dirty dirty solution, and seems to have broken the build on JRuby... So I'm very much open to other suggestions :smile:

boxofrad:
Closing this in favour of #864

PR #864, opened 2016-01-18
Ensure bundler installed rack is on the $LOAD_PATH before global rack fixes #859

### LINKED ISSUE #859: Prune bundler causes rack to be loaded twice

boxofrad:
This fixes #859 and I think is a better solution that the one I suggested in #863 as it doesn't rely on shelling out to bundler :smile_cat:

### DIFF
--- a/lib/puma/configuration.rb
+++ b/lib/puma/configuration.rb
@@ -128,6 +128,8 @@ module Puma
     # Load and use the normal Rack builder if we can, otherwise
     # fallback to our minimal version.
     def rack_builder
+      ensure_bundler_rack_on_load_path
+
       begin
         require 'rack'
         require 'rack/builder'
@@ -136,6 +138,23 @@ module Puma
         return Puma::Rack::Builder
       else
         return ::Rack::Builder
+      end
+    end
+
+    def ensure_bundler_rack_on_load_path
+      begin
+        require 'bundler'
+      rescue LoadError
+        return
+      end
+
+      begin
+        rack_spec = Bundler.definition.specs['rack'].first
+
+        if rack_spec
+          $LOAD_PATH.unshift File.join(rack_spec.full_gem_path, 'lib')
+        end
+      rescue Bundler::BundlerError
       end
     end
 


evanphx:
Fixed this via https://github.com/puma/puma/commit/c1b85b2b90d05e4d9cff2f83786e69b2bdcbdcdd instead. Thanks!

PR #873, opened 2016-02-01
Override REMOTE_ADDR header if set_remote_address has been used

mfrank01:
Hi,

This fixes an issue where `set_remote_address` is ignored if a proxy service adds the `REMOTE_ADDR` HTTP header but `set_remote_address` is configured to use a different HTTP header, i.e, `CLIENT_IP`

If this change is approved, when would it be possible to get a new gem version released? 

Please let me know what you think. 

Thanks.

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -495,7 +495,7 @@ module Puma
       # intermediary acting on behalf of the actual source client."
       #
 
-      unless env.key?(REMOTE_ADDR)
+      if !env.key?(REMOTE_ADDR) || @options[:remote_address_header] || @options[:remote_address_value]
         begin
           addr = client.peerip
         rescue Errno::ENOTCONN


evanphx:
Hm. Well, the intention of this feature is that it speeds up very high performance servers from having to resolve the remote address. If it's already set though, thats the fastest path.

Can you tell me why you'd want to set it when it's already set?

mfrank01:
When using a proxy server, i.e, Nginx, it automatically adds the `REMOTE_ADDR` field as the IP address of the host machine, and not of the client's IP address. (Source: http://blog.stevesmind.net/2012/10/nginx-not-reading-real-client-ip-address/ )

Typically, an HTTP header like `X-Forwarded-For` would be used to set the client's IP address.

Without this change, using nginx + puma will always result in 127.0.0.1 (or whichever interface to which puma has been bound)  for a `remote_ip` method call to an instance of `ActionDispatch::Request`. At least, this has been my experience.

If I'm overlooking something, please let me know. I'd rather minimize code changes.

evanphx:
That’s certainly true, but usually your application will check
X-Forwarded-For first. The intention of this feature is not to eliminate
people using X-Forwarded-For, it’s mostly about a small perf improvement.

That being said, if you’d like to use the feature this way, let’s do this.
Add an extra optional argument to `set_remote_address` called
`force=false`, If `force` is true, then your conditional logic would be
used. That should give you what you’re looking for.

mfrank01:
Sorry, I tried fixing the problem at the incorrect level. Puma is correctly passing headers along, it turns out my problem was with the RemoteIp middleware in rails. 

They extend the `custom_proxies` configuration variable to include all private IP ranges if the configuration setting is present but not in the type of `Array`. (https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/middleware/remote_ip.rb#L68)

Thanks for your time, I'll close this pull request for now.

PR #875, opened 2016-02-04
Fix server tests

schneems:
The tests all use a port which may not be available (it isn't always open on travis). Instead we can tell the tcp server to bind to port 0 which means it will find a free port, since the tests rely on sending data to that port, we must record the actual port connected to and retrieve that information from the server.


schneems:
Looks like this is still failing on jruby :(

I'm guessing the `addr[1]` interface isn't replicated. I'll try to repro locally.

schneems:
I made some mistakes in the tests, fixed in https://github.com/puma/puma/pull/877

PR #877, opened 2016-02-04
Fix jruby server tests

schneems:
We were accidentally connecting to a `nil` port, in my previous tests. MRI will interpret this the same as port 0. Jruby will not and throws an error

```
Error: test_very_large_return(TestPumaServer): TypeError: can't convert nil into String
org/jruby/ext/socket/RubyTCPServer.java:102:in `initialize'
org/jruby/RubyIO.java:851:in `new'
```

This tests makes sure we're using a port number instead of nil.


schneems:
All tests are passing

PR #883, opened 2016-02-08
Signal not full on worker exit #876

### LINKED ISSUE #876: Puma hangs with min-threads=0 and queue_requests=false
With a config of:

```
threads 0,1
queue_requests false
```

I found that Puma would stop responding after some time.  
We managed to get a stack trace out of the hung workers and found a nasty combination:

- The main server thread was stuck down in `ThreadPool#wait_until_not_full`
- The thread pool worker didn't have a worker thread alive

Presumably there's a race condition somewhere between trimming threads and `wait_until_not_full`; bumping our min-threads to 1 seems to have vanished the problem.

dadah89:
Fixes #876.

With:
```
threads 0,1
queue_requests false
```

Sequence:
* work is added ([calling `<<` in server.rb](https://github.com/puma/puma/blob/f048f38416be48ff6a7bb2a5b192c7d7a4a4e7d1/lib/puma/server.rb#L337))
* next [`wait_until_not_full` is executed](https://github.com/puma/puma/blob/f048f38416be48ff6a7bb2a5b192c7d7a4a4e7d1/lib/puma/server.rb#L338) (and it starts waiting on `@not_full`)
* in another thread, `trim` is executed (because `@waiting` is still > 0, `@trim_requested` gets incremented, marking the worker to be trimmed)
* finally, the worker thread wakes up, decrements `@waiting`, execute the work (removing the last item from `@todo`), finishes it, an then hits the `@trim_requested` check, exiting the loop before signaling `@not_full`, so `wait_until_not_full` never exits

Couldn't figure out a test for this case...

@evanphx what do you think?

/cc @osheroff

### DIFF
--- a/lib/puma/thread_pool.rb
+++ b/lib/puma/thread_pool.rb
@@ -77,6 +77,7 @@ module Puma
               if @trim_requested > 0
                 @trim_requested -= 1
                 continue = false
+                not_full.signal
                 break
               end
 


codehotter:
I think not_full.signal should be in the critical section that also changes spawned. It makes sense to signal there, because that also changes the condition in wait_until_not_full (it increases the right hand side by 1, so the check should run again).

In this location, we signal, then leave critical section and only later, in 2nd critical section decrease spawned. Something could happen in between those sections.

```
diff --git a/lib/puma/thread_pool.rb b/lib/puma/thread_pool.rb
index 11ddcf7..14bc0cc 100644
--- a/lib/puma/thread_pool.rb
+++ b/lib/puma/thread_pool.rb
@@ -111,6 +111,7 @@ module Puma
         mutex.synchronize do
           @spawned -= 1
           @workers.delete th
+          @not_full.signal
         end
       end
```

dadah89:
oh missed the notification, will check

evanphx:
Great work! I reviewed the locks and it's fine.

codehotter:
The race is probably more rare now, but can't it still occur?

1) work is added, 1 waiting
2) wait_until_not_full is executed
   @todo.size = 1
   @waiting = 1
   @max = 1
   @spawned = 1
   1 - 1 < 1 - 1 is false so we block
3) trim is requested. 
4) worker thread wakes up, decrements @waiting, executes the work.
5) trim is requested, so we set continue = false, signal not_full, then leave the mutex.synchronize
6) main server thread wakes up in wait_until_not_full:
   @todo.size = 0
   @waiting = 0
   @spawned = 1
   @max = 1
   0 - 0 < 1 - 1 is false so we block again
7) worker thread wakes up, decrements spawn and disappears. not_full is not signaled again and we have the deadlock again.

PR #897, opened 2016-02-22
Backlog parameter for unix sockets

vanchi-zendesk:
You specify the backlog similar to a TCP socket with "?backlog=1024". This way we can change net.core.somaxconn to a higher value if we want, and use an appropriate backlog to the socket.

### DIFF
--- a/lib/puma/binder.rb
+++ b/lib/puma/binder.rb
@@ -111,6 +111,7 @@ module Puma
 
             umask = nil
             mode = nil
+            backlog = nil
 
             if uri.query
               params = Util.parse_query uri.query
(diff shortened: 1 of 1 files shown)


vanchi-zendesk:
cc @zendesk/rules

vanchi-zendesk:
Just rebased against puma:master

vanchi-zendesk:
The test failure is unrelated. Also, I think caused by this:

```
      @launcher.events.fire_on_booted!

      begin
        server.run.join
```

server.run creates `@thread_pool` and the event is fired before that?

PR #901, opened 2016-02-26
fix uninitialized constant Puma::ControlCLI::StateFile

mitto:
```
$ RAILS_ENV=production SECRET_KEY_BASE=xxxx  bundle exec pumactl -F /var/www/app/current/config/puma.rb status

uninitialized constant Puma::ControlCLI::StateFile
```


corrupt952:
:+1:

PR #924, opened 2016-03-06
Fixes #922 - Correctly define file encoding as UTF-8

### LINKED ISSUE #922: "Incompatible character encodings: ASCII-8BIT and UTF-8" somewhere in Rack's libraries
Here is the Rackup file to reproduce:

```
# Here is the workaround:
=begin
require 'puma'
class Puma::Configuration
  def inspect
    "(not available)"
  end
end
=end

run(lambda { |env| ['200', {'Content-Type' => 'text/html'}, ["Hi!"]] })
```

This is in Ruby 1.9.3.

tbrisker:
Line 103 contains the unicode character ñ which causes this to break rake for ruby 1.9.3. 
Specifying the encoding for the file fixes this error.

### DIFF
--- a/lib/puma/const.rb
+++ b/lib/puma/const.rb
@@ -1,3 +1,4 @@
+#encoding: utf-8
 module Puma
   class UnsupportedOption < RuntimeError
   end


pglombardo:
Great on the quick fix.  I just hit this in my [travis tests](https://travis-ci.org/appneta/ruby-traceview/jobs/114077986) if you need an example.

deepj:
Unfortunately, Puma supports only Ruby 2.1 and later now.

pglombardo:
Ah ok that's understandable.  The gemspec should probably be updated to reflect that as well then.  Thanks!

PR #926, opened 2016-03-07
Set required ruby version to 2.1+ in gemspec

deepj:
See https://github.com/puma/puma/pull/924#issuecomment-193021059


evanphx:
Sorry, no plans to do this. Puma still works on older ruby versions, I'm just relegating my support of them to a lower tier.

deepj:
OK

PR #942, opened 2016-04-01
Socket activation cleanup, fixes, and a doc

dekellum:
Socket activation didn't work for SSL or any listeners specifying additional parameters.

These commits  begin with some logging and basic cleanup of the prior socket activation additions. 3030870 fixes socket activation for the additional listener types.  And finally I add a docs/systemd.md to give a more complete example of the whole setup.


evanphx:
Thanks!

PR #954, opened 2016-04-11
Update comment for Puma::Events#unknown_error (does not take an `env` anymore)

jf:
as per subject.


jf:
Ok, given that no actual code has changed, but only the text of a comment, I'm not expecting that this should fail. How do I request a restart of the checks?

PR #955, opened 2016-04-11
Add request info to standard error rescue

jf:
Error backtraces should really come with this context (makes pinpointing where we should look much easier). Would be open to some discussion about how best to do the printing. Puma::Events#unknown_error could handle it... but I'd have to pass in the `env` request object.

### DIFF
--- a/lib/puma/events.rb
+++ b/lib/puma/events.rb
@@ -106,15 +106,21 @@ module Puma
     end
 
     # An unknown error has occurred.
-    # +server+ is the Server object, +error+ an exception object, and
-    # +kind+ some additional info.
+    # +server+ is the Server object, +error+ an exception object,
+    # +kind+ some additional info, and +env+ the request.
     #
-    def unknown_error(server, error, kind="Unknown")
+    def unknown_error(server, error, kind="Unknown", env=nil)
       if error.respond_to? :render
         error.render "#{Time.now}: #{kind} error", @stderr
       else
-        @stderr.puts "#{Time.now}: #{kind} error: #{error.inspect}"
-        @stderr.puts error.backtrace.join("\n")
+        if env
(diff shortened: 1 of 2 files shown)


evanphx:
No, we don't touch STDERR directly here. You need to use `@events`.

jf:
Ok. Let me know if this is ok for you? I had to send `@events` the `env` as well (the alternative was to just specifically send `env['REQUEST_METHOD']` and `env['PATH_INFO']`). I also collapsed the `@stderr.puts` calls into one, so it writes all the lines in one call.

lib/puma/events.rb
```diff
@@ -106,15 +106,21 @@ def ssl_error(server, peeraddr, peercert, error)
...
-        @stderr.puts "#{Time.now}: #{kind} error: #{error.inspect}"
-        @stderr.puts error.backtrace.join("\n")
+        if env
+          string_block = [ "#{Time.now}: #{kind} error handling request { #{env['REQUEST_METHOD']} #{env['PATH_INFO']} }" ]
+          string_block << error.inspect
+        else
+          string_block = [ "#{Time.now}: #{kind} error: #{error.inspect}" ]
+        end
+        string_block << error.backtrace
+        @stderr.puts string_block.join("\n")
```
evanphx: I'd prefer you just call `@stderr.puts` multiple times rather than make an array and then join it.

jf: I'd prefer multiple calls too. Sorry for not elaborating; the only reason I did a single call is so that there wont be the possibility (?) of other output from other `@stderr.puts` calls being interleaved among these debug lines. Do you think this is a valid concern / makes sense?

jf: Any comments, @evanphx ? Would appreciate your input on this.

Issue #650, opened 2015-02-06
HTTP Parse Error


gshilin:
I'm using ```ruby 2.1.2p95```, ```rails 4.1.8```, ```puma 2.9.2```. For the following plain ```GET``` request 
``` http://localhost:3000/download/files/%D7%A4%D7%A8%D7%95%D7%99%D7%A7%D7%98%2020132%20personal%20ECG%20% ``` I'm getting
```
2015-02-05 17:36:30 +0200: HTTP parse error, malformed request (): #<Puma::HttpParserError: Invalid HTTP format, parsing fails.>
2015-02-05 17:36:30 +0200: ENV: {"rack.version"=>[1, 2], "rack.errors"=>#<IO:<STDERR>>, "rack.multithread"=>true, "rack.multiprocess"=>false, "rack.run_once"=>false, "SCRIPT_NAME"=>"", "CONTENT_TYPE"=>"text/plain", "QUERY_STRING"=>"", "SERVER_PROTOCOL"=>"HTTP/1.1", "SERVER_SOFTWARE"=>"2.9.2", "GATEWAY_INTERFACE"=>"CGI/1.2", "REQUEST_METHOD"=>"GET"}
---
```
Nothing fancy, just simple ```GET``` via browser

evanphx:
You have a trailing % with no entity which is a decoding error.

gshilin:
That's not me, that's come from outside. And I'm not getting it because of puma error.

evanphx:
Well, the request is invalid, how should puma handle it?

gshilin:
To supply configuration option: where to route invalid requests to
-- OR --
To reply with 404

PR #967, opened 2016-04-21
Add http logging to parse error

shicholas:
Sometimes I get errors like the one described in
https://github.com/puma/puma/issues/650. I made this PR with the
intention of printing the malformed request to STDOUT (at least for
MRI). I hope this accomplishes that.

I think this is a better alt to the error message and I don't think
there's a downside figuring the event is (hopefully) rare and log
management can easily handle these messages nowadays.


shicholas:
looks like #955 does this and more?

evanphx:
Well, I'm not sure what you're trying to do, but you didn't run this code because printf doesn't return a string and passing a pointer to %s is a crash.

Issue #968, opened 2016-04-23
Cannot bind SSL port due to missing verify_mode option


frankwong15:
Related to #832 

Reproduced in [frankwong15/rails-puma](https://github.com/frankwong15/rails-puma)

```
frank@frank-nb:~/projects/rails-puma$ puma -C config/puma.rb 
Puma starting in single mode...
* Version 3.4.0 (jruby 1.9.3), codename: Owl Bowl Brawl
* Min threads: 0, max threads: 16
* Environment: development
DEPRECATION WARNING:
Sass 3.5 will no longer support Ruby 1.9.3.
Please upgrade to Ruby 2.0.0 or greater as soon as possible.

* Listening on tcp://0.0.0.0:8080
* Listening on ssl://0.0.0.0:8443?keystore=keystore.jks&keystore-pass=password
Use Ctrl-C to stop
2016-04-23 15:23:43 +0800: Listen loop error: #<TypeError: can't convert nil into Integer>
org/jruby/puma/MiniSSL.java:162:in `initialize'
org/jruby/puma/MiniSSL.java:131:in `server'
/home/frank/.rvm/gems/jruby-1.7.19/gems/puma-3.4.0-java/lib/puma/minissl.rb:167:in `accept_nonblock'
/home/frank/.rvm/gems/jruby-1.7.19/gems/puma-3.4.0-java/lib/puma/server.rb:329:in `handle_servers'
org/jruby/RubyArray.java:1613:in `each'
/home/frank/.rvm/gems/jruby-1.7.19/gems/puma-3.4.0-java/lib/puma/server.rb:324:in `handle_servers'
/home/frank/.rvm/gems/jruby-1.7.19/gems/puma-3.4.0-java/lib/puma/server.rb:297:in `run'

```

PR #969, opened 2016-04-23
Fix puma/puma#968 Cannot bind SSL port due to missing verify_mode option

frankwong15:
Fix puma/puma#968

### DIFF
--- a/lib/puma/binder.rb
+++ b/lib/puma/binder.rb
@@ -194,7 +194,7 @@ module Puma
                                 MiniSSL::VERIFY_NONE
                               end
           else
-            ctx.verify_mode = MiniSSL::VERIFY_NONE
+            ctx.verify_mode = MiniSSL::VERIFY_PEER
           end
 
           if fd = @inherited_fds.delete(str)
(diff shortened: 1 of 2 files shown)


lib/puma/binder.rb
```diff
@@ -179,20 +179,22 @@ def parse(binds, logger)
...
+                              when "force_peer"
+                                MiniSSL::VERIFY_PEER | MiniSSL::VERIFY_FAIL_IF_NO_PEER_CERT
+                              when "none"
+                                MiniSSL::VERIFY_NONE
+                              else
+                                @events.error "Please specify a valid verify_mode="
+                                MiniSSL::VERIFY_NONE
+                              end
+          else
+            ctx.verify_mode = MiniSSL::VERIFY_NONE
```
reidmorrison: To prevent a security vulnerability from being flagged, would it be safer to default to `MiniSSL::VERIFY_PEER`?

frankwong15:
I would like to keep the existing default and only fix the issue I have. I have no opinion on other changing the default afterwards.

PR #974, opened 2016-04-28
Set worker thread name.

reidmorrison:
Supports multithreaded logging and problem diagnosis.


reidmorrison:
All tests pass aside from one pre-existing test failure on JRuby, which is unrelated to this change.

PR #1007, opened 2016-06-24
[ci skip]Update default number of threads

willnet:
Default number of threads is 16 in puma. So I think DEPLOYMENT.md needs to be fixed.

(English is not my native language, so I'm sorry if I misunderstood.)


willnet:
😃

Issue #957, opened 2016-04-11
Keep-alive Timeout


timabdulla:
How can we configure the keep-alive timeout? It seems that Puma is using the value contained in `PERSISTENT_TIMEOUT` to control this, but it doesn't seem to be accessible from the config DSL. is there a recommended way to set this variable?

evanphx:
@timabdulla I responded in gitter (btw, no need to ask in an issue and in there). It's not exposed now but I'd merge a PR to expose it. Can you tell me a little about why you want to set it?

timabdulla:
hey, thanks! sorry for asking twice.

we use amazon's ELB as a load-balancer, which routes directly to our puma instances. the idle timeout on our ELB is 60 seconds, which controls both the time the ELB will wait for a response from the server before issuing a 504, and controls the amount of time which the ELB will keep the connection open if it's a persistent connection. if the persistent timeout is lower on the backend web server, the ELB will error out and issue a 504 if the connection is closed before the ELB expects it to be when a new client attempts to initiate a connection with the ELB (see this link under the heading "Cause 2" for more info: http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/ts-elb-error-message.html#ts-elb-errorcodes-http504). we would like to adjust our puma config to make the persistent timeout higher than 60 seconds. if that makes sense to you, we can prepare a PR over here to expose this setting.

evanphx:
@timabdulla Sure, sounds like a cromulent reason. Happy to merge that PR!

dollschasingmen:
hi -- where is this PR and was it merged? we are seeing similar issues with amazons ELB

jorihardman:
@dollschasingmen The master branch now supports this. You can use `persistent_timeout` in your puma config to set the number of seconds.

PR #1017, opened 2016-07-15
Allow persistent_timeout to be configured via the dsl.

jorihardman:
@evanphx This is in response to #957. When using puma directly behind an Amazon ELB, ELB will issue a 504 if the server closes a connection sooner than the "idle timeout". This PR let's you configure puma's timeout for persistent connections.


jorihardman:
Tests are failing due to a bundler error: `An error occurred while installing rack (2.0.1), and Bundler cannot continue.`

PR #1025, opened 2016-07-21
adds network address

### LINKED ISSUE #1020: Access the server from other device
I'm trying to access my server from another device by visiting my IP address, but the server is not responding. This is probably related to https://github.com/puma/puma/issues/782 , but i can't find a way to fix this.

Any hints welcome, thanks! :)

walidvb:
This should close #1020, adding listening to the local network IP host in addition to localhost

### DIFF
--- a/lib/puma/binder.rb
+++ b/lib/puma/binder.rb
@@ -261,6 +261,10 @@ module Puma
         localhost_addresses.each do |addr|
           add_tcp_listener addr, port, optimize_for_latency, backlog
         end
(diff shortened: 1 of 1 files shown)


evanphx:
This just forces puma to listen on your devices public address. I'm not going to have puma always do that. If you want to do that, most folks will just have it bind to 0.0.0.0 so that it can be accessed remotely.

evanphx:
Er, your devices non-localhost addresses.

PR #1031, opened 2016-07-26
Rename History.txt to History.rdoc

swrobel:
History.txt appears to be in rdoc format, so it should have that extension. Not having an extension breaks changelog parsers like [vandamme](https://github.com/tech-angels/vandamme)


evanphx:
It's not actually rdoc format, it's adhoc. I'd rather convert it to
markdown actually.

swrobel:
@evanphx will do, closing and will submit and alternate PR

PR #1049, opened 2016-08-16
Rename History.md to CHANGELOG.md

WaKeMaTTa:
Because is the common.


evanphx:
No thanks. I like History.

WaKeMaTTa:
The truth is hard to find it, as History. :disappointed:

PR #1056, opened 2016-08-21
Use NIO to manage the reactor

evanphx:
If we go forward with the native websocket support, upgrading the reactor capacity is probably a good idea too.

@tarcieri: love to get your input here on using NIO.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,6 +10,6 @@ gem "test-unit", "~> 3.0"
 gem "rack", "< 2.0"
 gem 'minitest', '~> 5.8'
 
-gem "nio"
+gem "nio4r"
 
 gem "jruby-openssl", :platform => "jruby"
--- a/Gemfile
+++ b/Gemfile
@@ -10,4 +10,6 @@ gem "test-unit", "~> 3.0"
 gem "rack", "< 2.0"
 gem 'minitest', '~> 5.8'
 
+gem "nio"
+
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -10,4 +10,6 @@ gem "test-unit", "~> 3.0"
 gem "rack", "< 2.0"
 gem 'minitest', '~> 5.8'
 
+gem "nio"
```
tarcieri: I think you want `nio4r` here: https://rubygems.org/gems/nio

evanphx: Whoops! Thanks.

lib/puma/reactor.rb
```diff
@@ -16,17 +17,44 @@ def initialize(server, app_pool)
       @sleep_for = DefaultSleepFor
       @timeouts = []
 
-      @sockets = [@ready]
+      @selector = NIO::Selector.new
+      @selector.register(@ready, :r).value = nil
```
tarcieri: You can replace this with the `NIO::Selector#wakeup` mechanism, which is basically doing the same thing behind the scenes

tarcieri:
You'll probably want to add a `@selector.close` here:

https://github.com/puma/puma/blob/d8bf71a9ef09acd25932a4b528812a58bd8ac090/lib/puma/reactor.rb#L164

PR #1059, opened 2016-08-24
Add support for `verify_name`.

### LINKED ISSUE #1057: Fixed CommonName verify mode
I need Puma's client certificate authentication to allow me to verify against a fixed Common Name (i.e., the inverse of Nginx's `proxy_ssl_name`, a model that's possible with stunnel and PostgreSQL that we've used to great success).  The idea is: I don't know what host front-end servers will spring up at, but with a pre-configured CA and a valid client certificate with a common name I expect, I'm fine with puma accepting it (and otherwise not).

I'll be implementing this over the next few days into Puma/MiniSSL, but I was wondering if this is something that'd be considered useful to Puma proper, or is it too special case to upstream.  (e.g., I'm wondering if I should structure it as a PR against Puma or a monkey-patching Gem).  If it's something that'd be useful in Puma, what would it need to look like to get merged?

The way I'm thinking about it now, it'd work something like `verify_mode=peer&verify_name=something`.  Also, I'm not familiar with JRuby (other than knowing it's awesome), but would MRI-only support be a blocking issue for getting something like this merged?

Thoughts?

mieko:
- Using `verify_name=value` tests the commonName of the client leaf certificate against a fixed value
- `verify_name` implies VERIFY_PEER and FAIL_IF_NO_PEER_CERT
- verify_name value is technically compared against the commonName with === to allow more exotic uses (Regexp, Proc, etc)
- Add client_arbitrary.{crt,key} for tests
- Use the SSL app_data for passing around a ms_conn*, which now holds its failed_cert

Closes #1057

MRI only at the moment.  Not considering this complete, looking for general feedback.


nateberkopec:
I need to "learn to C" before I can review this properly, but the use case and the Ruby seems fine to me. Can you rebase please?

mieko:
Thanks for checking it out, @nateberkopec.  I should be able run another pass over this PR and rebase over the next week or so.

nateberkopec:
Ping!

nateberkopec:
tag @NickolasVashchenko

nateberkopec:
@mieko Can you rebase against master?

nateberkopec:
Closing as stale

PR #1061, opened 2016-08-29
Fix Puma init.d start

michaelsauter:
+start+ did not read the environment variables. However, e.g. +phased-restart+
did. Also +start+ was implemented quite differently from other actions.

Now +start+ is implemented like the other actions, it respects environment
variables and there's a lot less code duplication going on.


michaelsauter:
I see the tests have failed, but it looks like some kind of timeout. Probably unrelated to the change here?

Anything else I can do to get this merged?

PR #1064, opened 2016-08-31
Remove 100-199 status codes from constant.

frodsan:
The code actually checks if status is between 100 and 199:

```ruby
no_body ||= status < 200 || STATUS_WITH_NO_ENTITY_BODY.include?(status)
```

### DIFF
--- a/lib/puma/const.rb
+++ b/lib/puma/const.rb
@@ -76,12 +76,11 @@ module Puma
   # For some HTTP status codes the client only expects headers.
   #
 
-  no_body = {}
-  ((100..199).to_a << 204 << 205 << 304).each do |code|
-    no_body[code] = true
-  end
-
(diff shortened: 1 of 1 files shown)


evanphx:
But why should this be done instead of the current code?

frodsan:
I don't fully understand the question. But my changes are based on this commit https://github.com/puma/puma/commit/ba27787d0ddeec09330c1cd85d20f585a66b3b8e.

When Rack was removed as a dependency, 100-199 status codes were added again:
https://github.com/puma/puma/commit/537bc21593182cd9c4c0079a3936d05b1f91fe14#diff-2fedc25f1c4d43a510ad5e4f6a5c4ed3R82 but I don't think that is necessary when `status < 200` is present.

evanphx:
Ah ok. Your update makes much more sense! Thanks!

frodsan:
👍 thanks!

PR #1066, opened 2016-08-31
Fix warning: MiniTest::Unit::TestCase is now Minitest::Test.

frodsan:


evanphx:
How does this fix the warning?

frodsan:
This is the only test file where minitest's class is used. The other files use `Test::Unit::TestCase`. I think test/unit uses minitest behind the curtains. I also can change `MiniTest::Unit::TestCase` to `Minitest::Test`. Would you like to update the test suite to use only minitest classes?

PR #1070, opened 2016-09-01
Drop support for 1.8.7 and 1.9.2.

frodsan:
Puma requires 1.9.3 or higher (ref: https://github.com/puma/puma/pull/994)


frodsan:
Looks like Ruby 1.9.3 and 2.0.0 are not supported: https://github.com/puma/puma/commit/b4a2fded22299dd8a11fd64ea88aeb3b33a0846e

PR #1084, opened 2016-09-08
Use delete_if to remove dead workers.

frodsan:
```
allocated memory by file
-----------------------------------
544  /Users/frodsan/Code/frodsan/puma/lib/puma/thread_pool.rb (Before)
 40  /Users/frodsan/Code/frodsan/puma/lib/puma/thread_pool.rb (After)

retained memory by location
-----------------------------------
504  /Users/frodsan/Code/frodsan/puma/lib/puma/thread_pool.rb:184 (Before)
  0  /Users/frodsan/Code/frodsan/puma/lib/puma/thread_pool.rb:184 (After)
```

### DIFF
--- a/lib/puma/thread_pool.rb
+++ b/lib/puma/thread_pool.rb
@@ -181,7 +181,9 @@ module Puma
           @spawned -= 1
         end
 
-        @workers -= dead_workers
+        @workers.delete_if do |w|
+          dead_workers.include?(w)
+        end
       end
     end
 


frodsan:
Actually, something like this could work better:

```ruby
def reap
  @mutex.synchronize do
      @workers.each_with_index do |worker, idx|
        next if worker.alive?

        worker.kill

        @spawned -= 1

        @workers.delete_at(idx)
    end
  end
end
```

PR #1088, opened 2016-09-10
Minor refactor on Thread pool

frodsan:
- Puma::Threadpool uses only one constant from the constants file. At the beginning, I was thinking to add the missing require to const.rb to the ThreadPool, but I think it's better to just move the constant to where it's used.
- I removed the check for `timeout == 0` since the `times` block will not be executed if it's zero.

### DIFF
--- a/lib/puma/const.rb
+++ b/lib/puma/const.rb
@@ -113,11 +113,6 @@ module Puma
     # sending data back
     WRITE_TIMEOUT = 10
 
-    # How long, after raising the ForceShutdown of a thread during
-    # forced shutdown mode, to wait for the thread to try and finish
-    # up it's work before leaving the thread to die on the vine.
-    SHUTDOWN_GRACE_TIME = 5 # seconds
-
     # The original URI requested by the client.
     REQUEST_URI= 'REQUEST_URI'.freeze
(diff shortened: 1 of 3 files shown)


evanphx:
I believe this is wrong because it remove the call to `Thread#raise` when 0 is used.

frodsan:
This is the code now:

```ruby
when 0
  threads.each do |t|
    t.raise ForceShutdown
  end

  threads.each do |t|
    t.join Const::SHUTDOWN_GRACE_TIME
  end
else
  timeout.times do
    # ...
  end

  threads.each do |t|
    t.raise ForceShutdown
  end

  threads.each do |t|
    t.join Const::SHUTDOWN_GRACE_TIME
  end
end
```

And this is the proposed change:

```ruby
else
  # If timeout is `0`, it will skip this part,
  # and raise (force shutdown) immediately
  timeout.times do
    # ...
  end

  threads.each do |t|
    t.raise ForceShutdown
  end

  threads.each do |t|
    t.join Const::SHUTDOWN_GRACE_TIME
  end
end
```

I think in both scenarios #raise is called, when timeout is equal or greater than 0.

nateberkopec:
Makes sense to me - clever use of `0.times`.

PR #1105, opened 2016-09-27
Allow extra runtime deps to be defined when using prune_bundler

daveallie:
Extra gems are sometimes needed in the `before_fork` block or before Bundler gets loaded on the workers. This PR adds the ability to define any extra bundler gems to load when pruning bundler.

As an example, when using https://github.com/schneems/puma_worker_killer, it should be loaded in the before_fork of the `config/puma.rb` file like so:

``` ruby
before_fork do
  require 'puma_worker_killer'

  PumaWorkerKiller.config do |config|
    config.ram           = 1024
    config.frequency     = 5
    config.percent_usage = 0.98
    config.reaper_status_logs = false
  end
  PumaWorkerKiller.start
end
```

If you have installed the gem from the github repo, and are using the `prune_bundler` option, the `require 'puma_worker_killer'` will fail. The only way to get it to run is to require bundler, and then require the gem in the context of bundler. Which, if you're also attempting to prune bundler, completely defeats the purpose and means gems aren't reloaded across phased restarts.

### DIFF
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -604,8 +604,8 @@ module Puma
     # When using prune_bundler, if extra runtime dependencies need to be loaded to
     # initialize your app, then this setting can be used.
     #
-    # For each gem's name passed, that gem will be loaded when the environment
-    # is pruned.
+    # Before bundler is pruned, the gem names supplied will be looked up in the bundler
+    # context and then loaded again after bundler is pruned.
     # Only applies if prune_bundler is used.
     #
     # @example
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -601,6 +601,21 @@ module Puma
       @options[:raise_exception_on_sigterm] = answer
     end
 
+    # When using prune_bundler, if extra runtime dependencies need to be loaded to
+    # initialize your app, then this setting can be used.
+    #
+    # For each gem's name passed, that gem will be loaded when the environment
+    # is pruned.
+    # Only applies if prune_bundler is used.
+    #
+    # @example
+    #   extra_runtime_dependencies ['gem_name_1', 'gem_name_2']
+    # @example
+    #   extra_runtime_dependencies ['puma_worker_killer']
+    def extra_runtime_dependencies(answer = [])
+      @options[:extra_runtime_dependencies] = Array(answer)
+    end
+
     # Additional text to display in process listing.
     #
     # If you do not specify a tag, Puma will infer it. If you do not want Puma
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -584,6 +584,7 @@ module Puma
     # dictates.
     #
     # @note This is incompatible with +preload_app!+.
+    # @note This is only supported for RubyGems 2.2+
     def prune_bundler(answer=true)
       @options[:prune_bundler] = answer
     end
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -275,14 +275,12 @@ module Puma
         "#{d.name}:#{spec.version.to_s}"
       end
 
-      if @options[:extra_runtime_dependencies]
-        @options[:extra_runtime_dependencies].each do |d_name|
-          spec = Bundler.rubygems.loaded_specs(d_name) rescue nil
-          if spec
-            dirs += spec.require_paths.map { |x| File.join(spec.full_gem_path, x) }
-          else
-            log "* Couldn't to load extra dependency: #{d_name}"
-          end
+      @options[:extra_runtime_dependencies].each do |d_name|
+        spec = Bundler.rubygems.loaded_specs(d_name)
+        if spec
+          dirs += spec.require_paths.map { |x| File.join(spec.full_gem_path, x) }
+        else
+          log "* Could not load extra dependency: #{d_name}"
         end
       end
 
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -280,6 +280,7 @@ module Puma
           spec = Bundler.rubygems.loaded_specs(d_name) rescue nil
           if spec
             deps << "#{d_name}:#{spec.version.to_s}"
+            dirs += spec.require_paths.map { |x| File.join(spec.full_gem_path, x) }
           else
             log "* Couldn't to load extra dependency: #{d_name}"
           end
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -261,8 +261,8 @@ module Puma
 
     def prune_bundler
       return unless defined?(Bundler)
-      puma = Bundler.rubygems.loaded_specs("puma")
-      dirs = puma.require_paths.map { |x| File.join(puma.full_gem_path, x) }
+      puma = spec_for_gem("puma")
+      dirs = require_paths_for_gem(puma)
       puma_lib_dir = dirs.detect { |x| File.exist? File.join(x, '../bin/puma-wild') }
 
       unless puma_lib_dir
@@ -271,14 +271,13 @@ module Puma
       end
 
       deps = puma.runtime_dependencies.map do |d|
-        spec = Bundler.rubygems.loaded_specs(d.name)
-        "#{d.name}:#{spec.version.to_s}"
+        "#{d.name}:#{spec_for_gem(d.name).version}"
       end
 
       @options[:extra_runtime_dependencies].each do |d_name|
-        spec = Bundler.rubygems.loaded_specs(d_name)
+        spec = spec_for_gem(d_name)
         if spec
-          dirs += spec.require_paths.map { |x| File.join(spec.full_gem_path, x) }
+          dirs += require_paths_for_gem(spec)
         else
           log "* Could not load extra dependency: #{d_name}"
         end
@@ -295,6 +294,18 @@ module Puma
         args += [{:close_others => false}]
         Kernel.exec(*args)
       end
+    end
+
+    def spec_for_gem(gem_name)
+      Bundler.rubygems.loaded_specs(gem_name)
+    end
+
+    def require_paths_for_gem(gem_spec)
+      # full_require_paths is a RubyGems 2.2+ method
+      return gem_spec.full_require_paths if gem_spec.method_defined?(:full_require_paths)
+
+      # if full_require_paths doesn't exist, fallback to building the paths
+      gem_spec.require_paths.map { |x| File.join(gem_spec.full_gem_path, x) }
     end
 
     def log(str)
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -279,7 +279,6 @@ module Puma
         @options[:extra_runtime_dependencies].each do |d_name|
           spec = Bundler.rubygems.loaded_specs(d_name) rescue nil
           if spec
-            deps << "#{d_name}:#{spec.version.to_s}"
             dirs += spec.require_paths.map { |x| File.join(spec.full_gem_path, x) }
           else
             log "* Couldn't to load extra dependency: #{d_name}"
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -275,6 +275,15 @@ module Puma
         "#{d.name}:#{spec.version.to_s}"
       end
 
+      @options[:extra_runtime_dependencies].each do |d_name|
+        spec = Bundler.rubygems.loaded_specs(d_name) rescue nil
+        if spec
+          deps << "#{d_name}:#{spec.version.to_s}"
+        else
+          log "* Couldn't to load extra dependency: #{d_name}"
+        end
+      end
+
       log '* Pruning Bundler environment'
       home = ENV['GEM_HOME']
       Bundler.with_clean_env do
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -261,6 +261,9 @@ module Puma
 
     def prune_bundler
       return unless defined?(Bundler)
+
+      require_rubygems_min_version!(Gem::Version.new("2.2"), "prune_bundler")
+
       puma = spec_for_gem("puma")
       dirs = require_paths_for_gem(puma)
       puma_lib_dir = dirs.detect { |x| File.exist? File.join(x, '../bin/puma-wild') }
@@ -274,7 +277,7 @@ module Puma
         "#{d.name}:#{spec_for_gem(d.name).version}"
       end
 
-      @options[:extra_runtime_dependencies].each do |d_name|
+      Array(@options[:extra_runtime_dependencies]).each do |d_name|
         spec = spec_for_gem(d_name)
         if spec
           dirs += require_paths_for_gem(spec)
@@ -301,11 +304,7 @@ module Puma
     end
 
     def require_paths_for_gem(gem_spec)
-      # full_require_paths is a RubyGems 2.2+ method
-      return gem_spec.full_require_paths if gem_spec.method_defined?(:full_require_paths)
-
-      # if full_require_paths doesn't exist, fallback to building the paths
-      gem_spec.require_paths.map { |x| File.join(gem_spec.full_gem_path, x) }
+      gem_spec.full_require_paths
     end
 
     def log(str)
@@ -449,6 +448,13 @@ module Puma
       rescue Exception
         log "*** SIGHUP not implemented, signal based logs reopening unavailable!"
       end
+    end
+
+    def require_rubygems_min_version!(min_version, feature)
+      return if min_version <= Gem::Version.new(Gem::VERSION)
+
+      raise "#{feature} is not supported on your version of RubyGems. " \
+              "You must have RubyGems #{min_version}+ to use this feature."
     end
   end
 end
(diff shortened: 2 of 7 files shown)


daveallie:
Upon further testing it looks like this won't worker for bundler gem installed from git as https://github.com/puma/puma/blob/master/bin/puma-wild#L20 calls `gem 'gemname', 'gemversion'` which looks for rubygem gems.

daveallie:
Not running the extra dependencies through the `gem` call alleviates my previous concern, adding the path to $LOAD_PATH is enough.

nateberkopec:
Paging @schneems

schneems:
Hi, I wrote puma worker killer. Does it work if in a `before_block` ? https://github.com/schneems/puma_worker_killer/issues/43

I've not used `prune_bundler` before. Sounds like we want to do it if you're updating your gems so that bundler will update your load path. I would think that we want this to work with all git dependencies out of the box and not just special ones that we define. Does bundler have any ways to "refresh" a load path from an updated Gemfile?

daveallie:
`prune_bundler` runs before the `before_fork` block. It unloads all gems, except puma iirc, so that during your app boot, you can get bundler's context and load all the gems back in. I believe it is done this way so that individual workers hold the bundler context, allowing a rolling restart to reload all gems (and potentially use new ones if the gemfile has changed).

Before submitting it this way I attempted to load puma worker killer using bundler in the `before_fork` block, but to do it I needed to load bundler as well (as I was using a git ref for the gem). This meant that app workers were booted with a provided bundler context, meaning that rolling restarts didn't reload existing gems, unload old ones, or reload new ones.

I could have copy-pasted the logic behind `prune_bundler` into the top of my `before_fork` but thought it better to loop it into the existing logic and make it available for anyone else in the same situation.

daveallie:
Any update on this PR?

Dombo:
Hi, we've encountered a usecase for this inside one of our production applications - I notice that CI is failing on unrelated tests to this PR, has any of the maintainers had a chance to review the comment made by @daveallie on 07/12/2016.

Thanks

nateberkopec:
@Dombo can you expand on your use-case? Same as @daveallie's?

Dombo:
@nateberkopec sorry if I wasn't clear but yes it's the same use case as @daveallie's

daveallie:
@nateberkopec any chance of moving this PR forward? It's almost been a year and nothing has really happened.

kelvinoshea:
can we get this merged pls?

lib/puma/launcher.rb
```diff
@@ -262,6 +262,17 @@ def prune_bundler
         "#{d.name}:#{spec.version.to_s}"
       end
 
+      if @options[:extra_runtime_dependencies]
```
nateberkopec: This is always `true`.

lib/puma/launcher.rb
```diff
@@ -262,6 +262,17 @@ def prune_bundler
...
+        @options[:extra_runtime_dependencies].each do |d_name|
+          spec = Bundler.rubygems.loaded_specs(d_name) rescue nil
```
nateberkopec: Rather than a `rescue nil`, we should do a proper begin/rescue here, which will get rid of the the `if/else`.

lib/puma/launcher.rb
```diff
@@ -262,6 +262,17 @@ def prune_bundler
...
+          if spec
+            dirs += spec.require_paths.map { |x| File.join(spec.full_gem_path, x) }
+          else
+            log "* Couldn't to load extra dependency: #{d_name}"
```
nateberkopec: "Couldn't to"

lib/puma/dsl.rb
```diff
@@ -414,6 +414,15 @@ def prune_bundler(answer=true)
       @options[:prune_bundler] = answer
     end
 
+    # When using prune_bundler, if extra runtime dependencies need to be loaded to
+    # initialize your app, then this setting can be used.
+    #
+    # For each gem's name passed, that gem will be loaded when the environment
```
nateberkopec: "each gem name passed"

nateberkopec:
Can people accept a Bundler version requirement for this feature? It's currently bugged because the require paths are wrong. I'd rather just use `bundler`'s require_full_paths method but it's bundler 1.15+. 

`gemspec.require_paths` returns mixed absolute and relative paths for me on 1.17.2: `["/Users/nateberkopec/.gem/ruby/2.6.3/extensions/x86_64-darwin-18/2.6.0-static/puma-3.12.1", "lib"]`, which breaks the entire prune process.

nateberkopec:
Also there's no way we can merge this without tests.

nateberkopec:
Here's some ideas for refactoring this method while we're here. https://gist.github.com/nateberkopec/386db42c0eb3fe4fa91b009dd9f1f84b

lib/puma/launcher.rb
daveallie: There's actually no need for a rescue here at all, I don't know what I was thinking https://www.rubydoc.info/github/bundler/bundler/Bundler/RubygemsIntegration#loaded_specs-instance_method, `loaded_specs` is just a hash lookup. 🤦‍♂️

daveallie:
Apologies for the sloppy nature of this PR, will bring it up to scratch

daveallie:
So rubygem's `full_require_paths` didn't exist until rubygems v2.2.0 which isn't required by bundler until v2. Setting a bundler requirement of `>= 2` seems a little extreme for this. We could do something similar to [`Bundler.rubygems.loaded_gem_paths.loaded_gem_paths`](https://github.com/bundler/bundler/blob/50c9389c077ca4026b478dec1e812abdafe1118d/lib/bundler/rubygems_integration.rb#L150) and utilise `full_require_paths` if it's available, otherwise fallback to attempting to build full paths ourselves

nateberkopec:
>Apologies for the sloppy nature of this PR

Apologies for taking 4 years to leave a comment on it 😆

nateberkopec:
> similar to Bundler.rubygems.loaded_gem_paths and utilise full_require_paths if it's available, otherwise fallback to attempting to build full paths ourselves

Well first of all, the current method of building full paths is just straight up bugged, as I commented.

Second, if both code paths are correct, then we should just use the one we wrote rather than introduce a conditional to run someone else's equally correct code.

So we have to fix how we're building our own absolute paths.

daveallie:
The way to fix it would be to straight up rip out `full_require_paths` from rubygems and put in in Puma, which is looking like a worse and worse idea.

There are 2 main options I can see in order to move forward with this:
1. Enforce min rubygems version
2. Gate feature on min rubygems version

Neither have been done in Puma yet which kind of put me off them, but of the two I think the second option is the best.

I think you were pretty much spot on from the start, however not sure how you got the Bundler 1.15+ dep? afaict, it's just rubygems 2.2.0+ (which is bundler 1.3+, (or 1.6.1+ for root C extension installation) -- https://bundler.io/compatibility.html)

nateberkopec:
@schneems need a second opinion here

nateberkopec:
Did some more digging.

We can definitely use anything in RubyGems 2.2+, because that was released in 2013 and is older than our lowest supported Ruby version, 2.2. So I'm OK with gating this feature on it.

I too have no idea what I meant about a method only being in Bundler 1.15, so ignore that 😆 

What do you think @daveallie?

x-yuri:
@daveallie @nateberkopec Doesn't this fix gem versions? I mean, let's say I deployed a site (e.g. using [`mina`][1]), then changed the source code to use a newer version of `puma` and of some other gems that are part of `extra_runtime_dependencies`. On second deploy I won't get the new versions, will I?

As such, wasn't the decision to close #1295 way to rash?

[1]: https://github.com/mina-deploy/mina

Why can't `puma` just execute itself as is (w/o modifying the command)?

nateberkopec:
@x-yuri #1295 touched the exact same code and was running into the exact same bug as this PR, but in this PR I have an active contributor that's responding to reviews. So whatever #1295 was trying to fix is also going to have to work with this PR.

daveallie:
Hey @nateberkopec, sorry for my absence here, changed jobs in the last few weeks and it has been hectic. Yeah, as mentioned gating is my preferred and will be least intrusive, will have an update to the PR very soon with those changes.

daveallie:
@x-yuri, yes it will fix gem versions for the listed gems. In the same way you can't do a phased restart with a puma gem updated, the gems listed will need to full server stop/start. This feature should only be used for gems which are needed as part of the master puma process (i.e. in `before_fork`)

daveallie:
@nateberkopec - with the current changes, there is a change to the core requiring logic for `prune_bundler` setups. If a user has RubyGems 2.2+, they will now use `Gem::Specification#full_require_paths` (as in `require_paths_for_gem`) when determining the path for Puma's binary. Behaviour is unchanged for older versions.

It should be a direct improvement for RubyGems 2.2+ clients, and solve #1272 as #1295 was looking into. Unfortunately, there's not much we can do about super old clients (hopefully there aren't too many people still running the pre-2013 RubyGems).

lib/puma/launcher.rb
```diff
@@ -286,6 +294,18 @@ def prune_bundler
       end
     end
 
+    def spec_for_gem(gem_name)
+      Bundler.rubygems.loaded_specs(gem_name)
+    end
+
+    def require_paths_for_gem(gem_spec)
```
nateberkopec: So, we're actually requiring Rubygems 2.2+ *to use prune_bundler* at all now, is that correct?

I'm actually OK with that as well (again, rubygems 2.2+ is older than ruby 2.2) but we need to gate the code on that as well.

nateberkopec: Oh I see, you've done a fallback instead. 

I'm wondering if we should even do that. I think I sort of proved in the comments of this PR that the original implementation was bugged.

daveallie: I was just planning on straight-up raising on attempts to use the feature without rubygems 2.2+ (https://github.com/puma/puma/pull/1105/commits/85c7f21cff603b1e65e2e2dadb620ae47b72001e#diff-d9e00ab028817951ecfdbbe7e661a141R471) instead of silently failing. The fallback codepath here is because this method is now used to find the puma lib dir. Previously it was just the fallback line, run inline. Now it's the preferred codepath for clients with rubygems 2.2+ but we still support the old codepath.

nateberkopec: The problem is that the old codepath is clearly bugged.

Try this script and you'll see what I mean: 

```ruby
require 'bundler'
require 'puma'
gem_spec = Bundler.rubygems.loaded_specs("puma")
puts gem_spec.full_require_paths
puts gem_spec.require_paths.map { |x| File.join(gem_spec.full_gem_path, x) }
```

The latter line, for me, outputs nonsense like: 

```
"/Users/nateberkopec/.gem/ruby/2.6.3/gems/puma-4.0.1/Users/nateberkopec/.gem/ruby/2.6.3/extensions/x86_64-darwin-18/2.6.0-static/puma-4.0.1"
```

...which does not exist.

There's not really a point in maintaining a 2nd codepath that doesn't actually work, so we need to either remove the path or fix it.

MSP-Greg: Puma's requiring Ruby >= 2.2.0, which shipped with [RubyGems 2.4.5](https://github.com/ruby/ruby/blob/v2_2_0/lib/rubygems.rb#L12)

nateberkopec: I'm unfamiliar with what "shipped with" means _exactly_ in this context. Is it possible that _anyone_ is using Ruby 2.2 or greater with Rubygems 2.3 or less?

MSP-Greg: Sorry.  It's possible, but Ruby 2.2.0 included RubyGems 2.4.5 as a std-lib.

'Shipped with' <=> 'included'

nateberkopec: Right. Well I'd rather provide a feature that works over one that doesn't, so our choices are to: 

a) copy the entire implementation of `full_require_paths` to this file and then remove the rubygems gate 
b) remove the branch, and gate the entire feature of prune_bundler on Rubygems 2.2+

I am fine with either option.

daveallie:
I found it pretty difficult to test this using the test framework, happy to take suggestions on how to better do it.

Very basic test in there at the moment that ensures that the load path is adjusted with the new dependency.

lib/puma/launcher.rb
```diff
@@ -269,8 +272,16 @@ def prune_bundler
       end
 
       deps = puma.runtime_dependencies.map do |d|
-        spec = Bundler.rubygems.loaded_specs(d.name)
-        "#{d.name}:#{spec.version.to_s}"
+        "#{d.name}:#{spec_for_gem(d.name).version}"
+      end
+
+      Array(@options[:extra_runtime_dependencies]).each do |d_name|
```
nateberkopec: Hm, how would it be possible for this not to be an Array by this point? Is this coercion necessary?

nateberkopec:
@daveallie Looks great!

The easiest way to test would be to refactor the original method even further. Lines 265-285 could easily become their own method.

```ruby

def dependencies_and_files_to_require_after_prune
  puma = spec_for_gem("puma")

  deps = puma.runtime_dependencies.map do |d|
    "#{d.name}:#{spec_for_gem(d.name).version}"
  end
  
  [deps, require_paths_for_gem(puma) + extra_runtime_deps_directories]
end

def extra_runtime_deps_directories
  Array(@options[:extra_runtime_dependencies]).each do |d_name|
    if (spec = spec_for_gem(d_name))
      dirs += require_paths_for_gem(spec)
    else
      log "* Could not load extra dependency: #{d_name}"
    end
  end
end

def puma_wild_location
  puma = spec_for_gem("puma")
  dirs = require_paths_for_gem(puma)
  puma_lib_dir = dirs.detect { |x| File.exist? File.join(x, '../bin/puma-wild') }
  File.expand_path(File.join(puma_lib_dir, "../bin/puma-wild"))
end

def prune_bundler
  require_rubygems_min_version!(Gem::Version.new("2.2"), "prune_bundler")
  log "! Unable to prune Bundler environment, continuing" unless puma_wild_location
  
  deps, dirs = dependencies_and_files_to_require_after_prune
  
  home = ENV['GEM_HOME']
  Bundler.with_clean_env do
    ENV['GEM_HOME'] = home
    ENV['PUMA_BUNDLER_PRUNED'] = '1'
    args = [Gem.ruby, puma_wild_location, '-I', dirs.join(':'), deps.join(',')] + @original_argv
    # Ruby 2.0+ defaults to true which breaks socket activation
    args += [{:close_others => false}]
    Kernel.exec(*args)
  end
end
```

That's a start. Then you can just easily unit-test the stuff you've been touching rather than start up an entire server.

lib/puma/launcher.rb
daveallie: yeah it is, the value's default is `nil` and I think it's easier if we just accept `nil`, a single item, or an array

test/test_launcher.rb
```diff
@@ -0,0 +1,63 @@
+require_relative "helper"
+
+require "puma/configuration"
+require "puma/const"
+require "puma/launcher"
+
+class TestLauncher < Minitest::Test
+  def test_dependencies_and_files_to_require_after_prune_is_correctly_built_for_no_extra_deps
+    skip_on :appveyor, suffix: " - bundler not used in appveyor so prune bundler logic tests unavailable"
```
nateberkopec: @MSP-Greg there's no bundler on appveyor?

nateberkopec:
Need to give this one last review but it looks brilliant and adds tons of test coverage to prune_bundler that we didn't have before 🙏 Thank you @daveallie!

test/test_launcher.rb
```diff
@@ -0,0 +1,63 @@
...
+    assert_match(%r{puma-#{Puma::Const::PUMA_VERSION}$}, dirs[1]) # native extension dir
+    assert_match(%r{gems/rdoc-[\d.]+/lib$}, dirs[2]) # rdoc dir
+  end
+
+  def test_extra_runtime_deps_directories_is_empty_for_no_config
+    l = Puma::Launcher.new Puma::Configuration.new
+    assert_equal([], l.send(:extra_runtime_deps_directories))
+  end
+
+    def test_extra_runtime_deps_directories_is_correctly_built
```
nateberkopec: Is it just how github is showing this or is this def not correctly aligned?

test/test_launcher.rb
```diff
@@ -0,0 +1,63 @@
...
+    assert_match(%r{gems/rdoc-[\d.]+/lib$}, dep_dirs.first)
+  end
+
+  def test_puma_wild_location_is_an_absolute_path
+    skip_on :appveyor, suffix: " - bundler not used in appveyor so prune bundler logic tests unavailable"
+    l = Puma::Launcher.new Puma::Configuration.new
+    puma_wild_location = l.send(:puma_wild_location)
+    assert_match(%r{bin/puma-wild$}, puma_wild_location)
+    # assert no "/../" in path
+    assert(puma_wild_location !~ %r{/\.\./})
```
nateberkopec: refute_match

MSP-Greg:
@nateberkopec 

> there's no bundler on appveyor?

Give me a few days.  I'm still looking at tests...

test/test_launcher.rb
nateberkopec: @MSP-Greg 👋 just wondering as I don’t understand why this is a thing

PR #1110, opened 2016-10-12
Replace use of Process.getpgid which does not behave as intended on all platforms

montdidier:
This pull request has two small fixes for issues I experienced with puma 3.6.0
1. Using pumactl (3.6.0) on OSX it was possible to come cross uninitialised constant Puma::StateFile. I have rectified this.
2. On some platforms (specifically encountered on OpenBSD) Process.getpgid which is backed by the POSIX.1 getgpid implementation does not behave in a fashion which the original implementer expected. 

Process.getpgid is being used to confirm the existence of a running process before proceeding on to act upon said process. 

Using this man page as a [reference](http://man.openbsd.org/getpgid.2) one can see that unless getpgid is called from within the same session as the running process a permission error will be returned. Considering the main uses cases; running in the same session will not be the norm. The exception handling block surrounding Porcess.getpgid does not differentiate between an unknown process id and permission denied. Therefore the error "unknown pid" is returned and execution ceases.

Considering that the various kill methods in waiting result in the same exception I decided to simply change the method logic from "asking permission" to "asking forgiveness" and circumventing the need for Process.getpgid all-together.

### DIFF
--- a/lib/puma/control_cli.rb
+++ b/lib/puma/control_cli.rb
@@ -179,39 +179,39 @@ module Puma
       end
 
       begin
-        Process.getpgid @pid
+
+        case @command
+        when "restart"
+          Process.kill "SIGUSR2", @pid
+
+        when "halt"
+          Process.kill "QUIT", @pid
+
+        when "stop"
+          Process.kill "SIGTERM", @pid
+
+        when "stats"
+          puts "Stats not available via pid only"
+          return
+
+        when "reload-worker-directory"
+          puts "reload-worker-directory not available via pid only"
+          return
+
+        when "phased-restart"
+          Process.kill "SIGUSR1", @pid
+
+        else
+          message "Puma is started"
+          return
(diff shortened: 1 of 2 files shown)


montdidier:
Looks like it was a hung build?

nateberkopec:
@montdidier Can you remove the Puma version bump and Puma::StateFile commits (so leaving only b7a68791405729f5e). We're gonna fix the others separately. (cc #1138)

nateberkopec:
@montdidier LGTM but I don't think we have test coverage for the failure case? Can you add one?

montdidier:
Sorry for slow response. Will update as requested.

montdidier:
Ok. I've removed the Puma version bump and Puma::StateFile and added a test to cover the failure case. Anything else?

nateberkopec:
Brilliant! We'll take it from here.

PR #1111, opened 2016-10-12
Config first data timeout

alexlance:


alexlance:
There is an issue with the AWS ELB, where if the ELB's idle timeout is larger than the tcp keepalive timeout on the instance, then the ELB will sometimes report back 504s.

If you configure puma with this first_data_timeout setting, then you can circumvent that problem.

This patch exposes the first_data_timeout setting to the puma DSL. I have manually tested this patch and it works exactly as intended.

nateberkopec:
I'm going to say this is probably blocking on getting `test_timeout_in_data_phase` fixed, otherwise we have no coverage for this.

alexlance:
Right. I don't know how to fix that.

nateberkopec:
Heh me neither right now, sorry :) Not your fault/problem, we just need to fix that before merging.

alexlance:
So apparently if you add `travis_wait` in front of the `bundle exec` command in the .travis.yml file then you get 20 minutes of grace time instead of 10 minutes from travis ci. Would that help?

nateberkopec:
I don't think so, it's a deadlock that doesn't ever resolve.

frodsan:
I made some notes about that test here: https://github.com/puma/puma/commit/43ee6ddcacc1c5925a5e51059573dd80ccf10be0#commitcomment-20006908

nateberkopec:
Alright, you know what. This is just adding a config point, your use case is valid, not a big deal. Sorry this took so long to merge.

Issue #1103, opened 2016-09-27
TTIN not working?


michaelsauter:
### Steps to reproduce

1) Start Puma in clustered mode, e.g. with 2 workers

```
bind "tcp://0.0.0.0:3000"
workers 2
threads 1, 1
prune_bundler
environment 'development'
pidfile File.expand_path("../../tmp/puma/pid", __FILE__)
state_path File.expand_path("../../tmp/puma/state", __FILE__)
activate_control_app
```

2) `kill -s TTIN <PID of master process>`
### Expected behavior

I expected Puma to spawn a new worker.
### Actual behavior

Nothing seems to happen.
### System configuration

**Ruby version**: 2.2.4

I had a look at the implementation (https://github.com/puma/puma/blob/master/lib/puma/cluster.rb#L336) but I couldn't figure out why it wouldn't work. Also, TTOU works as expected.

Or is my initial assumption wrong that TTIN increases the worker count beyond the amount specified in the configuration?

steakknife:
TTIN and TTOU aren't working as expected. 

TTIN doesn't seem to do anything

TTOU appears to just kill a worker at random and then another one takes its place.

pchaganti:
👍

steakknife:
Why did you close this issue? It's still broken on HEAD.

nateberkopec:
Fixed by #1118

nateberkopec:
TTOU works as expected, but it isn't logging anything. I'll open a separate issue.

steakknife:
Ah that makes sense. Thanks for weekend FOSS hacking.

nateberkopec:
Reopening after the revert for 3.6.2.

fidalgo:
Any plans to get this?

nateberkopec:
Fixed on master, just forgot about this issue 👍

fidalgo:
Sorry @nateberkopec but can you point the commit or the PR?

If I start a Rails server `rails s` with:
 ```
Puma starting in single mode...
* Version 3.7.0 (ruby 2.3.3-p222), codename: Snowy Sagebrush
* Min threads: 5, max threads: 5
* Environment: development
* Listening on tcp://0.0.0.0:3000
```
I still can't see the Backtrace, and by the commit messages after 3.7.0 I can't infer the commit on this.

nateberkopec:
@fidalgo Sorry, can you expand on what you expect to happen? This issue is for TTIN signals, which should increase worker count.

[...]

PR #1118, opened 2016-10-21
fix access priorities of each level in LeveledOptions

hiroara:
Related to #1103 .


nateberkopec:
Excellent catch!

hiroara:
Gotcha! :)

PR #1121, opened 2016-10-24
plugins: listen_reaper add_plugin_restart_hook

steakknife:
Adds
- `plugin :listen_reaper` - stop all `Listen` instances on restart
- `plugin :add_plugin_restart_hook` - allow plugins to be triggered `on_restart`


steakknife:
This could also be done directly without donkey-punching the runtime.

nateberkopec:
I think this is probably more appropriate as a separate gem, since it's platform-specific and fixes problems caused by other gems.

PR #1129, opened 2016-11-10
Allow rack handler to accept ssl host

chtitux:
It then allow rails to start puma and pass the TLS/SSL configuration:
~~~
bin/rails server puma -b "ssl://0.0.0.0?key=/path/to/example.key&cert=/path/to/example.crt"
~~~

### DIFF
--- a/lib/rack/handler/puma.rb
+++ b/lib/rack/handler/puma.rb
@@ -35,6 +35,10 @@ module Rack
 
           if host && (host[0,1] == '.' || host[0,1] == '/')
             c.bind "unix://#{host}"
+          elsif host && host =~ /^ssl:\/\//
+            uri = URI.parse(host)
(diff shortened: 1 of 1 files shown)


chtitux:
@nateberkopec Should I add some lines to my patch? Or everything is fine for now?

lib/rack/handler/puma.rb
```diff
@@ -35,6 +35,10 @@ def self.run(app, options = {})
 
           if host && (host[0,1] == '.' || host[0,1] == '/')
             c.bind "unix://#{host}"
+          elsif host && host =~ /^ssl:\/\//
```
olleolleolle: Perhaps not use a regular expression, but a String instance method?

```ruby
elsif host && host.start_with?('ssl://')
```

PR #1135, opened 2016-11-18
Transfer encoding header should be downcased before comparison

jkraemer:
- OSX Finder sends 'Chunked' which resulted in failed WebDAV uploads due to
  matching against 'chunked'

### DIFF
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -223,7 +223,7 @@ module Puma
 
       te = @env[TRANSFER_ENCODING2]
 
-      if te == CHUNKED
+      if te && te.downcase == CHUNKED
         return setup_chunked_body(body)
       end
 
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -223,7 +223,7 @@ module Puma
 
       te = @env[TRANSFER_ENCODING2]
 
-      if te && te.downcase == CHUNKED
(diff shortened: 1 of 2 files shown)


nateberkopec:
Sounds good to me. [It's right there in the RFC:](https://tools.ietf.org/html/rfc2616#section-3.6)

> All transfer-coding values are case-insensitive

nateberkopec:
After a quick lookaround the code it looks like this is the only place we actually read transfer-encoding, so this LGTM.

lib/puma/client.rb
```diff
@@ -223,7 +223,7 @@ def setup_body
 
       te = @env[TRANSFER_ENCODING2]
 
-      if te == CHUNKED
+      if te && te.downcase == CHUNKED
```
nateberkopec: Actually, I have no idea if this is a performance sensitive path or not, but using `casecmp` [will be ever so slightly faster here](https://github.com/JuanitoFatas/fast-ruby#stringcasecmp-vs-stringdowncase---code) (and allocates one fewer object):

```ruby
CHUNKED.casecmp(te) == 0
```

jkraemer: good point, I changed that in my second commit.

nateberkopec:
Thanks for your contribution!

Issue #1137, opened 2016-11-20
Deadlock using restart plugin triggered from `rails s`


steakknife:
### Steps to reproduce

```shell
gem install rails
rails new puma-rails-rack-test --database=sqlite --skip-test --skip-turbolinks --skip-javascript  --skip-active-record --skip-action-mailer --skip-keeps --skip-git --skip-listen --skip-bundle
cd puma-rails-rack-test
echo "gem 'rack', github: 'rack', branch: 'pr/1127'" >> Gemfile
bundle
bin/rails s

# in another terminal
touch tmp/restart.txt
```
### Expected behavior

It should restart just like when run using `bin/bundle exec puma`

### Actual behavior

Rails hangs, CPU usage is high

```
=> Booting Puma
=> Rails 5.0.0.1 application starting in development on http://localhost:3000
=> Run `rails server -h` for more startup options
Puma starting in single mode...
* Version 3.6.0 (ruby 2.3.1-p112), codename: Sleepy Sunday Serenity
* Min threads: 5, max threads: 5
* Environment: development
* Listening on tcp://localhost:3000
Use Ctrl-C to stop
* Restarting...
=> Booting Puma
=> Rails 5.0.0.1 application starting in development on http://localhost:3000
=> Run `rails server -h` for more startup options
^CExiting
```

### System configuration

- Ruby 2.4.0
- Rails 5.0.1
- Rack master + pr/1127
- Puma master

nateberkopec:
I wonder if this is related to #1184

steakknife:
Still an issue:

```
=> Booting Puma
=> Rails 5.1.1 application starting in development on http://localhost:3000
=> Run `rails server -h` for more startup options
Puma starting in single mode...
* Version 3.8.2 (ruby 2.4.1-p111), codename: Sassy Salamander
* Min threads: 5, max threads: 5
* Environment: development
* Listening on tcp://0.0.0.0:3000
Use Ctrl-C to stop
* Restarting...
=> Booting Puma
=> Rails 5.1.1 application starting in development on http://localhost:3000
=> Run `rails server -h` for more startup options
```

steakknife:
Still an issue on 3.9.0

steakknife:
Closing this since it's never getting fixed.

stereobooster:
I wonder if this is the problem with `touch tmp/restart.txt` here too https://github.com/puma/puma/issues/1060. Need to try with `kill -USR2 pid`

[...]

PR #1138, opened 2016-11-20
Really fix `uninitialized constant Puma::StateFile`

steakknife:
```
$ bin/bundle exec pumactl -S tmp/puma.state restart
uninitialized constant Puma::StateFile
```

requiring puma doesn't require puma/state_file

Related: #901

### DIFF
--- a/lib/puma/control_cli.rb
+++ b/lib/puma/control_cli.rb
@@ -1,5 +1,5 @@
 require 'optparse'
-require 'puma'
+require 'puma/state_file'
 require 'puma/const'
 require 'puma/detect'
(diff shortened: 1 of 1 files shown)


steakknife:
(travis failure unrelated to this PR)

nateberkopec:
Can't repro, but I haven't tried Rails yet (which it looks like you're doing, using the Rails bundler binstub?)

steakknife:
bundler, Rails and springified

steakknife:
Ruby 2.3.1

steakknife:
Follow #1137 setup

PR #1140, opened 2016-11-20
Fix rbx build, trusty beta on Travis

nateberkopec:


nateberkopec:
> When using Rubinius, there’s currently an issue with selecting the correct version in RVM in our build environment, but only when specifying rbx as your version. As a workaround, specify rbx-2 instead.

Issue #1044, opened 2016-08-08
Tests failing on FreeBSD (race condition)


jlduran:
I've been experiencing defunct processes (one per worker) with puma in cluster mode on FreeBSD (10 & 11), although the app seems to run fine. On Debian, the same app doesn't show any zombies.

```
[app@example.com ~]$ ps aux
USER   PID %CPU %MEM    VSZ    RSS TT  STAT STARTED    TIME COMMAND
app  14799  0.4  0.7  13904   3648  0  Ss   06:15   0:00.05 -bash (bash)
app  14798  0.2  1.4  60372   7036  -  S    06:15   0:00.02 sshd: app@pts/0 (sshd)
app  14498  0.0  3.2  84084  15536  -  S    04:51   0:02.46 ruby: puma 3.6.0 (unix:///usr/local/www/apps/test_app/
app  14499  0.0 27.9 323400 137192  -  I    04:51   0:30.44 ruby: puma: cluster worker 0: 14498 [test_app] (ruby)
app  14500  0.0 29.9 329952 146652  -  I    04:51   0:33.67 ruby: puma: cluster worker 1: 14498 [test_app] (ruby)
app  14502  0.0  0.0      0      0  -  Z    04:52   0:00.00 <defunct>
app  14504  0.0  0.0      0      0  -  Z    04:52   0:00.01 <defunct>
app  14831  0.0  0.5  21156   2612  0  R+   06:15   0:00.01 ps aux
```

I decided to run the test suite and found these two errors, I wonder if they are related:

```
===================================================================================================================
Failure:
  </\{ "workers": 2, "phase": 0, "booted_workers": 0, "old_workers": 0, "worker_status": \[\{ "pid": \d+, "index": 0, "phase": 0, "booted": false, "last_checkin": "[^"]+", "last_status": \{\} \},\{ "pid": \d+, "index": 1, "phase": 0, "booted": false, "last_checkin": "[^"]+", "last_status": \{\} \}\] \}/> expected to be =~
[...]

PR #1143, opened 2016-11-20
Skip test_phased_restart_via_pumactl

nateberkopec:
Skip this until we can fix it, too racey right now.


nateberkopec:
See #1044

PR #1145, opened 2016-11-20
Travis: jruby-9.1.6.0 is latest stable

olleolleolle:
This PR changes to use latest stable jruby in Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ rvm:
   - 2.2.6
   - 2.3.2
   - ruby-head
-  - jruby-9.1.5.0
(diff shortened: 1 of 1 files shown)


nateberkopec:
😆  You still get your PR ;)

Merging as soon as it passes...

olleolleolle:
Heh, one day we'll all get tired of this and automate it once and for all. Until then, we can high-five at each patch release, like this: ✋

nateberkopec:
I've always found travis/rvm fuzzy matching to be super buggy, so I don't even bother with it.

PR #1158, opened 2016-11-23
Add tests for Puma::NullIO method comments.

frodsan:


lib/puma/null_io.rb
```diff
@@ -1,38 +1,27 @@
...
-    # Never yields
-    #
     def each
     end
 
-    # Mimics IO#read with no data
+    # Mimics IO#read with no data.
     #
-    def read(count=nil,buffer=nil)
+    def read(count = nil, _ = nil)
```
nateberkopec: I'd prefer `_buffer`

frodsan:
@nateberkopec fixed

Issue #1147, opened 2016-11-20
TTOU doesn't log anything


nateberkopec:
Looks confusing, even though when checking with `ps` it appears to be working correctly, killing the last spawned worker. Leads to output like:

```
[23860] Puma starting in cluster mode...
[23860] * Version 3.6.0 (ruby 2.3.2-p217), codename: Sleepy Sunday Serenity
[23860] * Min threads: 1, max threads: 1
[23860] * Environment: development
[23860] * Process workers: 2
[23860] * Phased restart available
[23860] * Listening on tcp://0.0.0.0:3000
[23860] Use Ctrl-C to stop
[23860] * Starting control server on unix:///var/folders/11/xdcy4xcj5vn8b202bzp1pym80000gn/T/puma-status-1479669005626-23860
[23863] + Gemfile in context: /Users/nateberkopec/Code/pumatest/Gemfile
[23864] + Gemfile in context: /Users/nateberkopec/Code/pumatest/Gemfile
[23860] - Worker 0 (pid: 23863) booted, phase: 0
[23860] - Worker 1 (pid: 23864) booted, phase: 0
[23871] + Gemfile in context: /Users/nateberkopec/Code/pumatest/Gemfile
[23860] - Worker 2 (pid: 23871) booted, phase: 0
[23893] + Gemfile in context: /Users/nateberkopec/Code/pumatest/Gemfile
[23860] - Worker 2 (pid: 23893) booted, phase: 0
```

In the above output I sent TTIN, TTOU, then TTIN again.

See #1103

dannyfallon:
This was fixed in #1165 and I think it can be closed. Here's the output in my local env running Puma and sending TTOU followed by 2xTTIN:

```
[4464] - Worker 1 (pid: 4470) booted, phase: 0
[4464] - Worker 2 (pid: 4474) booted, phase: 0
[4464] - Worker 3 (pid: 4478) booted, phase: 0
[4464] - Worker 0 (pid: 4466) booted, phase: 0
>> Send TTOU
[4464] - Worker 3 (pid: 4478) terminating
>> Send TTIN (twice)
[4464] - Worker 3 (pid: 4577) booted, phase: 0
[4464] - Worker 4 (pid: 4583) booted, phase: 0
```

nateberkopec:
Tada!

PR #1165, opened 2016-11-28
Move TTOU processing

sriedel:
This pull request changes a few things in the cluster:
- Move the check worker interval into a constant (please check to see if all occurrences of 5 were indeed meant to indicate the same interval)
- Move the killing of workers out of the signal handler into a #cull_workers method called from the #check_workers method
- Skip processing of #spawn_workers and #cull_workers if there is no difference in the number of workers to the expected number of workers (to deal with diff now being able to be < 0 in #spawn_workers)
- When a worker is culled, it is logged (takes care of #1147)

This of course changes the behaviour of puma in dealing with TTOU signals, as they are no longer processed immediately, but there is a delay until the next WORKER_CHECK_INTERVAL expires. On the upside, multiple TTOU signals can be handled at once and the signal handler is leaner and more symmetric to the TTIN signal handler.

Unfortunately I did not come up with a way to add tests for the changes (or the signal handlers themselves) without changing the internal structure of Puma::Launcher and Puma::Cluster. If you can give me pointers to how you'd like to have this tested, I'd be happy to add the tests.

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -115,6 +115,7 @@ module Puma
 
     def spawn_workers
       diff = @options[:workers] - @workers.size
+      return if diff < 1
 
       master = Process.pid
 
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -144,10 +144,12 @@ module Puma
     def cull_workers
       diff = @workers.size - @options[:workers]
       return if diff < 1
-      log "Culling #{diff.inspect} workers"
+
+      debug "Culling #{diff.inspect} workers"
 
       workers_to_cull = @workers[-diff,diff]
-      log "Workers to cull: #{workers_to_cull.inspect}"
+      debug "Workers to cull: #{workers_to_cull.inspect}"
+
       workers_to_cull.each do |worker|
         log "- Worker #{worker.index} (pid: #{worker.pid}) terminating"
         worker.term
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -6,6 +6,8 @@ require 'time'
 
 module Puma
   class Cluster < Runner
+    WORKER_CHECK_INTERVAL = 5
+
     def initialize(cli, events)
       super cli, events
 
@@ -152,7 +154,7 @@ module Puma
     def check_workers(force=false)
       return if !force && @next_check && @next_check >= Time.now
 
-      @next_check = Time.now + 5
+      @next_check = Time.now + WORKER_CHECK_INTERVAL
 
       any = false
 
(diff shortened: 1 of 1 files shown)


lib/puma/cluster.rb
```diff
@@ -178,6 +196,7 @@ def check_workers(force=false)
 
       @workers.delete_if(&:dead?)
 
+      cull_workers
```
nateberkopec: Is there really ever a case where we should cull *and* spawn?

lib/puma/cluster.rb
```diff
@@ -340,7 +359,6 @@ def setup_signals
 
       Signal.trap "TTOU" do
         @options[:workers] -= 1 if @options[:workers] >= 2
-        @workers.last.term
         wakeup!
```
nateberkopec: So this ends up delaying the response to TTOU until the next run of check_workers, right?

lib/puma/cluster.rb
nateberkopec: Maybe this should be rearchitected so that cull_workers and spawn_workers takes an integer argument, and then you calculate the value to give to those methods here. You could then call cull_workers(1) in your TTOU signal trap.

nateberkopec:
> multiple TTOU signals can be handled at once

Can you explain a little more about how/why this might occur?

lib/puma/cluster.rb
nateberkopec: Derp, you said that in your description.

sriedel:
It could happen if the signals are fired in quick succession (and maybe even by some automated process that wants to scale out by say 5 workers). If the signal handler isn't fast enough and the process is busy, signals can get dropped by the kernel, or worse, the signal handler can get interrupted by the same signal being sent again (thus we have a race condition in the signal handler). 

The more I think about it, it may actually be better to rearchitect the entire signal handling block to use a self pipe (q.v. https://www.sitepoint.com/the-self-pipe-trick-explained/) in order to minimize the race condition potential (that already exists with the worker counter increment/decrement code as it stands). But that may be too much change in one go. 

Would you rather have a second pull request introducing a self pipe, or should I try and rewrite everything at once?

nateberkopec:
> and maybe even by some automated process that wants to scale out by say 5 workers

Ah right, this is probably a fairly common case.

I have no opinion on whether or not a self pipe is better or worse here. *Not* including a self pipe with this PR would not make the code any worse than it already stands, so let's do the changes separately.

lib/puma/cluster.rb
evanphx: I think this approach is fine, no need to combine them for now.

evanphx:
:shipit:

nateberkopec:
You heard the man! MERGE!

PR #1169, opened 2016-12-04
Add eof? method to NullIO?

scbrubaker02:
Rack appears to be counting on eof? being defined on objects passed into its multipart parser, e.g. https://github.com/rack/rack/blob/master/lib/rack/multipart/parser.rb#L68 , but this method is not defined on Puma::NullIO.  This created a problem for rack-2.0.1 and puma-3.6.0.

Perhaps, I am missing the full context, but it seems that it should simply return true.

### DIFF
--- a/lib/puma/null_io.rb
+++ b/lib/puma/null_io.rb
@@ -26,6 +26,10 @@ module Puma
       0
     end
 
+    def eof?
+      true
+    end
+
     def sync=(v)
     end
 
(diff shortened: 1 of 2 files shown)


olleolleolle:
@nateberkopec This Rack commit began using `#eof?`: https://github.com/rack/rack/commit/3f6aee5544582abff3911e61dbe3c01506c63fbd

nateberkopec:
@olleolleolle I just added it to the commit message 👍

PR #1172, opened 2016-12-07
 Add background process configuration

mdesantis:
I found that the configuration proposed doesn't fit very well when Puma is daemonized, that is a common situation when Capistrano is used for deployment. I added my configuration which seems to work pretty well together with the default capistrano3-puma gem settings.

### DIFF
--- a/docs/systemd.md
+++ b/docs/systemd.md
@@ -168,3 +168,30 @@ Apr 07 08:40:19 hx puma[28320]: * Activated tcp://0.0.0.0:9233
 Apr 07 08:40:19 hx puma[28320]: * Activated ssl://0.0.0.0:9234?key=key.pem&cert=cert.pem
 Apr 07 08:40:19 hx puma[28320]: Use Ctrl-C to stop
 ~~~~
+
+## Alternative background process configuration
+
+If Capistrano and [capistrano3-puma](https://github.com/seuros/capistrano-puma) tasks are used you can use the following configuration. In this case, you would skip systemd Socket Activation, since Puma handles the socket by itself:
+
+~~~~
+[Service]
+# Background process configuration (use with --daemon in ExecStart)
+Type=forking
+
+# To learn which exact command is to be used to execute at "ExecStart" of this
+# Service, ask Capistrano: `cap <stage> puma:start --dry-run`. Your result
+# may differ from this example, for example if you use a Ruby version
+# manager. `<WD>` is short for "your working directory". Replace it with your
+# path.
+ExecStart=bundle exec puma -C <WD>/shared/puma.rb --daemon
+
+# To learn which exact command is to be used to execute at "ExecStop" of this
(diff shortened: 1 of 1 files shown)


mdesantis:
Does GitHub allow to squash the commits or I have necessarly to clone my fork and sqush them? :doughnut:

bdewater:
I think you should use systemd as it is meant to be, like written here: [Don’t Daemonize your Daemons!](http://www.mikeperham.com/2014/09/22/dont-daemonize-your-daemons/). According to the systemd manpage on writing unit files:

> If possible, do not use the Type=forking setting in service files. But if you do, make sure to set the PID file path using PIDFile=. See systemd.service(5) for details.

IMO the capistrano3-puma gem should be updated to use systemd now that all major Linux distributions have switched to it. I've adapted its default templates and overridden a couple of rake tasks to  make the gem work with systemd.

mdesantis:
@bdewater I agree wtih you; still I believe that the daemonized configuration is not an evil alternative and it could be worth to be mentioned.

docs/systemd.md
```diff
@@ -168,3 +168,24 @@ Apr 07 08:40:19 hx puma[28320]: * Activated tcp://0.0.0.0:9233
...
+
+If Capistrano and [capistrano3-puma](https://github.com/seuros/capistrano-puma) tasks are used you can use the following configuration. In this case, you would skip systemd Socket Activation, since Puma handles the socket by itself:
+
+~~~~
+[Service]
+# Background process configuration (use with --daemon in ExecStart)
+Type=forking
+
+# Use `cap <stage> puma:start -n` in order to spot the exact command,
+# which could change for example is a Ruby version manager is used
```
olleolleolle: Hi!

This text about how to define a systemd Service Unit file can be improved to avoid jargon.

```
+# Use `cap <stage> puma:start -n` in order to spot the exact command,
+# which could change for example is a Ruby version manager is used
```

After reading [a little about the systemd way of defining Services](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/System_Administrators_Guide/sect-Managing_Services_with_systemd-Unit_Files.html#sect-Managing_Services_with_systemd-Unit_File_Create) and Capistrano, I propose this different wording of the example text:  

```
# To learn which exact command is to be used to execute at "ExecStart" of this
# Service, ask Capistrano: `cap <stage> puma:start --dry-run`. Your result
# may differ from this example, for example if you use a Ruby version
# manager. `<WD>` is short for "your working directory". Replace it with your
# path.
```

olleolleolle: Thanks for improving Puma!

nateberkopec:
I'm not super familiar with systemd, but more documentation is usually better than less, even if it's not 100% on the cutting-edge. @bdewater if you've got something to add, would love to look at a PR.

PR #1114, opened 2016-10-19
Prune bundler should inherit fds

sj26:
Ruby 2.0 changed this behaviour so fds are closed by default during an exec. If using socket activation from systemd however this closes the fds before they can be activated (with some extra debug output added into `Puma::Binder#import_from_env`):

```
systemd[1]: Started puma.
puma[11696]: {"LISTEN_FDS"=>"1", "LISTEN_PID"=>"11696"}
puma[11696]: fd=3 sock=#<TCPServer:fd 3>
puma[11696]: % Registered unix:/app/tmp/sockets/puma.sock for activation from LISTEN_FDS
puma[11696]: * Pruning Bundler environment
puma[11696]: {"LISTEN_FDS"=>"1", "LISTEN_PID"=>"11696"}
puma[11696]: /app/vendor/bundle/ruby/2.3.0/gems/puma-3.6.0/lib/puma/binder.rb:64:in `for_fd': Bad file descriptor - not a socket file descriptor (Errno::EBADF)
puma[11696]:         from /app/vendor/bundle/ruby/2.3.0/gems/puma-3.6.0/lib/puma/binder.rb:64:in `block (2 levels) in import_from_env'
puma[11696]:         from /app/vendor/bundle/ruby/2.3.0/gems/puma-3.6.0/lib/puma/binder.rb:62:in `times'
puma[11696]:         from /app/vendor/bundle/ruby/2.3.0/gems/puma-3.6.0/lib/puma/binder.rb:62:in `block in import_from_env'
puma[11696]:         from /app/vendor/bundle/ruby/2.3.0/gems/puma-3.6.0/lib/puma/binder.rb:56:in `each'
puma[11696]:         from /app/vendor/bundle/ruby/2.3.0/gems/puma-3.6.0/lib/puma/binder.rb:56:in `import_from_env'
puma[11696]:         from /app/vendor/bundle/ruby/2.3.0/gems/puma-3.6.0/lib/puma/launcher.rb:55:in `initialize'
puma[11696]:         from /app/vendor/bundle/ruby/2.3.0/gems/puma-3.6.0/lib/puma/cli.rb:65:in `new'
puma[11696]:         from /app/vendor/bundle/ruby/2.3.0/gems/puma-3.6.0/lib/puma/cli.rb:65:in `initialize'
puma[11696]:         from /app/vendor/bundle/ruby/2.3.0/gems/puma-3.6.0/bin/puma-wild:29:in `new'
[...]

PR #1174, opened 2016-12-14
Launcher: Typo: argv should be args here

olleolleolle:
This PR amends a "probable typo". 

Everywhere ELSE in the file, it's `argv` but on the next line, we use `args`.

**Credit**: @parhs

   - see #1114


sj26:
Oh, nice catch. What a typo!

PR #1175, opened 2016-12-14
Raise error when `stdout` or `stderr` is not writable

jemiam:
This PR make you detect wrong arguments for `stdout_redirect` option easily.


evanphx:
This is incorrect as it's not uncommon to allow puma to make those files when it needs them.

evanphx:
Oh der. I misread the patch, it's checking that the directory exists.

PR #1179, opened 2016-12-20
set RACK_URL_SCHEME to https when X-FORWARDED-PROTO is https

jnimety:
This makes sure that `rack.url_scheme` is set correctly in ssl proxy environments.


nateberkopec:
Yeah. @jnimety let's try to line up more with what Rack does here: https://github.com/rack/rack/blob/c1437097dcdf92d53a692ca8135a3391791fbca3/lib/rack/request.rb#L186

nateberkopec:
Closing as dead in favor of #1491

PR #1186, opened 2017-01-04
Add option to reject connections when shutting down.

aquach:
Heya,

This is a change we recently made to make Puma more compatible with the way we're using nginx.

When nginx receives a connection error from an upstream proxy server, it temporarily removes the proxy server from the rotation and resends traffic to a different upstream. With the usual Puma behavior, requests to Puma while it's in shutdown mode hang until all threads are finished, at which point Puma shuts down. Only then does nginx send its requests to the next upstream. With this flag on, Puma actively rejects all connections so that nginx will immediately begin redirecting requests.

In other words, this change immediately cuts off clients that attempt to connect while Puma is shutting down rather than hanging until Puma finishes shutting down.

I'm all ears on feedback or if there's a better way to do this :)

### DIFF
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -161,6 +161,13 @@ module Puma
       @options[:drain_on_shutdown] = which
     end
 
+    # Immediately close connections when in graceful shutdown mode rather than
+    # not accepting them. This signals to downstream proxies that requests for
+    # the server should go somewhere else.
+    def reject_when_shutting_down(which=true)
+      @options[:reject_when_shutting_down] = which
+    end
+
     # Set the environment in which the Rack's app will run.
     def environment(environment)
       @options[:environment] = environment
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -860,6 +860,22 @@ module Puma
         @events.debug "Drained #{count} additional connections."
       end
 
+      if @options[:reject_when_shutting_down]
+        Thread.new do
+          while true
+            ios = IO.select @binder.ios
+            ios.first.each do |sock|
+              begin
+                if io = sock.accept_nonblock
+                  io.close
+                end
+              rescue SystemCallError
+              end
+            end
+          end
+        end
+      end
+
       if @thread_pool
         if timeout = @options[:force_shutdown_after]
           @thread_pool.shutdown timeout.to_i


nateberkopec:
I wonder why someone would even *want* the current behavior (i.e. why even make this configurable).

aquach:
Good question. It seems more correct for Puma to immediately reject to signify shutdown. I only added it as a configuration so that it wouldn't change existing behavior. Happy to make it the default too, if that's what you're interested in.

nateberkopec:
I'll defer to @evanphx on whether or not it's a good idea. IMO, if this doesn't break people's apps, it should be default and no need for a config point.

nateberkopec:
Paging @schneems too

evanphx:
Rather than accepting and closing in a loop (which is pretty rude and can easily confuse load balancers), the right thing to do is have the listening socket closed. I thought that was already done though....

aquach:
I'm unfamiliar with socket stuff, but would closing the listening socket cut off existing open connections that Puma is waiting for during the graceful shutdown period? If so, then that would break graceful shutdown.

Also, if I'm reading it right, https://github.com/puma/puma/blob/master/lib/puma/server.rb#L396 is the place where the sockets are closed. It's after the graceful shutdown on line 383.

nateberkopec:
> I'm unfamiliar with socket stuff, but would closing the listening socket cut off existing open connections that Puma is waiting for during the graceful shutdown period?

Not sure, but writing a test might help clarify :)

mainameiz:
+1 for rejecting new requests when shutting down.

schneems:
The concept is sound, i'm not sure about the implementation. I 100 % agree with this:

> Rather than accepting and closing in a loop (which is pretty rude and can easily confuse load balancers), the right thing to do is have the listening socket closed. I thought that was already done though....

At Heroku a thing i've seen before is that when a dyno is scaled down there seems like there might be a slight race condition where the the dyno gets told to shut down before the router gets told to stop sending requests to that dyno. 

In that case I want to avoid even accepting any connections. So rather than accepting and closing individual connections. 

We also need to be wary of other settings such as `drain_on_shutdown`. I think what I would ideally like is to first tell the socket to stop accepting new connections, then drain whatever is left in the queue. Unfortunately, I don't know if that's possible (via the socket API).

I'm fine with making that default behavior, but likely only after a release and some in-the-wild testing. Also assuming we can get this working the way we want.

I see this PR is quite stale.

mainameiz:
> Rather than accepting and closing in a loop (which is pretty rude and can easily confuse load balancers), the right thing to do is have the listening socket closed. I thought that was already done though....

After spending some time to investigate it, it seems to me that [`Binder#close`](https://github.com/puma/puma/blob/f7e1e58dcba1e19e470b6eb0c44c33a0722285f6/lib/puma/binder.rb#L49) is never called... There are two places where `Binder#close` is called - [one](https://github.com/puma/puma/blob/f7e1e58dcba1e19e470b6eb0c44c33a0722285f6/lib/puma/server.rb#L433) and [two](https://github.com/puma/puma/blob/f7e1e58dcba1e19e470b6eb0c44c33a0722285f6/lib/puma/server.rb#L274), but before `#close` is called it checks `@own_binder` variable... and it seems to me that this variable is always `false` because:

1) It is set to `false` in [`Server#inherit_binder`](https://github.com/puma/puma/blob/f7e1e58dcba1e19e470b6eb0c44c33a0722285f6/lib/puma/server.rb#L103)
2) Because all `Server` instances (for workers) are created only in one place, here

https://github.com/puma/puma/blob/f7e1e58dcba1e19e470b6eb0c44c33a0722285f6/lib/puma/runner.rb#L162-L165

and `Server#inherit_binder` always called for all this instances

mainameiz:
> I think what I would ideally like is to first tell the socket to stop accepting new connections, then drain whatever is left in the queue. 

I am not sure that it is possible, but it is 100% possible in reverse order: 1) drain the queue 2) stop accepting new connections.

I suppose that #1 is fast operation and gap between accepting `TERM` signal and stop accepting new requests is small.

mainameiz:
Could you take a look at my fix https://github.com/puma/puma/pull/1685?

mainameiz:
Is there any news?

calvinxiao:
We are running puma on Kubernetes, here is the work we do to avoid this problem.

1. add `readiness` http probe in Kubernetes
2. add `preStop` hook in Kubernetes, commad is `pkill --signal QUIT bundle && sleep 10`
3. add `get :readiness` to return 200 or 500 depending global variable `$app_ready`
4. set up `:QUIT` trap to update global variable `$app_ready`, so http call to `:readiness` will receive 500

In this way, Kubernetes will remove the `not ready` endpoints, stop sending traffic to the PODs that are about to close, one needs to make sure the `sleep` command is long enough for Kubernetes to fail the readiness probe.

you can apply check [health check(https://docs.nginx.com/nginx/admin-guide/load-balancer/http-health-check/) in nginx to avoid this problem.

---

I would like to have `before_server_close` hook [here](https://github.com/puma/puma/blob/v3.12.0/lib/puma/cluster.rb#L276) to handle the readiness variable change, then won't need the `preStop` hook.

mainameiz:
@nateberkopec @evanphx do you have any comments about this issue and [solution](https://github.com/puma/puma/pull/1685)?

evanphx:
Dealt with in #1685

Issue #1199, opened 2017-01-29
Parameters passed to Rack::Handler::Puma.run overwritten by config/puma.rb


twalpole:
Capybara provides support for using Puma as the server when testing an app.  
Puma is started by calling Rack::Handler::Puma.run and passing the :Host and :Port options to bind to.   Upgrading to Puma 3.7.0 is breaking things because the passed in host and port are overwritten by any settings in `config/puma.rb` if that file exists - https://github.com/teamcapybara/capybara/issues/1831.  I would expect the parameters passed to `run` to be the ones used. I think - https://github.com/puma/puma/commit/8509829729648f6b87e4e6c3fa611799e18e8353 - is the commit that changed the behavior.

richpeck:
Confirmed issue with new puma gem. Removing options set in config pass correctly

nateberkopec:
See #1200 for resolution.

Issue #1201, opened 2017-01-30
Puma 3.7.0 halt when used as Capybara server


oslyak:
### Steps to reproduce

1) bundle update
 to install Puma 3.7.0
2) rails test test/integration/

### Expected behavior
With Puma
Version 3.6.2 (ruby 2.3.3-p222), codename: Sleepy Sunday Serenity
Every thing work just fine.
Test just runs with selenium and I see Firefox window.

### Actual behavior
After update Puma to 
Version 3.7.0 (ruby 2.3.3-p222), codename: Snowy Sagebrush
When runs test.
Puma runs, but tests doesn't runs!
After few minutes exit with error:
**RuntimeError: Rack application timed out during boot**

### System configuration
OS: Ubuntu 16.10 x64
**Ruby version**: 2.3.3p222
**Rails version**: 5.0.1

### test/test_helper.rb
```
ENV['RAILS_ENV'] ||= 'test'
require File.expand_path('../../config/environment', __FILE__)
require 'database_cleaner'
require 'rails/test_help'
require 'capybara/rails'
require 'selenium-webdriver'
DatabaseCleaner.strategy = :truncation
DatabaseCleaner.clean
Capybara.javascript_driver = :selenium
Capybara.server = :puma
```

twalpole:
This is the same as issue #1199 and #1200 - There is a workaround in the Capybara issue about it - https://github.com/teamcapybara/capybara/issues/1831

nateberkopec:
See #1200 as resolution.

Issue #1200, opened 2017-01-30
Starting rails server with a different port


khpatel4991:
### Steps to reproduce

1) I have a simple rails 5 API app, and I updated puma to 3.7.0. I start my server on port 3090 with `rails server -p 3090` as I have frontend app running on 3000.

### Expected behavior
It should start listening on port 3090, not on the default 3000.
### System configuration

**Ruby version**: RVM 2.3.3-p222
**Rails version**: 5.0.1

Logs:
```
rs -p 3090
=> Booting Puma
=> Rails 5.0.1 application starting in development on http://localhost:3090
=> Run `rails server -h` for more startup options
Puma starting in single mode...
* Version 3.7.0 (ruby 2.3.3-p222), codename: Snowy Sagebrush
* Min threads: 5, max threads: 5
* Environment: development
* Listening on tcp://0.0.0.0:3000
Exiting
~/.rvm/gems/ruby-2.3.3/gems/puma-3.7.0/lib/puma/binder.rb:269:in `initialize': Address already in use - bind(2) for "0.0.0.0" port 3000 (Errno::EADDRINUSE)
```

twalpole:
This is basically the same as issue #1199

richpeck:
+1

michael-reeves:
Seeing the same thing trying to run port 3001.

ayghor:
see https://github.com/puma/puma/commit/42bec4600c51ab8a1c1ee5a0e1b738a4ffd82bf2, it was tagged `v3.6.2` and was __NOT__ merged on `v3.7.0` for some reason:

>     Revert "fix access priorities of each level in LeveledOptions"
>     
>     This reverts commit 15f0847289d0815284f87eeea4625100b581c026.
>     
>     See #1154, this commit unintentionally changed behavior regarding the
>     importance of command line options.

cant test reverting teh reversion right now

ayghor:
`v3.2.6` respect `-p` switch, `v3.7.0` does not.

andrew:
A similar thing happened back in November: https://github.com/puma/puma/issues/1154

ivanovaleksey:
@ayghor Do you mean `3.6.2`? Since `3.2.6' [doesn't exist](https://rubygems.org/gems/puma/versions).
And with `3.6.2` it works fine.

ayghor:
> @ayghor Do you mean 3.6.2? Since 3.2.6 doesn't exist.
> And with 3.6.2 it works fine.

[...]

PR #1203, opened 2017-01-31
Specified parameters should overwrite loaded config files

twalpole:
Fix issue #1199, #1200, #1201 by allowing specified parameters to overwrite those loaded from config files.

### DIFF
--- a/lib/puma/configuration.rb
+++ b/lib/puma/configuration.rb
@@ -30,6 +30,11 @@ module Puma
       @set << @cur
     end
 
+    def reverse_shift
+      @cur = {}
+      @set.unshift(@cur)
+    end
+
     def [](key)
       @set.reverse_each do |o|
         if o.key? key
@@ -201,10 +206,11 @@ module Puma
       end
 
       files.each do |f|
-        @options.shift
+        @options.reverse_shift
 
         DSL.load @options, self, f
       end
+      @options.shift
     end
 
     # Call once all configuration (included from rackup files)
--- a/test/config/settings.rb
+++ b/test/config/settings.rb
@@ -1,0 +1,2 @@
+port 3000
+threads 3, 5
--- a/test/test_config.rb
+++ b/test/test_config.rb
@@ -67,6 +67,17 @@ class TestConfigFile < Minitest::Test
     assert_equal conf.options[:workers], 4
   end
 
+  def test_parameters_overwrite_files
+    conf = Puma::Configuration.new(config_files: ['test/config/settings.rb']) do |c|
+      c.port 3030
+    end
+    conf.load
+
+    assert_match(/:3030$/, conf.options[:binds].first)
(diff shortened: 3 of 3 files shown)


ayghor:
this was already fixed on `v3.6.2` and not merged on `v3.7.0` for some reason.

ayghor:
see #1200

twalpole:
@ayghor It wasn't fixed on v3.6.2 - a fix for a different issue was reverted on 3.6.2 because while it fixed the previous issue it had created this new issue. So after reverting the previous issue still existed. This PR fixes both issues.

ayghor:
oops sry for that <3

twalpole:
@nateberkopec  Thoughts?

nateberkopec:
Wonderful, thanks @twalpole. Taking a look.

bikramwp:
In Rails 5.0.1, This change looks like it causes the options in `config/puma.rb` to be ignored:


### Scenario 1: config/puma.rb with custom bind option, running `rails s` command

`config/puma.rb` contains:
```
bind ENV.fetch("BIND") { "tcp://127.0.0.1:4444" }
```

In Puma 3.7.0, that generated an options array of:

```
[
  { ... , :Port=>3000, :Host=>"localhost"  },
  { ... , :binds=>["tcp://localhost:3000"] },
  { ... , :binds=>["tcp://127.0.0.1:4444"] }
]
```


In Puma 3.7.1, it generates:
```
[
  { ... , :binds=>["tcp://127.0.0.1:4444"] },
  { ... , :Port=>3000, :Host=>"localhost"  },
  { ... , :binds=>["tcp://localhost:3000"] }
]
```
### Scenario 2: config/puma.rb with custom bind option, running `rails s -p 9999` command

`config/puma.rb:` contains (same as Scenario 1):
```
bind ENV.fetch("BIND") { "tcp://127.0.0.1:4444" }
```

In Puma 3.7.0, that generated an options array of:
```
[
  { ... , :Port=>9999, :Host=>"localhost"  },
  { ... , :binds=>["tcp://localhost:9999"] },
  { ... , :binds=>["tcp://127.0.0.1:4444"] }
]
```


In Puma 3.7.1, it generates:
```
[
  { ... , :binds=>["tcp://127.0.0.1:4444"] },
  { ... , :Port=>9999, :Host=>"localhost"  },
  { ... , :binds=>["tcp://localhost:9999"] }
]
```

So, Rails' config (internal default or specified on `rails s` command) is always taking precedence
and `config/puma.rb` settings don't take effect.

I think this pull request is right in that `rails s` options should take priority over `config/puma.rb`,
but in the case where no `rails s` options are specified on the command line, `config/puma.rb` should
take precedence over Rails' internal defaults.

Unfortunately, I am not sure what the proper solution would be: is Rails configuring Puma incorrectly
or should Puma be working differently?

nateberkopec:
@bikramwp see #1229.

PR #1206, opened 2017-02-02
Bug fix for a race condition on MRI(all versions).

NickolasVashchenko:
I've stumbled into a pretty elusive bug in puma when I was working with it in my project. Most often it happened in the result of Interrupt signal to the process. Shortly, it leads to a possibility for the pipe to be closed here:
https://github.com/puma/puma/blob/482ea5a24abaccf33c49dc9238a22e2a9affe288/lib/puma/server.rb#L369
even before have been finished the write method calls here:
https://github.com/puma/puma/blob/482ea5a24abaccf33c49dc9238a22e2a9affe288/lib/puma/server.rb#L877
or here:
https://github.com/puma/puma/blob/482ea5a24abaccf33c49dc9238a22e2a9affe288/lib/puma/server.rb#L887
After pretty deep investigation I've found out that it becomes possible because of how MRI ruby handles its Global Interpretation lock.
Here what was happening:
1) The `@notify << STOP_COMMAND` is a system call that pauses the thread from where it has been done.
2) This system call get handled fast and successfully, but ruby GIL does not wake the thread yet.
3) After this system call it wakes another thread, the one that with `IO.select` waits for messages in `@check` pipe.
4) It gets it own 100ms chunk to do it's job and closes the pipe.
5) The first thread is woken up right in the middle of `<<` system call and finds out that pipe it used is already closed. Which leads to the bug that looked like this:
```
RuntimeError: can't modify frozen IOError
  /home/vagrant/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/puma-3.6.2/lib/puma/server.rb:877:in `write'
  /home/vagrant/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/puma-3.6.2/lib/puma/server.rb:877:in `<<'
  /home/vagrant/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/puma-3.6.2/lib/puma/server.rb:877:in `stop'
```
As a confirmation for this theory, the time of execution of `<<` method call is always a multiple of 100ms (0, 100, 200, 300, etc) because that is regular GLI time chunk when it changes the current thread: https://github.com/ruby/ruby/blob/trunk/thread_pthread.c#L1252 The more threads - the more likely to encounter it.
So I've made a small fix for that, in the same manner as already existent `rescue IOError` clause do it. The difference only on that IOError is thrown when pipe is already closed, an this case is when it being closed in the middle.
 
Unfortunately, it's very hard to meaningfully reproduce this case in tests, because it's caused by MRI's GIL assignment in presence of several threads(2 not enough). I've run current test suite and nicely tested it manually, but if it's necessary to cover that to make the PR mergeable, I can try to figure it out.

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -875,6 +875,8 @@ module Puma
       rescue IOError
         # The server, in another thread, is shutting down
       rescue RuntimeError => e
+        # The server, in another thread, has been shut down during the system call
+        # https://github.com/puma/puma/pull/1206
         raise e unless e.message.include?('IOError')
       end
     end
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -875,9 +875,10 @@ module Puma
       rescue IOError
         # The server, in another thread, is shutting down
       rescue RuntimeError => e
-        raise e unless e.message.include?('frozen')
+        raise e unless e.message.include?('IOError')
       end
     end
+    private :notify_safely
 
     # Stops the acceptor thread and then causes the worker threads to finish
     # off the request queue before finally exiting.
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -869,35 +869,31 @@ module Puma
       end
     end
 
-    # Stops the acceptor thread and then causes the worker threads to finish
-    # off the request queue before finally exiting.
-    #
-    def stop(sync=false)
+    def notify_safely(message)
       begin
-        @notify << STOP_COMMAND
+        @notify << message
       rescue IOError
         # The server, in another thread, is shutting down
+      rescue RuntimeError => e
+        raise e unless e.message.include?('frozen')
       end
+    end
 
+    # Stops the acceptor thread and then causes the worker threads to finish
+    # off the request queue before finally exiting.
+
+    def stop(sync=false)
+      notify_safely(STOP_COMMAND)
       @thread.join if @thread && sync
     end
 
     def halt(sync=false)
-      begin
-        @notify << HALT_COMMAND
-      rescue IOError
-        # The server, in another thread, is shutting down
-      end
-
+      notify_safely(HALT_COMMAND)
       @thread.join if @thread && sync
     end
 
     def begin_restart
-      begin
-        @notify << RESTART_COMMAND
-      rescue IOError
-        # The server, in another thread, is shutting down
-      end
+      notify_safely(RESTART_COMMAND)
     end
 
     def fast_write(io, str)


lib/puma/server.rb
```diff
@@ -869,35 +869,32 @@ def graceful_shutdown
...
-    #
-    def stop(sync=false)
+    def notify_safely(message)
       begin
-        @notify << STOP_COMMAND
+        @notify << message
       rescue IOError
         # The server, in another thread, is shutting down
+      rescue RuntimeError => e
+        raise e unless e.message.include?('IOError')
```
nateberkopec: Can you leave a comment here explaining this? Otherwise it looks pretty magical.

NickolasVashchenko: Sure.

nateberkopec: Maybe even just include a permalink to this pull, since your explanation is very complete here

NickolasVashchenko:
Done :)

NickolasVashchenko:
I guess tests failed because of some erratic issue, which not related to this PR.

nateberkopec:
We just discussed this. The team definitely thinks there could be a race condition here, but we're not sure we've got the mechanism correctly figured out.

NickolasVashchenko:
Okay, I think I can create some minimal implementation, that would produce an exception of such kind, give me couple of days. Also, if you have any particular questions about mechanism, you're welcome to ask me.

NickolasVashchenko:
tl;dr: this error, which I've reported, happens when `Puma::Server#stop` is called from any rescue clause. And it's ruby's internal bug.

How it happens:
- MRI has a special exception type which is thrown when one thread tries to access a file descriptor that is locked by some other thread: `IOError: stream closed`. 
Definition: https://github.com/ruby/ruby/blob/trunk/thread.c#L4823 
Locking: https://github.com/ruby/ruby/blob/trunk/thread.c#L1423
Scheduling interrupt when locked: https://github.com/ruby/ruby/blob/trunk/thread.c#L2191
- It is exactly the error that the author of the line `rescue IOError` in the `#stop` method was trying to catch.

- The problem is that it's an exception of a "special type" that is getting frozen(and even tainted) right after creation: https://github.com/ruby/ruby/blob/trunk/vm.c#L2078. I suppose that has been done because of security reasons and assumption that code, running in another thread, might have different SAFE level. 
- After that it can not be handled as a regular exception anymore, which ruby tries to do before throwing it. For example, here is the actual offender: https://github.com/ruby/ruby/blob/trunk/eval.c#L511
it's an attempt to assign the `#cause` field that should contain link to another exception caught by current rescue clause. It simply can do it because it's frozen and the first exception gets discarded and substituted with what we saw earlier: `RuntimeError: can't modify frozen IOError`. 
- If it happens outside the rescue clause, it throws the original exception.
- To sum up, it's 100% bug in ruby internals, that went unnoticed for a while because of not very often conditions when it happens. But I happened to have it failing for me in a production app with puma. I'm gonna to report it and maybe even try to submit a fix, but I'm not sure it will be accepted and backported soon, so puma in meanwhile actually may use this improvement.
- To reproduce it: 
1. Pull version of ruby where I've added one commit 5 sec sleep to imitate slow OS response or delay caused by rube scheduler: https://github.com/NickolasVashchenko/ruby/commits/special_exc_cause_bug
2. Compile it and use it to run this script: 
```
rd, wr = IO.pipe
Thread.new do
  IO.select [rd]
  wr.close
end

begin
  raise 'any exception'
rescue
  wr << 'A' 
end 
```

respire:
hello @nateberkopec @NickolasVashchenko 
i try to add mutex to guard the pipe, and after struggling fight (awww), it works now.
https://travis-ci.org/respire/puma/builds/242696832
i even run this build several times. looks like it's stable enough. 
the code fixed this issue is in my master branch. it contains some commits.
noted that this is not the final fix. i still need more time to analyze why it's working now.
if you have some ideas, please reply!  thanks! :)

jjb:
When this bug occurs does it leave puma in a bad state or does it only affect one http request?

NickolasVashchenko:
It leaves the thread in the bad state. So if it's somehow ended or killed, puma can keep up working.
However, I've got a simpler and more direct solution for the workaround and gonna submit it a day or two.

NickolasVashchenko:
Damn, rebasing the branch closed the the pull request. Is it possible to reopen it?

schneems:
Github won't let me do it. Says you closed the issue.

NickolasVashchenko:
Okay, thanks, I'll open a new on then.

schneems:
Doesn't the UI allow you reopen this PR?

schneems:
I am a maintainer

nateberkopec:
screen shot 2017-06-27 at 10 24 44 am

NickolasVashchenko:
Okay, I'll create a new one then.

Issue #1154, opened 2016-11-22
puma overides Port option from rack handler with default value from puma.config


chulkilee:
### Steps to reproduce

```sh
rails new --api --skip-listen --skip-active-record --skip-action-mailer foo1
rails new --api --skip-listen --skip-active-record --skip-action-mailer foo2

# in foo1
bin/rails s -p 3000

# in foo2
bin/rails s -p 3001
```

### Expected behavior

- foo1 runs on 3000
- foo2 runs on 3001

### Actual behavior

- foo1 runs on 3000 (correct)
- foo2 tried to run on 3000, and failed

```
=> Booting Puma
=> Rails 5.0.0.1 application starting in development on http://localhost:3001
=> Run `rails server -h` for more startup options
Puma starting in single mode...
* Version 3.6.1 (ruby 2.3.2-p217), codename: Sleepy Sunday Serenity
* Min threads: 5, max threads: 5
* Environment: development
* Listening on tcp://0.0.0.0:3000
Exiting
/myhome/.rbenv/versions/2.3.2/lib/ruby/gems/2.3.0/gems/puma-3.6.1/lib/puma/binder.rb:266:in `initialize': Address already in use - bind(2) for "0.0.0.0" port 3000 (Errno::EADDRINUSE)
	from /myhome/.rbenv/versions/2.3.2/lib/ruby/gems/2.3.0/gems/puma-3.6.1/lib/puma/binder.rb:266:in `new'
	from /myhome/.rbenv/versions/2.3.2/lib/ruby/gems/2.3.0/gems/puma-3.6.1/lib/puma/binder.rb:266:in `add_tcp_listener'
	from /myhome/.rbenv/versions/2.3.2/lib/ruby/gems/2.3.0/gems/puma-3.6.1/lib/puma/binder.rb:102:in `block in parse'
	from /myhome/.rbenv/versions/2.3.2/lib/ruby/gems/2.3.0/gems/puma-3.6.1/lib/puma/binder.rb:85:in `each'
	from /myhome/.rbenv/versions/2.3.2/lib/ruby/gems/2.3.0/gems/puma-3.6.1/lib/puma/binder.rb:85:in `parse'
	from /myhome/.rbenv/versions/2.3.2/lib/ruby/gems/2.3.0/gems/puma-3.6.1/lib/puma/runner.rb:133:in `load_and_bind'
[...]

PR #1215, opened 2017-02-17
Revert "fix access priorities of each level in LeveledOptions"

robertomiranda:
This reverts commit 15f0847.
See #1154, this commit unintentionally changed behavior regarding the
importance of command line options.

Similar to 6d1c157, Seems that 3.7.0 missed this revert

cc @nateberkopec

### DIFF
--- a/lib/puma/configuration.rb
+++ b/lib/puma/configuration.rb
@@ -36,7 +36,7 @@ module Puma
     end
 
     def [](key)
-      @set.reverse_each do |o|
+      @set.each do |o|
         if o.key? key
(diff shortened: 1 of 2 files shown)


evanphx:
@robertomiranda That's correct, it did cause #1154, but reverting it is not the correct fix. @schneems and I discussed the right way to handle it.

nateberkopec:
I think we're blocking on @schneems to implement that fix since he seems to understand the problem/solution the best, but IIRC it involved changes in Rails/Rack.

PR #1219, opened 2017-02-20
Prevent accumulation of clients being polled forever.

aasmith:
When a file descriptor for a client is returned from select, it is removed from the clients currently being monitored for a timeout. This removal happens independently from the code that removes the descriptor from the list of descriptors that are currently being polled, allowing a potential accumulation of descriptors being passed into by select.

### DIFF
--- a/lib/puma/reactor.rb
+++ b/lib/puma/reactor.rb
@@ -60,19 +60,14 @@ module Puma
                 end
               end
             else
-              # We have to be sure to remove it from the timeout
-              # list or we'll accidentally close the socket when
-              # it's in use!
-              if c.timeout_at
-                @mutex.synchronize do
-                  @timeouts.delete c
-                end
-              end
-
               begin
                 if c.try_to_finish
                   @app_pool << c
                   sockets.delete c
+
+                  @mutex.synchronize do
+                    @timeouts.delete c
+                  end
                 end
 
               # Don't report these to the lowlevel_error handler, otherwise
--- a/test/test_puma_server.rb
+++ b/test/test_puma_server.rb
@@ -296,8 +296,6 @@ class TestPumaServer < Minitest::Test
   end
 
   def test_timeout_in_data_phase
-    skip("Hangs too often, TODO: fix")
-
     @server.first_data_timeout = 2
     @server.add_tcp_listener @host, @port
     @server.run


aasmith:
The easiest way to reproduce this problem on demand is to make a simple clone of the test against an app with a reduced timeout (note the change in timeout is not needed, but helps prove the issue quicker).

Assuming a local instance of puma, running on port 80:

#### app.ru:
```ruby
Puma::Const::FIRST_DATA_TIMEOUT = 0.2

app = proc do |env|
  [ 200, {'Content-Type' => 'text/plain'}, ["a"] ]
end

run app
```

#### break.sh
```bash
while true
do
  curl -X POST -H "Host: example.com" -H "Content-Type: text/plain" -H "User-Agent:" -H "Accept:" -H "Content-Length: 5" 0:80
done
```

Start puma with `puma -w 1 -t 1:1 -e deployment -p 80 -v app.ru`, and then run the shell script. the shell script will pause forever when it hits the bug.

aasmith:
I'm not sure if my suggestion is the correct fix, but I'd like to hear from others who can repro this and what the best solution might be.

aasmith:
Also see the discussion on commit 43ee6ddca for more context.

nateberkopec:
Hey @aasmith - I'm trying your repro on latest master and it isn't working, even with `Puma::Const::FIRST_DATA_TIMEOUT = 0.1`. Can you take another look? Maybe we fixed this somewhere else.

aasmith:
I just repro'd this using the 3.9.1 gem, using the `ruby` docker image.

I have a small adjustment to the steps to repro this in docker (in case it's platform-specific?)

```console
$ docker run -it ruby bash

# gem install puma

<add app.ru as defined above>

# puma -w 1 -t 1:1 -e deployment -p 80 -d app.ru
# while true; do curl -v -X POST -H "Host: example.com" -H "Content-Type: text/plain" -H "User-Agent:" -H "Accept:" -H "Content-Length: 5" 0:80; done

<observe output will freeze>
```

After applying the patch, the problem is no longer observed.

Hope this helps!

nateberkopec:
Thanks!

stereobooster:
please rebase, tests were fixed in master

aasmith:
I was able to reproduce the problem using 3.10.0. However, I am no longer able to reproduce this with puma 3.11.0; it correctly gives a `408 Request Timeout` for each request.

Test system:

```console
root@af8662fe8d9c:/# puma --version
puma version 3.11.0
root@af8662fe8d9c:/# uname -a
Linux af8662fe8d9c 4.9.31-moby #1 SMP Sat Jun 24 06:29:23 UTC 2017 x86_64 GNU/Linux
root@af8662fe8d9c:/# ruby -v
ruby 2.5.0p0 (2017-12-25 revision 61468) [x86_64-linux]
```

nateberkopec:
Closing for now, but definitely willing to reopen if it can be reproduced on latest Puma.

PR #1223, opened 2017-02-20
Log parse_error's only in debug mode

sasso:
When doing AWS or Kubernetes health-checks in TCP mode this can produce a high number of log lines. Suggestions welcome if there is a better solution to silence these!

### DIFF
--- a/lib/puma/events.rb
+++ b/lib/puma/events.rb
@@ -91,8 +91,8 @@ module Puma
     # parsing exception.
     #
     def parse_error(server, env, error)
-      @stderr.puts "#{Time.now}: HTTP parse error, malformed request (#{env[HTTP_X_FORWARDED_FOR] || env[REMOTE_ADDR]}): #{error.inspect}"
-      @stderr.puts "#{Time.now}: ENV: #{env.inspect}\n---\n"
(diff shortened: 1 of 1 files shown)


schneems:
I think I want to know when an HTTP request is malformed. 

Do you know why health checks in TCP mode with kubernetes result in malformed requests? Can we fix it there instead?

sasso:
@schneems there is no way to fix this on AWS or Kubernetes (TCP checks are valid) and the only way is to silence it. The amount of health checks on big clusters overflows the logs and makes them unreadable.

nateberkopec:
I'm somewhat partial to reducing the noise here. I don't think malformed requests are errors which require the developer's attention most of the time.

evanphx:
The better way is to detect if HTTP parsing began at all, and if not, then silently return. If you're using a TCP check, then there should be no data at all.

nateberkopec:
@evanphx Can you expand on "detect if HTTP parsing began at all,"

nateberkopec:
I think I understand now. @sasso we can merge if, instead of always logging to debug, we just don't call parse_error when receiving a TCP check.

A test will be essential here as well.

PR #1225, opened 2017-02-23
Avoid waiting on idle connections when we have only 1 worker thread

davidor:
When serving non multithread safe applications, a single worker thread is used, possibly with quite a few worker processes.

When a keep-alive request is dispatched, Puma checks for a new request coming in. This check makes any one worker thread stall if no data is available, up to `FAST_TRACK_KA_TIMEOUT` seconds (set to 200ms currently).

This has the effect of neglecting other requests that are ready, waiting to be picked up for processing.

This patch fixes the case for max_threads == 1, where waiting 200ms can greatly affect performance. It probably affects it too with more threads, but further benchmarks are needed to consider when it pays to wait up to 200ms.

Reproducing this is easy when running Puma with 1 thread. I used a fast application that responds in a few ms. Even when the 99 perc of the response times was like 5ms, there were peaks of several seconds.

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -238,7 +238,7 @@ module Puma
       end
 
       queue_requests = @queue_requests
-      @fast_track_ka_timeout = fast_track_ka_timeout
+      ka_timeout = fast_track_ka_timeout
 
       @thread_pool = ThreadPool.new(@min_threads,
                                     @max_threads,
@@ -273,7 +273,7 @@ module Puma
           client.close
         else
           if process_now
-            process_client client, buffer
+            process_client client, buffer, ka_timeout
           else
             client.set_timeout @first_data_timeout
             @reactor.add client
@@ -402,7 +402,7 @@ module Puma
     # indicates that it supports keep alive, wait for another request before
     # returning.
     #
-    def process_client(client, buffer)
+    def process_client(client, buffer, ka_timeout)
       begin
 
         if client.env[HTTP_EXPECT] == CONTINUE
@@ -425,7 +425,7 @@ module Puma
 
             ThreadPool.clean_thread_locals if clean_thread_locals
 
-            unless client.reset(@status == :run && @fast_track_ka_timeout)
+            unless client.reset(@status == :run && ka_timeout)
               close_socket = false
               client.set_timeout @persistent_timeout
               @reactor.add client
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -238,6 +238,7 @@ module Puma
       end
 
       queue_requests = @queue_requests
+      @fast_track_ka_timeout = fast_track_ka_timeout
 
       @thread_pool = ThreadPool.new(@min_threads,
                                     @max_threads,
@@ -424,7 +425,7 @@ module Puma
 
             ThreadPool.clean_thread_locals if clean_thread_locals
 
-            unless client.reset(@status == :run)
+            unless client.reset(@status == :run && @fast_track_ka_timeout)
               close_socket = false
               client.set_timeout @persistent_timeout
               @reactor.add client
@@ -477,6 +478,16 @@ module Puma
         end
       end
     end
+
+    # Compute maximum time to wait for another request arriving on
+    # a client when keep-alive is enabled.
+    # If the pool we created has a maximum of one worker thread we
+    # can't afford to let it wait for the next request while other
+    # work might be pending.
+    def fast_track_ka_timeout
+      @max_threads > 1 ? FAST_TRACK_KA_TIMEOUT : 0.0
+    end
+    private :fast_track_ka_timeout
 
     # Given a Hash +env+ for the request read from +client+, add
     # and fixup keys to comply with Rack's env guidelines.
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -81,7 +81,7 @@ module Puma
       @timeout_at = Time.now + val
     end
 
-    def reset(fast_check=true)
+    def reset(fast_track_ka_timeout)
       @parser.reset
       @read_header = true
       @env = @proto_env.dup
@@ -101,8 +101,8 @@ module Puma
         end
 
         return false
-      elsif fast_check &&
-            IO.select([@to_io], nil, nil, FAST_TRACK_KA_TIMEOUT)
+      elsif fast_track_ka_timeout &&
+            IO.select([@to_io], nil, nil, fast_track_ka_timeout)
         return try_to_finish
       end
     end


unleashed:
Just a heads up: this is a bug that causes high spikes in latency if you use Puma in non multithreaded applications.

davidor:
Thanks for fixing the bug that I introduced in my second commit @unleashed

davidor:
I ran a quick test to show you the problem I’m trying to solve in this PR.

I set up a very basic web app served by Puma in clustered mode with 4 workers and 1 thread for each of them, to simulate a non thread-safe application. From a different machine, I sent around 1000 requests per second during 5 minutes.

Here are the charts with the response times (notice the non-linear scale on the Y axis).

Before the patch:
before


After the patch:
after


As you can see, the mean response time is what we would expect from a very basic application that does not much, a few ms. However, in the first chart, there are some requests that are really slow (> 1 second). I am aware that this test is very simple and we cannot draw too many conclusions from it, but I think it clearly shows the problem being addressed in the PR.

lib/puma/server.rb
```diff
@@ -478,6 +479,16 @@ def process_client(client, buffer)
       end
     end
 
+    # Compute maximum time to wait for another request arriving on
+    # a client when keep-alive is enabled.
+    # If the pool we created has a maximum of one worker thread we
+    # can't afford to let it wait for the next request while other
+    # work might be pending.
+    def fast_track_ka_timeout
+      Integer(@max_threads) > 1 ? FAST_TRACK_KA_TIMEOUT : 0.0
```
nateberkopec: Why wouldn't max_threads already be an integer?

lib/puma/server.rb
```diff
@@ -424,7 +425,7 @@ def process_client(client, buffer)
 
             ThreadPool.clean_thread_locals if clean_thread_locals
 
-            unless client.reset(@status == :run)
+            unless client.reset(@status == :run && @fast_track_ka_timeout)
```
nateberkopec: What's the purpose of an instance variable? Why not just call the accessor?

davidor: I added  `@fast_track_ka_timeout = fast_track_ka_timeout` inside the `run` method, but it's a mistake. I should have defined it in `initialize`. There isn't an accessor defined, but we could add it.

nateberkopec: I would say this: https://github.com/puma/puma/pull/1225/files/b373d26a37eaac35ed943c72b183b11f99406104#diff-ef4c610f0ce605eb90f2810856394fd5R487 is an accessor. I'm asking what the purpose of the instance variable is when you could just call `fast_track_ka_timeout`.

davidor: Avoid calculating the same thing multiple times, because `@max_threads` is static.

lib/puma/server.rb
davidor: You're right. In a previous Puma version that I tried, `@max_threads` was a String at that point, but it's no longer the case.

lib/puma/server.rb
nateberkopec: I'm loathe to cache anything unless we actually have a benchmark proving it is a significant speed difference. In this case, making a comparison against an integer.

I'm not even sure max_threads is always static in all cases.

davidor: OK. I can change this to just call `fast_track_ka_timeout`. The method would be called on every request, so I thought that it would be good to cache the result. However, I'm not 100% sure either that `@max_threads` is always static.

nateberkopec: Paging @evanphx for a quick gut check on my logic above.

evanphx: I don't mind pulling it out into a variable because that value will never change for the lifetime of the process. Shouldn't be an instance variable though, just use a local.

davidor:
@nateberkopec Thanks for your feedback. I addressed your two comments. I deleted the unnecessary conversion to Integer from my second commit. I also added a new commit to use a local var instead of an instance var to store the `fast_track_ka_timeout`. If you think this is OK, I can squash the last two commits to have a clean history.

nateberkopec:
I'm not sure setting this to 0.0 when max_threads=1 is a good idea. Couldn't slow clients basically get perma-locked-out of a worker if other, faster traffic is constantly ahead of them?

I'm OK with a faster timeout, but 0 seems like it could lead to a lot of edge cases. @evanphx

unleashed:
At this phase clients are potentially sending another request on keepalive. The value of this timeout makes Puma (with threads=1) sit idle waiting on this potential next request from this specific client leaving everything else unattended. That is currently 200ms of idling per worker process. The reset call on the client object will return falsey and further IO will be processed later on just like any other IO. This code is there just to handle the fast keepalive case in which data is ready to be processed. This is arguably an optimization, but not a good one when you are limited in threads. In fact, one single client per worker process can DoS Puma without multithreading by sending keepalive requests with the right timing: right under 200ms for the next request is enough to keep it busy with the same client all the time.

lib/puma/server.rb
```diff
@@ -424,7 +425,7 @@ def process_client(client, buffer)
 
             ThreadPool.clean_thread_locals if clean_thread_locals
 
-            unless client.reset(@status == :run)
+            unless client.reset(@status == :run && ka_timeout)
```
nateberkopec: Aren't we missing a comparison operator here?

nateberkopec:
@davidor Can you share your test setup from [here](https://github.com/puma/puma/pull/1225#issuecomment-283377542) so I can repro?

stereobooster:
I had something similar (spikes in latency), but in multi-threaded application. I deployed server with 8 workers instead of 2 (error in ansible recipe), my intention was to use number of workers equal to number of processors. And after this I started to notice 20-30 requests with timeouts (after 30s). But later when I noticed error in deployment script and lowered number of workers problem went away.

unleashed:
> At this phase clients are potentially sending another request on keepalive. The value of this timeout makes Puma (with threads=1) sit idle waiting on this potential next request from this specific client leaving everything else unattended. That is currently 200ms of idling per worker process. The reset call on the client object will return falsey and further IO will be processed later on just like any other IO. This code is there just to handle the fast keepalive case in which data is ready to be processed. This is arguably an optimization, but not a good one when you are limited in threads. In fact, one single client per worker process can DoS Puma without multithreading by sending keepalive requests with the right timing: right under 200ms for the next request is enough to keep it busy with the same client all the time.

Note that [CVE-2019-16770](https://github.com/puma/puma/commit/6baa4d8e1c88f2e4db2918df48416a5c49feec40) is exactly what I reported in the above comment 2.5 years ago. :(

PR #1234, opened 2017-03-06
Implement user_supplied_options behavior.

schneems:
This is the Puma side of rails/rails#28137 

When options are passed to the Puma rack handler it is unknown if the options were set via a framework as a default or via a user. Puma currently has 3 different sources of configuration, the user via command line, the config files, and defaults.

Rails 5.1+ will record the values actually specified by the user versus the values specified by the frameworks. It passes these values to the Rack handler and now it's up to Puma to do something with that information.

When only framework defaults are passed it will set

```
options[:user_supplied_options] = []
```

When one or more options are specified by the user such as `:Port` then those keys will be in the array. In that example it will look like this

```
options[:user_supplied_options] = [:Port]
```

This change is 100% backwards compatible. If the framework is older and does not pass this information then the `user_supplied_options` will not be set, in that case we assume all values are user supplied.

Internally we accomplish this separation by replacing `LeveledOptions` which was a generic way of specifying options with different priorities with a more explicit `UserFileDefaultOptions` this assumes only 3 levels of options and it will use them in the order supplied (user config wins over file based config wins over defaults).

Now instead of using 1 dsl to set all values, we use 3. A user dsl, a file dsl and a `Configuration.new` will return all 3 DSLs to the block. It's up to the person using the block to use the correct dsl corresponding to the source of data they are getting.

### DIFF
--- a/lib/puma/configuration.rb
+++ b/lib/puma/configuration.rb
@@ -151,11 +151,24 @@ module Puma
       user    = user_options[key]
       file    = file_options[key]
       default = default_options[key]
+
       user    = [user]    unless user.is_a?(Array)
       file    = [file]    unless file.is_a?(Array)
       default = [default] unless default.is_a?(Array)
 
+      user.compact!
+      file.compact!
+      default.compact!
+
       user + file + default
+    end
+
+    def force_defaults
+      @default_options.each do |k,v|
+        if v.respond_to? :call
+          @default_options[k] = v.call
+        end
+      end
     end
   end
 
@@ -189,9 +202,9 @@ module Puma
     end
 
     def initialize_copy(other)
-      @conf = nil
+      @conf        = nil
       @cli_options = nil
-      @options = @options.dup
+      @options     = @options.dup
     end
 
     def flatten
@@ -218,7 +231,7 @@ module Puma
         :worker_shutdown_timeout => DefaultWorkerShutdownTimeout,
         :remote_address => :socket,
         :tag => method(:infer_tag),
-        :environment => ENV['RACK_ENV'] || "development",
+        :environment => ->{ ENV['RACK_ENV'] || "development" },
         :rackup => DefaultRackup,
         :logger => STDOUT,
         :persistent_timeout => Const::PERSISTENT_TIMEOUT
@@ -247,7 +260,6 @@ module Puma
     # Call once all configuration (included from rackup files)
     # is loaded to flesh out any defaults
     def clamp
-      @options.shift
       @options.force_defaults
     end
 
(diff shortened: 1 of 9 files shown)


test/test_helper.rb
```diff
@@ -4,9 +4,11 @@
 begin
   require "bundler/setup"
 rescue LoadError
```
nateberkopec: Aha, yeah, this is probably the correct fix.

nateberkopec:
@schneems is this done now?

schneems:
Yes, please review. Ping me with any questions.

lib/puma/configuration.rb
```diff
@@ -151,14 +143,17 @@ def initialize(options={}, &blk)
     attr_reader :options, :plugins
 
     def configure(&blk)
-      @options.shift
-      DSL.new(@options, self)._run(&blk)
+      blk.call(@user_dsl, @file_dsl, @default_dsl)
```
nateberkopec: Can't we just `yield`?

test/test_rack_handler.rb
```diff
@@ -53,5 +53,114 @@ def test_handler_boots
...
+  end
+end
 
+class TestUserSuppliedOptionsIsNotPresent < Minitest::Test
+  def setup
+    @options = {}
+  end
+
+  def test_default_port_when_no_config_file
+    options = {}
```
nateberkopec: Just use @options?

test/test_rack_handler.rb
```diff
@@ -53,5 +53,114 @@ def test_handler_boots
...
+    conf = Rack::Handler::Puma.config(->{}, options)
+    conf.load
+
+    assert_equal ["tcp://0.0.0.0:9292"], conf.options[:binds]
+  end
+
+  def test_config_wins_over_default
+    file_port = 6001
+    @options = {}
```
nateberkopec: Already set

test/test_rack_handler.rb
```diff
@@ -53,5 +53,114 @@ def test_handler_boots
...
+    conf = Rack::Handler::Puma.config(->{}, @options)
+    conf.load
+
+    assert_equal ["tcp://0.0.0.0:#{user_port}"], conf.options[:binds]
+  end
+
+  def test_user_port_wins_over_config
+    user_port = 5001
+    file_port = 6001
+    @options = {}
```
nateberkopec: samesies

test/test_rack_handler.rb
schneems: That was a copy pasta error. Will fix

lib/puma/configuration.rb
schneems: Sure if you prefer that.

nateberkopec: no biggie, I just have a habit of using yield

allaire:
@schneems Hi Richard, I *think* this commit broke the correct loading of puma environment configuration file.

Before 3.8.0, I was able to run `RACK_ENV=production bundle exec puma start` and it was loading automatically `config/puma/production.rb`.

On 3.8.0, it's picking up the `production.rb` config file.

Might be related? https://github.com/puma/puma/pull/1269

schneems:
I'm not sure. Can you try with 3.9.0 and if it's still broken open a new issue with an example app?

allaire:
Yes 3.9.0 is still broken. I can open a new issue for sure !

--
Mathieu Allaire

PR #1235, opened 2017-03-07
Allow user supplied procline tags to be dynamic

tannerburson:
The current behavior of the `tag` config option allows the default value to be a callable (defaults to part of the master's working path), but doesn't allow user supplied tags to be dynamic. This patch allows tags to be a callable, making both default and user supplied behaviors the same.

The specific use case here is that I'd like to include the current SHA in the procline, specifically the procline of worker processes. This is helpful when monitoring phased restarts to tell which processes have rolled, which haven't, and what point in time they're from.

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -229,7 +229,12 @@ module Puma
 
     def worker(index, master)
       title = "puma: cluster worker #{index}: #{master}"
-      title << " [#{@options[:tag]}]" if @options[:tag] && !@options[:tag].empty?
+      if @options[:tag] && @options[:tag].respond_to?(:call)
+        title << " [#{@options[:tag].call}]"
+      elsif @options[:tag] && !@options[:tag].empty?
+        title << " [#{@options[:tag]}]"
+      end
+
       $0 = title
 
       Signal.trap "SIGINT", "IGNORE"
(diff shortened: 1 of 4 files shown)


schneems:
Default options are not allowed to be callable during runtime. They are finalized and turned into static values when `clamp` is called on the Configuration object.

`respond_to` isn't the fastest thing and calling procs has overhead. I'm leery to introduce slow-ish things that won't be widely used.

I don't use phased restarts, but based on this issue https://github.com/puma/puma/issues/770 config changes should be picked up between restarts? Is that not the case?

Shooting from the hip here, could you call the `tag` method from within the `before_fork` block?

tannerburson:
I have not been able to see the tag change during phased restarts, only when restarting the parent.

Totally agree that the performance of this is probably worse, but given that it only occurs during worker startup, it shouldn't be a measurable impact on actual web server performance.

All that said, it does look like re-setting it in `before_fork` works. It behaves similarly to setting it as a proc, in that the hook will be re-evaluated before each worker spins up. I'm honestly surprised it works, but definitely won't complain. 

I'll go ahead and close, as it doesn't sound like this is a common use case. I'll open a new issue if I'm able to reproduce and identify the default value for `tag` not being re-resolved for each worker though.

Thanks.

PR #1239, opened 2017-03-10
[close #1238] Don't call non-existent method

### LINKED ISSUE #1238: [3.8.0] undefined method `from_file' for Puma::Configuration:Class
### Steps to reproduce

1) bundle exec pumactl -S tmp/pids/puma.state -F config/puma.rb phased-restart

### Expected behavior

phased-restart  for server

### Actual behavior

undefined method `from_file' for Puma::Configuration:Class

### System configuration

**Ruby version**: 2.4.0
**Rails version**: 5.1.0.beta1

puma.rb
```ruby
# frozen_string_literal: true
#
threads_count = ENV.fetch('RAILS_MAX_THREADS') { 5 }

#
environment ENV.fetch('RAILS_ENV') { 'development' }

if ENV['RAILS_ENV'] == 'production'
  # Workers are forked webserver processes. If using threads and workers together
  # processes).
  #
  workers ENV.fetch('WEB_CONCURRENCY') { 2 }

  # (off by default)
  #
  prune_bundler

  # Bind unix domain socket
  bind ENV.fetch('PUMA_SOCK') { 'unix://tmp/sockets/puma.sock' }

  # Use `path` as the file to store the server info state. This is
  # used by `pumactl` to query and control the server.
  #
  state_path ENV.fetch('SERVER_STATE') { 'tmp/pids/puma.state' }

  # === Puma control rack application ===

  # Start the puma control rack application on `url`. This application can
  # be communicated with to control the main server. Additionally, you can
  # provide an authentication token, so all requests to the control server
  # will need to include that token as a query parameter. This allows for
  # simple authentication.
  #
  # Check out https://github.com/puma/puma/blob/master/lib/puma/app/status.rb
  # to see what the app has available.
  #
  activate_control_app ENV.fetch('PUMACTL_SOCK') { 'unix://tmp/sockets/pumactl.sock' }

  # Store the pid of the server in the file at `path`.
  #
  pidfile 'tmp/pids/puma.pid'

  # Change the default timeout of workers
  #
  worker_timeout 10
else
  #
  port ENV.fetch('PORT') { 3000 }

  # process behavior so workers use less memory. If you use this option
  # you need to make sure to reconnect any threads in the `on_worker_boot`
  # block.
  #
  # preload_app!
end

plugin :tmp_restart
```

schneems:
`Puma::Configuration._from_file` was removed in https://github.com/puma/puma/commit/89f37432deb499ce6cb72a7e7cba5a55430efacd. It was still being used by the `Puma::ControlCLI` class which was previously not being tested while being called with a `--config-file` option.

This PR reintroduces the previous behavior of loading a config file when specified via `pumactl` command.

cc/ @nateberkopec

### DIFF
--- a/lib/puma/control_cli.rb
+++ b/lib/puma/control_cli.rb
@@ -97,6 +97,7 @@ module Puma
 
     rescue => e
       @stdout.puts e.message
+      @stdout.puts e.backtrace
       exit 1
     end
 
@@ -230,6 +231,7 @@ module Puma
 
     rescue => e
       message e.message
+      message e.backtrace
       exit 1
     end
 
--- a/lib/puma/control_cli.rb
+++ b/lib/puma/control_cli.rb
@@ -78,11 +78,12 @@ module Puma
         end
 
         if @config_file
-          config = Puma::Configuration.from_file @config_file
-          @state ||= config.options[:state]
-          @control_url ||= config.options[:control_url]
+          config = Puma::Configuration.new({ config_files: [@config_file] }, {})
+          config.load
+          @state              ||= config.options[:state]
+          @control_url        ||= config.options[:control_url]
           @control_auth_token ||= config.options[:control_auth_token]
-          @pidfile ||= config.options[:pidfile]
+          @pidfile            ||= config.options[:pidfile]
         end
       end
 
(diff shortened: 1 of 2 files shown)


nateberkopec:
LGTM 🚢

PR #1252, opened 2017-03-22
Added case when stopping launcher not works

adis-io:
Run this test:
`TESTOPTS="--name test_handler_stops" bundle exec rake test`

P.S. Test fails because it's not implementation, but only test.

### DIFF
--- a/test/test_rack_handler.rb
+++ b/test/test_rack_handler.rb
@@ -161,3 +161,56 @@ class TestUserSuppliedOptionsIsNotPresent < Minitest::Test
   end
 end
 
+class TestServerStops < Minitest::Test
+  def app
+    proc { [200, {}, ['hello world']] }
+  end
+
+  def in_handler(app, options = {})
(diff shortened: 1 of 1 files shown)


adisos:
@schneems 

Hi, could you please look at this?

schneems:
Is this testing existing functionality, or a filing test for a bug? Master is failing on CI.

adisos:
@schneems 
Filling test for a bug.

schneems:
Thanks, master is failing so a failing CI on a PR isn't unusual right now.

nateberkopec:
The port_open method is incorrect. Replace with the following and everything passes:

screen shot 2017-06-04 at 8 14 39 am

nateberkopec:
[The problem is that binder.close is never called](https://github.com/puma/puma/blob/master/lib/puma/server.rb#L240).

evanphx:
Unsure what this is actually testing...

Issue #1258, opened 2017-03-31
Reset environment to initial state when restarting with USR1 or USR2


grosser:
### Steps to reproduce

```
# config/application.rb
puts "RAND #{ENV['TEST'] ||= rand(1212121212).to_s}"
```

```
kill -SIGUSR1 puma-pid
```

### Expected behavior

A new number is shown on every restart, showing that the env is reset

### Actual behavior

The same number is printed, showing that the env is not cleared.
This makes loading environment vars from a .env file fail since it won't update the previously set variables.

### System configuration

**Ruby version**: 2.3.1
**Rails version**: 5.0.1

### Monkey-patch

from https://github.com/zendesk/samson/pull/1862
```
added = ENV["SAMSON_ENV_ADDED_VIA_DOTENV"].to_s.split(',')
added.each { |e| ENV.delete(e) }

require 'dotenv'
before = ENV.keys
Dotenv.load(Bundler.root.join('.env'))
ENV["SAMSON_ENV_ADDED_VIA_DOTENV"] = (ENV.keys - before).join(",")
```

grosser:
see https://github.com/puma/puma/issues/1258

nateberkopec:
Does this happen with a USR2 restart as well?

grosser:
yes, update the title

PR #1260, opened 2017-04-01
restore environment to original state before restarting

grosser:
- prevents apps from forever poisoning the ENV by doing bad things (something overrides PATH/HOME etc)
 - allows use of Dotenv to load new ENV settings from disk (they do not override)

Test:

```
puts "THIS SHOULD CHANGE #{ENV["TEST"] ||= rand.to_s}"
```

then restart the server

fixes https://github.com/puma/puma/issues/1258

### DIFF
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -163,6 +163,8 @@ module Puma
 
     # Run the server. This blocks until the server is stopped
     def run
+      env = ENV.to_h
+
       @config.clamp
 
       @config.plugins.fire_starts self
@@ -178,6 +180,7 @@ module Puma
         graceful_stop
       when :restart
         log "* Restarting..."
+        ENV.replace(env)
         @runner.before_restart
         restart!
       when :exit
--- a/test/hello-env.ru
+++ b/test/hello-env.ru
@@ -1,0 +1,2 @@
+ENV["RAND"] ||= rand.to_s
+run lambda { |env| [200, {"Content-Type" => "text/plain"}, ["Hello RAND #{ENV["RAND"]}"]] }
--- a/test/test_integration.rb
+++ b/test/test_integration.rb
@@ -56,11 +56,7 @@ class TestIntegration < Minitest::Test
 
     @server = IO.popen("sh #{tf.path}", "r")
 
-    while (@server.gets) !~ /Ctrl-C/
-      # nothing
-    end
-
-    sleep 1
+    wait_for_server_to_boot
 
     @server
   end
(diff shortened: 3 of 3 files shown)


grosser:
@evanphx does this look sane to you ?

grosser:
@schneems maybe ?

nateberkopec:
This feels like the sort of change where you think "this couldn't possibly break anyone's apps!" and then it totally does.

grosser:
meh ... only 1 way to find out 😈

nateberkopec:
I know that's really vague (and it is strange that someone might expect the ENV to not be reverted on a USR signal restart), but it just seems like this could mess up apps.

grosser:
the default behavior should be a sane one ... everything else can be an option ...

schneems:
Interesting. I don't use the restart feature due to Heroku. So I have little operational opinions on it. This seems like reasonable behavior on the surface. You're telling the server to go away and a new one to come up. My only concerns oils be about why hooks have people used to load or set env vars. If it's something that wouldn't get called after a restart then this could break.

nateberkopec:
I definitely agree that this is reasonable behavior, but I just worry about people relying on the unreasonable behavior to set up their ENV.

schneems:
maybe we could do a diff of the environment and output it as a warning/debug-info?

grosser:
"hooks have people used to load or set env vars"
I think this does not matter since a restart means we do a new exec so
everything will be loaded anew.

On Fri, Apr 7, 2017 at 11:26 AM, Richard Schneeman <notifications@github.com

grosser:
I'll add a warn on the diff, should not be much ...

On Fri, Apr 7, 2017 at 11:41 AM, Richard Schneeman <notifications@github.com

grosser:
added ... is that what you had in mind ?

```
>> env['XXX'] = '222'
=> "222"
>> restore_environment env
Resoring process environment, updating:
XXX from nil to "222"
```

grosser:
on the other hand I'd rather avoid this overhead/noise since no app does that when it restarts ... pumas restart is already very clean, it only has this bug because it uses exec ... if it was a spawner then this would not happen ... and I think it's reasonable to not expect it to happen ...

grosser:
@schneems so you want it with warnings or without ?

schneems:
If you think without is the way to go, maybe leave a comment with that code in this issue so we could add it back if lots of people are confused by the behavior.

grosser:
sounds good, will work on tests now :)

grosser:
Warning patch:
```
@@ -180,7 +180,7 @@ module Puma
         graceful_stop
       when :restart
         log "* Restarting..."
-        ENV.replace(env)
+        restore_environment(env)
         @runner.before_restart
         restart!
       when :exit
@@ -204,6 +204,17 @@ module Puma

     private

+    def restore_environment(env)
+      a = ENV.to_a
+      b = env.to_a
+      diff = (a - b) + (b - a)
+      return if diff.empty?
+
+      diff.map! { |k, _| "#{k} from #{ENV[k].inspect} to #{env[k].inspect}" }
+      warn "Resoring process environment, updating:\n#{diff.join("\n")}"
+      ENV.replace(env)
+    end
+
     def reload_worker_directory
       @runner.reload_worker_directory if @runner.respond_to?(:reload_worker_directory)
     end
```

grosser:
great ... master is broken :/

grosser:
tests added ... extracted a few copy-paste parts into methods ... looks good ?

grosser:
not fixing the other errors on master ...
but integration test runs green locally

grosser:
@schneems mergeable ?

schneems:
Looks like master is failing with the same error. Thanks for the PR.

PR #1261, opened 2017-04-01
Don't omit empty header values.

jacksonrayhamilton:
Some servers might return empty HTTP header values.  For example:

```
Access-Control-Allow-Methods: GET, POST, PUT, DELETE, OPTIONS, PATCH
Access-Control-Expose-Headers:
Access-Control-Max-Age: 1728000
```

We are trying to write a test for a Rack proxy server, and in the tests we are using Puma.  Because Puma won't send headers with empty values, [our test is failing](https://github.com/ysbaddaden/prax.cr/blob/0517efd69dc6b67da1772948efb3e92a32bb7b93/test/proxy_test.rb#L67).  Our proxy server may be used with Rack configurations that use other handlers, such as WEBrick, which do send empty header values, so testing this case matters to us.

This change ensures that headers with empty values aren't omitted from the response.

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -687,7 +687,7 @@ module Puma
             next
           end
 
-          if vs.respond_to?(:to_s)
+          if vs.respond_to?(:to_s) && !vs.to_s.empty?
             vs.to_s.split(NEWLINE).each do |v|
               lines.append k, colon, v, line_ending
             end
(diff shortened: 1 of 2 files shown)


nateberkopec:
Looks like empty values are supported in the HTTP spec, so we should support them too. LGTM, can you add a test?

jacksonrayhamilton:
Sure, but can you point me to where the test should go?

nateberkopec:
Somewhere in https://github.com/puma/puma/blob/master/test/test_puma_server.rb?

jacksonrayhamilton:
Ok, added a test.

jacksonrayhamilton:
Thanks for merging this!  Can you make a release?

PR #1264, opened 2017-04-05
Expand the explanation for capistrano3-puma users

stevenchanin:
In the process of setting up a small, stand alone server for a Rails app on Ubuntu 16.04 LTS, I ran into a bunch of stuff I had to figure out to get Puma setup to auto-restart on crashes & auto-start on boot with systemd and capistrano3-puma.

I expanded the markdown docs to capture what I think is necessary and added some instructions for testing, etc.

If you think this might help other people, I'd be happy to have it merged.

Thanks for all the work on puma and capistrano3-puma (@seuros)

### DIFF
--- a/docs/systemd.md
+++ b/docs/systemd.md
@@ -171,9 +171,28 @@ Apr 07 08:40:19 hx puma[28320]: Use Ctrl-C to stop
 
 ## Alternative background process configuration
 
-If Capistrano and [capistrano3-puma](https://github.com/seuros/capistrano-puma) tasks are used you can use the following configuration. In this case, you would skip systemd Socket Activation, since Puma handles the socket by itself:
+If Capistrano and [capistrano3-puma](https://github.com/seuros/capistrano-puma) tasks are used you can use the following configuration. In this case, you would skip systemd Socket Activation, since Puma handles the socket by itself.
+
+First (as root, indicated by commands being prefixed with `#`), create a `puma.service`
+```
+# cd /etc/systemd/system
+# touch puma.service
+# chmod 664 puma.service
+```
+
+then pull that into an editor and insert the following setup. Then replace `<WD>` and do the dry-runs it suggests.
 
 ~~~~
+Unit]
+Description=Puma Rails Application/HTTP Server
+After=network.target
+
+# Preferably configure a non-privileged user (e.g. "deploy" if that's the user that owns the app and under which you want to run Puma)
+User=deploy
+
+# Specify the path to your puma application root
+WorkingDirectory=<WD>/current
+
 [Service]
 # Background process configuration (use with --daemon in ExecStart)
 Type=forking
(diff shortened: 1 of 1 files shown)


nateberkopec:
Totally. Need someone else to review this though as I am not a systemd user.

nateberkopec:
Literally anyone that uses capistrano and Puma, if you could just give this a read and let us know if it's sane, happy to merge.

In the long term, maybe we could move this stuff to wiki?

dekellum:
@nateberkopec I'll give this a review. Regarding the wiki suggestion, I would like to point out that [search] for the current docs/systemd.md is working well and it would be unfortunate to break that. Also I think pull-request formality for docs is appropriate given the cost of any misinformation accumulated. 

[search]: https://www.google.com/search?q=puma+systemd

nateberkopec:
Thanks!

dekellum:
I'm a regular puma and systemd user, but not a capistrano3-puma user.  Some comments on this PR:

Given the style of config generation in capistrano3-puma, and its current support for init.d (and others?), it would seem that a more complete solution would be to push support for systemd upstream to that project?  Cc: @seuros  

@mdesantis, the original contributor to ./docs/systemd alt. forking config, apparently had some luck using the default `pumactl` based stop/start/restart of capistrano3-puma, and as far I can tell, that is the most compelling reason to want to use the alt. forking config, and deal with its additional idiosyncrasies.   If as this PR suggests, @stevenchanin is able to make capistrano 3 use `systemctl` for stop/start/restart, then why not just use the `Type=simple` preferred configuration, with or without socket activation?

One important difference is that this is using/documenting systemd directive `Restart=always` where @mdesantis omitted this and the default is "Restart=no".  Restart=no makes a certain sense, if one is solely trying to get a systemd config to replace a legacy System V init.d config, for start on boot. As this PR points out, this wont have the advantage of systemd actively monitoring and restarting puma. In cluster mode however, the puma master also has the functionality to do that with its workers. (With legacy init.d, that was the only mechanism.)

I suspect the PID file warning is likely a timing issue with puma using `--daemon`: systemd expects that upon exit from the ExecStart command, the PID file is written and readable.  That might not be the case for example if the PID file is written by a subsequent, forked process.

@stevenchanin would you be willing to also cross review changes in #1329?

stevenchanin:
@dekellum -- thanks for the very thorough and detailed thinking on this. I'm traveling this week, so I'm not going to have time dig into this until next week when I'm home. But at that point, I'd be happy to cross review and to respond to / incorporate your comments (though I think you may understand systemd much more deeply than I do).

mdesantis:
@dekellum sorry for the delay. I add `WorkingDirectory` to the alternative forking configuration, since using `bundle exec` is almost required.

dekellum:
Folks, I think we have already extracted everything useful from this thread into the docs. 
@nateberkopec: suggest closing. Happy New Year!

stevenchanin:
@dekellum I'm sorry for not providing feedback on #1370, at the time I put up this PR, I was feeling my way through getting a systemd + puma + Capistrano setup working and I don't really have enough expertise with either systemd or puma to provide useful feedback. If you think all the valuable stuff has been included, I'm fine with this PR being closed.

mdesantis:
For me it's OK to close too, the alternative configuration is still working good in production. @dekellum thanks again for the review, you made my edit so much better. Happy 2018 to everyone!

PR #1266, opened 2017-04-10
test improvements

grosser:
- always use bundler
 - make tests run on their own (no `-Itest -Ilib` required)
 - helper is not a test
 - replace distracting test runner with standard colors

screen shot 2017-04-10 at 10 52 57 am

### DIFF
--- a/gemfiles/2.1-Gemfile
+++ b/gemfiles/2.1-Gemfile
@@ -8,5 +8,6 @@ gem "rake-compiler"
 
 gem "rack", "~> 1.6"
 gem "minitest", '~> 5.9'
+gem "minitest-rg"
 
 gem "jruby-openssl", :platform => "jruby"
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -4,37 +4,34 @@
 begin
   require "bundler/setup"
 rescue LoadError
+  warn "Failed to load bundler ... this should only happen during package building"
 end
 
 require "net/http"
 require "timeout"
 require "minitest/autorun"
-require "minitest/pride"
-require "puma"
-require "puma/detect"
+require "minitest/rg"
 
 $LOAD_PATH << File.expand_path("../../lib", __FILE__)
+
+require "puma"
+require "puma/detect"
 
 # Either takes a string to do a get request against, or a tuple of [URI, HTTP] where
 # HTTP is some kind of Net::HTTP request object (POST, HEAD, etc.)
 def hit(uris)
-  results = []
+  uris.map do |u|
+    response =
+      if u.kind_of? String
+        Net::HTTP.get(URI.parse(u))
+      else
+        url = URI.parse(u[0])
+        Net::HTTP.new(url.host, url.port).start {|h| h.request(u[1]) }
+      end
 
-  uris.each do |u|
-    res = nil
-
-    if u.kind_of? String
-      res = Net::HTTP.get(URI.parse(u))
-    else
-      url = URI.parse(u[0])
-      res = Net::HTTP.new(url.host, url.port).start {|h| h.request(u[1]) }
-    end
-
-    assert res != nil, "Didn't get a response: #{u}"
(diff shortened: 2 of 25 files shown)


grosser:
@schneems

test/helper.rb
```diff
@@ -1,40 +1,32 @@
 # Copyright (c) 2011 Evan Phoenix
 # Copyright (c) 2005 Zed A. Shaw
 
-begin
```
nateberkopec: I think this was added for a reason, I can't recall the issue.

test/test_app_status.rb
```diff
@@ -1,4 +1,4 @@
-require "test_helper"
```
nateberkopec: What's wrong with test_helper? That's a pretty common convention.

test/helper.rb
```diff
@@ -1,40 +1,32 @@
...
-  require "bundler/setup"
-rescue LoadError
-end
-
+require "bundler/setup"
 require "net/http"
 require "timeout"
 require "minitest/autorun"
-require "minitest/pride"
```
nateberkopec: No reason to remove IMO.

test/helper.rb
```diff
@@ -1,40 +1,32 @@
...
-    if u.kind_of? String
-      res = Net::HTTP.get(URI.parse(u))
-    else
-      url = URI.parse(u[0])
-      res = Net::HTTP.new(url.host, url.port).start {|h| h.request(u[1]) }
-    end
-
-    assert res != nil, "Didn't get a response: #{u}"
-    results << res
+  uris.map do |u|
```
nateberkopec: fine with this

grosser:
easier to see if it succeeded or failed ?
... I'd even prefer plain to pride since it's less distracting

grosser:
it's not a test and `rake test` requires it ... which means it gets loaded
twice ... which is weird

grosser:
let's kick it out and see what breaks then

nateberkopec:
Re: bundler rescue load-error, I mean I think that was added as a request by debian (?) or some other package manager team that wants to run the tests w/o bundler.

grosser:
kk, will add that as comment

grosser:
@nateberkopec merge ?

gemfiles/2.1-Gemfile
```diff
@@ -8,5 +8,6 @@ gem "rake-compiler"
 
 gem "rack", "~> 1.6"
 gem "minitest", '~> 5.9'
+gem "minitest-rb"
```
nateberkopec: Typo?

grosser: yep :D

grosser: ... should both use gemspec and gemspec should not be generated by hoe ...

grosser: fixed

nateberkopec:
The only change here I'm really ok with is the one to `def hit` and the Bundler rescue comment.

* I don't want to change the color output, minitest-pride is fine. Sorry if you disagree.
* `test_helper` is a common idiom and doesn't need to be changed.

grosser:
so given these 2 options, which you would chose the fabulously filed test ?

screen shot 2017-04-11 at 2 16 57 pm
screen shot 2017-04-11 at 2 16 33 pm

grosser:
extracted bundler changes to it's own PR

grosser:
so before I rebase this ...
 - you still like pride more than rg ?
 - is moving to helper ok ? (atm test_helper is treated as a test which is weird and could lead to load order issues)

nateberkopec:
@grosser 

> you still like pride more than rg ?

Yes

> is moving to helper ok ? (

Alright, it's OK.

grosser:
new clean PR instead of rebasing this mess :D 
https://github.com/puma/puma/pull/1283

PR #1268, opened 2017-04-10
sort configs/rackups/tests

grosser:
.. few of these are never user, still keeping them ... but makes the test folder a bit cleaner and encourages new testers to not just dump there ...


grosser:
@nateberkopec good to go ?

nateberkopec:
Thanks!

PR #1269, opened 2017-04-10
Fix #1265: Load environment-specific config file

### LINKED ISSUE #1265: What is the default configuration file path for a specified environment?
According to [Puma README: Configuration file][1]:

> If an environment is specified . . . through the RACK_ENV environment variable, the default file location will be config/puma/environment_name.rb.

Is this still true for Puma 3.8.2 with Ruby 2.4.1?

  [1]: https://github.com/puma/puma/blob/master/README.md#configuration-file

ma11hew28:
`@options[:environment]` is a lambda literal.

`@options[:environment].call` returns the environment string, eg, `"development"`.

### DIFF
--- a/lib/puma/configuration.rb
+++ b/lib/puma/configuration.rb
@@ -191,11 +191,8 @@ module Puma
       files = @options.all_of(:config_files)
 
       if files.empty?
-        imp = %W(config/puma/#{@options[:environment]}.rb config/puma.rb).find { |f|
-          File.exist?(f)
-        }
-
-        files << imp
+        names = %W(config/puma/#{@options[:environment].call}.rb config/puma.rb)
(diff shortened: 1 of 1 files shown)


nateberkopec:
👍  Needs a test though.

nateberkopec:
Ping, happy to merge with a test 👍

ma11hew28:
I'm sorry. I don't know how to write a test for this, but when I tried it out, it worked for me.

allaire:
Experiencing this issue as well

allaire:
I don't understand, going back to ruby 3.6.0 fix the environment config not loading bug, but it seems that this version is also using a proc for `@options[:environment]`, which got me thinking it might be something else?

nateberkopec:
I think #1340 is a more complete solution.

PR #1270, opened 2017-04-11
warn when bundler fails and avoid having to use -I to run tests

grosser:


grosser:
@nateberkopec this looking good ?

nateberkopec:
Sorry, what I meant was that the entire `require test_helper` line should remain the same. There's no reason to use require_relative here.

grosser:
The reason I changed that is to allow `ruby test/test_foo.rb` to work

grosser:
... that might not be an issue for senior devs, but it's not obvious that you have to add the magic `-Itest` incantation

nateberkopec:
Fair enough

Issue #1189, opened 2017-01-14
Issue with WebSocket through SSL


brunowego:
### Steps to reproduce

1) Open `puma.rb` and add SSL certificates:

```rb
...
ssl_key = File.expand_path '../certificates/server.key', __FILE__
ssl_cert = File.expand_path '../certificates/server.crt', __FILE__
bind "ssl://127.0.0.1:3000?key=#{ssl_key}&cert=#{ssl_cert}"
```

2) Set on `development.rb` to force SSL and allowed origin request for ActionCable:

```rb
...
config.force_ssl = true
config.action_cable.allowed_request_origins = ['https://lvh.me:4000']
...
```

3) In my settings I use redis:

```yml
development:
  adapter: redis
  url: redis://localhost:6379/1
...
```

### Expected behavior

The connection using ActionCable works seamlessly without  SSL.

### Actual behavior

If try use SSL, the warning message is shown:

```
WebSocket connection to 'wss://lvh.me:3000/cable' failed: WebSocket is closed before the connection is established.
```

### System configuration

**Ruby version**: ruby 2.3.2p217 (2016-11-15 revision 56796) [x86_64-darwin16]
**Rails version**: Rails 5.0.1

brunowego:
Similar issue reported at https://github.com/rails/rails/issues/27625

brunowego:
The issue is in Rails 5.0.1, reported here https://github.com/rails/rails/issues/27421

brunowego:
I solved with downgrade to version of Rails 5.0.0.1.

edwardmp:
I'm still seeing this issue on Rails 5.0.2. Will open issue on Rails github.

whitehat101:
I've reviewed the referenced rails issue, and this does appear to be from a quirk with Puma's minissl, so I thought I'd bump here instead of there. To use SSL+WebSockets I either need to lock rails to 5.0.0.1 or find someone other than Puma to do SSL termination, and I'm not excited about either. 

https://github.com/rails/rails/issues/28362#issuecomment-285903720
> My take on how to fix this would be for Puma to conform to TCPSocket's API, and to implement the write_nonblock method.

Do you agree that that is (part of) the long term solution?

I see @edwardmp mentions trying things with `write_nonblock`, but I wasn't able to find a branch or code. Could you share those efforts, or call out the blocks of code that need to be massaged?

[...]

PR #1274, opened 2017-04-14
Make write_nonblock available under SSL

evanphx:
This implementation of `write_nonblock` is not technically correct because it can block, but at least it gets past #1189 for the time being. 

The situation is that implementing it properly means having to be able to rewind the state of OpenSSL, which is hardly trivial.

### DIFF
--- a/lib/puma/minissl.rb
+++ b/lib/puma/minissl.rb
@@ -84,7 +84,7 @@ module Puma
       # is that it means we'd have to have the ability to rewind
       # an engine because after we write+extract, the socket
       # write_nonblock call might raise an exception and later
-      # code would pass the same data in, but the engine would thing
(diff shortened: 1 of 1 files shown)


lib/puma/minissl.rb
```diff
@@ -77,6 +79,19 @@ def write(data)
       alias_method :syswrite, :write
       alias_method :<<, :write
 
+      # This is a temporary fix to deal with websockets code using
+      # write_nonblock. The problem with implementing it properly
+      # is that it means we'd have to have the ability to rewind
+      # an engine because after we write+extract, the socket
+      # write_nonblock call might raise an exception and later
+      # code would pass the same data in, but the engine would thing
```
schneems: thing => think

PR #1277, opened 2017-04-27
[Close #1255] Prefer user supplied defaults

### LINKED ISSUE #1255: 3.8.0: rails s -b 0.0.0.0 will set port to default puma port
### Steps to reproduce

1) install rails with puma version 3.8.0/1/2
2) run `rails s -b 0.0.0.0`

### Expected behavior

Listening on tcp://0.0.0.0:3000

### Actual behavior

Listening on tcp://0.0.0.0:9292

### System configuration

In docker env:
**Ruby version**: 2.4.1
**Rails version**: 5.1.0rc1, not tested with 4.x

### Quick solution

Run `rails s -b 0.0.0.0 -p 3000`

schneems:
To build a "binds" we need a host IP (via Host) and a port. We were running into a problem where a Host was being explicitly set via user but the Port was being defaulted to by Rails. When this happened the Host was used, but Puma would accidentally use it's own default port 9292 instead of Rail's port of 3000.

The fix was to use the "default" port passed in from a framework (if available) when no explicitly set Port is provided.

### DIFF
--- a/lib/rack/handler/puma.rb
+++ b/lib/rack/handler/puma.rb
@@ -42,7 +42,10 @@ module Rack
             user_config.threads min, max
           end
 
-          self.set_host_port_to_config(options[:Host], options[:Port], user_config)
+          host = options[:Host] || default_options[:Host]
+          port = options[:Port] || default_options[:Port]
+
+          self.set_host_port_to_config(host, port, user_config)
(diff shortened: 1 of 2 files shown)


grosser:
release ?

Issue #1107, opened 2016-10-04
Cluster-mode: request blocked by busy worker even though other worker has idle threads


sbull:
I'm exploring cluster mode (with MRI, linux) with Rails 5, and I'm seeing some unfortunate behavior. I'm using 2 worker processes with 5 threads each (like the defaults in puma-heroku), with `preload_app!`. I've intentionally made a "slow" controller action to play with the thread & process operation and check how the requests are handled under load.

The mechanism by which puma chooses which worker in a cluster will handle a request is unclear to me (I'm thinking it might be up to the OS for whichever process grabs the IO first in Puma::Server#handle_servers?), but in my dev environment it seems that one of the workers is usually favored over the other - significantly different from a random distribution (I see the worker PID in my logs). So, I make 5 concurrent requests for the "slow" action, and on the 6th concurrent request to a "fast" action, it blocks until one of the "slow" requests finishes, and is handled by the same worker. But on my 7th concurrent request to a "fast" action, still before any of the first 6 requests have finished, it gets handled by the other worker without delay.

My conclusion is that worker#0 must be taking requests and handing them to its available threads until it determines that it has no more threads ready to handle them, but it doesn't figure that out until AFTER it's already grabbed the 6th request, preventing worker#1 from servicing it. So even though I've got a whole extra worker of capacity with 5 idle threads, my poor request 6 gets blocked until one of worker#0's threads frees up.

Is there a way to get around this? Maybe worker#0 could stop trying for the IO when all of its threads are busy - or maybe this is something related to the `queue_requests` option?

Here's the templated info:
### Steps to reproduce

1) Make a "slow" action along with a "fast" action:

``` ruby
class TesterController < ApplicationController

  def slow
    sleep 20
    render plain: 'slow'
  end

  def fast
    render plain: 'fast'
  end

end
```

2) Make 5 concurrent requests to /slow, then a concurrent request to /fast, and another concurrent request to /fast. Check which worker handles the requests (in log files, or debug messages?), and note the timing.
### Expected behavior

When the first 5 concurrent requests to /slow are handled by worker#0, both requests to /fast should be handled by worker#1.
### Actual behavior

When the first 5 concurrent requests to /slow are handled by worker#0, the 6th request to /fast is handled by worker#0 as well, and waits until a /slow request finishes before being processed. The 7th request to /fast appears to be correctly handled by worker#1.
### System configuration

**Ruby version**: MRI 2.3.1 (via rbenv)
**Rails version**: 5.0.0.1

[...]

PR #1278, opened 2017-04-27
Meter calling accept(2) with available pool capacity

evanphx:
Talking through this with @nateberkopec, @schneems, @FoobarWidget, and others at RailsConf , we realized that accepting new clients without taking account for the available capacity of the thread pool doesn't improve throughput, it only hurts it in the case of multiple workers. If a worker pauses (or starts up before other workers), then a worker can accidentally suck up a high number of
clients and leave unused capacity inside the other workers.

This change will smooth out this issue, with a minor penalty to maximum throughput.

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -363,7 +363,7 @@ module Puma
                     end
 
                     pool << client
-                    pool.wait_until_not_full unless queue_requests
+                    pool.wait_until_not_full
                   end
                 rescue SystemCallError
                   # nothing
--- a/lib/puma/thread_pool.rb
+++ b/lib/puma/thread_pool.rb
@@ -155,7 +155,15 @@ module Puma
 
     def wait_until_not_full
       @mutex.synchronize do
-        until @todo.size - @waiting < @max - @spawned or @shutdown
+        while true
+          return if @shutdown
+          return if @waiting > 0
+
+          # If we can still spin up new threads and there
+          # is work queued, then accept more work until we would
+          # spin up the max number of threads.
+          return if @todo.size < @max - @spawned
+
           @not_full.wait @mutex
         end
       end


nateberkopec:
Linking #1107

nateberkopec:
I think this could really improve performance for a lot of people, so I'm going to start looking at packaging up a new release.

ivoanjo:
@evanphx and @nateberkopec, thanks for looking into this and looking forward to a puma release :)

nateberkopec:
👍 since the Jruby bug on master was fixed, I think I can push a 3.9.0 soon.

waynerobinson:
This will also resolve some potential deadlocking issues for APIs that can sometimes end up calling themselves (i.e. authentication systems that use microservices that need authenticating). 

👍👍

joealba:
@evanphx @schneems Do you recommend this upgrade for apps that are not yet threadsafe?  And does that answer change if the app is also behind the Heroku router?

If I have N dynos with X workers each -- and the Heroku router picks a single busy worker in each of my N dynos, would that request fail due to the combination of this PR and a non-threadsafe app?

(I've tested this theory with `ab` and 4x higher concurrency than the number of workers, and I saw no failed requests.  But I thought it prudent to ask you folks as well.)

References:
- https://gist.github.com/schneems/c07d93d5a4ade679bbc3
- https://devcenter.heroku.com/articles/http-routing#request-distribution

schneems:
>  apps that are not yet threadsafe?

I actually don't recommend puma for apps that aren't threadsafe, unless it's a step to getting the app threadsafe. I recommend the `rack-freeze` gem for threadsafety in the middleware (it's a great gem). 

I've seen performance issues with Puma using X workers and 1 thread compared to Unicorn with X workers. I'm not totally sure why, but i've seen this several times. You could also try passenger.

I don't think this patch would hurt if you're only using 1 thread, but i've seen other perf issues with that setup. I would recommend either putting in leg work to get the app threadsafe, or moving to a different server.

lib/puma/server.rb
```diff
@@ -363,7 +363,7 @@ def handle_servers
                     end
 
                     pool << client
-                    pool.wait_until_not_full unless queue_requests
+                    pool.wait_until_not_full
```
danshultz: Was there a reason `queue_requests` was removed here?

nateberkopec: I think the reason is explained adequately in the OP?

lib/puma/thread_pool.rb
```diff
@@ -155,7 +155,15 @@ def <<(work)
...
       @mutex.synchronize do
-        until @todo.size - @waiting < @max - @spawned or @shutdown
+        while true
+          return if @shutdown
+          return if @waiting > 0
+
+          # If we can still spin up new threads and there
+          # is work queued, then accept more work until we would
+          # spin up the max number of threads.
+          return if @todo.size < @max - @spawned
```
ivanpesin: This is not an equivalent of initial condition. With this change puma serving a single-threaded app with "threads 1,1" attempts to pass a request to a worker that is already serving a request. We're seeing occasional deadlocks for the requests coming right after a long-running request  in our environment with "threads 1,1", "queue_requests false", and "workers 5" configuration. Rolling back this single change to the original condition solves the issue completely.

nateberkopec: In the case of a single-threaded worker, `@waiting` will be 0, `@todo.size` will be 1, `@max` will be 1 and `@spawned` will be 1, so the condition evaluates to `1 < 1 - 1`, which is zero, so Puma waits here.

ivanpesin: The problem is in line:
`+          return if @waiting > 0`

This is not how the original condition worked: `until @todo.size - @waiting < @max - @spawned or @shutdown`. The fact that `@waiting > 0` is not enough for decision not to do `@not_full.wait @mutex`.

I prepared a simple test scenario which demonstrates the problem: https://gist.github.com/ivanpesin/f40a76e51a60a83f1190113fdecddaa5

With this commit, second request gets *always blocked* until first request is served, even though there are available workers. Third requests goes to available worker.

With original condition, second request goes to next available worker and works as expected. The condition needs to be either reverted to original form or needs logic correction.

nateberkopec:
@eprothro 

> To clarify, my understanding is that there is still benefit in having the queue_requests option, as this allows requests to be received without blocking a worker thread from handling on another request that has already been received. This has not changed, simply how many requests could be in this state (being handled by the queue_requests thread before being handed to a worker thread) has changed.

That is correct.

lib/puma/thread_pool.rb
jasquat: We also had this issue.

Our setup is using 16 workers with 1 thread each running with an nginx server in front. We were getting high wait times between nginx and rails as measured by a middleware similar to the one at https://github.com/puma/puma/issues/1405#issuecomment-330806817. We partially undid the change in this pull request with the patch below and our wait times dropped as can be seen by the attached screenshot.

````ruby
module PumaMonkeypatch
  def wait_until_not_full
    @mutex.synchronize do
      until @todo.size - @waiting < @max - @spawned or @shutdown
        @not_full.wait @mutex
      end
    end
  end
end
Puma::ThreadPool.prepend(::PumaMonkeypatch)
````

image

Thoughts on how to proceed? Should we file a new issue?

nateberkopec: > Should we file a new issue?

Please do.

jasquat: Thanks. Ticket created: https://github.com/puma/puma/issues/1471

Issue #1072, opened 2016-09-02
Why does bin/puma-wild not have executable bit?


junaruga:
bin/puma-wild does not have executable bit, and is not included in puma.gemspec "s.executables", 
though it is included in the gem file.

Could you tell me why?

evanphx:
It's only launched internally, it's not meant to be run directly.

junaruga:
@evanphx okay. I got it. thanks.

PR #1279, opened 2017-04-29
Change permission to 0755 from 0644

yhirano55:
just executed `chmod +x bin/puma-wild`.


nateberkopec:
See #1072

yhirano55:
Thanks for reviewing. I got it.

PR #1283, opened 2017-05-03
make test_helper no longer be loaded as a test

grosser:
running `rake` loads all files starting with `test_` which is not supposed to also load the helper,
since the helper should be loaded by each test and is not a test itself.

@nateberkopec


grosser:
seeing this error on master too:

```
  1) Failure:
TestUserSuppliedOptionsIsEmpty#test_config_file_wins_over_port [/Users/mgrosser/Code/tools/puma/test/test_rack_handler.rb:121]:
Expected: ["tcp://0.0.0.0:6001"]
  Actual: ["tcp://0.0.0.0:5001"]
```

Issue #1181, opened 2016-12-27
Native extension build failure on MRI 2.4.0 linked against OpenSSL 1.1.0*


steakknife:
MRI 2.4.0 [changed](https://www.ruby-lang.org/en/news/2016/12/25/ruby-2-4-0-released/) OpenSSL compatibility, adding 1.1.\*.  It looks like the Puma native ext currently uses an OpenSSL ABI < 1.1. Probably worth supporting 0.98-1.1 as Ruby 2.4 does.

There's also a new OpenSSL cpp flag (ie `-DOPENSSL_API_COMPAT=0x10000000L`) for apps to ensure they are compiling against a compatible OpenSSL installation.

### Steps to reproduce (macOS example)

```
brew install openssl openssl@1.1 ruby-install chruby
mkdir -p /opt/rubies
ruby-install ruby-2.4.0 --no-install-deps -i /opt/rubies/ruby-2.4.0-ossl1.1 -- --without-X11 --without-tk --disable-install-doc --with-openssl-dir="$(brew --prefix openssl@1.1)" --enable-shared
chruby-exec ruby-2.4.0-ossl1.1 -- gem install puma -v 3.6.2 -N
# watch it crash
```

steakknife:
### Output
```
Building native extensions.  This could take a while...
ERROR:  Error installing puma:
	ERROR: Failed to build gem native extension.

    current directory: {{~}}/.gem/ruby/2.4.0/gems/puma-3.6.2/ext/puma_http11
/opt/rubies/ruby-2.4.0-ossl1.1/bin/ruby -r ./siteconf20161226-57195-1pnp5eq.rb extconf.rb
checking for BIO_read() in -lcrypto... yes
checking for SSL_CTX_new() in -lssl... yes
checking for openssl/bio.h... yes
creating Makefile

To see why this extension failed to compile, please check the mkmf.log which can be found here:

  {{~}}/.gem/ruby/2.4.0/extensions/x86_64-darwin-16/2.4.0/puma-3.6.2/mkmf.log

current directory: {{~}}/.gem/ruby/2.4.0/gems/puma-3.6.2/ext/puma_http11
make "DESTDIR=" clean

current directory: {{~}}/.gem/ruby/2.4.0/gems/puma-3.6.2/ext/puma_http11
make "DESTDIR="
compiling http11_parser.c
[...]

PR #1285, opened 2017-05-05
LibreSSL fix

fmauNeko:
This fixes running an SSL server built against LibreSSL.

Fixes parts of #1181


nateberkopec:
Hm, do we need a version guard here as well?

fmauNeko:
We don't, the problematic function, `DH_set0_pqg`, was added only to OpenSSL 1.1.0, and not to LibreSSL

ernie:
Popped in here to say this one worked for me today, only to find it just got merged. Thanks!

rafmagana:
This works for me. When is this going to be released?

PR #1289, opened 2017-05-12
DO NOT MERGE, WILL CLOSE: Add /gc URL to status server. Send GC stats as a result of GC request.

noahgibbs:
After talking to Nate, I did some tests to see how much of slow-request Rails performance was due to GC. The blog post should go up next week :-) As part of that, I built a customized Puma whose control server would do a full GC and dump GC stats.

Arguably the GC and the GC stats should be two separate URLs in the control server, but this seems useful enough to me to be worth a PR.

### DIFF
--- a/lib/puma/app/status.rb
+++ b/lib/puma/app/status.rb
@@ -60,7 +60,6 @@ module Puma
           return rack_response(200, OK_STATUS)
 
         when /\/gc-stats$/
-          stats = GC.stat
           json = "{" + GC.stat.map { |k, v| "\"#{k}\": #{v}" }.join(",") + "}"
           return rack_response(200, json)
 
--- a/lib/puma/app/status.rb
+++ b/lib/puma/app/status.rb
@@ -55,14 +55,14 @@ module Puma
             return rack_response(200, OK_STATUS)
           end
 
-	when /\/gc$/
-	  GC.start
-	  return rack_response(200, OK_STATUS)
+        when /\/gc$/
+          GC.start
+          return rack_response(200, OK_STATUS)
 
-	when /\/gc-stats$/
+        when /\/gc-stats$/
           stats = GC.stat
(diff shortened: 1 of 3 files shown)


nateberkopec:
All for it except the forced major on checking stats. There are times I want to do that, other times I don't.

noahgibbs:
So then you'd prefer two separate routes, maybe /gc and /gc-stats?

nateberkopec:
Yeah, either that or just remove the GC.start from the PR.

noahgibbs:
Okay. I've added a second route. Works for me locally. I see a bunch of CI failures, but no clear sign they're related to my PR. Did I break that, or was it already broken?

stereobooster:
master fixed, you can rebase

noahgibbs:
Huh. I "closed" this by pushing different commits to my repo. Hang on a sec...

nateberkopec:
Rubocop failures. Get rid of those filthy, filthy tabs.

stereobooster:
@nateberkopec do we have tests for those things?

nateberkopec:
For tabs? Yes, Rubocop fails otherwise.

stereobooster:
For URLs, like `/gc`

noahgibbs:
There are some tests in test/test_cli.rb, but currently no test for /gc or /gc-stats. I'll try to put something like that together quick.

Also, huh, I'm finding that these don't terminate for me locally. They take a very long time to run, possibly forever. At first I thought my new test-in-progress did that. But then I removed it, and nope, it still runs forever.

Okay, I'll look into this more. But I'm not sure I'm going to be able to usefully run the tests locally, even after reverting all the changes I made.

noahgibbs:
Okay, nope. I'll try to find a way to debug this test locally, then rejigger the commits and open a new PR.

PR #1294, opened 2017-05-15
Don't use Dir.exists? as it's deprecated and use Dir.exist? instead

masry707:


nateberkopec:
👍 Thanks for your contribution.

PR #1295, opened 2017-05-15
load realpath for require paths symlinked

### LINKED ISSUE #1272: Puma::Launcher#prune_bundler should load target path instead of symlink for puma_lib_dir
I found there was a periodic restart failed in my mina deployment. The main error could be found in puma error log:

> /xxx/.rvm/rubies/ruby-2.4.0/bin/ruby: No such file or directory -- /xxx/project/releases/38/vendor/bundle/ruby/2.4.0/gems/puma-3.8.1/bin/puma-wild (LoadError)

### Steps to reproduce

1) run `mina deploy` multiple times (depended on your `keep_releases` settings in mina config)

### Expected behavior

Puma should be restarted normally.

### Actual behavior
Puma failed to restart, and found error shown above.

### System configuration

**Ruby version**: 2.4.0
**Rails version**: 5.0.1
**Deployment Tool**: Mina
**puma**: 3.8.1
**mina-puma**: 1.0.1
**mina**: 1.0.6

### My Solution
According to the error message, it complained a missing file under a removed **release** directory, so I tried to debug and print some key variables and eventually found that the root reason for it: Puma::Launcher#prune_bundler load gem's required paths which are symlink.

```ruby
def prune_bundler
  puma = Bundler.rubygems.loaded_specs("puma")
  dirs = puma.require_paths.map { |x| File.join(puma.full_gem_path, x) }
  puma_lib_dir = dirs.detect { |x| File.exist? File.join(x, '../bin/puma-wild') }
```
`dirs` will be something like: 
```ruby
["/my_application_path/releases/194/vendor/bundle/ruby/2.4.0/gems/puma-3.8.1/my_application_path/releases/194/vendor/bundle/ruby/2.4.0/extensions/x86_64-linux/2.4.0/puma-3
.8.1",
"/my_application_path/releases/194/vendor/bundle/ruby/2.4.0/gems/puma-3.8.1/lib"]
```
But it'd better to use the target's absolute path directly:

```ruby
dirs = puma.require_paths.map { |x| File.join(Pathname(puma.full_gem_path).realpath, x) }
```
and then we will successfully get `dirs` like:
```
["/my_application_path/shared/vendor/bundle/ruby/2.4.0/gems/puma-3.8.1/my_application_path/releases/194/vendor/bundle/ruby/2.4.0/extensions/x86_64-linux/2.4.0/puma-3.8.1",
 "/my_application_path/shared/vendor/bundle/ruby/2.4.0/gems/puma-3.8.1/lib"]
```

I wonder if someone else meet the same problem? I will submit a PR if you guys think it is necessary.

Martin288:
fix #1272

### DIFF
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -239,7 +239,7 @@ module Puma
     def prune_bundler
       return unless defined?(Bundler)
       puma = Bundler.rubygems.loaded_specs("puma")
-      dirs = puma.require_paths.map { |x| File.join(puma.full_gem_path, x) }
+      dirs = puma.require_paths.map { |x| File.join(Pathname(puma.full_gem_path).realpath, x) }
       puma_lib_dir = dirs.detect { |x| File.exist? File.join(x, '../bin/puma-wild') }
 
       unless puma_lib_dir


grosser:
FYI trying to fix this whole bundler mess by unsetting bundler ENV vars on restart ... would that solve your issue too ?

```Ruby
# config/puma.rb
Puma::Runner.prepend(Module.new do
  def before_restart
    ENV.replace(Bundler. clean_env)
    super
  end
end)
```

Martin288:
@grosser thanks. but I don't know if there is any side effect that unsetting bundler ENV vars, have you tried it?

grosser:
yes, using that in prod since ~1 year
only issue is that restarts might not know how to restart so you have to use `--restart-cmd 'bundle exec puma'`

Martin288:
@grosser great! Will use it soon. But I still hope it could be solved from the puma repository, however, your solution is a good temporarily solution.

grosser:
once https://github.com/bundler/bundler/issues/5700 is resolved puma will work like this ... but there will be solve fallout with puma forgetting how to restart ... so ideally we'd fix that up properly until then :)

stereobooster:
@grosser by idea this is same as your code, but less hacky

```ruby
on_restart do
  ENV.replace(Bundler.clean_env)
end
```

x-yuri:
@grosser I'd rather avoid cleaning env. By the way, it's [deprecated][deprecated] for one reason or another. Additionally, it makes `restart_command` mandatory.

Also, I can confirm that PR by @Martin288 [fixes][test] it for cluster mode + `prune_bundler`. Since in this case `puma` restarts itself before starting workers. But unfortunately not for other cases. The solution might be to set instance variable on `Puma::Launcher.initialize` with `Pathname($0).realpath.to_s`. And use that instead of `$0` down the road.

[deprecated]: https://github.com/bundler/bundler/blob/v1.16.2/lib/bundler.rb#L267
[test]: https://gist.github.com/x-yuri/691d9c32dafe6b7165f3523213ac82ec

grosser:
yeah either way is fine, this whole realpath mess goes away when deploying using docker anyway, so meh :)

x-yuri:
With the following `puma` config (`prune_bundler` and `workers` are from the issue):

```rb
require 'puma/runner'
Puma::Runner.prepend(Module.new do
    def before_restart
        ENV.replace(Bundler.clean_env)
    super
  end
end)
workers 2
prune_bundler
```

It (`puma-3.12.0`) fails to restart but for another reason:

```rb
/home/user/app/current/config/puma/production.rb:19:in `before_restart': uninitialized constant #<Class:#<Puma::DSL:0x007fe32aeb7ef8>>::Bundler (NameError)
    from /home/user/app/releases/11/vendor/bundle/ruby/2.3.0/gems/puma-3.12.0/lib/puma/launcher.rb:194:in `run'
    from /home/user/app/releases/11/vendor/bundle/ruby/2.3.0/gems/puma-3.12.0/lib/puma/cli.rb:78:in `run'
    from /home/user/app/releases/11/vendor/bundle/ruby/2.3.0/gems/puma-3.12.0/bin/puma-wild:31:in `<main>
```

So, I remove `prune_bunder`, and this error goes away, but we're at back at [square one][1]. The issue is not resolved. I try it without `workers 2`, same result. The result is also identical with the following block:

```rb
on_restart do
  ENV.replace(Bundler.clean_env)
end
```

So, it must be safe to assume, that bundler/bundler#5700 has nothing to do with it. And if you look at the report [closely enough][2], you'll see that it has nothing to do with environment variables.

[1]: https://github.com/puma/puma/issues/1272
[2]: https://github.com/puma/puma/issues/1272#issuecomment-402392550

nateberkopec:
Closing this, as I think we're going to go with a different approach from #1105 to fix this code.

Issue #1293, opened 2017-05-15
IOError: stream closed from "@notify.close" when `close` is called from server.rb


henrik:
We're getting an error like this every few days in a small internal service app (running Padrino, deployed to Heroku) with low amounts of traffic:

```
IOError in at_exit
IOError: stream closed

[GEM_ROOT]/gems/puma-3.8.2/lib/puma/server.rb:393 :in `close`
 [GEM_ROOT]/gems/puma-3.8.2/lib/puma/server.rb:393 :in `ensure in handle_servers`
 [GEM_ROOT]/gems/puma-3.8.2/lib/puma/server.rb:396 :in `handle_servers`
 [GEM_ROOT]/gems/puma-3.8.2/lib/puma/server.rb:325 :in `block in run`
```

Line 393 being:

```
[GEM_ROOT]/gems/puma-3.8.2/lib/puma/server.rb:393:in `close'
391       ensure
392         @check.close
>>>393         @notify.close
394 
395         if @status != :restart and @own_binder
```

I haven't dug into this, but apparently `@notify` is some IOStream that is already closed when we're trying to close it, on server shutdown.

I would guess Puma could check if it's closed already to fix that error, but that's kind of sweeping it under the rug. Any ideas what the issue could be? Something we might be doing wrong on our end?

henrik:
Might it be the same issue as this one? https://github.com/puma/puma/pull/1206

vesan:
Just letting you know that we are seeing the same problem with puma 3.8.2 & Rails 5.0.2. I don't think we have anything special in the application that would cause it but you'll never know.

We also have quite low amounts of traffic (the application is still in development).

nateberkopec:
This error never occurs on our JRuby builds. Could be MRI specific, which does point to #1206, but merging that patch didn't fix the CRuby builds, so I'm not sure.

NickolasVashchenko:
Recent MRI versions do not even have the have the "stream closed" error message for couple of months anymore, it's called now "stream closed in another thread". Please update to the most recent patch version, it most likely will resolve your issue.

mezis:
@NickolasVashchenko: the error does persist with MRI `HEAD` as of writing:

```
...vendor/bundle/ruby/2.5.0/gems/puma-3.9.1/lib/puma/server.rb:393:in `close': stream closed in another thread (IOError)
```

(after receiving a `SIGTERM`, in certain circumstances)

[...]

PR #1302, opened 2017-05-29
WIP: Release fix

nateberkopec:


nateberkopec:
So @NickolasVashchenko's description of this bug in Ruby in #1206 was definitely on the money, but I'm not sure I've got a solution yet. Please check @evanphx to make sure we can actually do this, also I need a solution for why STDERR is getting closed, which is why I had to comment it out in this PR.

lib/puma/server.rb
```diff
@@ -386,11 +386,11 @@ def handle_servers
           @reactor.shutdown
         end
       rescue Exception => e
-        STDERR.puts "Exception handling servers: #{e.message} (#{e.class})"
-        STDERR.puts e.backtrace
+        # STDERR.puts "Exception handling servers: #{e.message} (#{e.class})"
```
nateberkopec: [Raises as if STDERR was closed](https://travis-ci.org/puma/puma/jobs/237251772#L453) if not commented out.

nateberkopec:
@evanphx said since only IO related errors can come out of io.close, we can actually just rescue nil.

NickolasVashchenko:
https://ruby-doc.org/core-2.2.0/Interrupt.html - this one can come virtually from anywhere. And this was exactly the case I've encountered in https://github.com/puma/puma/pull/1206

nateberkopec:
Interrupt doesnt inherit from standarderror tho, so `rescue nil` won't catch it.

nateberkopec:
Anyway @NickolasVashchenko want to take another swing? I'm trying to get rid of all the errors on master related to this :|

NickolasVashchenko:
Sure, why now. You're speaking about the failed tests on CI?

nateberkopec:
@NickolasVashchenko Yeah. As an example, look at ruby-head: https://travis-ci.org/puma/puma/jobs/237347708 and all of the "stream closed in another thread" errors. 

Your original patch doesnt seem to fix everything on CI any longer, and my patch here requires disabling the STDERR.puts stuff, so I'm not happy with either solution right now.

NickolasVashchenko:
Ok, I'll check what could be done there :thumbsup:

NickolasVashchenko:
Okay, clearly that's again a ruby bug with a complicated story. I'm still checking the details.

NickolasVashchenko:
Here are preliminary results.
It has been introduced here: https://github.com/ruby/ruby/commit/bc855ef1391bb79ca3554f67ae2d3aad19178d94
As an attempt to fix this race condition, described in this ticket: 
https://bugs.ruby-lang.org/issues/13076
It has been fixed by a busy waiting approach, when if a file descriptor is not released by a thread fast enough when this FD is closed in some other thread, then the "stream closed" exception gets repeatedly re-added in the thread internal interrupt queue. 
It leads to the case that it's possible for a thread to receive several hundred or even thousands of the interrupt of the same type during this operation. The main problem with that is that those exceptions being scheduled in bulk, but being dequeued one-by-one in very specific conditions. When a thread gets GVL back, on on any IO-related operation. 
But besides that is obviously a poor solution from the performance point of view, it also introduces the bug we've seen on the Puma's CI. What raised there is one of those excessively enqueued exceptions, that are being dequeued much-much later, on an IO operation performed in the affected thread. 
This is easily verifiable by, doing, for example, an IO.sysopen method call. After a couple of hundred calls to it, when the whole queue is processed, it starts working normally again.

I'm going to report this to the ruby-core team, but I'm still working on an example when it's consistently reproducible since it depends on how many interrupts has been scheduled during the waiting period. But that's gonna be too hard. If you're interested how and when it's gonna be resolved, I can post the link here when I'm done.
Here's what it probably means for Puma:
1. That is/will be recognized as a ruby bug.
2. It will be fixed somehow before the 2.5.0 release.
3. Thus, I don't think Puma needs a workaround for that because it's gonna be exceptionally ugly and short living.

nateberkopec:
> It does not affect any of the other Ruby versions that have ever been released.

I'm confused by this. We're seeing errors on 2.2, 2.3, 2.4 and ruby-head.

NickolasVashchenko:
My bad, sorry. I see they backported it:  https://github.com/ruby/ruby/blob/ruby_2_4/thread.c#L2199
In meanwhile I managed to get the script for reproducing the bug.

NickolasVashchenko:
Yeah, I think, fix can be tolerably ugly, but it would involve a native extension.

NickolasVashchenko:
Reported it: https://bugs.ruby-lang.org/issues/13632?next_issue_id=13627

NickolasVashchenko:
The guys fixed it blazingly fast 😄 
https://github.com/ruby/ruby/commit/59fb92971fa8e461430c35307a6b9eb0572ed4c5

NickolasVashchenko:
I also expect this to fix the https://github.com/puma/puma/issues/1293.

nateberkopec:
Wow, that was quick. Thanks to your thorough report, I think! I'll take another look.

NickolasVashchenko:
Okay, sorry for misinterpreting that, it's not yet backported, but just in the trunk.

Issue #1292, opened 2017-05-12
Release blocker: JRuby 9k failures


nateberkopec:
Seeing two new failures locally since 3.8:

```
  1) Error:
TestIntegration#test_restart_closes_keepalive_sockets:
Errno::ECONNREFUSED: Connection refused - connect(2) for "127.0.0.1" port 9998
    org/jruby/ext/socket/RubyTCPSocket.java:134:in `initialize'
    org/jruby/RubyIO.java:877:in `new'
    /Users/nateberkopec/Code/puma/test/test_integration.rb:218:in `test_restart_closes_keepalive_sockets'


  2) Error:
TestIntegration#test_restart_restores_environment:
Errno::ECONNREFUSED: Connection refused - connect(2) for "localhost" port 9998
    org/jruby/ext/socket/RubyTCPSocket.java:134:in `initialize'
    org/jruby/RubyIO.java:877:in `new'
    /Users/nateberkopec/Code/puma/test/test_integration.rb:248:in `test_restart_restores_environment'
```

FooBarWidget:
You sure this is only a JRuby bug? I remember running into macOS kernel bugs that look similar to this. Sometimes when you connect to a TCP server right after creating it, you get connection refused.

nateberkopec:
> macOS kernel bugs that look similar to this 

o great

nateberkopec:
[It looks like it happens on Travis too, so probably not a kernel bug](https://travis-ci.org/puma/puma/jobs/232411633)

nateberkopec:
[Actually, they seem to happen on CRuby too](https://travis-ci.org/puma/puma/jobs/232411625), so not sure why it passes for me locally.

nateberkopec:
Can confirm this started after #1260 was merged.

grosser:
reverting the test changes fixes the issue ... so something went wrong there :/

grosser:
found it ... https://github.com/puma/puma/pull/1304

PR #1304, opened 2017-05-30
fix jruby bug

grosser:
fixes https://github.com/puma/puma/issues/1292

### DIFF
--- a/test/test_integration.rb
+++ b/test/test_integration.rb
@@ -94,7 +94,7 @@ class TestIntegration < Minitest::Test
   end
 
   def wait_for_server_to_boot
-    true while @server.gets =~ /Ctrl-C/
+    true while @server.gets !~ /Ctrl-C/ # wait for server to say it booted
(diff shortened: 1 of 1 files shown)


nateberkopec:
Thanks for fixing this so quickly!

grosser:
maybe wait for travis before merging :D 

```
  2) Failure:
TestIntegration#test_restart_restores_environment [/home/travis/build/puma/puma/test/test_integration.rb:254]:
Expected # encoding: ASCII-8BIT
"Hello RAND 0.8985455927826598" to not be equal to # encoding: ASCII-8BIT
"Hello RAND 0.8985455927826598".
```

grosser:
I'll look into that one next ...

nateberkopec:
I did! that was failing before you, but if you want to fix that too I won't complain :D

grosser:
cannot fail before me since that's a test I added ;)

grosser:
here: https://github.com/puma/puma/pull/1305/files

PR #1305, opened 2017-05-30
fix jruby deleting env vars

grosser:
@nateberkopec

### DIFF
--- a/test/test_integration.rb
+++ b/test/test_integration.rb
@@ -239,6 +239,13 @@ class TestIntegration < Minitest::Test
 
   # It does not share environments between multiple generations, which would break Dotenv
   def test_restart_restores_environment
+    if Puma.jruby?
+      # jruby has a bug where setting `nil` into the ENV or `delete` do not change the
+      # next workers ENV
+      assert true
+      return
+    end
+
     server("-q test/rackup/hello-env.ru")
 
     s = connect


grosser:
hmm no I think this is even worse ... setting to "" will prevent things like Dotenv to load new env vars ... I'll have another look maybe just skip for now since it keeps the old behavior ...

nateberkopec:
That's annoying :( [It looks like it *should* work, maybe we should make a report.](https://github.com/jruby/jruby/blob/f7746d0be94f5d9f7cd27af610f1bc5dd2622794/spec/ruby/core/env/replace_spec.rb#L10)

grosser:
updated with a safer fixup ... better semi-broken then completely broken ...
otherwise we'd set `""` everywhere and things sure would break ...
also once jruby is fixed it will just work ...

grosser:
basically it will behave as before without this fix ... so not a degradation/blocker ... and once jruby is fixed it will work just fine

nateberkopec:
Works for me. I love how all the stars aligned and we didnt see any ioerrors on the CRuby parts of this built.

grosser:
tried reproducing it but without success ... so there must be something weird going on :/

```
ruby -e 'ENV["FOO"] = "BAR"; ENV.replace({}); exec "ruby -e \"puts ENV[%{FOO}]\""'
```

Issue #1308, opened 2017-06-01
Restarting via `SIGUSR2` broken in 3.9.0 works in 3.8.2


dguettler:
### Steps to reproduce

1) Updated application from puma 3.8.2 to 3.9.0

2) start puma server `bundle exec puma -p 3000`

3) send USR2 `kill -SIGUSR2 <pid>`

### Expected behavior

Puma server should be restarting and does so correctly in 3.8.2

### Actual behavior

With version 3.9.0 the outcome is
```
* Listening on tcp://0.0.0.0:3000
Use Ctrl-C to stop
* Restarting...
/usr/local/lib/ruby/2.3.0/rubygems/dependency.rb:319:in `to_specs': Could not find 'puma' (>= 0.a) among 19 total gem(s) (Gem::LoadError)
```

### System configuration

**Ruby version**: ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-linux]
**Rails version**: rails-5.0.3

nateberkopec:
pinging @grosser who recently touched this

grosser:
hmm I think this is because it restarts with an environment that does not have bundler
so should tweak the restart command to include bundle exec ... I though I tried that ... for now might be able to fix it by setting `--restart-command` manually will take a look shortly

grosser:
it works when running via `rails s` which hid the error ...
@restart_argv is `"/rbenv/versions/2.3.3/bin/ruby", "ruby/2.3.0/bin/puma", "-p", "3000", {:close_others=>true, 14=>14}]` ... so no mention of bundler ... so it won't load bundler ... so it cannot find puma ... 

the whole `@restart_command` is basically wrong and bundler used to override it by accident because it set a bunch of stuff in the ENV that was then inherited ... good times ...

grosser:
fixing here https://github.com/puma/puma/pull/1309

nateberkopec:
@grosser how could I repro this with a rails app? rails s, bin/puma, bundle exec puma, puma all work for me.

grosser:
`bundle exec puma` did not work for me

grosser:
the trick is to `gem uninstall puma -a` before doing it and having puma only installed via `bundle install --path vendor/bundle`

nateberkopec:
Still nothing.

grosser:
add gemspec to Gemfile and remove the dependencies like https://github.com/puma/puma/pull/1313 does then

```
gem uninstall puma -a
[...]

PR #1309, opened 2017-06-01
reinject RUBYOPT that bundler adds since we reset the environment bef…

grosser:
…ore restarting

we remove all bundler ENV state before restating since it includes PATH to gems, Gemfile location, the version of bundler ... all things that should be able to change when an app is redeployed

bundler adds `-rbundler/setup` to RUBYOPT but we remove that when restarting since we restore to pre-bundler ENV ... but we still want to restart with bundler

possible fallout:
 - `bundle _1.14.0_ puma` not being restarted automatically
 - `RUBYOPT` that existed before bundler get applied twice `ruby -rfoo -rbundler/setup puma` would require foo twice ... or whatever other RUBYOPT was given

fixes https://github.com/puma/puma/issues/1308

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,7 @@
 source "https://rubygems.org"
 
+gemspec
+
 gem "hoe"
 gem "hoe-git"
 gem "hoe-ignore"
--- a/Rakefile
+++ b/Rakefile
@@ -8,6 +8,7 @@ IS_JRUBY = defined?(RUBY_ENGINE) ? RUBY_ENGINE == "jruby" : false
 Hoe.plugin :git
 Hoe.plugin :ignore
 
+# Keep in sync with puma/gemspec
 HOE = Hoe.spec "puma" do
   self.readme_file    = "README.md"
   self.urls = %w!http://puma.io https://github.com/puma/puma!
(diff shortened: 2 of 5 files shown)


grosser:
@nateberkopec a bit of cleanup too ... hope you like it :D

grosser:
nvm ... got to rework that ... we cannot run tests without bundler or they will randomly fail :/

grosser:
getting better still need to iron out a few more kinks in the testing ... so many bugs in there that just worked by accident :D

grosser:
in summary:
 - make tests simpler ... they use `bundle exec` anyway since we run with bundler
 - ensure no other puma versions are around so we actually fail as described in the issue
 - reinject bundler since we remove it from the ENV by using `with_clean_env`

grosser:
extracted the basic logic into https://github.com/puma/puma/pull/1311 let's merge that first for a clean PR ... this here is getting messy

nateberkopec:
> ensure no other puma versions are around so we actually fail as described in the issue

CI is failing on this new check.

grosser:
rebased with my 2 other PRs (gemspec and test cleanup) ... so should merge them first ... I'll try to get em both green ... hopefully just randomness

grosser:
only change here is now the RUBYOPT re-injecting and using `bundle exec puma` to test ... but the injection is no longer needed since https://github.com/puma/puma/pull/1314 added all bundler variables back into the ENV ... so RUBYOPT is still present

could just keep it this way and release a patch version now ... but would mean Gemfile and bundler version will be stale during restarts (as in v38)

alternatively re-implement .clean_env and whitelist RUBYOPT to make Gemfile/bundler changes possible

grosser:
asking here to keep .clean_env around since idk how else to implement this feature without copy-pasting code from bundler https://github.com/bundler/bundler/pull/4702

grosser:
that will surely take a while to resolve ... so maybe release the patch version now and we can improve the env reset in the next minor ...

BrandonMathis:
Nice work on this fix!

I would +1 the suggestion to do a patch level release. Not being able to restart seems like quite a major defect for puma 3.9

grosser:
Did you find a bug that is fixed by using master ?

PR #1311, opened 2017-06-01
simplify test shellout and make sure no other version of puma is available so we know our tests work

grosser:
@nateberkopec 
/cc @schneems

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -143,6 +143,16 @@ else
   task :test => [:compile]
 end
 
+task :test => [:ensure_no_puma_gem]
+task :ensure_no_puma_gem do
+  Bundler.with_clean_env do
+    out = `gem list puma`.strip
+    if !$?.success? || out != ""
+      abort "No other puma version should be install to avoid false positives or lading it by accident but found #{out}"
+    end
+  end
+end
+
 namespace :test do
   desc "Run the integration tests"
   task :integration do
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -340,8 +340,6 @@ module Puma
 
       @restart_dir ||= Dir.pwd
 
-      require 'rubygems'
-
       # if $0 is a file in the current directory, then restart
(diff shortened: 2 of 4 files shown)


grosser:
updated ... realized I cannot remove bundler since then the gem would not be found (we already run the shellout with bundler since RUBYOPT is set) ... and switching to `bundle exec` does not work without the patch from https://github.com/puma/puma/pull/1309 ... so for now just some cleanup to make the other PRs cleaner

nateberkopec:
@grosser JRuby 9000 should have a green CI, it doesn't: https://travis-ci.org/puma/puma/jobs/238616761

grosser:
error seems to be:
```
  2) Error:
TestIntegration#test_restart_closes_keepalive_sockets:
#<Class:0x3cfe95b1>: execution expired
    org/jruby/RubyIO.java:2233:in `gets'
    /home/travis/build/puma/puma/test/test_integration.rb:93:in `wait_for_server_to_boot'
    /home/travis/build/puma/puma/test/test_integration.rb:82:in `restart_server'
    /home/travis/build/puma/puma/test/test_integration.rb:56:in `restart_server_and_listen'
    /home/travis/build/puma/puma/test/test_integration.rb:208:in `test_restart_closes_keepalive_sockets'
```
already restarted it twice ... reproduced it and turns out the `-Ilib` re-injection did not work on jruby ... so another bug fixed :D

grosser:
added more time to the timeout since it kept failing and on master the test took 59s ... on this branch it took 64 ... so I hope just randomness but now jruby has a bit more time ...

```
time ruby test/test_integration.rb -n 'TestIntegration#test_restart_closes_keepalive_sockets'
```

grosser:
somehow `TestThreadPool#test_trim_is_ignored_if_no_waiting_threads` failed on ruby 2.1 now ... but should be unrealted ...

grosser:
2.1 and jruby are green 🎉

PR #1313, opened 2017-06-01
simplify stub gemspec and remove invalid dev dependencies ... allow r…

grosser:
…unning bundle exec puma locally

@nateberkopec 
/cc @schneems

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -8,6 +8,7 @@ IS_JRUBY = defined?(RUBY_ENGINE) ? RUBY_ENGINE == "jruby" : false
 Hoe.plugin :git
 Hoe.plugin :ignore
 
+# Keep in sync with puma.gemspec
 HOE = Hoe.spec "puma" do
   self.readme_file    = "README.md"
   self.urls = %w!http://puma.io https://github.com/puma/puma!
@@ -17,13 +18,9 @@ HOE = Hoe.spec "puma" do
 
   spec_extras[:extensions]  = ["ext/puma_http11/extconf.rb"]
   spec_extras[:executables] = ['puma', 'pumactl']
-  spec_extras[:homepage] = self.urls.first
+  spec_extras[:homepage] = urls.first
 
   require_ruby_version ">= 1.9.3"
-
-  dependency "rack", [">= 1.1", "< 3.0"], :development
-
-  extra_dev_deps << ["rake-compiler", "~> 0.8"]
 end
 
 task :prerelease => [:clobber, :check_manifest, :test]
--- a/puma.gemspec
+++ b/puma.gemspec
@@ -1,52 +1,20 @@
 # -*- encoding: utf-8 -*-
 
-# This is only used when puma is a git dep from Bundler, so it's a little
-# weird.
+# This is only used when puma is a git dep from Bundler, keep in sync with Rakefile
 
-d = File.read(File.expand_path("../lib/puma/const.rb", __FILE__))
-if d =~ /VERSION = "(\d+\.\d+\.\d+)"/
-  version = $1
-else
-  version = "0.0.1"
-end
+version = File.read(File.expand_path("../lib/puma/const.rb", __FILE__))[/VERSION = "(\d+\.\d+\.\d+)"/, 1] || raise
 
 Gem::Specification.new do |s|
   s.name = "puma"
   s.version = version
-
-  s.required_rubygems_version = Gem::Requirement.new(">= 0") if s.respond_to? :required_rubygems_version=
   s.authors = ["Evan Phoenix"]
-  s.date = `git log --pretty="%ai" -n 1`.split(" ").first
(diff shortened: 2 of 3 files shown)


Rakefile
```diff
@@ -17,13 +18,9 @@ HOE = Hoe.spec "puma" do
...
   spec_extras[:extensions]  = ["ext/puma_http11/extconf.rb"]
   spec_extras[:executables] = ['puma', 'pumactl']
-  spec_extras[:homepage] = self.urls.first
+  spec_extras[:homepage] = urls.first
 
   require_ruby_version ">= 1.9.3"
-
-  dependency "rack", [">= 1.1", "< 3.0"], :development
-
-  extra_dev_deps << ["rake-compiler", "~> 0.8"]
```
grosser: these do not match what is in the Gemfile and used locally or travis ... useless anyway so just removing them ...

puma.gemspec
```diff
@@ -1,52 +1,19 @@
...
-# weird.
+# This is only used when puma is a git dep from Bundler, keep in sync with Rakefile
 
-d = File.read(File.expand_path("../lib/puma/const.rb", __FILE__))
-if d =~ /VERSION = "(\d+\.\d+\.\d+)"/
-  version = $1
-else
-  version = "0.0.1"
-end
+version = File.read(File.expand_path("../lib/puma/const.rb", __FILE__))[/VERSION = "(\d+\.\d+\.\d+)"/, 1] || raise
```
grosser: if this fails then something is wrong ... better fail and fix instead of having weird versions

puma.gemspec
```diff
@@ -1,52 +1,19 @@
...
-    else
-      s.add_dependency(%q<rdoc>, ["~> 4.0"])
-      s.add_dependency(%q<rake-compiler>, ["~> 0.8.0"])
-      s.add_dependency(%q<hoe>, ["~> 3.6"])
-    end
-  else
-    s.add_dependency(%q<rdoc>, ["~> 4.0"])
-    s.add_dependency(%q<rake-compiler>, ["~> 0.8.0"])
-    s.add_dependency(%q<hoe>, ["~> 3.6"])
-  end
```
grosser: don't need any of that to work locally or via git dependency in bundler

puma.gemspec
```diff
@@ -1,52 +1,19 @@
...
   s.description = "Puma is a simple, fast, threaded, and highly concurrent HTTP 1.1 server for Ruby/Rack applications. Puma is intended for use in both development and production environments. It's great for highly concurrent Ruby implementations such as Rubinius and JRuby as well as as providing process worker support to support CRuby well."
+  s.summary = "Puma is a simple, fast, threaded, and highly concurrent HTTP 1.1 server for Ruby/Rack applications"
   s.email = ["evan@phx.io"]
   s.executables = ["puma", "pumactl"]
   s.extensions = ["ext/puma_http11/extconf.rb"]
   s.files = `git ls-files`.split($/)
   s.homepage = "http://puma.io"
   s.license = "BSD-3-Clause"
-  s.rdoc_options = ["--main", "README.md"]
-  s.require_paths = ["lib"]
```
nateberkopec: isn't this a required attribute?

puma.gemspec
```diff
@@ -1,52 +1,19 @@
...
-  s.date = `git log --pretty="%ai" -n 1`.split(" ").first
   s.description = "Puma is a simple, fast, threaded, and highly concurrent HTTP 1.1 server for Ruby/Rack applications. Puma is intended for use in both development and production environments. It's great for highly concurrent Ruby implementations such as Rubinius and JRuby as well as as providing process worker support to support CRuby well."
+  s.summary = "Puma is a simple, fast, threaded, and highly concurrent HTTP 1.1 server for Ruby/Rack applications"
   s.email = ["evan@phx.io"]
   s.executables = ["puma", "pumactl"]
   s.extensions = ["ext/puma_http11/extconf.rb"]
   s.files = `git ls-files`.split($/)
   s.homepage = "http://puma.io"
   s.license = "BSD-3-Clause"
-  s.rdoc_options = ["--main", "README.md"]
```
nateberkopec: why remove?

puma.gemspec
```diff
@@ -1,52 +1,19 @@
...
-  s.require_paths = ["lib"]
-  s.required_ruby_version = Gem::Requirement.new(">= 1.9.3")
```
nateberkopec: why remove?

grosser:
not required.
removing all the things that are unnecessary

grosser:
please let me know which of the removed options is needed to run tests/use as git dependency ... or why else we'd need them would like to make this hacky gemspec as simple as possible

puma.gemspec
nateberkopec: from what I can tell removing this would change the default output for `gem install --document puma`.

puma.gemspec
nateberkopec: looked it up, this is the default value, so 👍

puma.gemspec
nateberkopec: removing this would remove the require ruby version check from bundler AFAICT, I don't want to do that.

grosser:
this is not the real gemspec, just a stub ... so nobody will use it for gem
install

grosser:
restored ruby requirement

grosser:
@nateberkopec good to go ?

Issue #1312, opened 2017-06-01
/usr/local/lib/ruby/gems/2.2.0/gems/puma-3.9.0/lib/puma/launcher.rb:166:in `run': undefined method `clean_env' for Bundler:Module (NoMethodError)


wangyuan99:
### Steps to reproduce

1) ...rails new myapp

2) ...cd myapp

3) ...rails server

### Expected behavior

should see:

=> Booting Puma
=> Rails 5.0.3 application starting in development on http://localhost:3000
=> Run `rails server -h` for more startup options
Puma starting in single mode...
* Version 3.8.2 (ruby 2.2.3-p173), codename: Sassy Salamander
* Min threads: 5, max threads: 5
* Environment: development
* Listening on tcp://0.0.0.0:3000
Use Ctrl-C to stop

### Actual behavior

got this:

=> Booting Puma
=> Rails 5.0.3 application starting in development on http://localhost:3000
=> Run `rails server -h` for more startup options
Exiting
/usr/local/lib/ruby/gems/2.2.0/gems/puma-3.9.0/lib/puma/launcher.rb:166:in `run': undefined method `clean_env' for Bundler:Module (NoMethodError)

### System configuration

**Ruby version**: 2.2.3
**Rails version**: 5.0.2

wangyuan99:
I was creating new rails applications, and start rails server, everything was working. Then I started creating new applications, then when i start rails server, I got that error message. All new application I created afterward, got that error message. Then  I tried copy the previous working application, the copied application works fine too.

nateberkopec:
Run `gem install bundler`, make sure `bundler -v` says 1.15.0, then try again.

nateberkopec:
What's the best way for us to solve this? You can't do optional dependencies, so I guess we just have to have a code path for "bundler is defined but is old"

matobinder:
I was too about to submit an issue on this, I'm stuck for a bit since I'm building apps in CloudFoundry and the ruby-buildpacks don't yet have the latest version of bundler.

Note I was just looking at bundler source, not that it would probably fix having an older version of bundler, but it looks like Bundler.clean_env is deprecated
https://github.com/bundler/bundler/blob/master/lib/bundler.rb#L257

grosser:
https://github.com/puma/puma/pull/1314

nateberkopec:
@wangyuan99 @matobinder Can either of you try again with Puma master and tell me if that fixes the problem for you? 

```ruby
[...]

PR #1314, opened 2017-06-02
support older and future versions of bundler that do not have clean_env

grosser:
fixes https://github.com/puma/puma/issues/1312

### DIFF
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -163,7 +163,7 @@ module Puma
 
     # Run the server. This blocks until the server is stopped
     def run
-      previous_env = (defined?(Bundler) ? Bundler.clean_env : ENV.to_h)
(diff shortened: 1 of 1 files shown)


grosser:
@nateberkopec @matobinder

lib/puma/launcher.rb
```diff
@@ -163,7 +163,7 @@ def phased_restart
 
     # Run the server. This blocks until the server is stopped
     def run
-      previous_env = (defined?(Bundler) ? Bundler.clean_env : ENV.to_h)
+      previous_env = (defined?(Bundler) ? Bundler::ORIGINAL_ENV : ENV.to_h)
```
nateberkopec: Why not `.original_env`?

nateberkopec: Ah I see, nvm. Should we be `clone`ing this like Bundler does though?

grosser:
we don't modify it, so no

grosser:
just being afraid that's not defined again on some version

PR #1315, opened 2017-06-02
if we use a symlinked folder restarting the app needs to pick up the …

grosser:
…new Gemfile

... had .clean_env for a reason ... need to bring this back to fix ... will add a test soon, just heads up that we still need 1 more fix

reproduction:
 - create a a folder
 - symlink to new location x
 - boot puma from x
 - swap x with new location y
 - SIGUSR2 puma

expected: puma picks up new Gemfile in y
actual: puma reloads Gemfile from x (possibly even deleted now)

@nateberkopec

### DIFF
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -163,7 +163,12 @@ module Puma
 
     # Run the server. This blocks until the server is stopped
     def run
-      previous_env = (defined?(Bundler) ? Bundler::ORIGINAL_ENV : ENV.to_h)
+      if defined?(Bundler)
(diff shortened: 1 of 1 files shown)


nateberkopec:
@grosser makes sense to me, waiting on the test.

nateberkopec:
@grosser Shouldn't we be replacing `with_clean_env` in prune_bundler as well?

grosser:
yeah, I am planing to turn this PR into a `cleanup bundler` ... just wanted to collect some proof first that we actually need to do that ... so playing with multiple gemfiles and seeing if not cleaning up all of bundler is good enough or if then load-paths etc stay around ... also still waiting for the bundler issue to either get resolved or clean_env to be restored

PR #1317, opened 2017-06-03
run tests with bundler since that is what our users run

grosser:
@nateberkopec would have caught https://github.com/puma/puma/issues/1308 early

does not change a lot since we previously ran the test with `bundle exec` and so already had all the load-paths etc setup with bundler

### DIFF
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -163,7 +163,16 @@ module Puma
 
     # Run the server. This blocks until the server is stopped
     def run
-      previous_env = (defined?(Bundler) ? Bundler::ORIGINAL_ENV : ENV.to_h)
+      previous_env =
+        if defined?(Bundler)
+          env = Bundler::ORIGINAL_ENV
+          # add -rbundler/setup so we load from Gemfile when restarting
+          bundle = "-rbundler/setup"
+          env["RUBYOPT"] = [env["RUBYOPT"], bundle].join(" ") unless env["RUBYOPT"].include?(bundle)
+          env
+        else
+          ENV.to_h
+        end
 
       @config.clamp
 
@@ -225,8 +234,8 @@ module Puma
       else
         redirects = {:close_others => true}
         @binder.listeners.each_with_index do |(l, io), i|
-        ENV["PUMA_INHERIT_#{i}"] = "#{io.to_i}:#{l}"
-        redirects[io.to_i] = io.to_i
+          ENV["PUMA_INHERIT_#{i}"] = "#{io.to_i}:#{l}"
+          redirects[io.to_i] = io.to_i
         end
 
         argv = restart_args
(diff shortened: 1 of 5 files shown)


nateberkopec:
This seems to reliably cause the Ruby 2.1 integration tests to fail.

grosser:
test takes 43s locally ... that seems a bit crazy ... will take a look why ...

grosser:
works!

grosser:
wrong issue :D

grosser:
runs fast locally now .... still times out on travis :(

grosser:
reproduction steps with bundler 1.15.0+
```
gem uninstall puma -a
bundle exec puma test/rackup/hello.ru
ps -ef | grep puma
kill -SIGUSR2 <puma-pid>
```

grosser:
idk what's going on in travis, but it's cancelling all builds ...

grosser:
I guess it cancels all builds when 1 fails ... which won't work here since we have undocumented  know failures ...

grosser:
found the ruby 2.1 bug ... the Gemfile had no gemspec directive :D 
... having a look at jruby next

grosser:
green baby :D

grosser:
only issue left is now the handling of non-standard Gemfile paths ... I"ll address that in another PR

lib/puma/launcher.rb
```diff
@@ -163,7 +163,15 @@ def phased_restart
 
     # Run the server. This blocks until the server is stopped
     def run
-      previous_env = (defined?(Bundler) ? Bundler::ORIGINAL_ENV : ENV.to_h)
+      previous_env =
+        if defined?(Bundler)
+          env = Bundler::ORIGINAL_ENV
+          # add -rbundler/setup so we load from Gemfile when restarting
+          env["RUBYOPT"] = (env["RUBYOPT"].to_s.split(" ") << "-rbundler/setup").uniq.compact.join(" ")
```
nateberkopec: Hmmm... are there scenarios where people's Rubyopt will get mangled by this?

grosser:
Don't know how ... Maybe could duplicate the required but that would not
matter ... maybe if you load puma from a gem and then bundlerBut after ...
But would not be defined at this point then

grosser:
@nateberkopec good to go ?

lib/puma/launcher.rb
nateberkopec: This still makes me nervous. Why aren't we just appending this?

nateberkopec:
I'm fine with the test changes but mangling RUBYOPT just seems like a recipe for disaster.

grosser:
isn't just appending more dangerous ... could lead to duplicates
this way it's added unless it's there ... could rewrite to be more ifs and
less split/push/join style ... yeah the uniq could do something weird if
there are multiple -r in there ...

grosser:
updated to a safer approach

nateberkopec:
Much better, thx.

Issue #1022, opened 2016-07-20
/etc/hosts can prevent puma from starting


AKovtunov:
After some coding got this error on running `rails s`:

```
puma-3.5.0/lib/puma/binder.rb:268:in `initialize': Address already in use - bind(2) for "127.0.0.1" port 3000 (Errno::EADDRINUSE)
```

My environment is:

```
$ rails -v       
Rails 5.0.0
$ ruby -v
ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-linux]
```

I've tried:
1. Create a new project - the same
2. Checked Rails 4.2 - problem **solved**
3. Reinstall rails 5 and ruby - the same problem
4. `lsof -wni tcp:3000` returns me nothing
5. ps aux | grep "rails" - nothing
6. ps aux | grep "puma" - nothing
7. ps aux | grep "ruby" -nothing
8. Use `puma` instead of `rails s` - problem **solved** 
9. Use `RAILS_ENV=production bundle exec rails s` - problem **solved**
10. Use `rails s -p 3001` - same problem, for other ports too

Monkey-patched https://github.com/puma/puma/blob/master/lib/puma/binder.rb and added pry.
Found that TCPServer.new is called few times with 127.0.0.1:3000

```
[3] pry(#<Puma::Binder>)> TCPSocket.gethostbyname "localhost"
=> ["localhost",
 [my hosts?],
 2,
 "127.0.0.1",
 "127.0.0.1",
 "127.0.0.1",
 "127.0.0.1",
 "127.0.0.1"]
```

I think this commit broke something for my machine https://github.com/puma/puma/commit/9b1de100044a45a810927c1da0ca3a4dea413890#diff-4fd9276d9f86d7f4827e6ca4ab6b82b8
And I think this is related to the https://github.com/puma/puma/issues/782

AKovtunov:
Downgrading to 3.4.0 works as charm

evanphx:
Oh geez. Ok, I’ll fix this asap and have 3.5.1 out.

evanphx:
@AKovtunov i'm curious, what does your `/etc/hosts` file look like? I'm wondering why `localhost` returned `127.0.0.1` multiple times..

AKovtunov:
After that I was wondering the same too and cleared a bit my hosts file.
I was working on a project with multi-services architecture and needed to bind some routes to 127.0.0.1 :) 

But the error is still here, here is my basic hosts that were installed by ubuntu

```
[...]

Issue #782, opened 2015-09-11
Unable to connect through 127.0.0.1 when bound to localhost


Tonkpils:
I'm not sure if this issue has been raised and I can't find any reference to it here but stemming from: https://github.com/rails/rails/issues/19815#issuecomment-111159194

Using `rails s` on WebRick allows using `localhost` and `127.0.0.1` to connect but when switching to `puma` this behavior changes and does not allow `127.0.0.1`

Is this intended behavior for Puma or would you be open for a fix?

ArturT:
I had this problem and I was looking for solution. I'm not sure what is exactly the reason of this behaviour but here are things I found out and solution to fix this in my app. 

When I start `rails server` then puma listens on localhost:3000

```
$ rails server
=> Booting Puma
=> Rails 4.2.3 application starting in development on http://localhost:3000
=> Run `rails server -h` for more startup options
=> Ctrl-C to shutdown server
Puma 2.13.4 starting...
* Min threads: 0, max threads: 16
* Environment: development
* Listening on tcp://localhost:3000
```
and http://127.0.01:3000 doesn't work.


When I start puma explicitly without config file then it listens on  tcp://0.0.0.0:3000 and http://127.0.01:3000 does work.

Read more https://github.com/puma/puma#configuration-file
```
$ puma -C '-'
Puma starting in single mode...
* Version 2.13.4 (ruby 2.2.3-p173), codename: A Midsummer Code's Dream
* Min threads: 0, max threads: 16
* Environment: development
* Listening on tcp://0.0.0.0:3000
```

When I created my custom config file then it works as well
`bundle exec puma -C config/puma.rb`
```ruby
# config/puma.rb
workers Integer(ENV['WEB_CONCURRENCY'] || 2)
threads_count = Integer(ENV['MAX_THREADS'] || 5)
threads threads_count, threads_count

preload_app!

rackup      DefaultRackup
port        ENV['PORT']     || 3000
environment ENV['RACK_ENV'] || 'development'

on_worker_boot do
  # Worker specific setup for Rails 4.1+
  # See: https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#on-worker-boot
  ActiveRecord::Base.establish_connection
end
```

Tonkpils:
[...]

PR #1318, opened 2017-06-03
Use Socket.ip_address_list to get loopback addresses

### LINKED ISSUE #1167: gethostbyname "localhost" is not a reliable solution for localhost addresses
### Steps to reproduce

Related [`jruby` issue](https://github.com/jruby/jruby/issues/4333).

```
class Server < Sinatra::Base
  configure do
    set :server, :puma
    set :bind, "localhost"
  end
end
```

1. `rvm use jruby-9.1.6.0`

2. Try to start `sinatra` app with jruby.

### Expected behavior

Server will be accessible on both `127.0.0.1` and `::1`.

### Actual behavior

Server is accessible only on `127.0.0.1`

### System configuration

**Ruby version**: jruby 9.1.6.0 (2.3.1) 2016-11-09 0150a76 OpenJDK 64-Bit Server VM 25.101-b13 on 1.8.0_101-b13 +jit [linux-x86_64]

### Fix

You are using `TCPSocket.gethostbyname "localhost"` [here](https://github.com/puma/puma/blob/master/lib/puma/binder.rb#L248) to get the list of localhost addresses.

In MRI ruby we have a [comment](https://github.com/ruby/ruby/blob/4043565023531d173478c216e2a9cc6228f44a7e/ext/socket/socket.c#L882):

```
 * Note that it is not guaranteed to be able to convert to IP address using gethostbyname, getaddrinfo, etc.
 * If you need local IP address, use Socket.ip_address_list.
```

We can use something like:
```
Socket.ip_address_list.select do |address|
  if address.ipv4?
    not address.ipv4_multicast? and not address.ipv4_private?
  else
    not address.ipv6_linklocal?
  end
end
.map do |address|
  address.ip_address
end
```

Thank you.

nateberkopec:
This will fix #1167 

[See JRuby bug](https://github.com/jruby/jruby/issues/4333)

Need some guidance from @evanphx here: is binder supposed to bind to *only loopback* interfaces by default (as I've done in this PR) or is it supposed to bind to *linklocal addresses* as well?

Here's what this method returned before (my machine, CRuby):

```
["::1", "fe80::1%lo0", "127.0.0.1"]
```

Now:

```
["127.0.0.1", "::1"]
```

Note the omission of an ipv6 link-local interface. The reason I removed it was that there are several other link-local interfaces on my machine which I don't think puma should bind to: 

```
irb(main):013:0> Socket.ip_address_list.select do |addrinfo|
irb(main):014:1* addrinfo.ipv6_linklocal?
irb(main):015:1> end
=> [#<Addrinfo: fe80::1%lo0>, #<Addrinfo: fe80::4b:c017:51b5:cd33%en0>, #<Addrinfo: fe80::989b:30ff:fe14:9e01%awdl0>, #<Addrinfo: fe80::f16e:6368:16b5:ff60%utun0>]
```

@evanphx if you can clarify the intent here, I'll add a test.

### DIFF
--- a/lib/puma/binder.rb
+++ b/lib/puma/binder.rb
@@ -245,9 +245,10 @@ module Puma
       end
     end
 
-    def localhost_addresses
-      addrs = TCPSocket.gethostbyname "localhost"
-      addrs[3..-1].uniq
+    def loopback_addresses
+      Socket.ip_address_list.select do |addrinfo|
+        addrinfo.ipv6_loopback? || addrinfo.ipv4_loopback?
+      end.map { |addrinfo| addrinfo.ip_address }.uniq
     end
 
     # Tell the server to listen on host +host+, port +port+.
@@ -259,7 +260,7 @@ module Puma
     #
     def add_tcp_listener(host, port, optimize_for_latency=true, backlog=1024)
       if host == "localhost"
-        localhost_addresses.each do |addr|
+        loopback_addresses.each do |addr|
           add_tcp_listener addr, port, optimize_for_latency, backlog
         end
         return
@@ -298,7 +299,7 @@ module Puma
       MiniSSL.check
 
       if host == "localhost"
-        localhost_addresses.each do |addr|
+        loopback_addresses.each do |addr|
           add_ssl_listener addr, port, ctx, optimize_for_latency, backlog
         end
         return


nateberkopec:
Ah, I think I get it now. Linking #782.

nateberkopec:
I'm also going to link #1022 here, because the original fix here caused that issue, maybe I can satisfy everyone.

nateberkopec:
Looks like it's in 1.9.3, so no.

nateberkopec:
Also, JRuby returns the same result as CRuby does here, though it prints the full ipv6 address rather than the short ::1.

Issue #1193, opened 2017-01-23
Setting no_token option causes crash


jlarigakis:
### Steps to reproduce

1) Put `{ no_token: true } ` as an option to `activate_control_app` in `puma/config.rb`, for example:
`activate_control_app 'unix:///tmp/pumactl.sock', { no_token: true }`

2) Attempt to start the server with
`pumactl start`

### Expected behavior

Puma should start

### Actual behavior

Puma crashes with the message ```undefined method `rindex' for :none:Symbol```

### System configuration

**Ruby version**: 2.4.0
**Puma version**: 3.6.2

## Notes

This issue can be worked around by using `{ auth_token: '' }`, however the example file suggests the `no_token` syntax, which is broken.

manishval:
@nateberkopec I have started looking at this and it seems to be an issue with `OptionParser` parsing the `:none` value  [being set](https://github.com/puma/puma/blob/992fd0d26fd2111b086da23a0bd6cbcdc9442fc4/lib/puma/dsl.rb#L96) for `control_auth_token`. `OptionsParser` is expecting a string instead of a symbol.

I'm not sure what the intended behavior should be in terms of the `no_token: true` option and the `control_auth_token`. But it looks like we can set the value of `control_auth_token` to an empty string instead of `:none` and it will be taken care of in the [runner](https://github.com/puma/puma/blob/master/lib/puma/runner.rb#L49).

nateberkopec:
@manishval Sounds like you're most of the way there. Open a PR?

PR #1321, opened 2017-06-03
Fix crash when no_token is set in config file #1193

manishval:
This fixes issue #1193 when puma would crash with `undefined method rindex for :none:Symbol` when `{ no_token: true }` is set as an option for `activate_control_app` in `puma/config.rb`.

### DIFF
--- a/lib/puma/runner.rb
+++ b/lib/puma/runner.rb
@@ -46,7 +46,7 @@ module Puma
       app = Puma::App::Status.new @launcher
 
       if token = @options[:control_auth_token]
-        app.auth_token = token unless token.empty? or token == :none
+        app.auth_token = token unless token.empty?
       end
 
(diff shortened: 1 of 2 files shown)


lib/puma/runner.rb
```diff
@@ -46,7 +46,7 @@ def start_control
       app = Puma::App::Status.new @launcher
 
       if token = @options[:control_auth_token]
-        app.auth_token = token unless token.empty? or token == :none
+        app.auth_token = token unless token.empty?
```
nateberkopec: why not "unless token = 'none'" and set the token to 'none'? a little more meaningful I think.

manishval: Good idea

nateberkopec:
Great! Needs a test.

manishval:
Sure. I will do that in a bit.

PR #1322, opened 2017-06-03
make tests fast by using readpartial and not blocking read

grosser:
takes restart tests from 40s -> 0.4s @nateberkopec 
there is 1 20s test left over ... but we don't run that on CI ... and it has a `sleep 20` will take a stab at that later if it bites me ...

### DIFF
--- a/test/test_integration.rb
+++ b/test/test_integration.rb
@@ -23,7 +23,7 @@ class TestIntegration < Minitest::Test
 
   def teardown
     File.unlink @state_path rescue nil
-    File.unlink @bind_path  rescue nil
+    File.unlink @bind_path rescue nil
     File.unlink @control_path rescue nil
 
     @wait.close
@@ -52,9 +52,9 @@ class TestIntegration < Minitest::Test
   def restart_server_and_listen(argv)
     server(argv)
     s = connect
-    initial_reply = s.read
+    initial_reply = read_body(s)
     restart_server(s)
-    [initial_reply, connect.read]
+    [initial_reply, read_body(connect)]
   end
 
   def signal(which)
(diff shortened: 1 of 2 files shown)


test/test_integration.rb
```diff
@@ -69,16 +69,8 @@ def wait_booted
   def restart_server(connection)
     signal :USR2
 
-    sleep 1
-
     connection.write "GET / HTTP/1.1\r\n\r\n" # trigger it to start by sending a new request
 
-    assert_raises Errno::ECONNRESET do
```
nateberkopec: this is weird, what was this even supposed to test?

grosser: idk ... making sure it restarts or so ...

grosser: the whole point of the soft restart is that the connection is not interrupted, so I don't think this made much sense ...

nateberkopec: I guess it's testing that the server *should not* respond within 2 seconds? That's...not really testing whether or not the server restarted lol.

nateberkopec: i guess the restarting is already tested by the wait_for_server_to_boot, because it looks for "Use Ctrl-C to stop" which won't appear if the server didn't restart.

nateberkopec:
New failures on JRuby and 2.1. I can repro the 2.1 failure locally sometimes:

screen shot 2017-06-03 at 5 16 34 pm

grosser:
I was expecting it to fail since it's non-blocking ... will take a look at
it

test/test_integration.rb
grosser: yeah I think this is basically a blocking read with a 2s timeout that then raises a `Errno::ECONNRESET` if it's not done

grosser:
fixed up ... now reading until the body appears ... jruby expired a bunch so I bumped the timeout back to 120 for it ... test says it took 10s now ... so idk why it failed before ... but should be fine either way.

grosser:
I think the issue was `PERSISTENT_TIMEOUT = 20` basically the `read` waited until the session was ended so every connection took 20s

PR #1324, opened 2017-06-04
make restart via pumactl more robust

grosser:
@nateberkopec 

not setting random global state and making variables a bit more limited in scope ... fails when changing 20s -> 19s ... so I'd guess that is why travis had issues ... let's see if this is more stable


grosser:
played more with it and also passes locally with 15 ... so that 20 just happens to  be the same value as the timeout, but is not correlated ... so the travis failures must also be random ... leaving the reafctoring since it's still more readable to me ...

PR #1331, opened 2017-06-13
fix #1214 Puma >= 3.6.1 + SSL + Persistent Connections. Puma Hangs :(

### LINKED ISSUE #1214: Puma >= 3.6.1 + SSL + Persistent Connections. Puma Hangs :(
Since we've upgraded to puma 3.7.0 from 3.6.0 we started to experience a very annoying issue in development. Puma would hang sometimes.

After an intense Friday afternoon we've managed to pin-point the problem.

### Steps to reproduce

1) Run any basic rack app using puma greater than or equal to 3.6.1. With the approriate configuration to run in SSL mode: puma -b 'ssl://0.0.0.0:3000?cert=/path/to/cert/localhost.crt&key=/path/to/key/localhost.key'

2) Request a page from an IRB session using a library with persistent connections like Typhoeus: Typhoeus.get('https://localhost:3000')

3) Press CTRL+C on puma process, puma says: "Gracefully stopping, waiting for requests to finish" but it never stops

4) Run the request again from the IRB console: Typhoeus.get('https://localhost:3000')

5) Puma unstuck itself

I wonder if it has been caused by this commit: https://github.com/puma/puma/commit/df18d9331dd9c07f76859e2605dbfb88e71fd92d 
just because it mentions SSL and it's in 3.6.1 tag.

I hope it helps.

respire:
see #1214 
one possible working fix.
but i'm not sure whether it's a correct solution.

### DIFF
--- a/lib/puma/minissl.rb
+++ b/lib/puma/minissl.rb
@@ -81,34 +81,20 @@ module Puma
         @socket.flush
       end
 
+      def read_and_drop(timeout = 1)
+        return unless IO.select([@socket], nil, nil, timeout)
+        read_nonblock(1024)
+      rescue Errno::EAGAIN
+        # do nothing
+      end
+
       def close
         begin
-          # Try to setup (so that we can then close them) any
-          # partially initialized sockets.
-          while @engine.init?
-            # Don't let this socket hold this loop forever.
-            # If it can't send more packets within 1s, then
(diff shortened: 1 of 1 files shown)


nateberkopec:
Please open PRs against master 👍

PR #1332, opened 2017-06-13
[Reactor] fix socket leaking problem

respire:
when fixing #1214 , i found a possible socket leaking problem in reactor.

### DIFF
--- a/lib/puma/reactor.rb
+++ b/lib/puma/reactor.rb
@@ -60,13 +60,11 @@ module Puma
                 end
               end
             else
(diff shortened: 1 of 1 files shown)


nateberkopec:
Please open all pull requests against master.

respire:
@nateberkopec i used to. but do you know why master's tests keeps failing now? tests even hang in my laptop. actually, that makes me feel like i'm forking a wrong branch so i pick 3.6.2 to state the problem.

PR #1335, opened 2017-06-13
[Reactor] fix socket leaking problem

respire:
if we remove it from the timeout list, we cannot delete that from `sockets' unless client closes this socket.


stereobooster:
Can you try to do testcase for this?

PR #1340, opened 2017-06-19
Fix environment file loader

### LINKED ISSUE #1327: Loading environment file is broken since 3.8.0
Loading the right environment file according to the rack environment is broken (https://github.com/puma/puma#configuration-file).

### Steps to reproduce

1) Make sure to create different `config/puma/production.rb`, `config/puma/development.rb` file (you can put `raise "foo"` in the production one)

2) Run `RACK_ENV=production bundle exec puma start`

3) Puma loads development configuration (or at least doesn't load the RACK_ENV's configuration file)

### Expected behavior

Puma should load the right environment file according to the RACK_ENV variable.

### Actual behavior

Puma isn't loading the environment file, even if it shows the right env in the start message. I think it's related to @schneems 's refactor here https://github.com/puma/puma/pull/1234

It works correctly on puma 3.7.x.

It works correctly when we manually specify the configuration file using `-C` switch.

### System configuration

**Ruby version**: 2.2.2
**Rails version**: 4 & 5

ViliusLuneckas:
This PR should fix #1327 issue. 

Could someone tell me if we need [this](https://github.com/ViliusLuneckas/puma/commit/fb3fc019d2ad68ea73f3e17e08c05257389c865e) kind of backward compatibility or just [CLI test](https://github.com/puma/puma/blob/master/test/test_cli.rb#L202) misuse configuration protocol?

### DIFF
--- a/lib/puma/configuration.rb
+++ b/lib/puma/configuration.rb
@@ -189,6 +189,14 @@ module Puma
     end
 
     def load
+      config_files.each do |f|
+        @file_dsl._load_from(f)
+      end
+      @options
+    end
+
+
+    def config_files
       files = @options.all_of(:config_files)
 
       if files.empty?
@@ -201,10 +209,7 @@ module Puma
         files = []
       end
 
-      files.each do |f|
-        @file_dsl._load_from(f)
-      end
-      @options
+      files
     end
 
     # Call once all configuration (included from rackup files)
--- a/lib/puma/configuration.rb
+++ b/lib/puma/configuration.rb
@@ -189,27 +189,22 @@ module Puma
     end
 
     def load
-      config_files.each do |f|
-        @file_dsl._load_from(f)
-      end
+      config_files.each { |config_file| @file_dsl._load_from(config_file) }
+
       @options
     end
-
 
     def config_files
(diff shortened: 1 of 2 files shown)


stereobooster:
Tests are failing. Does your branch based on latest master?

ViliusLuneckas:
@stereobooster, I forked puma more than a month ago when the master was failing. Merging latest commits from the master, tests should pass now.

olleolleolle:
This looks green and master builds green, are there any blockers to merging this change?

olleolleolle:
What's the relationship of this change to this change which closed that issue which this PR wants to fix?

https://github.com/puma/puma/commit/0584a3345e9d7ecd84494abc6ba7971309b5a338

ViliusLuneckas:
This PR fixes the process which loads puma config file based on `RACK_ENV`. The issue is well written in #1327. Meanwhile 0584a3345e9d7ecd84494abc6ba7971309b5a338 fixes some local problem which might or not be fixed with this PR. To be honest – no idea 😕

olleolleolle:
@nateberkopec Can you re-review this change, now that it's green?

PR #1345, opened 2017-06-27
temporary workaround for https://bugs.ruby-lang.org/issues/13632

NickolasVashchenko:
Okay, here is the promised workaround for https://bugs.ruby-lang.org/issues/13632 until it's backported everywhere. 
The problem with the solution of @respire was that it is actually far from being efficient because it iterates here: https://github.com/respire/puma/pull/3/files#diff-4fd9276d9f86d7f4827e6ca4ab6b82b8R53 until the whole interrupt queue of the thread is emptied. So it may happen, 10, 50 or even 500 times. It's much better to simply clean the whole queue at once in such cases.
So here in the PR there is a workaround with C extension that locates the internal interrupt queue of the thread and then instantly purges it. This code can safely be removed later.
@nateberkopec check this out, please.

### DIFF
--- a/ext/puma_http11/extconf.rb
+++ b/ext/puma_http11/extconf.rb
@@ -12,10 +12,4 @@ unless ENV["DISABLE_SSL"]
   end
 end
 
-BROKEN_VERSIONS = %w(2.2.7 2.3.4 2.4.1)
-if (BROKEN_VERSIONS.include? RUBY_VERSION)
-  $defs << '-DBROKEN_RUBY'
-  $defs << "-DVERSION_#{RUBY_VERSION.gsub('.', '_')}"
-end
-
 create_makefile("puma/puma_http11")
--- a/ext/puma_http11/extconf.rb
+++ b/ext/puma_http11/extconf.rb
@@ -12,4 +12,10 @@ unless ENV["DISABLE_SSL"]
   end
 end
 
+BROKEN_VERSIONS = %w(2.2.7 2.3.4 2.4.1)
+if(BROKEN_VERSIONS.include? RUBY_VERSION)
+  $defs << '-DBROKEN_RUBY'
+  $defs << "-DVERSION_#{RUBY_VERSION.gsub('.', '_')}"
+end
+
 create_makefile("puma/puma_http11")
--- a/ext/puma_http11/extconf.rb
+++ b/ext/puma_http11/extconf.rb
@@ -13,7 +13,7 @@ unless ENV["DISABLE_SSL"]
 end
 
 BROKEN_VERSIONS = %w(2.2.7 2.3.4 2.4.1)
-if(BROKEN_VERSIONS.include? RUBY_VERSION)
+if (BROKEN_VERSIONS.include? RUBY_VERSION)
   $defs << '-DBROKEN_RUBY'
   $defs << "-DVERSION_#{RUBY_VERSION.gsub('.', '_')}"
 end
--- a/ext/puma_http11/puma_http11.c
+++ b/ext/puma_http11/puma_http11.c
@@ -12,6 +12,8 @@
 #include <string.h>
 #include "http11_parser.h"
 
+#include "ruby/config.h"
+
 #ifndef MANAGED_STRINGS
 
 #ifndef RSTRING_PTR
@@ -63,6 +65,62 @@ struct common_field {
   int raw;
 	VALUE value;
 };
+
+enum enum4 {
+    status1,
+    status2,
+    status3,
+    status4
+};
+
+// Temporary workaround for https://bugs.ruby-lang.org/issues/13632
+// Can be removed when the fix is ported to all supported versions
+typedef struct half_thread {
+    void *vmlt_node1;
+    void *vmlt_node2;
+    void *self;
+    void *vm;
+    void *ec1;
+    void *ec2;
+    void *ec3;
+    int safe_level;
+    int raised_flag ;
+    void *last_status;
+    int state;
+    void *passed_block_handler;
+    void *passed_bmethod_me;
+    void *calling;
+    void *top_self;
+    void *top_wrapper;
+    void *root_lep;
+    void *root_svar;
+    void *thread_id;
+#ifdef NON_SCALAR_THREAD_ID
+    char thread_id_string[sizeof(void *) * 2 + 3];
+#endif
+    enum enum4 status;
+    int to_kill;
+    int priority;
+    void *ubf_list1;
+    void *ubf_list2;
+    pthread_cond_t cond;
+#ifdef HAVE_CLOCKID_T
+    clockid_t clockid;
+#endif
+    void *blocking_region_buffer;
+
+    void *thgroup;
+    void *value;
+
+    /* temporary place of errinfo */
+    void *errinfo;
+
+    /* async errinfo queue */
+    void *pending_interrupt_queue;
+    void *pending_interrupt_mask_stack;
+    int pending_interrupt_queue_checked;
+} rb_half_thread_t;
+
 
 /*
  * A list of common HTTP headers we expect to receive.
@@ -468,8 +526,18 @@ VALUE HttpParser_body(VALUE self) {
 void Init_io_buffer(VALUE puma);
 void Init_mini_ssl(VALUE mod);
 
+VALUE
+rb_thread_purge_queue(VALUE thread)
+{
+    rb_half_thread_t *th = (rb_half_thread_t*)DATA_PTR(thread);
+    th->pending_interrupt_queue_checked = 0;
+    rb_ary_clear(th->pending_interrupt_queue);
+    return Qnil;
+}
+
 void Init_puma_http11()
 {
+  rb_define_method(rb_cThread, "purge_interrupt_queue", rb_thread_purge_queue, 0);
 
   VALUE mPuma = rb_define_module("Puma");
   VALUE cHttpParser = rb_define_class_under(mPuma, "HttpParser", rb_cObject);
--- a/ext/puma_http11/puma_http11.c
+++ b/ext/puma_http11/puma_http11.c
@@ -73,12 +73,16 @@ enum enum4 {
     status4
 };
 
-// Temporary workaround for https://bugs.ruby-lang.org/issues/13632
-#if RUBY_VERSION == "2.2.7" || RUBY_VERSION == "2.3.4" || RUBY_VERSION == "2.4.1"
-  #define VERSION_WITH_BUG_13632
+#if defined(_WIN32)
+#include <windows.h>
+typedef HANDLE rb_nativethread_id_t;
+#else
+#include <pthread.h>
+typedef pthread_t rb_nativethread_id_t;
 #endif
 
-#ifdef VERSION_WITH_BUG_13632
+// Temporary workaround for https://bugs.ruby-lang.org/issues/13632
+#ifdef BROKEN_RUBY
 typedef struct half_thread {
     void *vmlt_node1;
     void *vmlt_node2;
@@ -97,7 +101,7 @@ typedef struct half_thread {
     void *passed_ci_or_calling;
     void *top_self;
     void *top_wrapper;
-    #if RUBY_VERSION != "2.4.1"
+    #ifndef VERSION_2_4_1
         void *base_block;
     #endif
     void *root_lep;
@@ -113,7 +117,7 @@ typedef struct half_thread {
     void *interrupt_event;
 #elif defined(HAVE_PTHREAD_H)
     void *ubf_list1_or_signal_thread_list;
-#if RUBY_VERSION != "2.2.7"
+#ifndef VERSION_2_2_7
     void *ubf_list2;
 #endif
     pthread_cond_t cond;
@@ -534,7 +538,7 @@ VALUE HttpParser_body(VALUE self) {
 
 void Init_io_buffer(VALUE puma);
 void Init_mini_ssl(VALUE mod);
-#ifdef VERSION_WITH_BUG_13632
+#ifdef BROKEN_RUBY
 VALUE
 rb_thread_purge_queue(VALUE thread)
 {
@@ -547,7 +551,7 @@ rb_thread_purge_queue(VALUE thread)
 
 void Init_puma_http11()
 {
-#ifdef VERSION_WITH_BUG_13632
+#ifdef BROKEN_RUBY
   rb_define_method(rb_cThread, "purge_interrupt_queue", rb_thread_purge_queue, 0);
 #endif
   VALUE mPuma = rb_define_module("Puma");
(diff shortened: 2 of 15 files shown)


NickolasVashchenko:
I see it fails with 2.3, since it has a bit different thread structure. Gonna be an easy fix.

nateberkopec:
Brilliant! Paging @evanphx.

evanphx:
So I would ask that the code in `rb_thread_purge_queue` have a version guard be added at the top of it, that exits early if the version doesn't need this fix. The reason for this is that you're making puma depend on the exact memory layout of MRI's thread struct and it could easily break in the future and segfault.

So if this was introduced in 2.3 and will be fixed in the release after 2.4, we should only run this code when one of those releases is running.

NickolasVashchenko:
Yeah, of course. I'll scope it to only affected versions when I'll be fixing it for 2.3.

NickolasVashchenko:
The status quo: has been introduced in 2.5.0, 2.4.1, 2.3.4 and 2.2.7, but has been only fixed for 2.5.0. So my fix affects the other three and it's working now. 
I also gonna check it also for windows, and if it's okay there as well, we're good to merge it.
@evanphx I could help you maintaining this code, monitor the status of backporting of the bug and its fix to adjust and remove when the fix backported everywhere. 
Also I can help you with maintaining other C code.

stereobooster:
This fix tests, which are broken in master 👍 . We can try https://www.appveyor.com/ for windows tests (travis supports MacOS).

NickolasVashchenko:
I can add a test for it, but it's gonna be erratic since it's pretty hard to reproduce this issue in lab environment.

NickolasVashchenko:
added a test to cover it. It mirrors the internal Ruby test for the same issue, but it's appropriate there.

test/test_http11.rb
```diff
@@ -183,4 +183,30 @@ def test_horrible_queries
     end
 
   end
+
+  def test_13632_workaround
+    a = []
```
stereobooster: `pipes = []`

test/test_http11.rb
```diff
@@ -183,4 +183,30 @@ def test_horrible_queries
...
+    t = []
```
stereobooster: threads

test/test_http11.rb
```diff
@@ -183,4 +183,30 @@ def test_horrible_queries
...
+    10.times do
+      r,w = IO.pipe
```
stereobooster: `r, w = IO.pipe`

test/test_http11.rb
```diff
@@ -183,4 +183,30 @@ def test_horrible_queries
     end
 
   end
+
+  def test_13632_workaround
```
stereobooster: I suppose you want to skip it if `purge_interrupt_queue` undefined

NickolasVashchenko: No, I don't. The `pending_interrupt?` is not expected to be true for all rubies that are not addressed specifically. We'd better see immediately that it started to happen in this case.

NickolasVashchenko: Improved all other naming

lib/puma/server.rb
```diff
@@ -896,11 +900,16 @@ def notify_safely(message)
       begin
         @notify << message
       rescue IOError
+        Thread.current.purge_interrupt_queue if Thread.current.respond_to? :purge_interrupt_queue
         # The server, in another thread, is shutting down
       rescue RuntimeError => e
         # The server, in another thread, has been shut down during the system call
         # https://github.com/puma/puma/pull/1206
-        raise e unless e.message.include?('IOError')
+        if e.message.include?('IOError')
```
stereobooster: this misleading: I suppose you want to guard against exact case (`RuntimeError: can't modify frozen IOError`) and not any error containing `IOError`

test/test_http11.rb
```diff
@@ -183,4 +183,30 @@ def test_horrible_queries
...
+      threads << Thread.new do
+        while r.gets
+        end rescue IOError
+        if Thread.current.pending_interrupt?
+          Thread.current.purge_interrupt_queue
+          assert !Thread.current.pending_interrupt?
+        end
+      end
+    end
+    pipes.each do |r,w|
```
stereobooster: `|r, w|`

lib/puma/server.rb
NickolasVashchenko: Not quite. I intended to maintain resemblance here with `rescue IOError` clause by not adding excessive words. To be frank, any other RuntimeError with `IOError` in the message is extremely unlikely.

test/test_http11.rb
stereobooster:
```
Error:
Http11ParserTest#test_13632_workaround:
NoMethodError: undefined method `purge_interrupt_queue' for #<Thread:0x0000000002ea02c0>
    /home/travis/build/puma/puma/test/test_http11.rb:197:in `block (2 levels) in test_13632_workaround'
```

lib/puma/server.rb
stereobooster: This reminds me old story: admin setup script to monitor server log, if there was fatal error in logs it restarted server. Later one user write to support complaining that every time one tries to login server was down. Username was "femme fatale"

test/test_http11.rb
NickolasVashchenko: It's because the version of 2.5.0 on CI is very old and does not include the fix. Please, disregard.

stereobooster: not sure I understand this situation. Is `purge_interrupt_queue` defined for any other ruby version? This test doesn't make sense if method undefined

lib/puma/server.rb
NickolasVashchenko: I'm calling a vanilla Ruby method, representing a syscall, so I don't not really expect them to be over-creative here :)However, I'm not insisting on keeping here a succint exception message, let's hear what other people say on that.

stereobooster: Just to be sure: I'm proposing to check against exact error `e.message == "can't modify frozen IOError"`

NickolasVashchenko: Got ya :) I just feel it'll be a bit uglier with no real purpose, actually. The `<<` method of IO throws only IOErrors of the StandardError by it's API, so just the`rescue RuntimeError` would be more than enough here. What we have there with `e.message.include?('IOError')` is already over-cautious and safe.

test/test_http11.rb
```diff
@@ -183,4 +183,30 @@ def test_horrible_queries
...
+    pipes = []
+    threads = []
+    10.times do
+      r, w = IO.pipe
+      pipes << [r, w]
+      threads << Thread.new do
+        while r.gets
+        end rescue IOError
+        if Thread.current.pending_interrupt?
+          Thread.current.purge_interrupt_queue
```
stereobooster: skip test for undefined `Thread.current.purge_interrupt_queue` or change to:
```ruby
```

NickolasVashchenko: That would not be right, because of the reasons I've described above. However, I've improved a bit reporting there, to show a clear message in this case.

test/test_http11.rb
```diff
@@ -183,4 +183,31 @@ def test_horrible_queries
...
+    pipes = []
+    threads = []
+    10.times do
+      r, w = IO.pipe
+      pipes << [r, w]
+      threads << Thread.new do
+        while r.gets
+        end rescue IOError
+        if Thread.current.pending_interrupt?
+          raise "Ruby bug 13632 detected for version #{RUBY_VERSION}, please open a bugreport for Puma." unless Thread.current.respond_to? :purge_interrupt_queue
```
stereobooster: technically speaking `Thread.current.respond_to? :purge_interrupt_queue` is defined only for `%w(2.2.7 2.3.4 2.4.1)`, so this test should fail for all other ruby versions. What is the sense?

test/test_http11.rb
NickolasVashchenko: If it's not defined and the bug is happened, than the test is supposed to fail. It's its purpose.

test/test_http11.rb
stereobooster: I finally got it. Disregard my previous comments

lib/puma/server.rb
stereobooster: > with no real purpose

I read the source code and my first thought was: why this strange way to detect IOError. Then I checked code above and saw `rescue IOError`. I got totally puzzled. I opened https://github.com/puma/puma/pull/1206, and only there I found that message reads as "can't modify frozen IOError"

stereobooster: I would place comment here like this:

> Workaround for https://bugs.ruby-lang.org/issues/13239

NickolasVashchenko: Okay.

NickolasVashchenko:
Checked out appveyor, it does not have those 3 versions of ruby that are interesting. With other, older ones, the build works, however.

ext/puma_http11/extconf.rb
```diff
@@ -12,4 +12,10 @@
   end
 end
 
+BROKEN_VERSIONS = %w(2.2.7 2.3.4 2.4.1)
```
nateberkopec: Shouldn't we be fixing this for 2.2.0 through 2.2.6, 2.3.0 through 2.3.3 and 2.4.0 as well?

ext/puma_http11/puma_http11.c
```diff
@@ -64,6 +66,75 @@ struct common_field {
...
+    char thread_id_string[sizeof(rb_nativethread_id_t) * 2 + 3];
+#endif
+    enum enum4 status;
+    int to_kill;
+    int priority;
+#if defined(_WIN32)
+    void *interrupt_event;
+#elif defined(HAVE_PTHREAD_H)
+    void *ubf_list1_or_signal_thread_list;
+#ifndef VERSION_2_2_7
```
nateberkopec: Is the reason we're not patching other Ruby versions the little weird subtleties like this one? Basically our struct here has to exactly match the thread struct of the Ruby version we're patching, right?

nateberkopec:
> The reason for this is that you're making puma depend on the exact memory layout of MRI's thread struct and it could easily break in the future and segfault.

Despite the excellent and heroic effort by @NickolasVashchenko on tracking down this bug and providing a fix, I'm starting to thing we should just wait for Ruby to backport and fix this.

From what I can tell, [2.2 will never get this backported](https://bugs.ruby-lang.org/projects/ruby/wiki/ReleaseEngineering) since normal maintenance finished in March. Hey @unak, when do you think 2.3.5 and 2.4.2 will be released, with the fix for https://bugs.ruby-lang.org/issues/13632? Please say November or earlier, so I don't have to do anything 😆

ext/puma_http11/extconf.rb
nateberkopec: Actually, because of the way this works, maybe we should only patch the latest teeny, because that's all we test on CI. This patch could potentially break on a teeny change (since we're so deep in the MRI internals here), so maybe targeting to *really specific* versions isn't a bad idea.

unak:
@nateberkopec We cannot promise when next versions of Ruby will be release, but I expect its in next few months (it's only my expectation, not official).
But, we also cannot promise that the fix for https://bugs.ruby-lang.org/issues/13632 will be included or not. In my feeling backporting the fix to 2.3 may be too difficult.

NickolasVashchenko:
Also managed to test it with version 2.4.1 on windows. The windows-specific pieces of memory layout do not depend on ruby version, so from this point, I feel that we're good to merge it.

ext/puma_http11/extconf.rb
NickolasVashchenko: No-no, the 2.2.0-2.2.6, 2.3.0-2.3.3 and 2.4.0 are good and unaffected by the bug, I've checked that pretty carefully.

ext/puma_http11/puma_http11.c
NickolasVashchenko: Yes, that's correct. The 3 mentioned with regards to configuration and OS.

nateberkopec:
Thanks, @unak! I understand that you are only estimating. That's all I needed!

Well, if it's even possible that a fix may not actually make it ever into 2.3, maybe we really do need to merge this.

@NickolasVashchenko in that case, where appropriate, please add windows to the build matrix: https://docs.travis-ci.com/user/multi-os/

stereobooster:
@nateberkopec but travis does not support windows. It is https://ci.appveyor.com/ that supports windows

unak:
@nateberkopec Now I'm testing the patch for https://bugs.ruby-lang.org/issues/13632 on Ruby 2.3 branch, and it seems to be no problem.  So, Ruby 2.3.5 will contain the fix.

NickolasVashchenko:
Spoke with @nateberkopec and we decided to move the C code in a separate gem, but leave hooks in the `rescue IOError` Puma blocks. 
I'll post an update here when ready.

NickolasVashchenko:
Okay, the gem is created and ready for usage. Tested for majority ruby versions both for Unix and windows.
https://rubygems.org/gems/stopgap_13632

NickolasVashchenko:
So only the hooks in rescue block stay for now.
@nateberkopec please, the thing :)

stereobooster:
@NickolasVashchenko tests failing (RuntimeError: can't modify frozen IOError). And before separate gem commit it was green

@nateberkopec if we enabled windows, maybe we want to enable MacOS in travis too?

NickolasVashchenko:
@stereobooster For this branch it's intended.

stereobooster:
@NickolasVashchenko can we add this to [Gemfile](https://github.com/puma/puma/blob/master/Gemfile)

```ruby
if %w(2.2.7 2.3.4 2.4.1).include? RUBY_VERSION
  gem "stopgap_13632", "~> 1.0.0", :platform => "mri"
end
```

nateberkopec:
@NickolasVashchenko Yeah, if we add it to the Gemfile, that'd be great. Would make our builds green again.

NickolasVashchenko:
Updated, it became better. 
I can also setup appveyor for Puma. @nateberkopec do we need it here?

stereobooster:
@NickolasVashchenko yeah latest commit fixed Ruby: 2.4.1, but not 2.2.7. I wonder why

NickolasVashchenko:
All the remaining failure messages seem unrelated to that Ruby bug.

stereobooster:
Isn't `RuntimeError: can't modify frozen IOError` should be fixed by this branch, it was all green before. I'm a bit far from underlying error, so please, tolerate my infinite questions

NickolasVashchenko:
Sure, no probs. 
I think it is maybe related to https://bugs.ruby-lang.org/issues/13239 and I'll look in it further in the next couple of days. But it's pretty safe to say that those issues are separate.

NickolasVashchenko:
Okay, looks like it was a combination of both since they did not port the other one to 2.2.7

stereobooster:
Awesome. Can't wait it to be merged. Well also would be nice to mention about stopgap and affected versions in readme and fix appveyor build, but this can go in other PR (this on hanging too long).

nateberkopec:
👍 Goodness it feels good to see those tests all pass again! 

I'm thinking we'll remove this code about ~6-months to 1 year after Ruby releases patches for this, so sometime in fall next year most likely.

nateberkopec:
Now I have to get those Appveyor tests actually working...

PR #1347, opened 2017-06-27
Mri gli pipe bug

NickolasVashchenko:
Refreshed pull request with changes from https://github.com/puma/puma/pull/1206, but with the merged master.


nateberkopec:
For future reference, @NickolasVashchenko and I discussed this change over Skype.

PR #1348, opened 2017-06-27
fix #1344 (bundle exec startup without RUBYOPT present)

### LINKED ISSUE #1344: "bundle exec puma" fails: NoMethodError: undefined method `include?' for nil:NilClass"
```
vendor/cache/puma-c264e6aba83f/lib/puma/launcher.rb:171:in `run'
vendor/cache/puma-c264e6aba83f/lib/puma/cli.rb:77:in `run'
vendor/cache/puma-c264e6aba83f/bin/puma:10:in `<top (required)>'
vendor/bundle/ruby/2.2.0/bin/puma:23:in `load'
vendor/bundle/ruby/2.2.0/bin/puma:23:in `<top (required)>'"; nil
```

bughit:


stereobooster:
Related https://github.com/puma/puma/issues/1344

bughit:
@stereobooster 
the investigation is here: https://github.com/puma/puma/issues/1344#issuecomment-313186304

bughit:
@stereobooster @nateberkopec 

My investigation (https://github.com/puma/puma/issues/1344#issuecomment-313186304) showed that in a `bundle exec puma` invocation if `puma` resolves to a ruby script rather than say an bash rbenv shim, then this puma ruby script will be loaded in process and Bundler::ORIGINAL_ENV["RUBYOPT"] will not be set.  In fact nothing set by the following will be in ORIGINAL_ENV

```ruby
    def set_bundle_environment
      set_bundle_variables
      set_path
      set_rubyopt
      set_rubylib
end
```

Is that not sufficient to proceed with the fix?  Or are there other expectations for what's in ORIGINAL_ENV than need to be dealt with?

nateberkopec:
Closing in favor of #1385

Issue #1371, opened 2017-07-17
If `user_supplied_variables` is an empty array, puma listens on two addresses


NickLaMuro:
Introduced in #1234 (really easier PR number to remember, I must say).

Reproduction steps defined below, but effectively when not providing a `port` option in the `config/puma.rb`, instead of using just the rack defaults of `:Host` and `:Port`, they get combined with the default `binds` of the options because `:user_supplied_variables` de-escalates the user options to "default options" in `lib/rack/handler/puma.rb` when provided, regardless if there are any actually supplied.

This is a bit of a perfect storm things lining up get to this point, and I am not covering the whole picture, but I don't think this the ideal behavior, and is quiet confusing to the end user (the Rails developer in this case).

### Steps to reproduce

1) `rails new --skip-spring new_rails_app`

2) `cd new_rails_app`

3) Delete `port` config in `puma.rb`

4) Run `bin/rails s`

### Expected behavior

The output of the `bin/rails s` startup should show that it is only listening on one address:

```console
$ bin/rails s
=> Booting Puma
=> Rails 5.1.2 application starting in development on http://localhost:3000
=> Run `rails server -h` for more startup options
Puma starting in single mode...
* Version 3.9.1 (ruby 2.3.3-p222), codename: Private Caller
* Min threads: 5, max threads: 5
* Environment: development
* Listening on tcp://localhost:3000
Use Ctrl-C to stop
```

### Actual behavior

Instead, `puma` is booted and listening on two ports:

```console
$ bin/rails s
=> Booting Puma
=> Rails 5.1.2 application starting in development on http://localhost:3000
=> Run `rails server -h` for more startup options
Puma starting in single mode...
* Version 3.9.1 (ruby 2.3.3-p222), codename: Private Caller
* Min threads: 5, max threads: 5
* Environment: development
* Listening on tcp://0.0.0.0:9292   <<<<<<
* Listening on tcp://localhost:3000
Use Ctrl-C to stop
```

### System configuration

**Ruby version**:  2.3.3 (probably irrelevant)
**Rails version**:  Rails >= 5.0.3 (when `user_supplied_variables` was added to `rails server`)

NickLaMuro:
cc @schneems

nateberkopec:
Fixed in #1383, I think @schneems just forgot to close.

schneems:
Hmm, it should have been closed automatically by github. Apparently I got my issue number wrong. Thanks for the eyes

schneems:
[...]

PR #1372, opened 2017-07-18
Fix empty :user_supplied_options causing duplicate puma binds

NickLaMuro:
This aims to work in the spirit of the original code from #1234, but doesn't allow the `default_options` inadvertently apply two different bind addresses.

What this change fixes:
-----------------------

When the `Rack::Handler::Puma` sets the `Puma::Configuration`, it checks to see if any `:user_supplied_options` exist in the options hash, and if that hash exists, it then takes any keys not existing in that hash, and pushes them down as `default_options`, giving them lesser priority when evaluated.

This works fine for most values, except the `:binds`, since that can sometimes be the combination of `:Host`/`:Port`.  When the `Puma::Configuration#configure` block is hit in the Rack handler, the `:binds` option already exists with a default from `Puma::Configuration.puma_default_options`, so the resulting config DSL will now append to that array, instead of creating a new value.

This is most likely not the desired behavior, as this will create two listening ports by default.

What this change does:
----------------------
This change effectively creates a `binds` key in the default_options, based off of the `:Host` and `:Port` values, which will then override the `Puma::Configuration.puma_default_options`'s `binds` key when the `Puma::Configuration` instance is instantiated from the rack handler.

To do this, the `set_host_port_to_config` has been updated to work with either a `Puma::DSL` (previous implementation), or a hash.  This allows us to call it prior to instantiating the `Puma::Configuration`, instead of during the configure block, preventing two listening ports from being defined.

### DIFF
--- a/lib/rack/handler/puma.rb
+++ b/lib/rack/handler/puma.rb
@@ -82,7 +82,7 @@ module Rack
           "Verbose"         => "Don't report each request (default: false)"
         }
       end
-    private
+
       def self.set_host_port_to_config(host, port, config)
         if host && (host[0,1] == '.' || host[0,1] == '/')
           config.bind "unix://#{host}"
--- a/lib/rack/handler/puma.rb
+++ b/lib/rack/handler/puma.rb
@@ -26,6 +26,8 @@ module Rack
           end
         end
 
+        self.set_host_port_to_config(default_options[:Host], default_options[:Port], default_options)
+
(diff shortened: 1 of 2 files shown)


NickLaMuro:
Fixes https://github.com/puma/puma/issues/1371

NickLaMuro:
I will be adding tests as well, but wanted to get a proposal up for https://github.com/puma/puma/issues/1371 to get some eyes on it incase it was the wrong approach.

nateberkopec:
@schneems fixed this via #1383

PR #1376, opened 2017-07-22
Updates for frozen string literal compatibility.

pat:
These changes allow the tests to run with `RUBYOPT="--enable-frozen-string-literal"` - well, except for the last three tests in `test/test_integration.rb`, which were hanging for me, so I added temporary skip statements in (though this happens without the env variable too - a quirk of my system, I guess).

At the current point in time, Rubocop is not frozen-string-literal friendly, as it depends on parser (which I've got a pending PR submitted for), but parser depends on racc (which has existing PRs from quite some time ago, not yet merged). So, if anyone knows anything about racc, would love your input there, as the maintainer of parser is happy enough with my patch but wants racc patched first. Due to the compiling nature of parser, it can't be used as a git reference in Gemfiles, so this is certainly more of a roadblock than I'd like.

All of that is to say: there's no point setting the `RUBYOPT` flag for MRI 2.4+ in the Travis configuration until these dependencies are sorted, which means there's the possibility of introducing regressions in the meantime.

https://github.com/whitequark/parser/pull/354
https://github.com/tenderlove/racc/pull/80
https://github.com/tenderlove/racc/pull/81

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -229,8 +229,8 @@ module Puma
     end
 
     def worker(index, master)
-      title = "puma: cluster worker #{index}: #{master}"
-      title << " [#{@options[:tag]}]" if @options[:tag] && !@options[:tag].empty?
+      title  = "puma: cluster worker #{index}: #{master}"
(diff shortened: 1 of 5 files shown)


pat:
Thanks Nate :)

Issue #1362, opened 2017-07-11
SIGINT or Ctrl+c have different outcomes when stopping server being run via puma and rails s


shayonj:
### Steps to reproduce

1) Clone a simple/barebones rails app: https://github.com/shayonj/test-repro-app

2) `rails s`

3) <kbd>ctrl</kbd> + <kbd>c</kbd>

### Expected behavior

```
rails s
=> Booting Puma
=> Rails 5.1.2 application starting in development on http://localhost:3000
=> Run `rails server -h` for more startup options
Puma starting in single mode...
* Version 3.9.1 (ruby 2.3.1-p112), codename: Private Caller
* Min threads: 5, max threads: 5
* Environment: development
* Listening on tcp://0.0.0.0:3000
Use Ctrl-C to stop
^C- Gracefully stopping, waiting for requests to finish
=== puma shutdown: 2017-07-10 20:45:22 -0700 ===
- Goodbye!
```
Notice, its gracefully stopping and logs shutdown timestamp.

### Actual behavior

```
rails s
=> Booting Puma
=> Rails 5.1.2 application starting in development on http://localhost:3000
=> Run `rails server -h` for more startup options
Puma starting in single mode...
* Version 3.9.1 (ruby 2.3.1-p112), codename: Private Caller
* Min threads: 5, max threads: 5
* Environment: development
* Listening on tcp://0.0.0.0:3000
Use Ctrl-C to stop
^CExiting
```

Perhaps, no graceful stop/exit?

### System configuration

**Ruby version**: `ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-darwin16]`
**Rails version**: `Rails 5.1.2`

### Further info
The graceful stop/exit, works fine from the cli version. For example:

1) Clone a simple/barebones rails app: https://github.com/shayonj/test-repro-app

2) `puma`

3) <kbd>ctrl</kbd> + <kbd>c</kbd>

 ```
 puma
Puma starting in single mode...
* Version 3.9.1 (ruby 2.3.1-p112), codename: Private Caller
* Min threads: 5, max threads: 5
* Environment: development
* Listening on tcp://0.0.0.0:3000
Use Ctrl-C to stop
^C- Gracefully stopping, waiting for requests to finish
=== puma shutdown: 2017-07-10 20:47:45 -0700 ===
- Goodbye!
```

----

### Proposal

[...]

PR #1377, opened 2017-07-27
Setup signal to trap SIGINT and gracefully stop server

shayonj:
This way, when you send `SIGINT` to a server, when running via
`rails s`, a graceful shutdown/stop will be performed.

Before this, not trapping `SIGINT` would mean that either `Rails::Server#start`
or `::Rack::Server#start` would be trapping `SIGINT` and exiting the process.

Fix for: https://github.com/puma/puma/issues/1362

**References**:
- [`::Rack::Server#start`](https://github.com/rack/rack/blob/5c36acac4d298dcf04c58a10014c849a478da53f/lib/rack/server.rb#L289)'s trapping of`SIGINT`.

**Replication of current/fixed behavior**:

```
rails s
=> Booting Puma
=> Rails 5.1.2 application starting in development on http://localhost:3000
* Version 3.9.1 (ruby 2.3.3-p222), codename: Private Caller
* Environment: development
* Listening on tcp://0.0.0.0:3000
Use Ctrl-C to stop
=== puma shutdown: 2017-07-26 19:53:48 -0700 ===
- Goodbye!
Exiting
```

### DIFF
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -399,6 +399,20 @@ module Puma
       end
 
       begin
+        Signal.trap "SIGINT" do
+          if Puma.jruby?
+            @status = :exit
+            graceful_stop
+            exit
+          end
+
+          stop
+        end
+      rescue Exception
+        log "*** SIGINT not implemented, signal based gracefully stopping unavailable!"
+      end
+
+      begin
         Signal.trap "SIGHUP" do
           if @runner.redirected_io?
             @runner.redirect_io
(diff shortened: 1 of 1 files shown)


lib/puma/launcher.rb
```diff
@@ -399,6 +399,14 @@ def setup_signals
       end
 
       begin
+        Signal.trap "SIGINT" do
```
stereobooster: code lower in this file

```ruby
if Puma.jruby?
  Signal.trap("INT") do
```

move those two together

shayonj: fixed

PR #1384, opened 2017-08-03
Add /gc URL to status server to do a major GC. Add a /gc-status URL for GC.stats.

noahgibbs:
Should now be filthy-tab-free with an appropriate test in test_cli.rb.


nateberkopec:
giphy

PR #1385, opened 2017-08-05
do not blow up when RUBYOPT is not defined

grosser:
tried to run from master ... kaboom :D

```
NoMethodError: undefined method `include?' for nil:NilClass
  puma-0d7a8bb5f377/lib/puma/launcher.rb:171:in `run'
```

### DIFF
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -165,10 +165,10 @@ module Puma
     def run
       previous_env =
         if defined?(Bundler)
-          env = Bundler::ORIGINAL_ENV
+          env = Bundler::ORIGINAL_ENV.dup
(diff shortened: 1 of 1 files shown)


grosser:
@nateberkopec

stereobooster:
I suppose this PR shadows this one https://github.com/puma/puma/pull/1348/files.

grosser:
kinda ... was not sure what that did in addition to this "just fix it"
change ... so I made a new PR

nateberkopec:
Yeah, this looks like it does the same thing as #1348 with a much smaller change.

@grosser why don't we .dup ORIGINAL_ENV

grosser:
added .dup for good measure ... might have produced weird errors if someone shells our from a puma process

Issue #1133, opened 2016-11-14
Status Codes, Status Codes everywhere


HoneyryderChuck:
Check this out:

https://github.com/puma/puma/blob/f5f23aaac7aaccff1b6b138d93dd4b1755ebf1c2/lib/puma/server.rb#L747

Works quite well, although it's a repetition of [rack](https://github.com/rack/rack/blob/4b33af1c80c822cbcbb69113ff1e54f9454921c1/lib/rack/utils.rb#L585-L592)

I've also checked webrick, and it also has it's own map of status code to reason. In fact, I'd say that every server implementation has it.

This is quite tricky, as I don't know theoretically which one should be the source of knowledge on that. It definitely doesn't seem to belong to `rack`, as it shouldn't have any knowledge of http semantics (http reason seems to be server-only). However, repeating what's already available on the standard library seems to be not a good approach. Problem is, [webrick has less keys than rack and puma](https://github.com/nahi/ruby/blob/webrick_trunk/lib/webrick/httpstatus.rb#L56-L97). 

This does seem to be something that should/could be centralized, maybe in ruby standard library. But what's the best approach on ensuring that the list is up-to-date?

nateberkopec:
Hm, I dunno. The datastructure isn't [that large](https://github.com/puma/puma/blob/3f66b3d7d4413f843e4e541c4d282238318c4cd2/lib/puma/const.rb#L16).

For comparison, look at the mime-types library, which is basically a huge list of constants for use by other libraries. It's enormous!

I just don't think the size of the constant is worth the effort and extraction.

nateberkopec:
Aha! I figured out why.

See https://github.com/puma/puma/commit/537bc21593182cd9c4c0079a3936d05b1f91fe14 and #705.

TiagoCardoso1983:
@nateberkopec my concern isn't memory, just correctness. Not only based on API-maintenance supports, I don't think that rack is the best place to keep http internals info, so I actually welcome that commit. As puma is the http server, that seems fine.

However, my argument was making it more "central" to ruby. I was comparing with nodejs, where `http.STATUS_CODES` is available to anyone in the runtime. Ruby also has the benefit of having an http server implementation in the standard library (and that module with status codes seems abstracted enough), but it has been slowly ignored by everyone, so that updates to new http status codes seem to get more love in rack.

nateberkopec:
Yeah, that already exists in Rack::Utils (which is where Puma's status codes are directly copy-pasted from).

[...]

PR #1392, opened 2017-08-11
Implement rfc2324 section 2.3.2, Status 418: I'm a teapot.

hoffm:
Reference: https://www.ietf.org/rfc/rfc2324.txt

### DIFF
--- a/lib/puma/const.rb
+++ b/lib/puma/const.rb
@@ -53,6 +53,7 @@ module Puma
     415 => 'Unsupported Media Type',
     416 => 'Range Not Satisfiable',
     417 => 'Expectation Failed',
+    418 => 'I\'m A Teapot',
     422 => 'Unprocessable Entity',
     423 => 'Locked',
     424 => 'Failed Dependency',


nateberkopec:
Linking #1133, because we've have to add 418 back when importing the stdlib's status codes.

hoffm:
@nateberkopec The Lord's work is never done.

nateberkopec:
Merging this, because [*at least one* Ruby web framework supports 418](https://github.com/hanami/controller/blob/master/lib/hanami/http/status.rb#L20), so if they send down a 418 via Rack we want to append the correct message. Our HTTP status code hash should basically reflect whatever the popular frameworks will send down to us. (I'm not going to just add everything Hanami does though, they've got a few in there that actually conflict with current IANA codes, e.g. 451: https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml) 

This change has been made to ensure the highest standards of scalability, accuracy and reliability in teapot deliverability.

PR #1403, opened 2017-08-29
Add early hints feature

eileencodes:
Worked with @tenderlove on adding early hints. This is still a work in progress as there are a few things we need to address, but this PR is to start the conversation.

This commit adds the early hints lambda to the headers hash. Users can
call it to emit the early hints headers. For example:

```
class Server
  def call env
    if env["REQUEST_PATH"] == "/"
      env['rack.early_hints'].call([{ link: "/style.css", as: "style" }, { link: "/script.js" }])
      [200, { "X-Hello" => "World" }, ["Hello world!"]]
    else
      [200, { "X-Hello" => "World" }, ["NEAT!"]]
    end
  end
end

run Server.new
```

In this example, the server sends stylesheet and javascript early hints
if the proxy supports it, it will send H2 pushes to the client.

Of course not every proxy server supports early hints, so to enable the
early hints feature with puma you have to pass the configuration variable,
`--early-hints`.

### DIFF
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -241,6 +241,10 @@ module Puma
       @options[:mode] = :tcp
     end
 
+    def early_hints(answer=true)
+      @options[:early_hints] = answer
+    end
+
     # Redirect STDOUT and STDERR to files specified.
     def stdout_redirect(stdout=nil, stderr=nil, append=false)
       @options[:redirect_stdout] = stdout
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -81,7 +81,7 @@ module Puma
       @precheck_closing = true
     end
 
-    attr_accessor :binder, :leak_stack_on_error
+    attr_accessor :binder, :leak_stack_on_error, :early_hints
 
     forward :add_tcp_listener,  :@binder
     forward :add_ssl_listener,  :@binder
@@ -594,6 +594,24 @@ module Puma
 
       env[RACK_INPUT] = body
       env[RACK_URL_SCHEME] =  env[HTTPS_KEY] ? HTTPS : HTTP
+
+      if @early_hints
+        env[EARLY_HINTS] = lambda { |headers|
+          fast_write client, "HTTP/1.1 103 Early Hints\r\n".freeze
+
+          headers.each_pair do |k, vs|
+            if vs.respond_to?(:to_s) && !vs.to_s.empty?
+              vs.to_s.split(NEWLINE).each do |v|
+                fast_write client, "#{k}: #{v}\r\n"
+              end
+            else
+              fast_write client, "#{k}: #{v}\r\n"
+            end
+          end
+
+          fast_write client, "\r\n".freeze
+        }
+      end
 
       # A rack extension. If the app writes #call'ables to this
       # array, we will invoke them when the request is done.
--- a/test/test_puma_server.rb
+++ b/test/test_puma_server.rb
@@ -159,6 +159,61 @@ class TestPumaServer < Minitest::Test
     assert_equal "HTTP/1.0 200 OK\r\nContent-Length: 0\r\n\r\n", data
   end
 
+  def test_early_hints_works
+    @server.app = proc { |env|
+     env['rack.early_hints'].call("Link" => "</style.css>; rel=preload; as=style\n</script.js>; rel=preload")
+     [200, { "X-Hello" => "World" }, ["Hello world!"]]
+    }
+
+    @server.add_tcp_listener @host, @port
+    @server.early_hints = true
+    @server.run
+
+    sock = TCPSocket.new @host, @server.connected_port
+    sock << "HEAD / HTTP/1.0\r\n\r\n"
+
+    data = sock.read
+
+    expected_data = (<<EOF
+HTTP/1.1 103 Early Hints
+Link: </style.css>; rel=preload; as=style
+Link: </script.js>; rel=preload
+
+HTTP/1.0 200 OK
+X-Hello: World
+Content-Length: 12
+EOF
+).split("\n").join("\r\n") + "\r\n\r\n"
+
+    assert_equal true, @server.early_hints
+    assert_equal expected_data, data
+  end
+
+  def test_early_hints_is_off_by_default
+    @server.app = proc { |env|
+     assert_nil env['rack.early_hints']
+     [200, { "X-Hello" => "World" }, ["Hello world!"]]
+    }
+
+    @server.add_tcp_listener @host, @port
+    @server.run
+
+    sock = TCPSocket.new @host, @server.connected_port
+    sock << "HEAD / HTTP/1.0\r\n\r\n"
+
+    data = sock.read
+
+    expected_data = (<<EOF
+HTTP/1.0 200 OK
+X-Hello: World
+Content-Length: 12
+EOF
+).split("\n").join("\r\n") + "\r\n\r\n"
+
+    assert_nil @server.early_hints
+    assert_equal expected_data, data
+  end
+
   def test_GET_with_no_body_has_sane_chunking
     @server.app = proc { |env| [200, {}, []] }
 
--- a/lib/puma/cli.rb
+++ b/lib/puma/cli.rb
@@ -181,6 +181,10 @@ module Puma
             user_config.tcp_mode!
           end
 
+          o.on "--early-hints", "Enable early hints support" do
+            user_config.early_hints
+          end
+
           o.on "-V", "--version", "Print the version information" do
             puts "puma version #{Puma::Const::VERSION}"
             exit 0
--- a/lib/puma/const.rb
+++ b/lib/puma/const.rb
@@ -221,5 +221,7 @@ module Puma
     HIJACK_P = "rack.hijack?".freeze
     HIJACK = "rack.hijack".freeze
     HIJACK_IO = "rack.hijack_io".freeze
+
+    EARLY_HINTS = "rack.early_hints".freeze
   end
 end
--- a/lib/puma/runner.rb
+++ b/lib/puma/runner.rb
@@ -161,6 +161,10 @@ module Puma
         server.tcp_mode!
       end
 
+      if @options[:early_hints]
+        server.early_hints = true
+      end
+
       unless development?
         server.leak_stack_on_error = false
       end


tenderlove:
@eileencodes and I are working on pushing [early hints](https://datatracker.ietf.org/doc/draft-ietf-httpbis-early-hints/) support in to the Rack spec for version 2 (version 2 of the SPEC, not version 2 of the gem).  I'd like to try experimentally implementing this in Puma first, then push support to Rails, then upstream to the Rack SPEC.  IOW, I don't want this to be done in a vacuum.

Does this seem like an acceptable approach for the Puma team?  I can understand if you don't want to be used for experimentation, but I would like to get this in to production systems sooner rather than later.  😊

Also, we've verified this to work with Puma + h2o.

tenderlove:
I think we're using Ruby 2.3 specific features in the test in this PR.  If the overall concept seems OK, then we'll fix it up.

nateberkopec:
:clap: Give me a while to look at it, headed to Japan soon so not sure how much time I'll have. Last resort we can talk about it at Kaigi?

tenderlove:
@nateberkopec sounds good. 😊

nateberkopec:
>  I can understand if you don't want to be used for experimentation, but I would like to get this in to production systems sooner rather than later.

also I think I speak for the whole Puma team when I say that we have no problem with experimentation. hell Evan wanted to add native websockets and rewrite the entire reactor. so adding an optional feature...pretty tame 😆

lib/puma/server.rb
```diff
@@ -595,6 +600,24 @@ def handle_request(req, lines)
...
+            if as = link[:as]
+              fast_write client, "; as=#{as}"
+            end
+            fast_write client, "\r\n".freeze
+          end
+
+          fast_write client, "\r\n".freeze
+        }
+      else
+        env[EARLY_HINTS] = lambda { |_| }
```
nateberkopec: Why is this necessary? I guess I don't understand.

lib/puma/server.rb
```diff
@@ -595,6 +600,24 @@ def handle_request(req, lines)
       env[RACK_INPUT] = body
       env[RACK_URL_SCHEME] =  env[HTTPS_KEY] ? HTTPS : HTTP
 
+      if @early_hints
+        env[EARLY_HINTS] = lambda { |links|
```
nateberkopec: It would be nice to have this lambda live somewhere other than the body of this method.

lib/puma/dsl.rb
```diff
@@ -241,6 +241,10 @@ def tcp_mode!
       @options[:mode] = :tcp
     end
 
+    def early_hints!
```
nateberkopec: I know our DSL style is really inconsistent, but for futureproofing I'd prefer:

```ruby
def early_hints(answer=true)
    @options[:early_hints] = answer
end
```

lib/puma/server.rb
```diff
@@ -595,6 +600,24 @@ def handle_request(req, lines)
...
+          fast_write client, "HTTP/1.1 103 Early Hints\r\n".freeze
+
+          links.each do |link|
+            fast_write client, "Link: <#{link[:link]}>; rel=preload"
+            if as = link[:as]
+              fast_write client, "; as=#{as}"
```
nateberkopec: Should we auto-detect this part? I did some of this in [rack-http-preload.](https://github.com/nateberkopec/rack_http_preload/blob/master/lib/rack_http_preload.rb#L25).

lib/puma/server.rb
```diff
@@ -97,6 +98,10 @@ def tcp_mode!
       @mode = :tcp
     end
 
+    def early_hints!
```
nateberkopec: I think you were copying tcp_mode!, but that's a `!` method because it doesn't make sense for a server to be in tcp_mode and then *not* be in tcp mode later. Although I can't think of why anyone ever would (restarting and disabling early hints as the config changes?), it's more plausible that a server could toggle its support of early_hints. So this should probably just be treated like a straight-up accessor.

test/test_puma_server.rb
```diff
@@ -159,6 +159,32 @@ def test_GET_with_empty_body_has_sane_chunking
...
+    @server.add_tcp_listener @host, @port
+    @server.early_hints!
+    @server.run
+
+    sock = TCPSocket.new @host, @server.connected_port
+    sock << "HEAD / HTTP/1.0\r\n\r\n"
+
+    data = sock.read
+
+    assert_equal <<~eos.split("\n").join("\r\n") + "\r\n\r\n", data
```
nateberkopec: Rubocop is barfing on your squiggly heredoc.

lib/puma/server.rb
tenderlove: Agree.  @eileencodes and I were trying to move it out of here, but the lambda body needs to access the io object (`client`) and `fast_write` is a private method.  I'd love to move the lambda out of here, but I am not sure how.

lib/puma/server.rb
tenderlove: Unsure.  h2o pushes the resource regardless.  I think this is something we need to iterate on. (IOW, I don't have an answer)

lib/puma/server.rb
tenderlove: This is so that app code can run without changes even if the server is being run without early hints support.  The example app that @eileencodes put in the PR description can be run whether early hints is enabled or not (it prevents the app code from writing a `if env[EARLY_HINTS]` conditional)

test/test_puma_server.rb
tenderlove: I saw the error, and I think RuboCop needs to be fixed.  This is new syntax in Ruby 2.3.  But also, since this is new syntax in Ruby 2.3, and you probably want to run the Puma tests on older versions of Ruby anyway, we'll fix it. 😉

lib/puma/server.rb
nateberkopec: It's also possible we may not want to auto-detect `as` at the *puma* level. I just wanted to ask the question.

lib/puma/server.rb
eileencodes: Moved to an `attr_accessor` and deleted this method.

lib/puma/dsl.rb
eileencodes: Fixed 👍

test/test_puma_server.rb
eileencodes: Fixed the tests here too. :+1: I also added a test to test that early hints is off by default.

nateberkopec:
Need to look at this one more time when I'm not jetlagged but I think this is ready.

tenderlove:
Hey folks.  I've been talking to @matthewd about this feature.  He convinced me that the lambda should take a header hash just like the header hash from a rack response.  This would alleviate webservers from knowing how the `Link` header is formatted, and may allow us to send other headers (like `X-` headers) for debugging purposes.

IOW, we should change the API to be something like this:

```
env[EARLY_HINTS].call(“Link”: [“<...>; rel=preload; as=..”, “..”])
```

WDYT?

nateberkopec:
@tenderlove If I'm reading the HTTP 103 spec correctly:

> A server MUST NOT include Content-Length, Transfer-Encoding, or any
   hop-by-hop headers ([RFC7230], section 6.1) in the informational
   response using the status code.
> A client MAY speculatively evaluate the headers included in the
   informational response while waiting for the final response.  For
   example, a client may recognize the link header of type preload and
   start fetching the resource.  However, the evaluation MUST NOT affect
   how the final response is processed; the client must behave as if it
   had not seen the informational response.

So, 103 status CANNOT contain a Connection, Content-Length or Transfer-Encoding header, but MAY contain *any other header*?

If I'm getting that right, your change makes sense. But should anyone be responsible for omitting the prohibited headers? Puma's job or someone elses?

tenderlove:
> If I'm getting that right, your change makes sense. But should anyone be responsible for omitting the prohibited headers? Puma's job or someone elses?

I think if you wanted to omit them in Puma it would be fine.  However, I'd expect the proxy to complain or raise an error if it got those headers, so maybe Puma shouldn't care?  It would be nice for the user to get an exception closer to where the error is occurring, but I don't think it's necessary.

eileencodes:
I've updated the PR so that now the lambda now takes a header hash and tested this in the Rails implementation for our test app. Let me know if you'd like other changes but I think this is good to go now.

jumph4x:
ilu guys

lib/puma/server.rb
```diff
@@ -595,6 +595,24 @@ def handle_request(req, lines)
...
+        env[EARLY_HINTS] = lambda { |headers|
+          fast_write client, "HTTP/1.1 103 Early Hints\r\n".freeze
+
+          headers.each_pair do |k, vs|
+            if vs.respond_to?(:to_s) && !vs.to_s.empty?
+              vs.to_s.split(NEWLINE).each do |v|
+                fast_write client, "#{k}: #{v}\r\n"
+              end
+            else
+              fast_write client, "#{k}: #{v}\r\n"
```
PikachuEXE: Should this be `fast_write client, "#{k}: #{vs}\r\n"` instead?
I don't see `v` defined in this level

eileencodes: v is defined right here https://github.com/puma/puma/pull/1403/files/ea37ada76937ff6b3988e952f598c26cf373904e#diff-ef4c610f0ce605eb90f2810856394fd5R604

eileencodes: Oh i see what you're saying - I will fix it

ioquatix:
Was this ever integrated into the Rack SPEC? I had a look but couldn't see anything on the latest master branch.

What was the final design of the API?

I'm implementing this in `falcon` and support H2 server push directly so it would be fun to compare.

ioquatix:
I implemented the design that was merged into puma.

https://github.com/socketry/falcon/commit/248d0a458294914f0d694b26540afccfea7b5b18

Here is my brief feedback:

- It's very HTTP/1 centric (i.e. using headers). For HTTP/2, I need to parse the headers and turn that into push requests. I guess it's okay. I don't know how I'd handle other headers correctly.
- I only add the lambda to the rack env when it's possible to send early hints. So, it's conditionally enabled. Is that okay? Because I don't bother to support `103 Early Hints` yet, but only support H2 push promises.
- I wonder if exposing this in the application layer is really a good idea. I feel like a sufficiently intelligent server could figure out what things to send and when, and additionally cache what the client has/has not received.

The only real point is - is this interface to generic? Would it be better to limit it to push-promise functionality, or is it good to have it allow any header?

ioquatix:
I wrote up a brief summary regarding implementing this feature.

https://www.codeotaku.com/journal/2019-02/falcon-early-hints/index

PR #1410, opened 2017-09-14
Travis: Use latest bundler

olleolleolle:
This PR updates the `before_install` step to not peg to a specific version of Bundler, but to use latest stable release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ cache: bundler
 before_install:
   # bundler installation needed for jruby-head
   # https://github.com/travis-ci/travis-ci/issues/5861
-  - gem install bundler -v 1.11.2
+  - gem install bundler
 script: "TESTOPTS=-v bundle exec rake"
 branches:
   only:


olleolleolle:
A pedant's note:

Could it be reasonable to compile the C extensions using a compiler version which does not emit warnings?

Sketch of a configuration to make this happen:

```yaml
      addons:
        apt:
          sources:
            - ubuntu-toolchain-r-test
          packages:
            - g++-6
      env:
        - global
          - CC=gcc-6
          - CXX=g++-6
```

nateberkopec:
> Could it be reasonable to compile the C extensions using a compiler version which does not emit warnings? 

Sorry what are you getting at here? That we don't currently emit all possible warnings and we should?

b/c if we're not omitting warnings on CI I agree, we should.

olleolleolle:
@nateberkopec Hi Nate, I was looking at the compile output, and it's using an ancient version of GCC, which has C90 as standard. The emitted warnings are... old-style.

olleolleolle:
@nateberkopec I thought long about it, and wondered if we should be compiling with something v. old (proving that it works) OR if the right solution is to compile it with a newer version of GCC (avoiding compilation warnings about C90 code style).

olleolleolle:
(I'm sorry to make you have this low-value, pedantic conversation.)

nateberkopec:
> (I'm sorry to make you have this low-value, pedantic conversation.)

I'm a maintainer, my job is to have many low-value conversations over and over UNTIL THE END OF TIME 😆 

I don't think there's any value compiling with things people don't use anymore.

olleolleolle:
If _usage_ is a guiding thing, then "what Travis defaults to" could be said to constitute usage. If that's so, then perhaps we ought to follow this old-style way of moving declarations to the top of the function (C90 recommendation)? 🌗 Tricky.

nateberkopec:
> then "what Travis defaults to" could be said to constitute usage

Meh, but our true usage is "what people bundle/gem install on".

PR #1417, opened 2017-09-20
fixed typo in --control-url args

holamendi:
`--control` is supossed to be `--control-url`


nateberkopec:
Merged in a different PR, but thanks for submitting.

Issue #1428, opened 2017-10-05
gemspec addition for windows RubyInstaller2 builds


MSP-Greg:
Ruby on windows has a new build/extension gem system based on MSYS2.  The previous system was called RubyInstaller, the new system is RubyInstaller2.

A common issue is building a gem with a package requirement.  RubyGems is hooked so it allows RubyInstaller2 runtime files to install the package before compile.

See [eventmachine/eventmachine](https://github.com/eventmachine/eventmachine/blob/master/eventmachine.gemspec#L16) and [ruby/opensll](https://github.com/ruby/openssl/blob/master/openssl.gemspec#L25).  Also [RubyInstaller2 Wiki page - For gem developers](https://github.com/oneclick/rubyinstaller2/wiki/For-gem-developers).

Hence, coould someone please add the following line to `puma.gemspec`?

```ruby
s.metadata["msys2_mingw_dependencies"] = "openssl"
```

I'd do a PR, but I don't want to fork for this one issue..

PR #1434, opened 2017-10-10
Add msys2_mingw_dependencies to puma.gemspec

jumbosushi:
Add a line to puma.gemspec as specified in #1428

### DIFF
--- a/puma.gemspec
+++ b/puma.gemspec
@@ -13,6 +13,7 @@ Gem::Specification.new do |s|
   s.email = ["evan@phx.io"]
   s.executables = ["puma", "pumactl"]
   s.extensions = ["ext/puma_http11/extconf.rb"]
+  s.metadata["msys5_mingw_dependencies"] = "openssl"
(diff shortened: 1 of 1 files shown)


juliancheal:
@jumbosushi aw you beat me to it.
screen shot 2017-10-10 at 22 39 26

puma.gemspec
```diff
@@ -17,5 +17,6 @@ Gem::Specification.new do |s|
             %w[History.md LICENSE README.md]
   s.homepage = "http://puma.io"
   s.license = "BSD-3-Clause"
+  s.metadata["msys2_mingw_dependencies"] = "openssl"
```
juliancheal: I think this would be better on line 16 underneath `s.extensions = ["ext/puma_http11/extconf.rb"]` seems to be more related to extensions than `homepage` & `metadata`

jumbosushi: Just changed it to your way. I wasn't 100% sure about where to put the line so this was helpful. Thanks for the tip!

PR #1435, opened 2017-10-10
Added two new HTTP Status Codes 421, and 451.

juliancheal:
Ran the generator to update the list of HTTP Codes.

The generator removed `418 'I'm A Teapot'` which I had to add back in. Being British, I need a good cuppa.

Of note is HTTP Status Code 451 named after Ray Bradbury's Fahrenheit 451.


juliancheal:
@nateberkopec All 🍏 except Travis seems to be failing on Ruby 2.2.8, but I think that's unrelated.

nateberkopec:
LGTM

Issue #1386, opened 2017-08-05
Sending SIGTERM to a server process when running via `rails s` (in clustered mode) fails to stop the server successfully


shayonj:
### Steps to reproduce

1) Clone a simple/barebones rails app: https://github.com/shayonj/test-repro-app

2) `rails s`

3) From a different shell run`lsof -i tcp:3000` and grab `pid` of the server

4) `kill -TERM <pid>`

5) `ps <pid>`

### Expected behavior

Server and workers both should be stopped. `ps <pid>` should return empty.

Ex:
```
 ps 47308
  PID   TT  STAT      TIME COMMAND
```

### Actual behavior

Server continues to be running, but workers are stoped.

```
rails s
=> Booting Puma
=> Rails 5.1.2 application starting in development on http://localhost:3000
=> Run `rails server -h` for more startup options
[47287] Puma starting in cluster mode...
[47287] * Version 3.9.1 (ruby 2.3.3-p222), codename: Private Caller
[47287] * Min threads: 5, max threads: 5
[47287] * Environment: development
[47287] * Process workers: 2
[47287] * Phased restart available
[47287] * Listening on tcp://0.0.0.0:3000
[47287] Use Ctrl-C to stop
[47287] - Worker 0 (pid: 47307) booted, phase: 0
[47287] - Worker 1 (pid: 47308) booted, phase: 0
[47287] - Gracefully shutting down workers...
```

```
ps 47287
PID   TT  STAT      TIME COMMAND
47287 s004  S+     0:01.15 puma 3.9.1 (tcp://0.0.0.0:3000) [test-repro-app]  
```

### Important Note

This only happens if you run via `rails s` and not when running via `puma` bin.

### System configuration

**Ruby version**: `ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-darwin16]`
**Rails version**: `Rails 5.1.2`

nateberkopec:
Looks like you fixed this on master

shayonj:
hi @nateberkopec, I can still reproduce this issue on master (3.10.0)

shayonj:
this is solved in https://github.com/puma/puma/pull/1464

PR #1437, opened 2017-10-11
Appropriately wait for worker child process when shutting down via SIGTERM

shayonj:
This way when a SIGTERM is issued to a puma cluster (master process)
the shutting down of workers/child processes successfully completes.

Before this, it meant the master process will wait for every child process, not just
puma specific child process. Which meant even though the workers were to
successfully terminate, the server never exits as intended.

Issue: https://github.com/puma/puma/issues/1386

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -24,7 +24,9 @@ module Puma
       @workers.each { |x| x.term }
 
       begin
-        Process.waitall
+        @workers.each do |w|
+          Process.waitpid(w.pid)
+        end
       rescue Interrupt
         log "! Cancelled waiting for workers"
       end


shayonj:
@nateberkopec @MSP-Greg is this good to close (and so the mentioned issue)? Looks the changes were rolled into a single PR. :)

MSP-Greg:
@shayonj 

Please accept my apology, as I should have mentioned you in the PR.  It was kind of a 'if all three are rolled into one, this should work' thing...

nateberkopec:
Yes - thanks you for opening, @shayonj.

shayonj:
not at all @MSP-Greg. Thanks all!

PR #1439, opened 2017-10-16
Add ssl support to Windows builds, Appveyor

MSP-Greg:
Also, dummy def for RDoc & YARD in mini_ssl.c

### DIFF
--- a/lib/puma/minissl.rb
+++ b/lib/puma/minissl.rb
@@ -43,7 +43,19 @@ module Puma
           output = engine_read_all
           return output if output
 
-          data = @socket.read_nonblock(size)
+          if /mswin|mingw/ !~ RUBY_PLATFORM
+            data = @socket.read_nonblock(size)
+          else
+            begin
+              data = @socket.read_nonblock(size)
+            rescue IO::WaitReadable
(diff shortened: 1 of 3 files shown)


lib/puma/minissl.rb
```diff
@@ -43,7 +43,19 @@ def read_nonblock(size, *_)
...
+            begin
+              data = @socket.read_nonblock(size)
+            rescue IO::WaitReadable
+              IO.select([@socket.to_io])
+              retry
+            rescue IO::WaitWritable
+              IO.select(nil, [@socket.to_io])
+              retry
+            end
+          end
```
Jesus: This method name is `#read_nonblock`, shouldn't we raise `Errno::EAGAIN` rather than blocking?

It should be the callers responsibility to retry.

MSP-Greg: Maybe have a look at https://github.com/puma/puma/commit/716577597091ed4fd161cb2ec25aab338c74f6c9?

MSP-Greg: I don't recall. Since the error is handled in `read_and_drop`, if the change works,...

PR #1457, opened 2017-11-15
change Reactor @sockets in-place instead of copying

eprothro:
The copying involved with the current implementation hides the value of the `sockets` variable local to the `run_internal` method (e.g. `@sockets` becomes a stale, unused value). 

I don't believe this copying/hiding is intentional, and it makes debugging/tracing more difficult.


evanphx:
Moving away from this code in master anyway.

PR #1459, opened 2017-11-16
Catch more IOErrors in 2.2

nateberkopec:


nateberkopec:
Ugh, I might just move 2.2 to allowed failures.

PR #1444, opened 2017-10-20
minissl read_nonblock fix

MSP-Greg:
Travis is having intermittent errors across various releases.  This seems to fix issue.  Just went thru a bit of nonblock code variations in ruby core.

Please review, as I haven't really been outside of this code/file...

PR #1463, opened 2017-11-19
Fixup test/helper.rb to match commit 998882e

MSP-Greg:


MSP-Greg:
@nateberkopec

Given the results here, I'm wondering about a combination of this and #1444.

PR #1464, opened 2017-11-19
Combine PR's #1437, 44, & 63

MSP-Greg:


nateberkopec:
Ah, wonderful! Thank you!

MSP-Greg:
I'm certainly glad it passed the 2nd time, as the failures were intermittent.

Re Appveyor and ruby-head (trunk), it is currently not 'allow-failure'.  It will always return a build, but it may not be the most current, as it is tested.  Said another way, if a build from yesterday passed, and subsequent builds fail, yesterday's build is what is used.  From memory, I don't recall builds being broken for more than 24 hours.

But, from what I can tell, Travis does ruby-head builds once a day, and they are **not tested**.  So, the possibility exists for a Travis build to be broken.  Also, they currently aren't installing bundled gems.

Hence, for those reasons, Appveyor ruby-head can often be included in the normal build test result, but Travis may best be left 'allow-failure'.  If it becomes an issue, feel free to change Appveyor (or ping me).   Thanks.

PR #1470, opened 2017-11-27
Fix stat & TestCLI#test_control_for_tcp

MSP-Greg:


MSP-Greg:
@nateberkopec 

Thought an explanation might be helpful, as I don't know puma that well.  Most of the intermittent failures originally fail with something like the following:
```diff
TestCLI#test_control_for_tcp = [MinitestRetry] retry 'test_control_for_tcp' count: 1,  msg: --- expected
+++ actual
@@ -1 +1,2 @@
-"{ \"backlog\": 0, \"running\": 0 }"
+# encoding: ASCII-8BIT
+"{ \"backlog\": , \"running\":  }"
```

The output shown at the bottom of the log is the retry info.  So, from the above, we see two things:

1. The split command is not working in `/test/test_cli.rb`.  Hence, the change from `"\r\n"` to `/\r?\n/` for the delimiter.

2. The last line has no zeros.  The backlog and running attributes from `lib/puma/server.rb` are shown below:
```ruby
    def backlog
      @thread_pool and @thread_pool.backlog
    end

    def running
      @thread_pool and @thread_pool.spawned
    end
```

That code shows that `@thread_pool` may be nil, and hence, the return will either be nil or a number >= 0.  Hence, the added `|| 0` code to `lib/puma/cluster.rb` and `lib/puma/single.rb` for their 'stats' info.

This fixes the issue with the test, but I don't know if the test code should generate a non nil value for `@thread_pool`, and hence there really is an issue...  Thanks, Greg

PR #1325, opened 2017-06-04
Let's use Rubocop!

palkan:
Closes #1149.

I've added very basic, all-off-by-default, Rubocop config, which passes. It contains mostly Performance and Layout cops.

Also, I've added a TODO config which contains cops producing a small number of offenses – so, we can easily fix them and make the code style more consistent.

There is the output of the `rubocop -f o -c .rubocop_todo.yml` command:

```
29   Layout/SpaceAroundOperators
21   Layout/SpaceInsideBlockBraces
18   Metrics/LineLength
16   Layout/SpaceAroundEqualsInParameterDefault
15   Layout/SpaceBeforeBlockBraces
15   Layout/SpaceInsideHashLiteralBraces
8    Layout/EmptyLines
6    Layout/EmptyLinesAroundMethodBody
     Layout/EmptyLinesAroundModuleBody
5    Layout/IndentationWidth
4    Layout/EmptyLinesAroundClassBody
3    Layout/AccessModifierIndentation
2    Style/WhileUntilModifier
1    Metrics/ParameterLists
1    Performance/RedundantBlockCall
1    Performance/RedundantMatch
1    Performance/StringReplacement
1    Style/TernaryParentheses
--
152  Total
```

PR #1472, opened 2017-11-28
Use Rubocop 0.50

palkan:
That's the small follow-up for #1325.

Initially I decided not to correct `SpaceBeforeBlockBraces` cop offences but update the cop itself to comply with Puma style (more precisely this syntax `->{}` for no-op lambdas).

The patch has been released in 0.50.0.

There is also new cop `Performance/HashEachMethods`, which detected one typo (?) and one false negative)

By the way, that's the last version of Rubocop which supports 1.9.3.

@nateberkopec Maybe it make sense to reduce the number of TODO items within along with Rubocop upgrade? (I.e. within this PR)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -15,6 +15,10 @@ Layout/SpaceAfterColon:
 Layout/SpaceAroundKeyword:
   Enabled: true
 
+Layout/SpaceBeforeBlockBraces:
+  EnforcedStyleForEmptyBraces: no_space
+  Enabled: true
+
 Layout/SpaceBeforeFirstArg:
   Enabled: true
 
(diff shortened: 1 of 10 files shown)


nateberkopec:
Hm, `SpaceBeforeBlockBraces` failing on 2.1?

palkan:
Wow, for some reason it uses Rubocop 0.49( Even though I set `~>0.50.0` in Gemfile. That's strange.
Could you, please, try to restart the job without cache?

nateberkopec:
I believe you have to edit gemfiles/2.1-Gemfile.

palkan:
Oh, my bad, didn't notice that 2.1 has its own gemfile. Updated

nateberkopec:
Thanks!

Issue #996, opened 2016-06-08
Specifing SSL Ciphers to use


bararchy:
Is there a way to pass on a String like in Ruby's SSL Cipher Context class that will make the server use only specific settings ? 

for example in Ruby:

``` ruby
ssl_context = OpenSSL::SSL::SSLContext.new
ciphers = '!ADH:!RC4:!aNULL:!MD5:!EXPORT:!SSLv2:HIGH'
ssl_context.ciphers = ciphers
```

evanphx:
At present, there is not, no.

mirceal:
Would you consider taking a PR for this?

nateberkopec:
> Would you consider taking a PR for this?

I think so, yes.

mirceal:
Can we please reopen the issue?

mirceal:
PR is being prepared and should land shortly

mirceal:
May we get some feedback on https://github.com/puma/puma/pull/1476 and on the test failures?
From the looks of it it does not seem to be related to the code that is introduced. How can we run the whole test suite locally before submitting the PR to ensure

PR #1474, opened 2017-11-29
Dev/ssl cipher support

eallison91:
https://github.com/puma/puma/issues/996

Support specifying which ssl cipher suites the server should support. Default behavior is unchanged.


eallison91:
Didn't mean to include those other commits.

PR #1476, opened 2017-11-29
Add support for specifying ssl ciphers via :binds parameters

eallison91:
https://github.com/puma/puma/issues/996

Allows optional specifying of which ssl ciphers the server should support. Default behavior is unchanged.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,6 +8,7 @@ gem "rake-compiler"
 gem "rack", "< 3.0"
 gem "minitest", "~> 5.9"
 gem "minitest-retry"
+gem "pry-byebug"
 
 gem "jruby-openssl", :platform => "jruby"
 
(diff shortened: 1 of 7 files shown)


nateberkopec:
Thanks. I need to look at the implementation, but if we're gonna merge this, it will need docs and a test.

eallison91:
I am in the process of adding tests, but need some guidance on how to make them succeed in travis and appveyor. Based on the existing tests in test_binder.rb, it looks like things pertaining to SSL are skipped when running on appveyor. Is that accurate?

I also have one test that specifically *should* be run when the engine is JRuby (possibly still not on appveyor?). I am using `skip unless Puma.jruby?` Is that the correct way to achieve that?

eallison91:
Attempt to rebase changes resulted in branch getting recreated... lesson learned.

Yet another PR since I can't reopen it. https://github.com/puma/puma/pull/1478

PR #1416, opened 2017-09-20
Control CLI Bug Fix

hiimtaylorjones:
As per recent comments on [this commit](https://github.com/puma/puma/commit/a8f54f74cde910b2bf93260a113b057ba6ec40b5) and a [request](https://twitter.com/nateberkopec/status/910520447279693825) by @nateberkopec on Twitter, this PR addresses a bug that was in control-cli. 

The commit is fairly self explanatory, but there was an error in adding in the control-url arg. This PR fixes that typo.

Issue #1477, opened 2017-11-30
invalid option: --control-url


sigmike:
### Steps to reproduce

1) Install puma 3.11.0

2) Create a config with only `activate_control_app 'unix:///tmp/puma.sock'`

3) Run `pumactl -F <config file> start`

### Expected behavior

Starting an complaining about application not configured.

### Actual behavior

```
invalid option: --control-url
/home/mike/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/puma-3.11.0/lib/puma/cli.rb:47:in `initialize'
/home/mike/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/puma-3.11.0/lib/puma/control_cli.rb:258:in `new'
/home/mike/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/puma-3.11.0/lib/puma/control_cli.rb:258:in `start'
/home/mike/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/puma-3.11.0/lib/puma/control_cli.rb:223:in `run'
/home/mike/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/puma-3.11.0/bin/pumactl:8:in `<top (required)>'
/home/mike/.rbenv/versions/2.4.2/bin/pumactl:23:in `load'
/home/mike/.rbenv/versions/2.4.2/bin/pumactl:23:in `<main>'
```

### System configuration

**Ruby version**: 2.4.2
**Rails version**: N/A
**Puma version**: 3.11.0

### Comments

Seems to be caused by commit https://github.com/puma/puma/commit/459ab196369be9355432d43097e43a5e0761b8e9

Puma 3.10.0 works fine.

crevete:
I've got the same issue and go back to 3.10 as workaround for now.

jlduran:
https://github.com/puma/puma/commit/459ab196369be9355432d43097e43a5e0761b8e9

jxa:
I get the same. The problem stems from an option mismatch between pumactl and puma cli. On puma cli the option is `--control` and in pumactl the option is called `control-url`. 

Putting together a PR as soon as I figure out how to write a test for this.

chickenham:
I noticed that a new version of puma was released but this is still not fixed.  Is this going to be fixed at some point - or is there a workaround?

crevete:
[...]

PR #1487, opened 2017-12-15
Add support for --control-url in cli

jxa:
# Add support for --control-url

- deprecate `--control`
- add new option `--control-url`
- https://github.com/puma/puma/pull/1416 introduced a bug in pumactl the puma cli does not respond to `control-cli` option. On puma it is called `control`
- this commit https://github.com/puma/puma/commit/a8f54f74cde910b2bf93260a113b057ba6ec40b5 was correct, but it looked like a typo since the names of the options are not consistent across versions
- the best option for fixing this seems to be to support both options in the cli
- we don't want to stop supporting `control` in the cli for backwards-compatibility

### DIFF
--- a/lib/puma/control_cli.rb
+++ b/lib/puma/control_cli.rb
@@ -220,7 +220,7 @@ module Puma
     end
 
     def run
-      start if @command == "start"
+      return start if @command == "start"
 
       prepare_configuration
 
--- a/README.md
+++ b/README.md
@@ -167,7 +167,7 @@ $ puma -b 'ssl://127.0.0.1:9292?key=path_to_key&cert=path_to_cert'
 Puma has a built-in status/control app that can be used to query and control Puma itself.
 
 ```
-$ puma --control tcp://127.0.0.1:9293 --control-token foo
+$ puma --control-url tcp://127.0.0.1:9293 --control-token foo
 ```
 
 Puma will start the control server on localhost port 9293. All requests to the control server will need to include `token=foo` as a query parameter. This allows for simple authentication. Check out [status.rb](https://github.com/puma/puma/blob/master/lib/puma/app/status.rb) to see what the app has available.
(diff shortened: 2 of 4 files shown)


jxa:
Also reported here https://github.com/puma/puma/issues/1477

jxa:
an alternative approach is to simply revert #1416, and write a spec for pumactl. downside being that the options do not match across both interfaces.

jlduran:
The `README.md` also references `--control`. But that can be easily addressed in a separate commit.

lib/puma/control_cli.rb
```diff
@@ -220,7 +220,7 @@ def send_signal
     end
 
     def run
-      start if @command == "start"
+      return start if @command == "start"
```
jxa: Prior to this change, `send_request` was being called with `@command == "start"`. The control server does not respond to this command, so returns a 404, which then ultimately calls `exit 1`, making it very difficult to figure out why my test was failing silently. This early return short-circuits all that mess.

jxa:
I'm open to options on how to actually fix the CLI @evanphx @nateberkopec @schneems.

Whatever the ultimate decision, I think this is an extremely valuable test which will prevent similar accidental breakage in the future.

jxa:
/sigh that is, it will be valuable if I can get it to pass reliably on CI

jxa:
Travis build is failing for ruby 2.1, but not just on this PR. These other PRs have the same failure: 
* https://travis-ci.org/puma/puma/builds/325036450 
* https://travis-ci.org/puma/puma/builds/323477393

nateberkopec:
@jxa Sorry for the delay here, I've been taking a break from Ruby.

Can you rebase? I think we fixed the build issues.

jxa:
@nateberkopec sorry I missed your mention somehow! I have rebased, but the appveyor build has failed on 2 versions while trying to compile with ragel. I doubt these are related to my changes. Any advice on how to deal with this issue?

nijikon:
Can this be merged?

PR #1491, opened 2017-12-21
RACK_URL_SCHEME calculation caring about more headers

olleolleolle:
This PR is about #1179, as a way of pushing the discussion onwards.

- add new constants to `const.rb` and use them
- `RACK_URL_SCHEME` is now read from a series of HTTP headers (just like Rack does it)


nateberkopec:
This definitely looks like the right approach, just needs the things you already pointed out to get it over the line.

evanphx:
Can this be merged? Is anything waiting on it?

olleolleolle:
@evanphx I apologize, this had been sitting - I'll turn these TODO items into a checkbox list and then do them.

When visible and less obtuse, review becomes much easier.

olleolleolle:
In order to get this code mergeable, I offer it as-is to review. After re-reading this change, I think it's ready.

I had one question to my earlier self, who was borrowing that code.

## Q: Why am I splitting `X-Forwarded-Proto` on first comma?

Would anyone add more content to _that_ header? Are there any examples?

### A: Example from the code author

The answer is: "yes there are examples!"

https://github.com/rack/rack/commit/85ca454e6143a3081d90e4546ccad602a4c3ad2e

### Documentation about the header, in general

None of these mention any of the practical examples of "multiple proxies after each other", but they can be nice sources of learning about the header, in general.

[Mozilla MDN explains the X-Forwarded-Proto](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto)

<details>

> Cloudflare also appends an `X-Forwarded-Proto` header, which can either be HTTP or HTTPS depending on the protocol the user used to visit the site, like this:
> 
>     "X-Forwarded-Proto: https"

[Cloudflare support docs on the header](https://support.cloudflare.com/hc/en-us/articles/200170986-How-does-Cloudflare-handle-HTTP-Request-headers-)

<summary>Example: Cloudflare explains X-Forwarded-Proto</summary>

</details>

<details>

> The `X-Forwarded-Proto` request header helps you identify the protocol (HTTP or HTTPS) that a client used to connect to your load balancer. Your server access logs contain only the protocol used between the server and the load balancer; they contain no information about the protocol used between the client and the load balancer. To determine the protocol used between the client and the load balancer, use the X-Forwarded-Proto request header. Elastic Load Balancing stores the protocol used between the client and the load balancer in the X-Forwarded-Proto request header and passes the header along to your server.
> 
> Your application or website can use the protocol stored in the X-Forwarded-Proto request header to render a response that redirects to the appropriate URL.
> 
> The X-Forwarded-Proto request header takes the following form:
> 
>     X-Forwarded-Proto: originatingProtocol
> 
> The following example contains an X-Forwarded-Proto request header for a request that originated from the client as an HTTPS request:
> 
>     X-Forwarded-Proto: https

[AWS ELB documentation on the header](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/x-forwarded-headers.html#x-forwarded-proto)

<summary>Example: AWS explains X-Forwarded-Proto

</details>

---

**Addendum:** TIL: the `Forwarded` header

There's a standardized header `Forwarded` (2014) and it has structure. We're not managing that in Puma yet.

- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded
- https://en.wikipedia.org/wiki/X-Forwarded-For#Alternatives_and_variations

lib/puma/server.rb
```diff
@@ -588,8 +588,8 @@ def normalize_env(env, client)
     end
 
     def default_server_port(env)
-      return PORT_443 if env[HTTPS_KEY] == 'on' || env[HTTPS_KEY] == 'https'
-      env['HTTP_X_FORWARDED_PROTO'] == 'https' ? PORT_443 : PORT_80
+      return PORT_443 if env[HTTPS_KEY] == 'on' || env[HTTPS_KEY] == HTTPS
```
nateberkopec: this is a bit hard to read as-is. Can we do this instead:

```ruby 
  if ['on', HTTPS].include?(env[HTTPS_KEY]) || env[HTTP_X_FORWARDED_PROTO] == HTTPS 
    PORT_443
  else
    PORT_80
  end
```

lib/puma/server.rb
```diff
@@ -627,7 +627,18 @@ def handle_request(req, lines)
...
-      env[RACK_URL_SCHEME] =  env[HTTPS_KEY] ? HTTPS : HTTP
+      env[RACK_URL_SCHEME] =
+        if env[HTTPS_KEY]
+          HTTPS
+        elsif env[HTTP_X_FORWARDED_SSL] == 'on'
+          HTTPS
+        elsif env[HTTP_X_FORWARDED_SCHEME]
+          env[HTTP_X_FORWARDED_SCHEME]
+        elsif env[HTTP_X_FORWARDED_PROTO]
+          env[HTTP_X_FORWARDED_PROTO].split(',', 2)[0]
```
nateberkopec: These magic numbers could use a comment

olleolleolle: Adding a comment!

lib/puma/server.rb
olleolleolle: Thanks for example!

olleolleolle:
@nateberkopec I tried to interpret your comments as commits.

olleolleolle:
## Note: To run one test file, saving yourself time, use `TEST`:

```
bundle exec rake test TEST=test/test_puma_server.rb
```

olleolleolle:
@nateberkopec Thanks for asking again, for tests. I sat down and had a go.

lib/puma/server.rb
```diff
@@ -588,8 +588,11 @@ def normalize_env(env, client)
     end
 
     def default_server_port(env)
-      return PORT_443 if env[HTTPS_KEY] == 'on' || env[HTTPS_KEY] == 'https'
-      env['HTTP_X_FORWARDED_PROTO'] == 'https' ? PORT_443 : PORT_80
+      if ['on', HTTPS].include?(env[HTTPS_KEY]) || env[HTTP_X_FORWARDED_PROTO].to_s[0...5] == HTTPS || env[HTTP_X_FORWARDED_SCHEME] == HTTPS || env[HTTP_X_FORWARDED_SSL] == "on"
```
nateberkopec: OK - now that we've refactored this out a bit, line 591 and the if/if/else in like 634 look really similar. Is there any possibility for re-use here?

test/test_puma_server.rb
```diff
@@ -104,7 +104,64 @@ def test_respect_x_forwarded_proto
...
+    end
+
+    @server.add_tcp_listener @host, @port
+    @server.run
+
+    req = Net::HTTP::Get.new("/")
+    req['HOST'] = "example.com"
+    req['X_FORWARDED_SSL'] = "on"
+
+    res = Net::HTTP.start @host, @server.connected_port do |http|
```
nateberkopec: Hmmm. This doesn't match our current style of just writing onto a TCP socket. Thoughts on that?

test/test_puma_server.rb
```diff
@@ -104,7 +104,64 @@ def test_respect_x_forwarded_proto
...
+      http.request(req)
+    end
+
+    assert_equal "80", res.body
+  end
+
+  def test_respect_x_forwarded_scheme
+    @server.app = proc do |env|
```
nateberkopec: Once my other comment is resolved - the setup in all of these is so similar, we should extract to a helper method.

test/test_puma_server.rb
```diff
@@ -104,7 +104,64 @@ def test_respect_x_forwarded_proto
...
+
+    req = Net::HTTP::Get.new("/")
+    req['HOST'] = "example.com"
+    req['X_FORWARDED_SSL'] = "off"
+
+    res = Net::HTTP.start @host, @server.connected_port do |http|
+      http.request(req)
+    end
+
+    assert_equal "80", res.body
```
nateberkopec: I wonder - why not just test the actual methods modified rather than test res.body? This test is testing a lot more than it has to.

test/test_puma_server.rb
olleolleolle: I'd be glad to!

test/test_puma_server.rb
olleolleolle: Eh, I only repeated test bodies which were there, tests that checked similar things, to have something to start with.

When I refactor this, I'll try to narrow down the tests to be more about the methods I changed.

olleolleolle:
@nateberkopec I took a closer look, and made follow-up commits.

nateberkopec:
🚢 🚢 🚢

nateberkopec:
Thanks for all your hard work here, @olleolleolle !

PR #1498, opened 2017-12-31
 CI against ruby 2.5.0

ttanimichi:


olleolleolle:
The failing build is fixed by #1506 - which removes 2.1.

PR #1500, opened 2018-01-04
Skip SSL tests instead silently passing them.

voxik:
If SSL is not used, the tests might be silently passing, while they are
not executed at all.

Let me explain.

The think is that you might not be testing some functionality of Puma and you might not even notice it, because the behavior of test suite apparently depends on load order.

E.g. the test suite is failing for me when I try to execute the test suite like this:

~~~
$ ruby -e 'Dir.glob "./test/**/test_*.rb", &method(:require)' -- -v

... snip ...

  1) Failure:
TestPumaServerSSL#test_ssl_v3_rejection [/builddir/build/BUILD/puma-3.11.0/usr/share/gems/gems/puma-3.11.0/test/test_puma_server_ssl.rb:119]:
Expected /wrong\ version\ number/ to match # encoding: ASCII-8BIT
"OpenSSL error: error:1417D102:SSL routines:tls_process_client_hello:unsupported protocol - 337105154".

177 runs, 372 assertions, 1 failures, 0 errors, 0 skips
~~~

But if you use the right load order, the test suite pass, although it ignores some tests:

~~~
$ ruby -r./test/test_puma_server_ssl -e 'Dir.glob "./test/**/test_*.rb", &method(:require)' -- -v

... snip ...

177 runs, 351 assertions, 0 failures, 0 errors, 0 skips
~~~

But with the patch applied, you can see that some tests were actually skipped:

~~~
$ ruby -r./test/test_puma_server_ssl -e 'Dir.glob "./test/**/test_*.rb", &method(:require)' -- -v

... snip ...

177 runs, 351 assertions, 0 failures, 0 errors, 4 skips
~~~

### DIFF
--- a/test/test_puma_server_ssl.rb
+++ b/test/test_puma_server_ssl.rb
@@ -21,7 +21,7 @@ DISABLE_SSL = begin
 class TestPumaServerSSL < Minitest::Test
 
   def setup
-    return if DISABLE_SSL
+    skip("SSL isn't used.") if DISABLE_SSL
     @port = 3212
     @host = "127.0.0.1"
 
@@ -55,8 +55,6 @@ class TestPumaServerSSL < Minitest::Test
   end
 
   def test_url_scheme_for_https
-    return if DISABLE_SSL
-
     body = nil
     @http.start do
       req = Net::HTTP::Get.new "/", {}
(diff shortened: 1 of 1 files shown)


voxik:
Now I noticed that the "TestPumaServerSSLClient" tests suffers the same issue. BTW this was introduced in d421a538252751b0cccf4638af

olleolleolle:
This looks like a good update to the test suite!

schneems:
This broke master, the scope of the variable @host was changed to host in another commit. Going to try working on a fix on #1542

Issue #1504, opened 2018-01-19
Segfault in puma 3.11.1


vincentwoo:
After @evanphx released 3.11.1, I just got the following segfault during normal server operation:

```
/Users/vwoo/.gem/ruby/2.5.0/gems/puma-3.11.1/lib/puma/client.rb:293: [BUG] Segmentation fault at 0x0000000000000008
ruby 2.5.0p0 (2017-12-25 revision 61468) [x86_64-darwin17]

-- Crash Report log information --------------------------------------------
   See Crash Report log file under the one of following:
     * ~/Library/Logs/DiagnosticReports
     * /Library/Logs/DiagnosticReports
   for more details.
Don't forget to include the above Crash Report log file in bug reports.

-- Control frame information -----------------------------------------------
c:0006 p:---- s:0042 e:000041 CFUNC  :execute
c:0005 p:0064 s:0035 e:000034 METHOD /Users/vwoo/.gem/ruby/2.5.0/gems/puma-3.11.1/lib/puma/client.rb:293
c:0004 p:0035 s:0030 e:000029 METHOD /Users/vwoo/.gem/ruby/2.5.0/gems/puma-3.11.1/lib/puma/client.rb:350
c:0003 p:0036 s:0026 e:000025 BLOCK  /Users/vwoo/.gem/ruby/2.5.0/gems/puma-3.11.1/lib/puma/server.rb:280 [FINISH]
c:0002 p:0125 s:0016 E:0002e0 BLOCK  /Users/vwoo/.gem/ruby/2.5.0/gems/puma-3.11.1/lib/puma/thread_pool.rb:120 [FINISH]
c:0001 p:---- s:0003 e:000002 (none) [FINISH]

-- Ruby level backtrace information ----------------------------------------
/Users/vwoo/.gem/ruby/2.5.0/gems/puma-3.11.1/lib/puma/thread_pool.rb:120:in `block in spawn_thread'
/Users/vwoo/.gem/ruby/2.5.0/gems/puma-3.11.1/lib/puma/server.rb:280:in `block in run'
/Users/vwoo/.gem/ruby/2.5.0/gems/puma-3.11.1/lib/puma/client.rb:350:in `eagerly_finish'
/Users/vwoo/.gem/ruby/2.5.0/gems/puma-3.11.1/lib/puma/client.rb:293:in `try_to_finish'
/Users/vwoo/.gem/ruby/2.5.0/gems/puma-3.11.1/lib/puma/client.rb:293:in `execute'

[...]

PR #1505, opened 2018-01-19
Fix segfault in HttpParse_execute

vincentwoo:
Fixes https://github.com/puma/puma/issues/1504

I've chosen to have HttpParse_execute simply early terminate if passed nil as `data` rather than trying to ensure that it is never passed `nil` from outside, since there doesn't seem to be existing tests for `client.rb`, but it's also totally reasonable to just have `parser.execute` only be invoked when you know you have data for it.

I've also added a couple QOL things to fix a compiler warning, whitespace, and make it so you can just run the http11 tests without the whole suite.

### DIFF
--- a/ext/puma_http11/http11_parser.c
+++ b/ext/puma_http11/http11_parser.c
@@ -41,14 +41,11 @@ static const int puma_parser_start = 1;
 static const int puma_parser_first_final = 47;
 static const int puma_parser_error = 0;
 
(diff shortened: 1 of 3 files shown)


evanphx:
I fixed this via 49ed71ad3ea3e173a05de5c621c712549b496587 instead.

vincentwoo:
Thanks!

Issue #1480, opened 2017-12-05
Incorrect chunked transfer encoding handling


petr666:
Hi,
version 3.7.0 (probably change "Pickup any remaining chunk data as the next request") introduced bug with consequent chunked requests.
It seems puma implementation does not conform to https://tools.ietf.org/html/rfc7230#section-4.1. Valid chunked body is:

chunked-body = *chunk
                           last-chunk
                           trailer-part
                           CRLF

but Puma does not count with the last CRLF which is wrongly appended to consequent request, while printing:

HTTP parse error, malformed request (): #<Puma::HttpParserError [047a99aabe6323c0020b9e290f19d997]: Invalid HTTP format, parsing fails.>

I've tested this via Wireshark. When last CRLF is not present, puma sends response without printing any error. But Wireshark does not display request, because it is incomplete/invalid. In case last CRLF is present, Wireshark prints request, but puma prints error and consequent request within same connection fails.

ademars94:
+1 Also experiencing issues with chunked request.

100cm:
its's still not work in puma 3.12.0

![image](https://user-images.githubusercontent.com/12691638/55848437-86ad7680-5b7f-11e9-8f84-a9fff93b6823.png)

and the error is 

![image](https://user-images.githubusercontent.com/12691638/55848452-9036de80-5b7f-11e9-8d4e-6fbe06b5f62d.png)

May rack still read content_Length?

![image](https://user-images.githubusercontent.com/12691638/55848479-a47adb80-5b7f-11e9-8a5d-47375ae5dbb2.png)

PR #1508, opened 2018-01-24
Fix 3 corner cases when request body is chunked

florin555:
I have found and fixed 3 cases when puma does not read a chunked body correctly.

In the first case it hangs.

In the second and third cases it adds extra bytes to the body.

Please see the individual commit messages for more information.

### DIFF
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -147,6 +147,7 @@ module Puma
         if @partial_part_left <= chunk.size
           @body << chunk[0..(@partial_part_left-3)] # skip the \r\n
           chunk = chunk[@partial_part_left..-1]
+          @partial_part_left = 0
         else
           @body << chunk
           @partial_part_left -= chunk.size
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -171,6 +171,7 @@ module Puma
           if len == 0
             @body.rewind
             rest = io.read
+            rest = rest[2..-1] if rest.start_with?("\r\n")
             @buffer = rest.empty? ? nil : rest
             @requests_served += 1
             @ready = true
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -211,7 +211,7 @@ module Puma
       while true
         begin
           chunk = @io.read_nonblock(4096)
-        rescue Errno::EAGAIN
+        rescue IO::WaitReadable
           return false
         rescue SystemCallError, IOError
           raise ConnectionError, "Connection error detected during read"
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -145,7 +145,9 @@ module Puma
     def decode_chunk(chunk)
       if @partial_part_left > 0
         if @partial_part_left <= chunk.size
-          @body << chunk[0..(@partial_part_left-3)] # skip the \r\n
+          if @partial_part_left > 2
+            @body << chunk[0..(@partial_part_left-3)] # skip the \r\n
+          end
           chunk = chunk[@partial_part_left..-1]
           @partial_part_left = 0
         else
--- a/test/test_puma_server.rb
+++ b/test/test_puma_server.rb
@@ -658,6 +658,38 @@ EOF
     assert_equal "hello", body
   end
 
+  def test_chunked_request_pause_between_cr_lf_after_size_of_second_chunk
+    body = nil
+    @server.app = proc { |env|
+      body = env['rack.input'].read
+      [200, {}, [""]]
+    }
+
+    @server.add_tcp_listener @host, @port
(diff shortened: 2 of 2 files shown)


olleolleolle:
Hi, Florian, cool changes!

I went and read the AppVeyor failed-build output.

This is the output message from the failed AppVeyor build: does it tell you anything?

```
  1) Error:
TestPumaServer#test_chunked_request_pause_between_cr_lf_after_size_of_second_chunk:
Errno::ECONNABORTED: An established connection was aborted by the software in your host machine.
    C:/projects/puma/test/test_puma_server.rb:686:in `read'
    C:/projects/puma/test/test_puma_server.rb:686:in `test_chunked_request_pause_between_cr_lf_after_size_of_second_chunk'
```

<details>

```
TestPumaServer#test_chunked_request_pause_between_cr_lf_after_size_of_second_chunk = [MinitestRetry] retry 'test_chunked_request_pause_between_cr_lf_after_size_of_second_chunk' count: 1,  msg: Errno::ECONNABORTED: An established connection was aborted by the software in your host machine.
    C:/projects/puma/test/test_puma_server.rb:686:in `read'
    C:/projects/puma/test/test_puma_server.rb:686:in `test_chunked_request_pause_between_cr_lf_after_size_of_second_chunk'
[MinitestRetry] retry 'test_chunked_request_pause_between_cr_lf_after_size_of_second_chunk' count: 2,  msg: Errno::ECONNABORTED: An established connection was aborted by the software in your host machine.
    C:/projects/puma/test/test_puma_server.rb:686:in `read'
    C:/projects/puma/test/test_puma_server.rb:686:in `test_chunked_request_pause_between_cr_lf_after_size_of_second_chunk'
[MinitestRetry] retry 'test_chunked_request_pause_between_cr_lf_after_size_of_second_chunk' count: 3,  msg: Errno::ECONNABORTED: An established connection was aborted by the software in your host machine.
    C:/projects/puma/test/test_puma_server.rb:686:in `read'
    C:/projects/puma/test/test_puma_server.rb:686:in `test_chunked_request_pause_between_cr_lf_after_size_of_second_chunk'
2.05 s = E
```

</details>

florin555:
Hi! Thanks for the feedback.

I made some exploratory commits and I was able to isolate the source of the problem to the fact that in that particular test I used 2 sleeps and wrote to the socket 2 times. All other tests sleep and write to the socket at most once.

Otherwise the test follows the same pattern like all the other tests.

I was not able to reproduce the failure on my computer, and also Travis CI is passing, so I believe it must be something specific to Appveyor, but I don't have a way to debug further.

The test needs the 2 sleeps because the bug that the test tries to reproduce happens only in that case, so I can't just change the test because then it becomes meaningless.

olleolleolle:
@florin555 _Perhaps_ you know how to make a focused query to https://help.appveyor.com/discussions/problems ?

nateberkopec:
Is this related to #1480?

florin555:
https://github.com/puma/puma/issues/1480 is different. This pull request fixes other issues.

mubashirhanif:
@florin555 and I think that the failing test is because of a potential bug in ruby on windows.

```
# https://github.com/preisanalytics/puma/blob/382cdaf2cdbb781f91d7e0ee8a36c6d8b4c88c66/lib/puma/client.rb#L203
def read_chunked_body
      ...
        rescue Errno::EAGAIN
          return false
        rescue SystemCallError, IOError
          raise ConnectionError, "Connection error detected during read"
        end
      ...
end
```
On Linux and Mac the exception `Errno::EWOULDBLOCK` is just an alias to `Errno::EAGAIN`.
On windows, these exceptions point to 2 different objects/exceptions, which causes the code to raise a wrong exception instead of returning false(the expected behaviour).

Also note `[Errno::EAGAIN::Errno, Errno::EWOULDBLOCK::Errno]` outputs
`[11, 11]` on linux, `[35, 35]` on Mac and `[11, 10035]` on windows(verified on 1.9.3 and 2.3.3). 

In this ruby-doc it says that both `Errno::EWOULDBLOCK` and `Errno::EAGAIN` are extended by `IO::WaitReadable`(verified on both Windows and Linux). And also since 1.9.3 is support is dropped by Puma. So I think we ca we can safely replace `Errno::EAGAIN` and `Errno::EWOULDBLOCK` with `IO::WaitReadable` in case of exception from `IO::read_nonblock` method.

mubashirhanif:
The current failing builds on Travis are inherited from the current state of master branch.

florin555:
@olleolleolle, @nateberkopec could you please merge this, now that the build is green?

pschrammel:
hi, sorry pushing: Are there any merge blockers? How can this go upstream?

Issue #911, opened 2016-02-28
SIGTERM fails to kill puma running as a rack handler


veganstraightedge:
I'm running Ruby 2.3.0, Rails 4.2.5.1 and Puma 3.0.2 on Mac OS X 10.11.3 (15D21).

# Repro Steps

1. Open Terminal.app
2. Start Puma in a (pretty vanilla Rails) with `rails s`. (Starts on port 3000. All systems nominal.)
3. Quit Terminal.app without explicitly stopping Puma (with control-c).
4. Re-open Terminal.app
5. Re-start Puma (same as step 2)

```
=> Booting Puma
=> Rails 4.2.5.1 application starting in development on http://localhost:3000
=> Run `rails server -h` for more startup options
=> Ctrl-C to shutdown server
A server is already running. Check /Users/s/Dropbox/Projects/darkmatter/org-charter-server-rails/tmp/pids/server.pid.
Exiting
```

PR #1527, opened 2018-03-09
allow app to also handle sighup and call super

grosser:
our app wants to reopen rails logs on sighup
to do that it needs to trap sighup too
but calling pumas trap after does not work since it crashes the server if redirect_io was not used

with this change there is 1 less trap used and our app can only call pumas trap if it was set

see https://github.com/zendesk/samson/pull/2637

/cc @ragurney @jonmoter

### DIFF
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -414,16 +414,14 @@ module Puma
         log "*** SIGINT not implemented, signal based gracefully stopping unavailable!"
       end
 
-      begin
-        Signal.trap "SIGHUP" do
-          if @runner.redirected_io?
+      if @runner.redirected_io?
+        begin
+          Signal.trap "SIGHUP" do
             @runner.redirect_io
-          else
-            stop
           end
+        rescue Exception
+          log "*** SIGHUP not implemented, signal based logs reopening unavailable!"
         end
-      rescue Exception
-        log "*** SIGHUP not implemented, signal based logs reopening unavailable!"
       end
     end
   end


nateberkopec:
Doesn't this change mean SIGHUP will no longer call `stop`?

grosser:
it will kill the process (default behavior of SIGHUP) ... not sure if that's bad ... would certainly stop the server :D

```
ruby -e sleep
kill -HUP 39052
Hangup: 1
```

lib/puma/launcher.rb
```diff
@@ -414,16 +414,14 @@ def setup_signals
...
       end
 
-      begin
-        Signal.trap "SIGHUP" do
-          if @runner.redirected_io?
+      if @runner.redirected_io?
+        begin
+          Signal.trap "SIGHUP" do
             @runner.redirect_io
-          else
```
nateberkopec: @grosser What I meant was you eliminated this branch of code, and I don't understand why.

grosser: I was basically thinking "if you send SIGHUP to a server that does not have any reload mechanic then wtf are you doing" ... but maybe it's valid usecase :(

nateberkopec: I think it was important to #1377?

grosser:
that PR is about SIGINT not SIGHUP

stop for SIGHUP was added in
https://github.com/puma/puma/commit/eb6468d5585c01287b72a92e797773bcb123c1af
to fix https://github.com/puma/puma/issues/911 which is about SIGTERM ...
idk kinda confusing :D

nateberkopec:
SIGWHYDONTWECALLTHEWHOLETHINGOFF

grosser:
#911 has reproduction steps, so if they still work then it's ok ?
it's not really an urgent/important change, just 1-more-weirdness in the codebase 🤷‍♂️

evanphx:
I'm going to reject this PR because it changes the meaning of SIGHUP for existing users.

PR #1529, opened 2018-03-11
Freebsd rc.d jungle for puma

desnudopenguino:
I'm a FreeBSD Rails user, and ran into the issue of not having the service start at (re)boot. I only have recent experience with rbenv, but someone else should be able to make sense and add other environment systems like rvm.

It kind of takes a different approach in the config file by using json, but I felt it would be a little less cumbersome to process that instead of writing overly complex shell logic to parse configs.


nateberkopec:
Thanks for sharing.

PR #1532, opened 2018-03-12
Expose top level Puma.stats API

schneems:
Right now to get stats for a puma process the only way is to go through the control server. The puma-stats-logger does this by reading in a state file then querying the control server manually.

https://github.com/hired/puma-stats-logger/blob/7ad7798e9d06ff44e047ac56d0307c4ff9c73751/lib/puma_stats_logger/middleware.rb#L28

Instead, I’m proposing adding a top level `Puma.stats` method that will allow anyone inside of the same process to get access to the stats.

This could be instrumented by a gem to theoretically export these stats to say a Heroku dashboard where we could list out backlog or thread count.

The format of stats is a hash, and will change depending on if the server is in “single” or “clustered” mode. 

Clustered:

```
{ "workers": 2, "phase": 0, "booted_workers": 2, "old_workers": 0, "worker_status": [{ "pid": 19832, "index": 0, "phase": 0, "booted": true, "last_checkin": "2018-03-12T16:03:12Z", "last_status": { "backlog":0, "running":5 } },{ "pid": 19833, "index": 1, "phase": 0, "booted": true, "last_checkin": "2018-03-12T16:03:12Z", "last_status": { "backlog":0, "running":5 } }] }
```

Single: 

```
{ "backlog": 0, "running": 2 }
```

Alternatively if we could somehow enable another process to get these stats from Puma via pumactl by default without requiring any additional in app config, that would also work.


lib/puma/launcher.rb
```diff
@@ -86,6 +86,7 @@ def initialize(conf, launcher_args={})
       else
         @runner = Single.new(self, @events)
       end
+      Puma.stats_object = @runner
```
twalpole: There is no `Puma.stats_object` so this causes an error

twalpole: Nevermind - requiring 'rack/handler/puma' doesn't trigger a require of 'puma'  so `Puma.stats_object=` wasn't being defined.

schneems: Is there an issue that needs to be fixed? Are we not testing something with the handler that we should?

twalpole: @schneems Not really sure -- Capybara was just requiring 'rack/handler/puma' when setting up puma, which used to be fine, but with this change resulted in an error because `Puma.stats_object=` wasn't defined.  If Capybara requires 'puma' as well the error goes away.  Whether or not just requiring `rack/handler/puma` should be enough is up to you.

schneems: This should fix it https://github.com/puma/puma/pull/1546

jjb:
Hi @schneems  - have you seen my work here?

https://github.com/puma/puma/issues/1512#issuecomment-362669497

I accessed stats by looking through memory for the instance of `Puma::Launcher`. 😆 

The work you are doing is going to make this much less hacky -- awesome!

schneems:
@jjb great! I'm trying to log backlog as an optional metric at Heroku https://github.com/heroku/barnes/pull/14. I took a look at puma-stats-logger originally and was surprised to find out the hoops that had to be jumped through to get the info. Seems like this patch already satisfies a few use cases :)

lykaios:
Thanks for this work. 

Has it already made it into a release? When I try using the API I get these error: 
```
NameError: uninitialized constant Puma::DSL::PumaStats
  config/puma.rb:19:in `block in _load_from'
  /Users/nate.allen/.rbenv/versions/2.1.10/lib/ruby/gems/2.1.0/gems/puma-3.11.4/lib/puma/configuration.rb:276:in `call'
  /Users/nate.allen/.rbenv/versions/2.1.10/lib/ruby/gems/2.1.0/gems/puma-3.11.4/lib/puma/configuration.rb:276:in `block in run_hooks'
  /Users/nate.allen/.rbenv/versions/2.1.10/lib/ruby/gems/2.1.0/gems/puma-3.11.4/lib/puma/configuration.rb:276:in `each'
  /Users/nate.allen/.rbenv/versions/2.1.10/lib/ruby/gems/2.1.0/gems/puma-3.11.4/lib/puma/configuration.rb:276:in `run_hooks'
  /Users/nate.allen/.rbenv/versions/2.1.10/lib/ruby/gems/2.1.0/gems/puma-3.11.4/lib/puma/cluster.rb:451:in `run'
  /Users/nate.allen/.rbenv/versions/2.1.10/lib/ruby/gems/2.1.0/gems/puma-3.11.4/lib/puma/launcher.rb:184:in `run'
  /Users/nate.allen/.rbenv/versions/2.1.10/lib/ruby/gems/2.1.0/gems/puma-3.11.4/lib/puma/cli.rb:78:in `run'
  /Users/nate.allen/.rbenv/versions/2.1.10/lib/ruby/gems/2.1.0/gems/puma-3.11.4/bin/puma:10:in `<top (required)>'

```

schneems:
Where is `Puma::DSL::PumaStats` coming from? Can you give me an example app that reproduces the problem and open a new issue?

lykaios:
You can disregard - I ended up going with a different approach.

jjb:
Final got around to using the new API. Thanks @schneems!

😂 

screen shot 2018-07-16 at 4 52 03 pm

yob:
Thanks for this - I was able to leverage it to build a small puma plugin that passes the numbers to statsd every few seconds: https://github.com/yob/puma-plugin-statsd

sj26:
Yes!!! Thank you!!!

PR #1533, opened 2018-03-13
Add warning about init.d being replaced by systemd

j15e:
Using init.d script in a systemd environment will work for jobs start/stop, but restart will not be managed as expected with old init.d system. This may cause confusion, pain and a lot of systemd search time.

Also move `init.d` last in the jungle documentation.


nateberkopec:
Thanks!

PR #1534, opened 2018-03-13
Add a notice for systemd with socket path in release path

j15e:
I think this would be good to know for systemd users, we lost a lot of time on this problem before pin-pointing the real issue.


nateberkopec:
Thanks

PR #1538, opened 2018-03-17
Appveyor fixes

MSP-Greg:
This PR stabilizes Appveyor testing, along with removing Ruby 2.1, and adding Ruby 2.5.  I realize that using Puma in a Windows production environment is probably uncommon, but it may be used for development.

Also, current windows Ruby builds (2.5 & trunk) are compiled with gcc 7.3.0 and using OpenSSL 1.1.0.  Since Travis is using an older toolset, Appveyor testing may show issues due to the newer 'toolset' used to build & test Puma.

Since:
1. I am not familiar with using Puma in a production environment
2. Testing may not show all issues, and often cannot duplicate production loading 
3. Some of my previous PR's required patching (sorry @evanphx )
4. I do use Puma locally, but normally without SSL.

I thought I'd describe what issues were involved and my solutions.  They are probably not optimal.

### Issue 1 - SSLv3 Rejection
```
TestPumaServerSSL#test_ssl_v3_rejection`:
TimeoutEveryTestCase::TestTookTooLong: execution expired
C:/projects/puma/lib/puma/server.rb:940:in `join'
C:/projects/puma/lib/puma/server.rb:940:in `stop'
C:/projects/puma/test/test_puma_server_ssl.rb:54:in `
```

This issue was consistent with both Ruby 2.5 & trunk, both of which build using OpenSSL 1.1.0g.  The solution chosen was to use `@server.thread.kill` instead of `@server.close` for this test, along with a test for `windows?`.  A boolean `@no_teardown` variable is used in `test_puma_server_ssl.rb` to bypass `teardown` when this is done.

### Issue 2 - IOError - stream already closed
```
TestConfigFile#test_config_files_with_rack_env = C:/projects/puma/lib/puma/minissl.rb:252:in `close': closed stream (IOError)
	from C:/projects/puma/lib/puma/minissl.rb:252:in `close'
```

Currently, `Puma::MiniSSL::Server#close` consistes of `@socket.close`.  Changed to `@socket.close unless @socket.closed?`.

### Issue 3 - Intermittent Errno::EADDRINUSE

These errors originate in `add_ssl_listener` or `add_tcp_listener`, although tcp seems to be affected only when ssl tests were run previously.  Fixes were windows specific changes to `test_puma_server_ssl.rb`.

### Issue 4 - Intermittent ragel build failures

Added the ragel package to the MSYS2 system before compiling, this was done in appveyor.yml.

### Notes:

From working with ruby trunk testing for well over a year, Windows sockets do have some differences from *nix sockets.  For quite a while, both Windows MinGW & MSWin builds have been passing all Ruby Core tests.

Cleaned up caching on AV, the last test on my Puma fork showed all gems coming from the cache.

As mentioned, these fixes may not be optimal, but even it they're a starting point, they do seem to consistently pass AV tests.

The only build warning from OpenSSL 1.1.0g is:
```
compiling ../../../../ext/puma_http11/mini_ssl.c
../../../../ext/puma_http11/mini_ssl.c: In function 'engine_init_client':
../../../../ext/puma_http11/mini_ssl.c:211:3: warning: 'DTLSv1_method' is deprecated [-Wdeprecated-declarations]
   conn->ctx = SSL_CTX_new(DTLSv1_method());
   ^~~~
In file included from C:/msys64/mingw64/include/openssl/ct.h:13:0,
                 from C:/msys64/mingw64/include/openssl/ssl.h:61,
                 from ../../../../ext/puma_http11/mini_ssl.c:15:
C:/msys64/mingw64/include/openssl/ssl.h:1642:1: note: declared here
 DEPRECATEDIN_1_1_0(__owur const SSL_METHOD *DTLSv1_method(void)) /* DTLSv1.0 */
 ^
```
If I need to change anything or resubmit this PR with changes, I'd be happy to.  I would just like to get Appveyor testing to be consistent and passing.

Thanks, Greg

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,8 @@
+# cache cleanup 2018-03-16
+
 init:
   - set PATH=C:\ruby%ruby_version%\bin;C:\Program Files\7-Zip;C:\Program Files\AppVeyor\BuildAgent;C:\Program Files\Git\cmd;C:\Windows\system32
+  # Download current trunk, install OpenSSL via trunk_pkgs.cmd file
   - if %ruby_version%==_trunk (
       appveyor DownloadFile https://ci.appveyor.com/api/projects/MSP-Greg/ruby-loco/artifacts/ruby_trunk.7z -FileName C:\ruby_trunk.7z &
       7z x C:\ruby_trunk.7z -oC:\ruby_trunk &
@@ -7,7 +10,11 @@ init:
     )
 
 install:
-  # Download RI OpenSSL Knapsack package
+  # Install ragel
+  - if "%ri_file%"=="x64_2" ( C:\msys64\usr\bin\pacman -S --noconfirm mingw-w64-x86_64-ragel )
+  - if "%ri_file%"=="x86_2" ( C:\msys64\usr\bin\pacman -S --noconfirm mingw-w64-i686-ragel   )
+  
+  # Install ragel & download RI OpenSSL Knapsack package
   # RI DevKit is only installed in Ruby23 and Ruby23-x64 folders
   # RI2 MSYS2/MinGW OpenSSL package is standard Appveyor item
   - if "%ri_file%"=="x86" (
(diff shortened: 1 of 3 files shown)


appveyor.yml
```diff
@@ -1,35 +1,57 @@
+# cache cleanup 2018-03-16
```
schneems: What's this comment referring to? Is it extra?

MSP-Greg: It's a note I left in for my fork testing.  With Appveyor, there is no way of resetting a project's cache from the web ui, so, over time, the cache may contain old gem dependencies, etc.  I can delete the line if you'd like.

MSP-Greg:
@schneems

> It looks like you put some code into that method, would your fixes be related to that? It looks like all your appveyor tests are passing.

Well, when posting those changes, I was focused on 'green', and overlooked an error in not allowing for all possible return values.  Again, apologies for that.

Some of the failures fixed in this PR are intermittent, and from working with trunk testing in Appveyor, there are some differences re sockets, etc.  As to whether they're windows, gcc 7.3.0, or OpenSSL 1.1.0 related, I'm not totally sure.

So, with the exception of the patch to `minissl.rb`, I left all the changes in the test files.  I've run it several times on my fork, and the tests are stable for all Ruby versions.

Thanks, Greg

lib/puma/minissl.rb
```diff
@@ -249,7 +249,7 @@ def accept_nonblock
       end
 
       def close
-        @socket.close
+        @socket.close unless @socket.closed?       # closed? call is for Windows
```
nateberkopec: 👍

test/test_puma_server_ssl.rb
```diff
@@ -20,38 +20,46 @@ def ssl_error(server, peeraddr, peercert, error)
 
 class TestPumaServerSSL < Minitest::Test
 
+  # Only @events, @http, and @server need to be instance variables
```
nateberkopec: 👍 but the comment isn't necessary

MSP-Greg: Confused.  Would you like it removed?  I added it for people that may not read the PR/commit description.

MSP-Greg: Just rebased from today's commits and deleted comment...

schneems:
Thanks for the PR!

PR #1545, opened 2018-03-21
jungle: Using `ps -p` will check correctly whether $PID exists

hoshinotsuyoshi:
I think that It is appropriate to use `ps -p` in that place.

I think that under certain conditions, existing code ``[ "`ps -A -o pid=
| grep -c $PID`" -eq 0 ]`` may behave incorrectly.

For example, if the content of $PID happens to be a short string(such as `199`), and another process includes the string(process such as `2199`), then unfortunately `grep -c` will return
positive number even if `199` is not running.

### DIFF
--- a/tools/jungle/init.d/puma
+++ b/tools/jungle/init.d/puma
@@ -48,7 +48,7 @@ do_start_one() {
   if [ -e $PIDFILE ]; then
     PID=`cat $PIDFILE`
     # If the puma isn't running, run it, otherwise restart it.
-    if [ "`ps -A -o pid= | grep -c $PID`" -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


nateberkopec:
Paging previous contributors @vipulnsward @michaelsauter @snow @kwilczynski

michaelsauter:
Thanks for paging! We never had the issue described here, but I'd agree with @hoshinotsuyoshi that this needs fixing.

kwilczynski:
Hi @hoshinotsuyoshi, thank you fixing this!

A side note - this whole script is almost crying out loud to be re-written and maintained, somewhat.

hoshinotsuyoshi:
Thank you!

PR #1546, opened 2018-03-21
Ensure Puma.stats_object available using handler

schneems:
When requiring only the handler puma.rb is not loaded. This defines a method that is used in the launcher https://github.com/puma/puma/pull/1532#discussion_r176171578.


twalpole:
LGTM

PR #1540, opened 2018-03-20
Travis - update Ruby 2.2 thru 2.4 versions, add 2.5.0

MSP-Greg:

PR #1552, opened 2018-03-30
Use Ruby 2.2.9 on Travis again

eugeneius:
Followup to https://github.com/puma/puma/pull/1540 and https://github.com/puma/puma/pull/1542.

The test failures we saw after upgrading last time were caused by the stopgap_13632 gem not being activated on the new version.

I would have jumped straight to 2.2.10 here, but the stopgap_13632 gem hasn't been updated to cover that version yet - I've opened https://github.com/NickolasVashchenko/stopgap_13632/pull/6 to do that.


NickolasVashchenko:
Merged your PR and released:
https://rubygems.org/gems/stopgap_13632/versions/1.2.0

PR #1559, opened 2018-04-12
Add ruby-head with enabled JIT option into Travis matrix

deepj:
I think Puma would be tested against Ruby JIT since Puma is one of the most important Ruby projects. And it'd be nice to have it prepared to work with JIT in the advance.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,9 @@ rvm:
   - rbx-3
 matrix:
   fast_finish: true
+  include:
+    - rvm: ruby-head
+      env: RUBYOPT="--jit"
   allow_failures:
     - rvm: jruby-head
     - rvm: rbx-3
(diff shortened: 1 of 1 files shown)


nateberkopec:
I agree, but this should be in `allow-failures`.

deepj:
@nateberkopec Done. But it failed on jruby problem https://travis-ci.org/puma/puma/jobs/377248677

deepj:
@nateberkopec I'd be in to test it against JIT option with compilation and testing on the Windows. Unfortunately, I don't get how to set it up for Appveyor since I've never used Appveyor.

MSP-Greg:
@deepj

Re Appveyor, I'll add it to one or another of the versions...  Today's result (build with OpenSSL 1.1.0h)
```
183 runs, 351 assertions, 0 failures, 0 errors, 18 skips
ruby 2.6.0dev (2018-07-11 trunk 63941) +JIT [x64-mingw32]
```

schneems:
Thanks for the PR!

Issue #980, opened 2016-05-12
Drop TLS 1.0?


chewi:
We're trying to pass a PCI DSS audit and we noticed that Puma is hardcoded to allow TLS 1.0. It's not completely outlawed yet (2 years left) but you do have to state why you are using it and have mitigations in place. We'd rather just rule it out entirely. Given that you generally shouldn't have Puma directly exposed to the Internet, I don't see why old clients need to be supported. The headline casualties are Android 4.3 (Jelly Bean) and IE7. Of course, it could be made configurable but that's non-trivial. :wink:

evanphx:
I'll likely add code to remove it being an option in another year. There was someone asking about the ability to override the cipher settings as well. I might do that.

chewi:
Happy new year! Our next audit is coming up. What's the current feeling about this?

Issue #1561, opened 2018-04-15
Puma is using MiniSSL. What is the difference between MiniSSL and OpenSSL?


BlackNightFury:
Hi. I am using the Puma.

But I can see that Puma is using MiniSSL and it hasn't some functionalities which OpenSSL has.

Why is Puma using MiniSSL?

perlun:
@BlackNightFury MiniSSL is not a concept in itself, more than the class name of Puma's SSL implementation which can be found here: https://github.com/puma/puma/blob/master/lib/puma/minissl.rb

I don't have much experience with it myself, but is currently looking into start using it for development & production usage. To the best of my knowledge "it works" (even though I've seen some exceptions & weird behavior; will file issues about them if I can reproduce it.)

What are the _specific_ things in "OpenSSL" (a C library used for implementing SSL/TLS support) that are missing in Puma? If you make a list and be very specific, I am sure you will get more feedback from the project maintainers. (I am not a maintainer myself.)

BlackNightFury:
@perlun Thanks for your response.

So we have been using OpenSSL and our project needs strong security.
We planned to move from thin to puma because of concurrency issue but we also could see that Puma is using MiniSSL not OpenSSL.

So can you guarantee that MiniSSL is same as OpenSSL?

perlun:
@BlackNightFury 

> So we have been using OpenSSL and our project needs strong security.

Please define "strong security".

> We planned to move from thin to puma because of concurrency issue but we also could see that Puma is using MiniSSL not OpenSSL.
> 
> So can you guarantee that MiniSSL is same as OpenSSL?

I can most certainly guarantee that MiniSSL is _not_ the same as OpenSSL. [OpenSSL](https://www.openssl.org/) is a library which implements the SSL/TLS protocols. [LibreSSL](http://www.libressl.org/) is another library which was forked from OpenSSL a few years ago.

Can you guarantee that LibreSSL is the same as OpenSSL? Probably mostly yes, since they share the same ancestry, but there might still be subtle differences.

Can you guarantee that Puma's MiniSSL implementation is the same as OpenSSL? No, it's a different implementation of its own.

I guess what you're really asking is whether Puma's MiniSSL imlpementation is _secure enough_. To that, I cannot really answer - @evanphx or @nateberkopec are probably much better suited to answer that. (I would also be personally interested in the answer btw 😄)

BlackNightFury:
@perlun Thanks for your answers.
[...]

PR #1562, opened 2018-04-17
add support for disabling TLSv1.0

skrobul:
Many organizations run their applications using in environments that fall into scope of PCI-DSS compliance audits. One of the requirements set out by standard is to migrate to more secure protocols if possible.

PCI Security Standards council has advised to migrate away from TLSv1.0 over last few years and recently set a migration deadline of 30 June 2018 (see [this article](https://blog.pcisecuritystandards.org/are-you-ready-for-30-june-2018-sayin-goodbye-to-ssl-early-tls) for more details).

Change proposed in this commit gives an user option to disable `TLSv1.0` during bind, while still leaving the `TLSv1.1` and `TLSv1.2` enabled. `SSLv2` and `SSLv3` are permanently disabled (as they should).

Default behaviour is not changed if the `no_tls` option is not defined.

Related: #980

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,12 @@ $ puma -b 'ssl://127.0.0.1:9292?keystore=path_to_keystore&keystore-pass=keystore
 ```
 See https://www.openssl.org/docs/man1.0.2/apps/ciphers.html for cipher filter format and full list of cipher suites.
 
+Don't want to use insecure TLSv1.0 ?
+
+```
+$ puma -b 'ssl://127.0.0.1:9292?key=path_to_key&cert=path_to_cert&no_tlsv1=true'
+```
+
 ### Control/Status Server
 
 Puma has a built-in status/control app that can be used to query and control Puma itself.
--- a/ext/puma_http11/mini_ssl.c
+++ b/ext/puma_http11/mini_ssl.c
@@ -142,6 +142,7 @@ VALUE engine_init_server(VALUE self, VALUE mini_ssl_ctx) {
   VALUE obj;
   SSL_CTX* ctx;
   SSL* ssl;
+  int ssl_options;
 
   ms_conn* conn = engine_alloc(self, &obj);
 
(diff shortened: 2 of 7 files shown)


evanphx:
This looks fine. I merged a similar PR that caused this to need to be rebased. Can you do that and then I'll merge it.

themikejr:
Gentle bump :-) My team would benefit from this.

skrobul:
Sorry I forgot to comment, this has been rebased in e142b9f and it should be ready to merge

MikaelSmith:
I'd love to see this fixed, although a more forward-ready implementation would be nice: the ability to list what TLS versions to allow would support disabling TLSv1.1 in the future.

HoneyryderChuck:
Why does puma have their own "minissl", instead of using ruby's "openssl" library?

skrobul:
Tiago,
Ha! I was thinking the same, but after looking into code, the minissl is
just a class name that is used to wrap SSL related functionality. Upon
compilation it still uses OpenSSL. See more details in
https://github.com/puma/puma/issues/1561

HoneyryderChuck:
@skrobul I get that, `minissl` is just an OpenSSL binding. However, the ruby std library already provides one, and with a stable albeit clunky API, which allows you to do the things this PR aims at in a future-proof way. So my question was really: why does puma feel the need to support their own openssl binding which has no API parity with `ruby-openssl` and has no clear advantage besides arguably "leaner" (and therefore more limited, hence this PR)? There must be a reason, right?

MSP-Greg:
> with a stable albeit clunky API

Given that:
1. Ruby OpenSSL has deprecated `ssl_version` and replaced it with two attributes
2. RubyGems is now only supporting TVSv1.2 connections
3. OpenSSL 1.1.1 with TVSv1.3 should be released sometime in the next 6 months

Rather than add an `no_tlsv1` option, might it be better to add a `min_vers(ion)` option?  People may want a `no_tlsv1_1` option in a year or so...

mmizani:
gentle bump, would be great to have this feature.

hshyk:
Has there been any updates with this? This would be great if we could disable TLSv1. Also, would be nice for configuration to disable TLSv1.1.

schneems:
Not clear based off of the diff, is it possible to trigger this without explicitly supplying a `bind` in the dsl?

If not could we add this as a new dsl option?

PR #1567, opened 2018-04-22
Fix ssl_v3_rejection test hanging

deivid-rodriguez:
I wanted to run the tests locally so I can understand how `puma` works better. But I would consistently get a hang on the `server.stop(true)` call in the teardown of the `ssl_v3_rejection` test.

The change in https://github.com/puma/puma/commit/d03b125640efa46e539f053996515b28bc62c13f seems to fix it.

In order to reproduce the problem in CI, so you can see the error, I figured I could remove all the Windows specific tweaks in the test, since they seemed to be written specifically to avoid the issue from happening. I came to think that because the hang was happening on the test teardown, and those tweaks seemed to try to prevent that specific teardown from being run. My guess was right and the hang reproduced on Windows after doing that (see [here](https://ci.appveyor.com/project/deivid-rodriguez/puma/build/1.0.9/job/4rm4bos4866h32xr)). With the patch, closing the server seems to consistently never hang on any platform, at least not in that test.

Note that I'm actually getting the hang on MRI 2.5.1 on Linux, but for some reason Travis wouldn't hit it.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -43,13 +43,25 @@ install:
   - bundle install --without documentation --path C:/av_bundle
 
   # Download & install current OpenSSL package for later RubyInstaller2 version(s)
-  - set openssl=mingw-w64-x86_64-openssl-1.1.0.g-1-any.pkg.tar.xz
+  - if %ruby_version%==25-x64 (
+      set openssl=mingw-w64-x86_64-openssl-1.1.0.g-1-any.pkg.tar.xz
+    )
+  - if %ruby_version%==25 (
+      set openssl=mingw-w64-i686-openssl-1.1.0.g-1-any.pkg.tar.xz
+    )
   - set  dl_uri=https://dl.bintray.com/msp-greg/ruby_trunk
   - if %ruby_version%==25-x64 (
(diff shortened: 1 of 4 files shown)


MSP-Greg:
@deivid-rodriguez 

Glad to see you're able to take a break from byebug.  Re ragel, I added it due to intermittent failures, often in just one job (ruby version) of a build.  See 

https://ci.appveyor.com/project/MSP-Greg/puma/build/7/job/17o5qjunlbva5q5j

Greg

deivid-rodriguez:
@MSP-Greg But did you notice that you were installing the 32 bit package during the 64 bit jobs, and viceversa?

MSP-Greg:
@deivid-rodriguez 

Thanks for pointing that out.  Stupid mistake on my part, apologies.

The RubyInstaller2 builds are done correctly on lines 14 & 15, but RubyInstaller builds (Ruby 2.3 and older) are incorrect.  I've seen intermittent ragel failures on both, so if you could fix the issue rather than removing the code, it would be appreciated.

The MB on my home development system is not booting, so I'm behind...

Thanks, Greg

deivid-rodriguez:
@MSP-Greg I changed the code to swap the packages-

Issue #1541, opened 2018-03-20
Request handling time metrics


rianmcguire:
tl;dr: Collect per-request threadpool wait time and other interesting timings, and expose them via the Rack environment.

We've just started using Puma in production and we're still in the process of tuning it. We're collecting Puma's backlog metric, as well as measuring the time spent inside our Rack app.

We occasionally have large bursts of requests, and we can see from the backlog that requests are waiting for Puma's threadpool, but we don't have a measure for how long they are waiting. We have haproxy in front of Puma, so we can get the total request time from there, but that's distorted by the time taken for slow clients to send request bodies.

I'm considering modifying Puma to track various times in the request lifecycle and calculate the threadpool wait time for each request, as well as the time spent waiting for the request body. These times could be exposed via Rack environment as, eg. `puma.threadpool_wait` and `puma.request_wait` and then collected by middleware.

Is this a feature the maintainers would be interested in having in Puma?

nateberkopec:
> Is this a feature the maintainers would be interested in having in Puma?

👍 Absolutely.

stereobooster:
> Collect per-request threadpool wait time

If you use `nginx` you can set

```
proxy_set_header X-Request-Start "${msec}";
```

and then you can create Rack middleware which will read this value and calculate difference between this value and current time - this is time between request got to webserver and actually started to execute

[Example of Rack middleware](https://github.com/blacklane/kiev/blob/master/lib/kiev/rack/request_logger.rb#L97)

rianmcguire:
That's a great idea, thanks @stereobooster. Looks like something similar is possible with haproxy:
```
http-request set-header X-Request-Start t=%Ts%
```
I'll still need to subtract the request body read time, but that's an easy wait to measure the socket backlog + threadpool wait times. I was a bit stuck on measuring the time in the socket backlog, since puma stops accepting connections when the pool is full, and there's no socket API for determining how long a connection has been waiting.

PR #1569, opened 2018-04-24
Request body read time metric

rianmcguire:
Measure the time spent reading the HTTP request body and expose it to the Rack app as `env['puma.request_body_wait']`.

This can be combined with a timestamp from a upstream proxy to get an indication of how long a request was waiting for a Puma thread to become available.

Fixes https://github.com/puma/puma/issues/1541

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -20,6 +20,7 @@ Clustered mode is shown/discussed here. Single mode is analogous to having a sin
 * By default, a single, separate thread is used to receive HTTP requests across the socket.
   * When at least one worker thread is available for work, a connection is accepted and placed in this request buffer
   * This thread waits for entire HTTP requests to be received over the connection
+    * The time spent waiting for the HTTP request body to be received is exposed to the Rack app as `env['puma.request_body_wait']` (milliseconds)
   * Once received, the connection is pushed into the "todo" set
 * Worker threads pop work off the "todo" set for processing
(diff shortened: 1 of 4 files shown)


rianmcguire:
Looks like I have a test failure on JRuby. I'll investigate that in the next couple of days.

nateberkopec:
This feature should be optional.

@schneems do you have any thoughts about we should organize stuff like this in general? Add to Puma.stats?

schneems:
Puma.stats is mostly for high level server metrics. Also everything in there is pretty cheap to compute.

This would be a per request metric. The only good way to expose that would be either through the env hash like this or some kind of more architected notification API where people could subscribe to certain events.


Check out https://github.com/puma/puma/issues/1577 and the attached PR. The goal is to give a high level indication that the app needs to scale out with more servers. Is that kinda what you’re trying to achieve with these measurements?

rianmcguire:
@schneems Yes, that's exactly what I'm after.

I'm using the same approach as SamSaffron described with the upstream timestamp header. haproxy can't buffer the entire request body like nginx, so I wanted to account for the time spent in puma's reactor waiting for the body.

If there's a broader plan to make the `Puma.stats` backlog metric accurate, I'd prefer that to this very specific (and complicated) approach of calculating the wait time.

I like your "negative backpressure" idea. The disadvantage versus a backlog or wait time metric is it doesn't let you quantify how far over capacity you are, but I'm not sure how much that matters in practice.

PR #1573, opened 2018-04-27
Fix Puma::DSL.load

hugoarpin:
The comment in DSL.load says that files loaded with this method
will get loaded later via Configuration.load. I find this to be not the
case, because by the time DSL.load is called, Configuration.load has
already been called so it does nothing. The only way I can load inside
my config file is to call DSL._load_from directly.

Please correct me if I'm wrong, but I would like a fix, thanks.

### DIFF
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -104,10 +104,8 @@ module Puma
     end
 
     # Load additional configuration from a file
-    # Files get loaded later via Configuration#load
     def load(file)
-      @options[:config_files] ||= []
-      @options[:config_files] << file
+      _load_from(file)
     end
 
     # Adds a binding for the server to +url+. tcp://, unix://, and ssl:// are the only accepted


nateberkopec:
@schneems

schneems:
I added it when I made this change https://github.com/puma/puma/commit/995ed8bd4585574543977a2fd3f80de55cff2236. Though it broke some stuff and got modified later like in https://github.com/puma/puma/pull/1340/files. The comment might not be true anymore. It’s hard to keep track of how all the config works which is why I tried leaving all the comments. This change doesn’t break tests so that’s a good sign. Can you give me an example of what exactly you’re trying to do and how it’s not working? In general I think the dsl load is not a very public facing feature. 

Is it possible for you to write a test that fails on master but passes on your branch?

hugoarpin:
Ok basically what I want is a command like `include` for nginx.

Let's say I always want the number of workers to be the number of processors. I do not want to add this line into all my config files for every environments because I would have to maintain this code at multiple places. So what I do is something like this:

```ruby
# config/puma/_shared.rb
workers(Etc.nprocessors)
```

```ruby
# config/puma/production.rb or config/puma/staging.rb

# Does not work
load(File.expand_path('../_shared.rb', __FILE__))

# I must use _load_from instead
_load_from(File.expand_path('../_shared.rb', __FILE__))
```

I do not have experience writing tests, I would probably be able if I tried, but would be painful. Let me know. Thanks!

schneems:
We need a test to merge otherwise we risk accidentally reverting this capability.

For your use case you can maybe build it into a plugin. 

Also be careful with this measurement, if you’re running on a shared host like Heroku’s 1x dynos then it will show 8 processors even though none of them are dedicated to your app.

hugoarpin:
Well it was just to give to an example (my servers run on physical machines), I have a lot more code that is shared.

From what I see 995ed8bd4585574543977a2fd3f80de55cff2236 broke DSL.load, I'm pretty sure it must have been working before. Now this method is useless inside config files.

Anyway I just wanted to let you know that it didn't work, but frankly this is a bit trivial and I do have a personal workaround in _load_from (or I could simply use instance_eval), but surely someday someone else will be annoyed by this.

I can close this pull request now. I suggest that you simply revert (or tweak) 995ed8bd4585574543977a2fd3f80de55cff2236.

Thanks for your time!

PR #1579, opened 2018-05-04
[close #1577] Negative Backpressure Metric

### LINKED ISSUE #1577: Accurate Backpressure Metrics from Puma Server
**TLDR;** The existing backlog metric is flawed as a backpressure mechanism and may soon be basically meaningless with #1563. 

I want to be able to measure whether a Ruby web app is being under-provisioned for resources. One way that this can be accomplished is by measuring the number of requests that are fully formed that a server is not able to process. 

Currently as a proxy I'm using Puma's "backlog" (which is different than than TCP backlog). 

## Backlog Behavior

For a request to appear in the backlog metric it must go through this flow:

- A request comes into the server, perhaps it is queued at the socket
  - A `Puma::Server` instance gets notified that there is capacity in its thread pool
  - The `Puma::Server` accepts the request from the socket and puts it into the Reactor
  - The reactor ensures that the header and body are fully formed
  - The reactor passes the request to the thread pool
  - At this point the `@todo` array size has been incremented by one

If there are requests in `@todo` that have not been poped off and are being worked on then they show up as "backlog". 

At first this looks like an ideal metric, except for the `Puma::Server` meters request based on threadpool capacity. In an ideal world it will never try to add a request if there is not capacity. There are cases where a two slower clients might get their requests accepted and then there is a temporary backlog, otherwise the backlog metric should be zero.

### 1) TCP backlog metric

We could measure the number of requests that are queued at the TCP socket and report on that.

Unfortunately: There doesn't seem to be an API for getting this data. We can see the max size of the TCP backlog, but not actually how many clients are connected and awaiting being processed. Also even if we could, then we're risking weird conditions for example if someone is slow client attacking us, there could be tons of things in the backlog that will never be fully requests that are ready to be processed by the server.

###  2) 1 Reactor, many servers

 Change the architecture so instead of 1 reactor for every process, we have one reactor on the parent process (or another sidecar process). It tries to buffer requests as fast as possible with no waiting. It can pass the fully buffered requests via some mechanism. Perhaps a unix domain socket though I don't know what it would look like for windows and jruby (though they can't run in clustered mode anyway). 

If we did this than the backpressure metric would be requests that have been fully buffered by the reactor but have not been accepted by a cluster or threadpool. Assuming we can measure that somehow.

### 3) Any other ideas? 

???

cc/ @jjb

schneems:
This PR introduces the `pool_capacity` stat which can be used as a “negative back pressure metric”. 

What does a “negative backpressure metric” mean? It means when this number is low, we have a higher amount of backpressure. When it is zero it means that our worker has no ability to process additional requests. This information could be used to scale out by adding additional servers. When that happens the requests should be re-distributed between the extra server and the “pool capacity” number for each individual server should go up.

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -292,7 +292,8 @@ module Puma
           begin
             b = server.backlog || 0
             r = server.running || 0
-            payload = %Q!#{base_payload}{ "backlog":#{b}, "running":#{r} }\n!
+            t = server.pool_capacity || 0
+            payload = %Q!#{base_payload}{ "backlog":#{b}, "running":#{r}, "pool_capacity": #{t} }\n!
             io << payload
           rescue IOError
             Thread.current.purge_interrupt_queue if Thread.current.respond_to? :purge_interrupt_queue
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -293,7 +293,7 @@ module Puma
             b = server.backlog || 0
             r = server.running || 0
             t = server.pool_capacity || 0
-            payload = %Q!#{base_payload}{ "backlog":#{b}, "running":#{r}, "pool_capacity": #{t} }\n!
+            payload = %Q!#{base_payload}{ "backlog":#{b}, "running":#{r}, "pool_capacity":#{t} }\n!
             io << payload
           rescue IOError
             Thread.current.purge_interrupt_queue if Thread.current.respond_to? :purge_interrupt_queue
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -169,8 +169,15 @@ module Puma
     end
 
 
+    # This number represents the number of requests that
+    # the server is capable of taking right now.
+    #
+    # For example if the number is 5 then it means
+    # there are 5 threads sitting idle ready to take
+    # a request. If one request comes in, then the
+    # value would be 4 until it finishes processing.
     def pool_capacity
-      @thread_pool and @thread_pool.waiting
+      @thread_pool and @thread_pool.pool_capacity
     end
 
     # Lopez Mode == raw tcp apps
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -168,6 +168,11 @@ module Puma
       @thread_pool and @thread_pool.spawned
     end
 
+
+    def pool_capacity
+      @thread_pool and @thread_pool.waiting
+    end
+
     # Lopez Mode == raw tcp apps
 
     def run_lopez_mode(background=true)
--- a/lib/puma/single.rb
+++ b/lib/puma/single.rb
@@ -14,7 +14,8 @@ module Puma
     def stats
       b = @server.backlog || 0
       r = @server.running || 0
-      %Q!{ "backlog": #{b}, "running": #{r} }!
+      t = @server.pool_capacity || 0
+      %Q!{ "backlog": #{b}, "running": #{r}, "pool_capacity": #{t} }!
     end
 
     def restart
--- a/lib/puma/thread_pool.rb
+++ b/lib/puma/thread_pool.rb
@@ -58,7 +58,7 @@ module Puma
       @clean_thread_locals = false
     end
 
-    attr_reader :spawned, :trim_requested
+    attr_reader :spawned, :trim_requested, :waiting
     attr_accessor :clean_thread_locals
 
     def self.clean_thread_locals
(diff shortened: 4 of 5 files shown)


mimen:
@schneems Is this ready to be merged?

schneems:
Yep!

jjb:
sweet! a few questions

- Where did this idea come from? Discussion in another ticket/channel/IRL?
- Does this resolve the "we can't access a meaningful metric given the current design" conundrum discussed in #1577, or is it more of a stopgap solution?
- What is the highest value that this can be?
- (related to above) How (if at all) is it affected by `0:N` vs. `N:N` thread configuration?

schneems:
> Where did this idea come from? Discussion in another ticket/channel/IRL? Does this resolve the "we can't access a meaningful metric given the current design" conundrum discussed in #1577, or is it more of a stopgap solution?

I think this is the thing TM that "fixes" #1577. Though if you disagree, i'm happy to re-open. 

Ahh, i see i missed some of your comments over there, sorry about that. I had intended conversation about this specific interface to happen on this PR. I'm using this in production right now and it is providing good feedback. 

> What is the highest value that this can be?

That's already been asked, I can add that in as another metric. Seems worthwhile. Per worker it would be max-threads. It's a good idea to let the stat tell us that directly. While i'm doing that I can also expose the current number of threads per worker. The two values are related.

> (related to above) How (if at all) is it affected by 0:N vs. N:N thread configuration?

Going down to 0 in "pool capacity" is bad in the N:N config. There may be a race condition with 0:N case where all your workers drop to zero (because you've not yet hit N threads) and then the stat fires and reports that you have 0 capacity, then they all dynamically create a new thread and the capacity would actually be 1 times the number of processes.

I think this is mitigated by the fact that we would need to use a stream of this metric rather than just one value. I.e. even if it's showing 0 right now, under the same sustained load it would show a positive number the next time it reports. Even if we took some scaling action based on the 0 indicator, a future stream of positive numbers would perhaps indicate we could safely scale back down.

sj26:
Yesss, thank you!

Now we can publish a "Utilization" metric for sensible autoscaling, at least in the N:N configuration. :100:

schneems:
Let me know what you think!

The other major question to answer with metrics is "am I using a good number for my thread count?".

I'm not totally sure about that one. My current best suggestion is to pick a magic number and live with it. Alternatively, adjust it up or down and note if the average response time increases or decreases.

I guess I want some combination of a CPU metric (want to be close to 100% utilization) but also some kind of a "contention" metric. For example, you can certainly saturate an app with 1000 threads, but your contention will be through the roof.

Maybe contention could be time spent idle per thread. I don't know if it's possible to get that from Ruby, it might require a patch to ruby/ruby and wouldn't be useful for at least a year until it gets released.

I'm open to other thoughts/suggestions. Maybe we should open another issue on it for some brainstorming.

jjb:
I've been thinking about this problem for years. Ideally one would like to try different thread counts under similar circumstances and see how performance and CPU load are affected (assuming that processes are scaled up to use available ram, and the amount of ram more threads take is not an issue).

But that's hard to do because it's hard to get directly comparable traffic circumstances.

But... I just now had this idea: One could have, on the same machine,  Puma processes with thread counts. Then the behavior of each can be compared. Most important is probably simply throughput. If a process with 4 threads serves almost 2 times as many requests as a process with 2 threads, 4 threads is worth the DB connections. If a process with 8 threads serves only 20% more requests than a process with 4 threads, 8 threads is too many and a waste of DB connections.

Doing this would require that the logic which distributes requests to processes is very accurate in its determination of if a process has available capacity. I don't know to what extent this is possible.

If it is possible, then we can add features to Puma to facilitate this sort of experiment.

We could even eventually create some sort of autoscaling.

jjb:
(above comment was edited a lot after posting, please read on website)

schneems:
Interesting. That could be really cool. There's lots of apps that can't run that many workers. I also worry about large apps that might need to tightly control DB connection counts.

I think this is a good experiment. We could maybe manually do this in a "on_worker_boot" block of an app, though would need a way to determine thread multiplier for each process.

I'm not sure how to log and record throughput of each individual worker in that scenario.

> Doing this would require that the logic which distributes requests to processes is very accurate in its determination of if a process has available capacity. I don't know to what extent this is possible.

Puma 3.12 does a good job of this AFAIK after a recent patch. You can read the docs about how this all works https://github.com/puma/puma/pull/1576.

Let me know if you've got any other ideas. The value of a brainstorming session usually comes from the number of ideas. With quantity comes quality.

PR #1580, opened 2018-05-09
Friendlier dev setup

deivid-rodriguez:
I was really surprised that puma would ask me to remove every single gem matching /puma/ from my system, just to be able to run puma's tests.... Whatever problem this was fixing, it feels like the wrong solution for it... I mean, this is what `bundler` is for, right?

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -73,16 +73,6 @@ else
   task :test => [:compile]
 end
 
-task :test => [:ensure_no_puma_gem]
-task :ensure_no_puma_gem do
-  Bundler.with_clean_env do
-    out = `gem list puma`.strip
-    if !$?.success? || out != ""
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Failure looks unrelated, culprit could be #1487.

deivid-rodriguez:
I rebased this PR since master is now green.

deivid-rodriguez:
New year, new rebase! :stuck_out_tongue_closed_eyes:

deivid-rodriguez:
In particular, I was asked not only to remove all versions of puma from my system, but also [`capistrano3-puma`](https://github.com/seuros/capistrano-puma). That seems like a pretty high barrier for running puma's tests.

deivid-rodriguez:
Thanks! :)

PR #1583, opened 2018-05-23
Fix @notify.close can't modify frozen IOError (RuntimeError)

chwevans:
Let me know if there is anything else you need for this contribution.


lib/puma/server.rb
```diff
@@ -256,7 +256,8 @@ def handle_servers_lopez_mode
           Thread.current.purge_interrupt_queue if Thread.current.respond_to? :purge_interrupt_queue
         end
 
-        @notify.close
+        # Prevent can't modify frozen IOError (RuntimeError)
+        @notify.close rescue nil
```
phstc: @chwevans I don't know exactly about the fix, but I'm often getting `RuntimeError: can't modify frozen IOError` on Heroku after scaling down.

evanphx:
Can you provide a backtrace for this? Never seen a frozen IO.

lib/puma/server.rb
phstc: @evanphx sure. I get that almost every day when Heroku [auto restart/recycle dynos](https://devcenter.heroku.com/articles/dynos#automatic-dyno-restarts).

image

Issue #1588, opened 2018-05-25
Thread work needs to be inside mutex?


jjb:
Looks like the thread doing its work is inside the mutex which figures out if there is work to do

https://github.com/puma/puma/blob/master/lib/puma/thread_pool.rb#L128-L132

Could this instead go directly after the close of the mutex block?

I'm guessing i'm missing something, because otherwise this would be a huge performance problem that someone would have noticed.

nateberkopec:
I'm not sure what this is referring to, and the link is now broken.

Issue #1586, opened 2018-05-25
why local variables in threads?


jjb:
Why are these local variables created?

https://github.com/puma/puma/blob/master/lib/puma/thread_pool.rb#L86-L90

Thanks!

alexdowad:
It's a performance optimization.

Look at `git blame` on that file. (https://github.com/puma/puma/blame/master/lib/puma/thread_pool.rb) You can see some commits with commit message: "Add missing localvars. **Accessing localvars is faster than accessing ivars.**"

PR #1589, opened 2018-05-25
Improved thread pool code

jjb:
These are various small and large changes to the `ThreadPool` code. 

My main goal here is to work toward code that is easier to reason about and easier to change.

More specific goals that this code may facilitate:
* understand which objects need to be threadsafe together -- we can possibly have more than one mutex, each wrapping a smaller amount of code
* related to this: the `@todo` array would probably be better implemented as a `Queue`. `Queue`s come with built in concurrency-facilitating behavior (they have an internal mutex so no outside mutext is needed and `#pop` sleeps when the queue is empty and returns `nil` when the `Queue` has been shut down), which I believe will make the code nicer and possibly more performant. See an example of how this can be used here: https://github.com/patrickxb/stathat/pull/10/commits/b47dd80503a059225d8f9eca9c8d1f97bb7c9827
* try to move thread coordination and thread work into different places. currently the main thread loop both checks to see if it's supposed to reap itself and also checks for more available work. i think it will be possible and beneficial to move the reaping code to a higher level, and only have the thread cease when it sees "shutdown"

There are several `Thread` methods that `ThreadPoolWorker` simply passes down to the `Thread` that it manages. This could be a smell... but I think it can lead to further improvements in which coordination code goes where. Perhaps we can keep iterating so that a `shutdown` command does several of the things needed for managing an individual thread, without need for fine-grained thread coordination at the top `ThreadPool` level.

related: #1577, #1586, #1588

It's a work in progress. I'd love to get some feedback to direct my continued work. I would recommend looking at the commits one by one. "hide whitespace changes" will help with some of them.

image

I'm guessing that you will like some of the initial aesthetic changes. If so, I can make another PR with just those so we can get those into master and simplify this PR.

### DIFF
--- a/lib/puma/thread_pool.rb
+++ b/lib/puma/thread_pool.rb
@@ -61,12 +61,6 @@ module Puma
     attr_reader :spawned, :trim_requested
     attr_accessor :clean_thread_locals
 
-    def self.clean_thread_locals
-      Thread.current.keys.each do |key| # rubocop: disable Performance/HashEachMethods
-        Thread.current[key] = nil unless key == :__recursive_key__
-      end
-    end
-
     # How many objects have yet to be processed by the pool?
     #
     def backlog
--- a/lib/puma/thread_pool.rb
+++ b/lib/puma/thread_pool.rb
@@ -295,7 +295,7 @@ module Puma
     # Tell all threads in the pool to exit and wait for them to finish.
     #
     def shutdown(timeout=-1)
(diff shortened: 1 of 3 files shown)


jjb:
@schneems ever get a chance to take a look at this?

evanphx:
I'm going to say no to this change. The usage of moving to a big set of Hash's for configuration coupled with moving the most important part of the file to a different file doesn't help with understanding, imho.

PR #1598, opened 2018-06-18
Add Appveyor building fat-binary gem to testing

MSP-Greg:


MSP-Greg:
The Build system in the PR creates two Appveyor jobs, one for the 64 bit gem & testing, and one for the 32 bit.

From experience, when Appveyor is busy, builds may have problems verifying GPG keys, which are used by two of the OpenSSL packages.  You'll notice in the 64 bit job the [bottom line in the log](https://ci.appveyor.com/project/puma/puma/build/1.0.81/job/1dbmq747yulhu4dg#L118) is a key error.

I did a build locally in my repo, and forgot that a file is added to the gem for looking up the correct so file in the fat-binary gem.  That took me two revisions to fix (sorry), so there were three Appveyor tests ran.  The first two passed.

As to the JRuby error in Travis, that seems to be shared by most recent commits also.

Greg

MSP-Greg:
Revised version created with shared script files downloaded.  See #1601

PR #1606, opened 2018-07-04
test_ssl_v3_rejection another error string

MSP-Greg:
Found with Ruby 2.4.4 & OpenSSL 1.0.2o


MSP-Greg:
See [openssl.txt](https://github.com/openssl/openssl/blob/master/crypto/err/openssl.txt#L2844)

PR #1611, opened 2018-07-09
Update url for binder parsing test for JRuby

jkutner:
This change will fix the [failing JRuby CI build](https://travis-ci.org/puma/puma/jobs/399748817). I suspect this test never worked, and something in how event errors were handled has changed.

In any case, I've added a keystore-pass url param to make MiniSSL happy (the `examples/puma/keystore.jks` has an empty password, which is fine).

I've also added a port to the URL string to fix a `no implicit conversion of nil into String` error that cam after fixing the keystore-pass (further supporting my suspicion that this never worked).

### DIFF
--- a/test/test_binder.rb
+++ b/test/test_binder.rb
@@ -49,7 +49,7 @@ class TestBinder < Minitest::Test
 
     keystore = File.expand_path "../../examples/puma/keystore.jks", __FILE__
     ssl_cipher_list = "TLS_DHE_RSA_WITH_DES_CBC_SHA,TLS_DHE_RSA_WITH_3DES_EDE_CBC_SHA"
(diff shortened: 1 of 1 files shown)


schneems:
Tests for travis are good. Windows tests seem borked for a totally unrelated reason. Thanks a ton for the PR!

Windows error:

```
bundle --version
C:/Ruby24/bin/bundle:23:in `load': cannot load such file -- C:/Ruby24/lib/ruby/gems/2.4.0/specifications/gems/bundler-1.16.1/exe/bundle (LoadError)
	from C:/Ruby24/bin/bundle:23:in `<main>'
Command exited with code 1
ruby -v
```

PR #1601, opened 2018-06-23
Appveyor - Create fat-binary gems and test

MSP-Greg:
Two jobs on Appveyor, 64 bit & 32 bit:
1. Builds Windows fat binary gems - currently for 2.2 thru 2.5. 64 bit also includes trunk.
2. Loops thru all Ruby versions, installs the gem, and tests.
3. Tests are shown on the Appveyor job test tab.
4. Gems and a 7z of test logs are saved as artifacts.
5. Current commit info (date, short SHA, summary) is stored in gemspec as metadata.
6. Shared Powershell files are downloaded from Appveyor.  Only three files added to repo.

Built three times on my [fork](https://ci.appveyor.com/project/MSP-Greg/puma/history), builds 10-12. Approx time is 11 minutes.

PR #1613, opened 2018-07-12
appveyor.yml - fix casing, OpenSSL 1.1.0h, convert to ps

MSP-Greg:
Several updates to Appveyor.yml

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,96 +1,97 @@
-# cache cleanup 2018-03-16
-
 init:
-  - set PATH=C:\ruby%ruby_version%\bin;C:\Program Files\7-Zip;C:\Program Files\AppVeyor\BuildAgent;C:\Program Files\Git\cmd;C:\Windows\system32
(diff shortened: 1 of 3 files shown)


MSP-Greg:
@schneems

Finally.  I forgot, the test that's been intermittently failing on Travis will run on Windows.

Please have a look at [Test Update - skip handling, windows ](https://github.com/puma/puma/pull/1601/commits/b1141ae4361f3cbeb6c34897ec69456d1f14c55e).  I've got to add `TestPumaControlCli#test_control_url` to it, but I've run it plenty of times testing the Widows pre-compiled gem PR.

I can remove that and do a separate PR for the test update.  Also, an update to `control_cli.rb`; it's set up fine for a cli app, but could have some stability issues when used in testing...

Thank, Greg

MSP-Greg:
Re the commit `Fix test_pumactl.rb for CI`, might the issue really be that the following two lines should be flipped?

https://github.com/puma/puma/blob/f7e1e58dcba1e19e470b6eb0c44c33a0722285f6/lib/puma/binder.rb#L104-L105:

schneems:
This is great! Thanks!

I’ll take a look at the other PR shortly. I like the change for skipping certain suites. We merged in a change that makes the required version of Ruby 2.2 or higher so we could make the str a kwarg.

PR #1616, opened 2018-07-15
Updates test skips & RuboCop

MSP-Greg:
Updates helper.rb skip methods & RuboCop to Ruby 2.2

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,6 +1,6 @@
 AllCops:
   DisabledByDefault: true
-  TargetRubyVersion: 1.9.3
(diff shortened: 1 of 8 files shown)


olleolleolle:
The new skip methods improve the legibility of the tests! The RuboCop update seems to have worked nicely.

MSP-Greg:
I just rebased and changed the PR to allow showing the real location of the skip.  Previous tests and the original commit would show the location of the skip as somewhere in helper.rb.

The update fixes that...

And I had a typo (I thought I changed something from an instance variable to a local, but I forgot one `@`).  Updated twice...

PR #1617, opened 2018-07-15
Travis: use 9.2.0.0

olleolleolle:
- this is the latest stable


MSP-Greg:
@olleolleolle

Apologies, maybe I went a bit nuts with 'Make ci work' 72882f2, but 9.2.0.0 was added along with JRuby-head, and both are passing (head is allow failure).  Thanks, Greg

PR #1612, opened 2018-07-10
Fix markdown syntax [ci skip]

jiikko:
#### before
image

#### after
image

PR #1618, opened 2018-07-16
Fix broken formatting in README [ci skip]

y-yagi:
**before**

before

**after**

after


y-yagi:
Duplicated with #1612.

Issue #1615, opened 2018-07-13
Fix Flakey Test


schneems:
Some of the network tests that actually spin up puma and then make requests against it are flakey. When the build is restarted they pass. In addition to these specific tests failing they also cause other future tests to fail since future tests are expecting the port to be open. This makes it hard to find out what original test was flakey. 

The overall goal is to make the test suite more robust. This includes:

- [ ] Making tests independent so one fails, it does not cause a cascade. 
- [ ] Making tests that spin up a process more consistent. 
  - [ ] Possibly adding some retry behavior inside of the suite
  - [ ] Possibly involve finding hidden race conditions and making them less dependent on timing (for example actually checking that a server has booted before sending it requests.

I don't know if it will be helpful or not, but i have a gem https://github.com/schneems/wait_for_it that I use to spin up other puma processes in other tests such as https://github.com/schneems/puma_worker_killer/blob/master/test/puma_worker_killer_test.rb#L6-L12

Here's an example of a failed travis test: https://gist.github.com/schneems/6197f97650d8f58b95bb7f053e8d2057

MSP-Greg:
@schneems

See https://ci.appveyor.com/project/MSP-Greg/puma/history.  Certainly seems to be reliable.

Also, above mentioned PR #1609 and the Travis PR #1620.  I'd run tests on Travis with my fork, but since it's 'branches: only: "master"', It's a PITA to run another branch in a fork...

I'm happy to help, but when CI fails all the time, it's difficult to see what intermittent issues exist...

nateberkopec:
We good now

PR #1619, opened 2018-07-17
control_cli.rb updates

MSP-Greg:
1. Add ensure before server.close
2. Change @server to method variable
3. Change puma requires to require_relative

In the spirit of Issue #1615, having `server.close` not called with a CLI app is normally ok, but in testing it may be an issue...

### DIFF
--- a/lib/puma/control_cli.rb
+++ b/lib/puma/control_cli.rb
@@ -1,8 +1,8 @@
 require 'optparse'
-require 'puma/state_file'
-require 'puma/const'
-require 'puma/detect'
-require 'puma/configuration'
+require_relative 'state_file'
+require_relative 'const'
+require_relative 'detect'
+require_relative 'configuration'
 require 'uri'
 require 'socket'
 
@@ -129,7 +129,7 @@ module Puma
       uri = URI.parse @control_url
 
       # create server object by scheme
-      @server = case uri.scheme
+      server = case uri.scheme
                 when "tcp"
                   TCPSocket.new uri.host, uri.port
                 when "unix"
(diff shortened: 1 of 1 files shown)


MSP-Greg:
@schneems

Boy, I really hate these keyserver errors, which is the reason that the Appveyor build [failed](https://ci.appveyor.com/project/puma/puma/build/1.0.108/job/rgkn37txykdiyl2v#L146).  I built this twice locally, and it passed.  That included building 2.2 thru trunk and trunk with JIT...

I can't really fix the issue by using only an Appveyor.yml file.  It seems to work consistently in the 'pre-compiled' gem build code, and it works fine with ruby-loco/trunk.  Both of those have PowerShell script code with functions for a proper 'retry' function.  Also, I have seen Travis issues with keyservers timing out, so it's not just Appveyor.

In case you're wondering, 'new Windows Ruby' (MSYS2/MinGW and RubyInsatller2) uses the MSYS2 build tools, and they have only released OpenSSL 1.0.2 packages.  I may be partially to blame, but a long time ago, I started using OpenSSL 1.1.0 for ruby-loco trunk testing, so both trunk and 2.5 are built with 1.1.0.  Ruby 2.2 thru 2.4 use 1.0.2.  The 1.1.0 package is a custom package, so that's why a key must be imported, as there's no way to install a package with pacman without one.

Sorry for the long explanation, but if it was trivial, I would have fixed it already.  And, the custom keys are required for both trunk and 2.5, so even setting trunk to 'allow failure' may still result in false failures on 2.5...

Thanks, Greg

MSP-Greg:
Closed.  See a3ac7b9

PR #1620, opened 2018-07-18
travis.yml - remove gem install bundler

MSP-Greg:
Last three travis build have failed...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,8 @@ language: ruby
 cache: bundler
 before_install:
   # https://github.com/travis-ci/travis-ci/issues/9383#issuecomment-377680108
-  - gem install bundler
+  # not needed for RubyGems 2.7.7...
+  #- gem install bundler
 branches:
   only:
     - "master"


.travis.yml
```diff
@@ -5,7 +5,8 @@ language: ruby
 cache: bundler
 before_install:
   # https://github.com/travis-ci/travis-ci/issues/9383#issuecomment-377680108
-  - gem install bundler
+  # not needed for RubyGems 2.7.7...
+  #- gem install bundler
```
deivid-rodriguez: Maybe we can change it to `gem update --system 2.7.7` so that further updates of `rubygems` don't break the build?

.travis.yml
```diff
@@ -5,7 +5,8 @@ language: ruby
 cache: bundler
 before_install:
   # https://github.com/travis-ci/travis-ci/issues/9383#issuecomment-377680108
```
deivid-rodriguez: Since the line is no longer run, we probably no longer need the comment either.

deivid-rodriguez:
This PR fixes the Travis build, can we merge it?

.travis.yml
deivid-rodriguez: Not sure when 3.x will be released, but ok with that! Hopefully new releases won't break anything... :)

MSP-Greg:
Closed , see 72882f2

PR #1621, opened 2018-07-19
appveyor - remove gpg check

MSP-Greg:
The build system used on Appveyor (MSYS2/MinGW) needs an update.  Because of that, retrieving keys from keyservers is intermittent, and causes false failures.  Keys are not needed, but I prefer to use them since people may be running code locally.

For the time being, this PR removes GPG checks of OpenSSL packages used in testing.  This, combined with #1620, should make CI testing stable...


MSP-Greg:
Closed , see 72882f2

Issue #880, opened 2016-02-05
Logging stops after a while 


grepruby:
Hi 

I am using puma on production. After a while, logging does stop. It doesn't write anything in log file. Issue is random but I know it has something to do with puma.  

I hope I dont have to remind how important logs are for any live application. Aren't puma is production candidate ?

Please advice.

YuriSolovyov:
does app itself continue running? I guess rack env is set to production?

codehotter:
Are you using Rails.logger.silence? This became thread-safe in rails only very recently (4.2.6), and if you're using an older version, a race condition can permanently set the log level to ERROR in your application, even if you've configured it higher. Are errors still logged?

mwpastore:
I just ran into this. Puma stops writing to the logs after they're rotated. The solution is to sighup the top-level Puma process after logrotate does its thing. This tells Puma to close and reopen the log files.

I'm running Puma through Upstart, so I modified my puma config.rb to write out a pid file, and added these lines to /etc/logrotate.d/upstart:

```
        sharedscripts # only run the postrotate script once
        postrotate
                /bin/kill -HUP `cat /srv/rack/example.com/puma.pid 2>/dev/null` 2>/dev/null
        endscript
```

Your solution may not be exactly identical but hopefully this gives you a place to work from.

evanphx:
That's correct, if you move the log files, you need to tell puma so.

grepruby:
I am having same problem with production log with puma. However I am not using upstart instead I am using god script to start/stop puma. Also we are not rotating log of god/puma , so not sure in our case what is causing issue. 

My sidekiq workers keep writing in log file, regardless what puma does. So clear it has something with puma. 

Please advice.

mwpastore:
@grepruby:

1. What operating system?
2. How did you install God? `gem install` or yum or apt or something, e.g. `apt-get install ruby-god`?
3. Are you redirecting stderr and stdout through puma (e.g. `puma --redirect-stdout ..`) or through god (e.g. `w.log_cmd = ..`)?

grepruby:
@evanphx 
1 Ubuntu 14
2 Installed God through  gem. I am using rvm to manage ruby.
3 No. Not redirecting stderr or stdout

My God script looks like below
```

RAILS_ROOT='/var/www/apps/wl-event/current'
SHARED_ROOT='/var/www/apps/wl-event/shared'
STATE = "#{SHARED_ROOT}/log/puma.state"
[...]

PR #1626, opened 2018-08-01
Reopen logs working with renaming of the log file

AnatolyShirykalov:
Reopen logs of rack application after renaming log file


test/rackup/hello-logs.ru
```diff
@@ -0,0 +1,15 @@
+require 'logger'
+class AppWithLogger
+  def initialize
+    @logger = Logger.new('test/log.log')
+  end
+
+  def call(env)
+    puts "write to log"
+    @logger.info "hello"
+    #p @logger.instance_variable_get(:@logdev)
```
olleolleolle: Is this debugging which you want to keep in the test file?

AnatolyShirykalov: removed

nateberkopec:
Linking #1527 #880 #561

nateberkopec:
Please ask to re-open if you feel this is still needed in Puma.

However, this PR as stands currently reimplements log rotation via SIGHUP without a) removing the old code path or b) providing a use case as to why this way is better.

jeremy:
I was reminded of this recently when switching from Unicorn to Puma. Unicorn reopens all loggish-looking File objects on `USR1`, so any `Logger.new(path)` you have floating around your app (debug stuff, plugins/engines, gems?) that get log-rotated (a common route is a wildcard rule that catches an unanticipated logfile) will Just Work if you have logrotate `kill -USR1 /path/to/unicorn.pid`, because most log files don't need special handling and are simple to detect. Relying on this is convenient because the opposite is not: painstaking, error-prone, order-dependent, infuriating Signal.trap management.

Which suggests an alternative! Offer some kind of `on_reload/HUP` so apps can enlist their own behavior. Trigger a config refresh, reopen logfiles, freshen up, whatever. If someone wants, Unicorn-style log reopening, then `on_HUP { Unicorn::Util.reopen_logs }`

Anyway, back to the beginning:
> providing a use case as to why this way is better

Filing this as a data point that while "better" is arguable, "easier to reliably migrate" from Unicorn is certain!

nateberkopec:
`Signal.trap` wrangling definitely sucks. `on_HUP` sounds like a good alternative - are you suggesting that instead of the Unicorn behavior because it isn't a breaking change?

jeremy:
> are you suggesting that instead of the Unicorn behavior because it isn't a breaking change?

Yep! Not everyone's a fan of the clever Unicorn behavior, either. 😈

I think it's possible to provide as a Puma plugin too though. I'll give that a shot to demonstrate.

jeremy:
Here's what I set up, if you're looking for a quick way to get logrotate signaling Puma to reopen logfiles in your app.

Caveats
* You must redirect stderr or stdout to a file. Otherwise, the default Puma HUP behavior is to non-gracefully terminate the process—the opposite of what we want.

```ruby
# Gracefully reopen logfiles on HUP using Unicorn's reopener. Chains to
# Puma's HUP handler which signals its workers and reopens stdout/stderr.
module LogReopener
  extend self

  def trap(signal)
    require "unicorn"
    require "benchmark"

    # Funky construction here: Signal.trap returns the previous trap, so we
    # can refer to our own return value to chain handlers. (Puma's default
    # handler broadcasts HUP to workers and reopens stdout/stderr in each.)
    previous = Signal.trap(signal) do
      yield reopen_logs
      previous.call
    end
  end

  def reopen_logs
    timed { Unicorn::Util.reopen_logs }
  end

  private
    def timed
      result = nil
      ms = 1000 * Benchmark.realtime { result = yield }
      return result, ms
    end
end

# Install the HUP handler as early as we can, before the whole app boots.
# Mimic Puma-esque log output indicating logfile reopens and delays.
LogReopener.trap "SIGHUP" do |count, elapsed|
  puts "[#{Process.pid}] ! SIGHUP: reopened #{count} logfiles in #{elapsed.round}ms"
end
```

PR #1629, opened 2018-08-04
Fix  running test_puma_server_ssl.rb alone

MSP-Greg:
Puma::MiniSSL.check must be defined for DISABLE_SSL to be false, it is defined for 'MRI' ruby in puma_http11, assuming gem was built with SSL.

Test file cannot be run alone without patch.


MSP-Greg:
Closed.  See 6a7112d

PR #1636, opened 2018-08-21
Expose puma stats as prometheus metrics

harmjanblok:
We use prometheus for monitoring our application. We would like to be able to get the `Puma.stats` in prometheus metrics format, so we can include this in our dashboards (and alarmings).

This is a proposal implementation, I think at least following should be added:

- [ ] extract metric boilerplate into method/class, for example `Metric::gauge`
- [ ] add `#metrics` for clustered mode
- [ ] add example in `config.rb` for `activate_control_app` on `tcp...`

Instead of using the `prometheus-client` gem, I've chosen to use the prometheus text-based format as defined on https://prometheus.io/docs/instrumenting/exposition_formats/.

### DIFF
--- a/lib/puma/single.rb
+++ b/lib/puma/single.rb
@@ -19,6 +19,24 @@ module Puma
       %Q!{ "backlog": #{b}, "running": #{r}, "pool_capacity": #{t}, "max_threads": #{m} }!
     end
 
+    def metrics
+      metrics = <<~METRICS
+        # HELP puma_backlog Number of established but unaccepted connections in the backlog
+        # TYPE puma_backlog gauge
+        puma_backlog #{@server.backlog || 0}
+        # HELP puma_running State of the server
+        # TYPE puma_running gauge
+        puma_running #{@server.running || 0}
+        # HELP puma_pool_capacity Number of allocatable worker threads
+        # TYPE puma_pool_capacity gauge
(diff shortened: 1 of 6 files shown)


lib/puma/single.rb
```diff
@@ -19,6 +19,24 @@ def stats
       %Q!{ "backlog": #{b}, "running": #{r}, "pool_capacity": #{t}, "max_threads": #{m} }!
     end
 
+    def metrics
+      metrics = <<~METRICS
```
MSP-Greg: Puma is currently testing with Ruby 2.2.x, and I believe `<<~` is only available with Ruby 2.3.x

harmjanblok:
Adding the metrics for the `single` mode was the easy part, adding the metrics for the `clustered` mode turned out to be more difficult. For example the logic to render multiple labels for the same metric. Searching for solutions I discovered Puma supports for `Puma::Plugin`s with custom logic. I think it makes more sense to put all this logic into a dedicated gem as a `Puma::Plugin`.

That would keep the `puma` gem itself free from this logic, but still allows users to enhance puma with prometheus metrics.

The initial approach for this plugin is on https://github.com/harmjanblok/puma-metrics/pull/1.

Please let me know if it would make sense to include this logic into the `puma` gem instead.

Issue #1254, opened 2017-03-24
Performance degradation with multithreading and connection pools


mwpastore:
I maintain several of the [Ruby framework samples](https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/Ruby) for [TechEmpower's FrameworkBenchmarks](https://www.techempower.com/benchmarks/). The latest [Round 14 Preview 1.1](https://www.techempower.com/benchmarks/previews/round14/) numbers reflect very favorably on all the perf work that's been going on in the Ruby community over the past few months and years. However, there is an anomaly in Puma's numbers that may be worth digging into!

### Context

Round 14 Preview 1.1 was executed on a machine with 80 hardware threads (and some ungodly amount of RAM). 

* Puma, Unicorn, and Passenger were configured to run with 100 processes. 
* Puma was additionally configured to run with four threads (min and max) per process. 
* Under Puma, Sequel and ActiveRecord were configured to use connection pools with four slots. 
* Under Unicorn and Passenger, Sequel and ActiveRecord were configured to run single-threaded. 
* The `wrk` benchmark was executed against these instances with a concurrency of 256. 

I'll be talking about YARV/MRI 2.4 throughout the rest of this report unless otherwise specified. (There is a JRuby/TorqueBox benchmark but it has some other undiagnosed performance issues.) 

### The Good

On most of the tests, Puma wins handily thanks to its increased concurrency, even on CPU-bound tasks. For example, [JSON serialization under Sinatra](https://www.techempower.com/benchmarks/previews/round14/#section=data-r14&hw=ph&test=json&l=8vmx33&d=e5&a=2&f=zijxtr-zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-mh33):

<img width="937" alt="sinatra-json" src="https://cloud.githubusercontent.com/assets/8540936/24311597/df5fe206-10a2-11e7-87e0-8889b813cd06.png">

And [single SELECT queries against PostgreSQL under Roda/Sequel](https://www.techempower.com/benchmarks/previews/round14/#section=data-r14&hw=ph&test=db&l=8vmx33&d=e3&a=2&f=zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-lrsv) (note the FE column):

<img width="930" alt="roda-sequel-postgres-db" src="https://cloud.githubusercontent.com/assets/8540936/24311769/910e41fa-10a3-11e7-8da5-c511a31845c5.png">

[...]

PR #1646, opened 2018-09-12
Balance incoming requests across processes

wjordan:
This is my first pass at a minimal solution to #1254. It adds an extra heuristic to `wait_until_not_full` that tries to balance incoming requests across all idle workers first before accepting requests for additional threads in a busy (but not full) worker. The theory is that minimizing thread contention through a more balanced request load should optimize multi-thread, multi-process performance in some use-cases.

The load-balancing heuristic should work something like this:

1. If the worker is idle (zero threads have work), always accept a request.
2. If the worker is busy (some but not all threads have work), perform a non-blocking (timeout = 0) [`select`](https://ruby-doc.org/core-2.5.1/IO.html#method-c-select) on the socket. If the socket has work available, then accept another request.
    - The socket should only ever have work available to this non-blocking `select` when all workers are already busy with at least one request, because step 1 has all idle workers waiting on the socket with a blocking `accept`.
3. If the worker is full (all threads have work), wait for a thread to finish, then repeat.
4. If some threads are busy and the socket is empty, wait either for a thread to finish its work or 1 second (whichever comes first), then repeat.
    - The timeout duration controls the minimum time between `select`-polling when none of the busy threads finish their work sooner. Maybe 1 second is a bit high- not sure yet what would work best for most cases.

So far I've only run this PR through some manual load-testing on a single Rails route with some potentially-promising results, but I haven't done exhaustive testing/validation yet. I plan to eventually test out these changes on my production workload described in https://github.com/puma/puma/issues/1254#issuecomment-420112799. If anyone has any self-contained benchmark scripts or concurrency-test harnesses that might be useful to test this PR against please let me know.

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -396,7 +396,7 @@ module Puma
                     end
 
                     pool << client
-                    pool.wait_until_not_full
+                    pool.wait_until_not_full(sock)
                   end
                 rescue SystemCallError
                   # nothing
--- a/lib/puma/thread_pool.rb
+++ b/lib/puma/thread_pool.rb
@@ -21,6 +21,12 @@ module Puma
     # up its work before leaving the thread to die on the vine.
     SHUTDOWN_GRACE_TIME = 5 # seconds
 
+    # Duration between non-blocking socket polls
+    # when extra threads are available but no work.
+    # This setting helps load-balance work across multiple processes
+    # before balancing across multiple threads.
+    WORK_AVAILABLE_TIMEOUT = 1 # seconds
+
     # Maintain a minimum of +min+ and maximum of +max+ threads
     # in the pool.
     #
@@ -190,20 +196,28 @@ module Puma
     # then the `@todo` array would stay the same size as the reactor works
     # to try to buffer the request. In tha scenario the next call to this
     # method would not block and another request would be added into the reactor
-    # by the server. This would continue until a fully bufferend request
+    # by the server. This would continue until a fully buffered request
     # makes it through the reactor and can then be processed by the thread pool.
-    def wait_until_not_full
+    def wait_until_not_full(sock)
       @mutex.synchronize do
         while true
           return if @shutdown
 
-          # If we can still spin up new threads and there
-          # is work queued that cannot be handled by waiting
-          # threads, then accept more work until we would
-          # spin up the max number of threads.
-          return if @todo.size - @waiting < @max - @spawned
+          busy_threads = @spawned - @waiting + @todo.size
+          threads_available = @max - busy_threads > 0
 
-          @not_full.wait @mutex
+          # Accept more work if no threads are busy,
+          # or if extra threads and non-blocking work are available.
+          return if busy_threads.zero? ||
+            (threads_available && IO.select([sock], nil, nil, 0))
+
+          # If threads are available but no work,
+          # check for work again after a short timeout.
+          # Otherwise if threads are busy, block without timeout
+          # until a thread finishes its work.
+          timeout = threads_available ? WORK_AVAILABLE_TIMEOUT : nil
+
+          @not_full.wait @mutex, timeout
         end
       end
     end


evanphx:
This is super good and a great idea!

Thinking through it, I wonder if the main issue here is that if all workers have at least one thread working and a client connects, it might take up to 1 second before any worker realizes there is a new client and goes off to do the accept. I could see that happening in the case of a well loaded server, with the workers sometimes happening to check only a few milliseconds after the client tries to connect and then random it taking a full second. People would see observer that externally as a random 1 second latency in requests, which would be pretty odd.

This is a really good idea and we just need to think through the logic to be sure that new clients are served quickly.

wjordan:
Here's a first pass at benchmarking this PR using the [TechEmpower Ruby Sinatra Benchmarking Test](https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/Ruby/sinatra) mentioned in https://github.com/puma/puma/issues/1254#issue-216906287:

I ran these tests on a server with 40 vCPUs, configured with 1, 5, or 5* (running this PR) threads, concurrency 40, running the 'query' test with a query-level of 20. Here is the command-line used:

```
./tfb --test sinatra --concurrency-levels 40 --type query --query-levels 20
```

Results:

## Summary
- 1 thread: **3458.32** req/sec
- 5 threads: **2125.11** req/sec (39% less than 1-thread)
- 5* threads: **2993.21** req/sec (13% less than 1-thread, 41% more than 5-thread)

## Full data

<details>
<summary>1 thread</summary>

```
sinatra: [10] Puma starting in cluster mode...
sinatra: [10] * Version 3.12.0 (ruby 2.4.4-p296), codename: Llamas in Pajamas
sinatra: [10] * Min threads: 1, max threads: 1
sinatra: [10] * Environment: production
sinatra: [10] * Process workers: 40

 Queries: 20 for query
 wrk -H 'Host: tfb-server' -H 'Accept: application/json,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 40 --timeout 8 -t 40 "http://tfb-server:8080/queries?queries=20"
---------------------------------------------------------
Running 15s test @ http://tfb-server:8080/queries?queries=20
  40 threads and 40 connections
    Latency    11.35ms    2.78ms  71.50ms   80.34%
    Req/Sec    87.03      9.24   121.00     73.40%
  Latency Distribution
     50%   10.46ms
     75%   11.77ms
     90%   15.20ms
     99%   20.50ms
  52222 requests in 15.10s, 39.07MB read
  Socket errors: connect 0, read 52220, write 0, timeout 0
Requests/sec:   3458.32
Transfer/sec:      2.59MB
STARTTIME 1536794976
ENDTIME 1536794991
Benchmark results:
{'results': [{'connect': 0,
              'endTime': 1536794991,
              'latencyAvg': '11.35ms',
              'latencyMax': '71.50ms',
              'latencyStdev': '2.78ms',
              'read': 52220,
              'startTime': 1536794976,
              'timeout': 0,
              'totalRequests': 52222,
              'write': 0}]}
```
</details>

<details>
<summary>5 threads</summary>

```
 Queries: 20 for query
 wrk -H 'Host: tfb-server' -H 'Accept: application/json,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 40 --timeout 8 -t 40 "http://tfb-server:8080/queries?queries=20"
---------------------------------------------------------
Running 15s test @ http://tfb-server:8080/queries?queries=20
  40 threads and 40 connections
    Latency    18.58ms    7.01ms  57.54ms   70.36%
    Req/Sec    53.41      9.63    80.00     71.38%
  Latency Distribution
     50%   17.28ms
     75%   22.53ms
     90%   27.46ms
     99%   41.22ms
  32087 requests in 15.10s, 24.01MB read
  Socket errors: connect 0, read 32087, write 0, timeout 0
Requests/sec:   2125.11
Transfer/sec:      1.59MB
STARTTIME 1536794526
ENDTIME 1536794541
Benchmark results:
{'results': [{'connect': 0,
              'endTime': 1536794541,
              'latencyAvg': '18.58ms',
              'latencyMax': '57.54ms',
              'latencyStdev': '7.01ms',
              'read': 32087,
              'startTime': 1536794526,
              'timeout': 0,
              'totalRequests': 32087,
              'write': 0}]}
```
</details>

<details>
<summary>5* threads</summary>

```
sinatra: [9] Puma starting in cluster mode...
sinatra: [9] * Version 3.12.0 (ruby 2.4.4-p296), codename: Llamas in Pajamas
sinatra: [9] * Min threads: 5, max threads: 5
sinatra: [9] * Environment: production
sinatra: [9] * Process workers: 40

 Queries: 20 for query
 wrk -H 'Host: tfb-server' -H 'Accept: application/json,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 40 --timeout 8 -t 40 "http://tfb-server:8080/queries?queries=20"
---------------------------------------------------------
Running 15s test @ http://tfb-server:8080/queries?queries=20
  40 threads and 40 connections
    Latency    13.17ms    3.97ms  50.42ms   74.66%
    Req/Sec    75.13      9.54   111.00     72.93%
  Latency Distribution
     50%   12.14ms
     75%   15.14ms
     90%   18.62ms
     99%   25.85ms
  45199 requests in 15.10s, 33.82MB read
  Socket errors: connect 0, read 45197, write 0, timeout 0
Requests/sec:   2993.21
Transfer/sec:      2.24MB
STARTTIME 1536794378
ENDTIME 1536794393
Benchmark results:
{'results': [{'connect': 0,
              'endTime': 1536794393,
              'latencyAvg': '13.17ms',
              'latencyMax': '50.42ms',
              'latencyStdev': '3.97ms',
              'read': 45197,
              'startTime': 1536794378,
              'timeout': 0,
              'totalRequests': 45199,
              'write': 0}]}
```
</details>

wjordan:
I've now tested this PR on my production workload (see https://github.com/puma/puma/issues/1254#issuecomment-420112799 for earlier results) and confirmed that this patch brings Puma's multithreaded performance significantly closer to single-threaded Unicorn in my tests.

> Thinking through it, I wonder if the main issue here is that if all workers have at least one thread working and a client connects, it might take up to 1 second before any worker realizes there is a new client and goes off to do the accept. I could see that happening in the case of a well loaded server, with the workers sometimes happening to check only a few milliseconds after the client tries to connect and then random it taking a full second. People would see observer that externally as a random 1 second latency in requests, which would be pretty odd.

Yeah, it looks like the tradeoff is frequency of `select`-polling on the socket (and associated overhead of re-acquiring the GVL and thread-pool mutex) versus max queue-latency when all workers are busy (averaging to something like `timeout / workers` when requests take longer than `timeout` to respond). It's possible this timeout could be reduced to 100ms or even 10ms without significant overhead, maybe a bit more testing/benchmarking could determine the best default for this value.

evanphx:
@wjordan That's great news! So I took a crack at improving the logic a bit and pushed the change to a f-even-accept branch. Check out the commit https://github.com/puma/puma/commit/e326b75f638ef881a9ae0bb536702952a6de5d3d.

Would you be open to trying out my change on your production system and reporting back on the numbers?

dmfallak:
I tested with a patch from your f-even-accept branch, and it seems to improve request time, specifically I notice that the time spent in Redis and DB is reduced.

nateberkopec:
I would say @evanphx's branch needs a benchmark + probably a test and we can move forward with this.

wjordan:
> Would you be open to trying out my change on your production system and reporting back on the numbers?

Sorry, I haven't found the time to try out `f-even-accept` in my production environment yet - once I got the migrated system running stable on this PR I moved on to other things. In any case, I agree with @nateberkopec that an isolated benchmark (perhaps a simplified/extracted variant of the test described in https://github.com/puma/puma/pull/1646#issuecomment-420845786 would suffice) would be most useful in the long run to verify the improvement from this approach and prevent any future regressions.

nateberkopec:
I took a stab at trying to make a simpler reproducing benchmark and couldn't quite come up with something. This was what I tried, but it benchmarks the same on master and on f-even-accept.

```ruby
QUERIES_PER_REQUEST = 20
TIME_TO_USE_GVL_IN_SEC = 0.005
TIME_TO_UNLOCK_GVL_IN_SEC = 0.002

run lambda { |env| 
  QUERIES_PER_REQUEST.times do 
    t = Time.now
    spin_with_gvl_until = t + TIME_TO_USE_GVL_IN_SEC
    while Time.now <= spin_with_gvl_until
      # spin baby spin
    end
    sleep(TIME_TO_UNLOCK_GVL_IN_SEC)
  end

  [
    200, 
    {"Content-Type" => "text/plain"}, 
    ["Hello World"]
  ]
}
```

leandromoreira:
Just out of curiosity how does Puma (the master worker) choses which process should work handle a specific req? (in the worker config (wiht thread pool) for MRI)

Does it stick or always does the lb? (i couldn't find this on code)

nateberkopec:
> Just out of curiosity, today how does Puma (the master worker) choses which process should work handle a specific req?

Depends on what OS you're on. See the nio4r docs. On Linux, I think you're using epoll by default.

More or less, it's random, and the OS decides, not us.

leandromoreira:
> > Just out of curiosity, today how does Puma (the master worker) choses which process should work handle a specific req?
> 
> Depends on what OS you're on. See the nio4r docs. On Linux, I think you're using epoll by default.
> 
> More or less, it's random, and the OS decides, not us.

Thanks, I was under the impression that this code https://github.com/puma/puma/blob/2668597ec1dd9546d83db9f2ec5ad092add483e6/lib/puma/cluster.rb#L490-L500 was doing that but it doesn't seem the case, I'll read the https://www.rubydoc.info/gems/nio4r to try to understand a little bit!

nateberkopec:
Closing and moving to #1920

PR #1650, opened 2018-09-18
fix(jungle): change pid existence check's condition branches

adam101:
Pull request https://github.com/puma/puma/pull/1545 broke the ability to stop and restart running instances. The old version of checking $PID existence was truthy **on non-existence**, now it's truthy **on existence**.

Example in case of existing PID 1209:

```sh
$ if [ "`ps -A -o pid= | grep -c 1209`" -eq 0 ]; then echo true; else echo false; fi;
true
```

```sh
$ if ps -p 1209 > /dev/null; then echo true; else echo false; fi;
false
```


p8:
This fixes an issue we had with stopping and restarting. 👍

PR #1655, opened 2018-09-24
Don't call .stop on a server that doesn't exist

### LINKED ISSUE #1523: Puma::Single#stop raises undefined method `stop' for nil:NilClass
idk when or how this happens ... just found that in our error reporter ... also not sure how bad that is, but smells like something is going wrong :(

```
Undefined method `stop' for nil:NilClass
--
ruby/2.5.0/gems/puma-3.8.2/lib/puma/single.rb:18
ruby/2.5.0/gems/puma-3.8.2/lib/puma/single.rb:18 in stop
ruby/2.5.0/gems/puma-3.8.2/lib/puma/launcher.rb:146 in stop
ruby/2.5.0/gems/puma-3.8.2/lib/puma/launcher.rb:384 in block in setup_signals
```

patching it with this in the hope it makes some strange hang-on-restart errors we saw go away ...

```
Puma::Single.prepend(Module.new do
  def stop
    super
  rescue NoMethodError
  end
end)
```

mipearson:
Ignore the branch name - this turns out to work! Fixes #1523 for us on 3.12.0.

Issue we had was a misconfiguration was causing puma to be SIGTERM'd by ECS before it could set `@server` to anything, which meant that as it tried to graceful stop it hit #1523, confusing our debugging efforts.

### DIFF
--- a/lib/puma/single.rb
+++ b/lib/puma/single.rb
@@ -24,7 +24,7 @@ module Puma
     end
 
     def stop
-      @server.stop false
+      @server.stop(false) if @server
     end
 
     def halt
@@ -34,7 +34,7 @@ module Puma
     def stop_blocked
       log "- Gracefully stopping, waiting for requests to finish"
       @control.stop(true) if @control
-      @server.stop(true)
+      @server.stop(true) if @server
     end
 
     def jruby_daemon?


mipearson:
@cpinto I don't, sorry. I remember it was something specific to our code. I'd recommend sourcing this branch into your `Gemfile` if only to see where the source error lies, and then reverting to the gem version once fixed.

PR #1667, opened 2018-10-29
Update README.md

sahilda:
Fixes header formatting


evanphx:
Rebased and picked up the existnig fix!

Issue #1663, opened 2018-10-25
Non preload worker boot failure infinite loop


stefansedich:
When running in clustered mode without preloading and if my worker fails to boot (unable to connect to database in this example) it will forever create a new worker and never fail, ideally I would like to be able to say to terminate if a worker fails to boot in the same way that the entire process will terminate if I were using preloading.

Becomes an issue as in my case it sat there spamming my logs without ever terminating, when running with preload or without clustrering obviously it just terminates the entire process right away, is it possible to achieve what I need at the moment?

stefansedich:
Having a play around with this, I experimented by adding a new `max_worker_terminations` configuration parameter which can control the number of worker terminations we accept before we want to nuke the entire process.

Any feedback or thoughts would be appreciated and I can wrap it up into a PR if everyone agrees this would be valuable.

evanphx:
Unsure what a better way to handle this is. The point of the cluster mode is that the workers are restarted because they might have crashed.

PR #1668, opened 2018-10-30
Add ability to control the maximum number of worker terminations

stefansedich:
This is an initial cut at fixing #1663 by adding a new configuration parameter that will enable you to set the maximum number of worker terminations you will accept before nuking the entire parent process.

The other idea instead of a count was to simply set if you accept worker terminations or not when running in clustered mode which would probably simplify things even more.

### DIFF
--- a/lib/puma/cli.rb
+++ b/lib/puma/cli.rb
@@ -100,6 +100,8 @@ module Puma
     def setup_options
       @conf = Configuration.new do |user_config, file_config|
         @parser = OptionParser.new do |o|
+          o.set_summary_width 38
+
           o.on "-b", "--bind URI", "URI to bind to (tcp://, unix://, ssl://)" do |arg|
             user_config.bind arg
           end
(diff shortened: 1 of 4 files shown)


lib/puma/cli.rb
```diff
@@ -100,6 +100,8 @@ def configure_control_url(command_line_arg)
     def setup_options
       @conf = Configuration.new do |user_config, file_config|
         @parser = OptionParser.new do |o|
+          o.set_summary_width 38
```
evanphx: Probably best to remove this in this PR.

stefansedich:
@evanphx that was the idea, the issue in our case is that if running in clustered mode and something happens to cause workers to restart they will reboot in a super tight loop causing logs to balloon and it gives me no real way to know that bad things are happening.

If the process exists it at least causes our containers to be restarted on a slower loop which reduces the issue and also gives me a way to see that the containers themselves are constantly rebooting which is something that I can pick up on easier.

lib/puma/cli.rb
stefansedich: It did cause the output to look odd which was the reason for adding, will remove after the main discussion after the PR in general.

evanphx:
In that case, we'd need to have some pretty explicit documentation about why a person would use it.

And also decide if the parent process should be performing cleaning, such as terminating the other workers and closing sockets, etc.

stefansedich:
@evanphx I am going to close this for now as my focus has shifted and I will get back to it when I can again.

Issue #1670, opened 2018-11-01
SSL Error with self-signed certificate in dev when using Chrome 70


eric-norcross:
### Steps to reproduce

- Setup a dev environment on macOS 10.14.1
- Setup and configure a self-signed certificate 
- Try to connect with Chrome 70

### Expected behavior

It should connect with Chrome 70

### Actual behavior

Puma throws the error: `SSL error, peer: 127.0.0.1, peer cert: , #<Puma::MiniSSL::SSLError: OpenSSL error: error:141F7065:SSL routines:final_key_share:no suitable key share - 337604709>`

When attempting to connect on Chrome 70, Chrome shows the following error 
<img width="685" alt="screen shot 2018-11-01 at 9 14 21 am" src="https://user-images.githubusercontent.com/1047774/47864405-be391b00-ddb6-11e8-81d1-cf9baba31beb.png">

When accessing `https://localhost:8080` on `FireFox 63.0 (64-bit)` I'm prompted to add a one-time security exception in order connect. `Safari Version 12.0.1` has no issue connecting. 

### System configuration

- `ruby 2.5.3`
- `rails 5.1.6`
- `puma 3.12.0`
- `macOS 10.14.1`
- `Chrome Version 70.0.3538.77 (Official Build) (64-bit)`
- I've created a self-signed certificate following the Heroku instructions [here](https://devcenter.heroku.com/articles/ssl-certificate-self)
- I have the following in my `hosts` file:

```
    127.0.0.1	localhost
    255.255.255.255	broadcasthost
    ::1             localhost
    127.0.0.1 localhost.ssl
```

- Added the self-signed certificate to my `system` keychain and always trusted it
- In my `puma` config file, i'm starting puma with: `ssl_bind '127.0.0.1', '8080', { key: 'localhost_ssl/server.key', cert: 'localhost_ssl/server.crt', verify_mode: 'none' }`

I've also added [an issue](https://bugs.chromium.org/p/chromium/issues/detail?id=899994) in Chrome bug tracker

davidben:
Looks like the issue is this line here:
https://github.com/puma/puma/blob/72882f2319e65b371e1458069723279b3196a220/ext/puma_http11/mini_ssl.c#L193

[...]

PR #1671, opened 2018-11-02
Implemented NID_X9_62_prime256v1 (P-256) curve over P-521

eric-norcross:
Implemented NID_X9_62_prime256v1 (P-256) curve over P-521 in order to support Chrome 70 and Edge


eric-norcross:
Sorry, I forgot to reference the issue this resolves:
https://github.com/puma/puma/issues/1670

PR #1685, opened 2018-12-08
Do not accept new requests on shutdown

mainameiz:
Related to https://github.com/puma/puma/pull/1186

**Problem:**

When puma process receives `TERM` signal it tries to finish all current requests and terminate. But... it also accepts "new requests" and these requests are hanged. When the processing of all current requests (which was made before shutdown signal is received) are finished, all "new requests" are closed with error `Connection reset by peer`.
This error prevents load balancers (like nginx) to forward these requests to other web-server, because this error does not guarantee that the request was not processed, but this guarantee is required to forward not idempotent HTTP request (like POST).

Seems like puma should refuse new requests by closing a socket that accepts new connections.

Read https://github.com/puma/puma/pull/1186#issuecomment-445462578 for implementation details.

This solution is simple and fast. But probably there is a better solution.

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -79,6 +79,7 @@ module Puma
       @first_data_timeout = options.fetch(:first_data_timeout, FIRST_DATA_TIMEOUT)
 
       @binder = Binder.new(events)
+      @own_binder = true
 
       @leak_stack_on_error = true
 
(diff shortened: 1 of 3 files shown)


schneems:
Just now seeing this, looks like this handles the case for threaded only servers, I ended up having to do something extremely similar for multi-process workers.

PR #1688, opened 2018-12-18
Fix formatting for SSL cipher suite sections

michaelherold:
Some of the formatting for this section was messed up and it made it hard to read through the documentation. This change fixes that formatting and makes it easier to read.


michaelherold:
It looks like I'm fixing the same thing here as in #1667. Please feel free to take that one if you wish.

evanphx:
Thanks! looks like we picked up the same fix elsewhere.

PR #1689, opened 2018-12-19
Allow mutual TLS CA to be set using `ssl_bind` DSL

michaelherold:
When using mutual TLS, you must specify the CA certificate chain to use
for verifying the peer. Using Puma's `ssl_bind` DSL did not give you the
option of doing so, which lead to confusing errors when attempting to
use it.

Now, when specifying the `verify_mode` as either `peer` or `force_peer`,
you can use the DSL to set the `ca` value [as needed within the `Binder`][binder].
This allows you to use the DSL instead of falling back to the default
`bind` syntax via the URI-style configuration pattern.

[binder]: https://github.com/puma/puma/blob/2668597ec1dd9546d83db9f2ec5ad092add483e6/lib/puma/binder.rb#L180-L184

### DIFF
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -296,13 +296,14 @@ module Puma
     def ssl_bind(host, port, opts)
       verify = opts.fetch(:verify_mode, 'none')
       no_tlsv1 = opts.fetch(:no_tlsv1, 'false')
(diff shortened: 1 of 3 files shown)


evanphx:
Could you rebase this on master? There have been a few fixes around this code merged lately.

michaelherold:
Done!

PR #1690, opened 2018-12-19
Add option to suppress SignalException on SIGTERM

mic-kul:
What:
This PR adds config option to disable raising SignalException when SIGTERM is received.

Why:
When running Puma inside Docker container that is orchestrated by Kubernetes/Docker swarm/ECS or other container platform SIGTERM is something expected, usually instructing app to shut down gracefully. Rolling restart is handled by infrastructure / orchestrator itself.
SignalException should not be raised in this case.


evanphx:
Is this still a WIP?

mic-kul:
Thanks for the reminder. Committed my recent changes and it's ready for another look @evanphx

schneems:
> SignalException should not be raised in this case.

The exception is used by many libraries to clean up on SIGTERM. Firing that exception forces all the ensure blocks of the application to fire, for example telling libraries to safely close references to files etc. I don't have a problem with this being a configurable option, but I do worry about the unintended side effects of enabling this for long term system stability.

mic-kul:
@schneems I agree it has side effects - ie. Kafka buffers might not be flushed properly.

I just couldn't understand why clustered mode raises it, while single mode does not.
In my current project we are overriding signal to SIGINT and it works without triggering on-error hooks for Airbrake/Rollbar libraries.

schneems:
I think in single mode the exception would be raised directly by the process since it’s what gets the sigterm versus the code in clustered is designed to propagate the sigterm. I think anyway.

PR #1691, opened 2018-12-20
re-implement (native) IOBuffer for JRuby

kares:
... to match up JRuby's ext with the C ext

extensions now provide the same functionality - no need for branching (around `Puma::IOBuffer`)

this has been motivated by warnings on Java 9+ [due](https://github.com/puma/puma/blob/master/lib/puma/java_io_buffer.rb#L10) `field_reader :buf` having protected access
since JRuby does not generate native classes for Java sub-classes (atm): https://github.com/jruby/jruby/issues/5532


evanphx:
@headius Would you mind giving this a review?

headius:
Oops, never saw this. @enebo or I would be fine, but I'm traveling now.

ext/puma_http11/org/jruby/puma/IOBuffer.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @JRubyMethod
+    public RubyInteger capacity(ThreadContext context) {
+        return context.runtime.newFixnum(buffer.unsafeBytes().length);
+    }
+
+    @JRubyMethod
+    public IRubyObject reset() {
+        buffer.setRealSize(0);
```
enebo: Not sure this is actionable but this won't reduce/remove backing array of bytelist.  I would not worry about it but if someone loads a massive IOBuffer and it is reused it will just drag around a massive backing array as long as it lives.

kares:
sorry about that I removed the comment - was half asleep and did not notice the approval first.
probably as they changed the UI style and there wasn't a green mark next to your comment.

enebo:
@kares no worries.  I agree my comment would be confusing to read even with the approval.  I should have probably mentioned it after the merge.

enebo:
@kares @olleolleolle I know this was motivated by Java 9+ warnings but do you anticipate any perf gains now that this landed?

kares:
not really, except for some byte copy-ing ... maybe slightly less mem use (not sure what the buff is used for)

adurgin:
@evanphx @kares @enebo  Hi. Would it be possible to release a gem with this fix included?

PR #1695, opened 2018-12-28
Avoid Ruby warnings

olleolleolle:
This PR avoids two **Ruby warnings** output when running the test suite.

```
/Users/olle/opensource/puma/lib/puma/minissl.rb:5: warning: mismatched indentations at 'rescue' with 'begin' at 3
```

and

```
/Users/olle/opensource/puma/test/test_puma_server_ssl.rb:150: warning: assigned but unused variable - ssl_listener
```

### DIFF
--- a/lib/puma/minissl.rb
+++ b/lib/puma/minissl.rb
@@ -2,7 +2,7 @@
 
 begin
   require 'io/wait'
-  rescue LoadError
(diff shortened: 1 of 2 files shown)


olleolleolle:
This should build fine now, after MSP-Greg's commit.

PR #1706, opened 2019-01-11
Fix restart for Ruby 2.6 and Bundler 2

exviva:
Ruby 2.6 ships with Bundler 1.17, and we're requiring it early
by putting `-rbundler/setup` in `ENV['RUBYOPT']`. If an application
is using Bundler 2, the app's Gemfile will cause an error
for Bundler 1.17, something along the lines of "You must use
Bundler 2 or greater with this lockfile. (Bundler::LockfileError)".

The code in PR makes sure that the process spawned during a restart
will require the same version of Bundler as is used by the current process.

### DIFF
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -367,6 +367,18 @@ module Puma
       lib = File.expand_path "lib"
       arg0[1,0] = ["-I", lib] if [lib, "lib"].include?($LOAD_PATH[0])
 
+      # Ruby 2.6 ships with Bundler 1.17, and we're requiring it early
+      # by putting `-rbundler/setup` in ENV['RUBYOPT']. If an application
+      # is using Bundler 2, the app's Gemfile will cause an error
+      # for Bundler 1.17, something along the lines of "You must use
+      # Bundler 2 or greater with this lockfile. (Bundler::LockfileError)".
+      #
+      # The code below makes sure that the process spawned during a restart
+      # will require the same version of Bundler as is used by the current process.
+      if spec = Gem.loaded_specs['bundler']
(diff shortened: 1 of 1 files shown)


lib/puma/launcher.rb
```diff
@@ -367,6 +367,18 @@ def generate_restart_data
...
+      # Ruby 2.6 ships with Bundler 1.17, and we're requiring it early
+      # by putting `-rbundler/setup` in ENV['RUBYOPT']. If an application
+      # is using Bundler 2, the app's Gemfile will cause an error
+      # for Bundler 1.17, something along the lines of "You must use
+      # Bundler 2 or greater with this lockfile. (Bundler::LockfileError)".
+      #
+      # The code below makes sure that the process spawned during a restart
+      # will require the same version of Bundler as is used by the current process.
+      if spec = Gem.loaded_specs['bundler']
+        arg0[1,0] = spec.full_require_paths.flat_map {|p| ['-I', p] }
```
olleolleolle:
```suggestion
        arg0[1, 0] = spec.full_require_paths.flat_map { |path| ['-I', path] }
```

This is great, with lots of clarity in the comment.

My suggestion was stylistic (`Kernel#p` is a method.)

exviva:
Thanks for the review @olleolleolle

exviva:
Turns out updating rubygems to 3.0.2 fixed the issue, so closing this.

mortik:
Still got that issue with rubygems 3.0.3 and bundler 2.0.1, ruby 2.6.1

mortik:
tried to remove prune_bundler from my puma config and now it seems to work... am i supposed to remove prune_bundler when using bundler 2?

PR #1709, opened 2019-01-19
Remove URI backport for Ruby 1.9.3

bdewater:
Ruby 2.2+ has been a requirement since https://github.com/puma/puma/pull/1506


bdewater:
The single test failure is unrelated to the change.

PR #1716, opened 2019-01-25
Min worker timeout

### LINKED ISSUE #1714: worker_timeout <= Cluster::WORKER_CHECK_INTERVAL causes Terminations
### Steps to reproduce

1) Set worker_timeout to Cluster::WORKER_CHECK_INTERVAL or lower.

2) Run puma in cluster mode.

### Expected behavior

It should only terminate processes that don't report in within the worker_timeout.

### Actual behavior

If you set worker_timeout to < Cluster::WORKER_CHECK_INTERVAL it will endlessly boot and terminate workers:

```
[ActionCable] allowed origins: ["https://beta5"]
[14944] - Worker 0 (pid: 20045) booted, phase: 0
[14944] ! Terminating timed out worker: 19785
[14944] - Spawned worker: 20155
[ActionCable] allowed origins: ["https://beta5"]
[14944] - Worker 1 (pid: 20155) booted, phase: 0
[14944] ! Terminating timed out worker: 20045
[14944] - Spawned worker: 20273 
```
if you set worker_timeout == Cluster::WORKER_CHECK_INTERVAL the behaviour gets a bit more interesting, you'll get occasional terminations as you run just over the limit. Either from the worker being slow to report or the cluster being slow to check the reporting - typically happening under load. From first hand experience these are super confusing to debug!

The immediate solution is two-fold, updating the config docs to mention the effective minimum timeout and actually aborting on startup if the worker_timeout is too low.

### System configuration

**Ruby version**: 2.5.0
**Rails version**:  6.0.0.alpha
**Puma version**: 3.12.0

mdkent:
Fixes #1714.

Abort on startup if the worker_timeout is too low and mention the limit.

6 is a bit awkward though, might be cleaner to set the minimum at 2 * the WORKER_CHECK_INTERVAL.

### DIFF
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -434,6 +434,13 @@ module Puma
     # that have not checked in within the given +timeout+.
     # This mitigates hung processes. Default value is 60 seconds.
     def worker_timeout(timeout)
+      timeout = Integer(timeout)
+      min = Cluster::WORKER_CHECK_INTERVAL
+
+      if timeout <= min
+        raise "The minimum worker_timeout must be greater than the worker reporting interval (#{min})"
+      end
+
       @options[:worker_timeout] = Integer(timeout)
     end
 
--- a/examples/config.rb
+++ b/examples/config.rb
@@ -174,7 +174,8 @@
 # the given timeout. If not the worker process will be restarted. This is
 # not a request timeout, it is to protect against a hung or dead process.
 # Setting this value will not protect against slow requests.
-# Default value is 60 seconds.
+#
+# The minimum value is 6 seconds, the default value is 60 seconds.
 #
 # worker_timeout 60
 


lib/puma/dsl.rb
```diff
@@ -434,6 +434,13 @@ def tag(string)
...
     # This mitigates hung processes. Default value is 60 seconds.
     def worker_timeout(timeout)
+      timeout = Integer(timeout)
+      min = Cluster::WORKER_CHECK_INTERVAL
+
+      if timeout <= min
+        raise "The minimum worker_timeout must be greater than the worker reporting interval (#{min})"
+      end
+
       @options[:worker_timeout] = Integer(timeout)
```
bogn83: Shouldn't this use the already casted timeout here now? Why cast again?

PR #1728, opened 2019-02-20
Switch IO reactor to nio4r

evanphx:
This moves away from IO.select to using nio4r to allow the reactor to
scale beyond 1024 active clients. This happens when folks are using
websockets usually.


MSP-Greg:
@evanphx

nio4r uses libev, which isn't Windows compatible (would need to change to libuv).  Thoughts?  I think Appveyor will probably break with this.

evanphx:
@MSP-Greg Happy to consider other options, for sure. On Windows it will fall back to IO.select I believe, which means Windows would stay as-is. Would be nice to allow Windows to take advantage of this change, but it's net neutral otherwise.

MSP-Greg:
> On Windows it will fall back to IO.select I believe

Sorry, I should have looked at it more.  Passed on Appveyor.

nio4r is aware of the issue, but they thought switching to libuv is (understandably) something they're not interested in changing at this time.  I'm not a c type, so...

lib/puma/reactor.rb
```diff
@@ -122,36 +129,48 @@ def initialize(server, app_pool)
...
+      selector = @selector
 
       while true
         begin
-          ready = IO.select sockets, nil, nil, @sleep_for
+          ready = selector.select @sleep_for
         rescue IOError => e
           Thread.current.purge_interrupt_queue if Thread.current.respond_to? :purge_interrupt_queue
-          if sockets.any? { |socket| socket.closed? }
+          if sockets.any? { |socket| m.value.closed? }
```
nateberkopec: `m.value`?

nateberkopec: You've used this all over the place put I'm totally confused on where `m` comes from.

lib/puma/reactor.rb
```diff
@@ -122,36 +129,48 @@ def initialize(server, app_pool)
...
           else
             raise
           end
         end
 
-        if ready and reads = ready[0]
-          reads.each do |c|
-            if c == @ready
+        if ready
+          ready.each do |m|
```
nateberkopec: maybe we can think of something more descriptive than `m`

lib/puma/reactor.rb
```diff
@@ -182,7 +204,8 @@ def run_internal
                 c.write_500
                 c.close
 
-                sockets.delete c
+                selector.deregister m
```
nateberkopec: maybe this could be it's own method

HoneyryderChuck:
> You can avoid allocating an array each time you call #select by passing a block to select. 

From https://github.com/socketry/nio4r/wiki/Selectors#selecting-io-objects-for-readiness , if it's a nice-to-have.

lib/puma/reactor.rb
evanphx: Yup, done.

lib/puma/reactor.rb
```diff
@@ -121,37 +128,69 @@ def initialize(server, app_pool)
...
+                        # There is a bug where we seem to be registering an already registered
+                        # client. This code deals with this situation but I wish we didn't have to.
+                        monitors.delete_if { |submon| submon.value.to_io == c.to_io }
+                        selector.deregister(c)
+                        mon = selector.register(c, :r)
+                      end
+                    rescue IOError
+                      # Means that the io is closed, so we should ignore this request
+                      # entirely
+                    else
```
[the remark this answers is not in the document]
evanphx: One of my fav underused constructs!

MSP-Greg:
The Appveyor build **after** the rebase is a false positive.  Apologies, I never allowed for gem installation to fail.  It should work now...

ioquatix:
As the maintainer of nio4r, I'd be happy to evaluate other options. Or even build something entirely new on top of `libuv`. I don't think there has been any decision like we shouldn't rebuild `nio4r` on top of `libuv` just FYI. The surface area of `libuv` is huge compared to `nio4r`.

I think `nio4r` is fine for what it is. I don't think it will still be useful in 10 years. Things are moving rapidly in this space, e.g. `liburing`.

The actual problem space is pretty large though. For example, `nio4r` is primarily concerned with exposing non-blocking network sockets, but there are many areas where blocking behaviour still exists: disk IO (e.g. `read`, `write`, `readdir`), database IO (e.g. `libpq`), name resolution (e.g. DNS) and a whole ton of other stuff.

If your only issue is blocking network IO, then `nio4r` is fine and a solid solution. It would be nice to make sure it is as good as possible, e.g. on Windows too, but honestly that's not a huge priority for me.

PR #1736, opened 2019-03-07
JRuby - move to Xenial, set all to 'allow_failures'

MSP-Greg:
Recent Travis 'CRON' jobs are failing on JRuby 9.2.5.0.

I don't use JRuby.  I looked at `ruby/psych`, both of it's JRuby jobs are 'allow failure'.  Couldn't find a repo/gem using OpenSSL that tests on JRuby.

The PR allows Travis to pass.  Note that jruby-head normally passes.  I don't know enough about JRuby to know why 9.2.5.0/9.2.6.0 is failing.  Also, http://rubies.travis-ci.org/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ before_install:
     r_eng="$(ruby -e 'STDOUT.write RUBY_ENGINE')";
     rv="$(ruby -e 'STDOUT.write RUBY_VERSION')";
     if [ "$r_eng" == "ruby" ]; then
-      if   [ "$rv" \< "2.3" ]; then gem update --system 2.7.8 --no-document
+      if   [ "$rv" \< "2.3" ]; then gem update --system 2.7.9 --no-document
       elif [ "$rv" \< "2.6" ]; then gem update --system --no-document --conservative
       fi
     fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,12 @@ cache: bundler
 before_install:
   # rubygems 2.7.8 and greater include bundler, leave 2.6.0 untouched
   - |
+    r_eng="$(ruby -e 'STDOUT.write RUBY_ENGINE')";
     rv="$(ruby -e 'STDOUT.write RUBY_VERSION')";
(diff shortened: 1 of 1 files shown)


ahorek:
I can't replicate the failure locally, but I saw this on different repositories a few times after jruby-openssl 0.10.2 was released.
It might be something wrong on travis, broken caches or something... @kares have you seen this before?

MSP-Greg:
@ahorek thanks for checking it locally.  I figured it was something unrelated to Puma, as jruby-head seems stable.

The Appveyor error is caused by a failed lookup at two keyservers, which is used for the MinGW OpenSSL 1.0.2 package for Ruby 2.4, which is no longer the current MinGW package (which is 1.1.1).  The lookup succeeded for the 32 bit build, but failed for the 64 bit.  Hence, not related to Puma.  I'll see if I can find a reliable way to install the package.  This isn't the first time it's happened...

MSP-Greg:
I updated the PR to disable the rubygems update code for jruby.  It passed in my fork, and it's passed here.  When a few more builds have been completed, jruby 9.2.6.0 should be removed from 'allow failure'...

kares:
haven't seen the failures before but if they all look like [this one](https://travis-ci.org/puma/puma/jobs/502230265#L1230) than its very likely smt to do with RubyGems and/or Bundler. there were issues with platform gems before, the new jossl 0.10.2 did not change that much in terms of gemspec its pretty much ~ as 0.10.1 (and previous versions). 

maybe a decent resolution would be not to update the jruby-openssl default gem, just remove : 
```
gem "jruby-openssl", :platform => "jruby"
```
... from the Gemfile esp. since you're only testing recent JRuby versions (was necessary before 9K)

MSP-Greg:
@kares

Thanks for looking into it.

> haven't seen the failures before but if they all look like this one than its very likely smt to do with RubyGems and/or Bundler.

I guessed the same, that's why I forced an update bypassing the `gem update --system` code for the two JRuby jobs.  The code solved some issues with MRI ruby on Travis.  Both JRuby jobs are now passing, but the fact that `gem update --system` breaks the jobs is interesting...

Issue #1674, opened 2018-11-08
Puma hangs on shutting down workers when received SIGINT


atitan:
### Steps to reproduce
```
bundle exec puma -C config/puma.rb

# in another terminal
kill -SIGINT {puma pid}
```

### Expected behavior
```
[72484] Use Ctrl-C to stop
[72484] - Worker 0 (pid: 72516) booted, phase: 0
[72484] - Worker 1 (pid: 72517) booted, phase: 0
[72484] - Gracefully shutting down workers...
[72484] === puma shutdown: 2018-11-08 17:19:54 +0800 ===
[72484] - Goodbye!
```

### Actual behavior

Workers are already shut down, but the cluster master hanged on `Process.waitpid`.

However, if I press ctrl+c after puma hanged , it terminated correctly.
```
[72484] Use Ctrl-C to stop
[72484] - Worker 0 (pid: 72516) booted, phase: 0
[72484] - Worker 1 (pid: 72517) booted, phase: 0
[72484] - Gracefully shutting down workers...
```

### System configuration

**Operating System**: macOS 10.14.1
**Ruby version**: 2.5.3
**Rails version**: 5.2.1
**Puma version**: 3.12.0
**Bundle version**: 1.16.6

ioquatix:
I seem to start experiencing this with 2.6.0-rc1 on travis - maybe an odd coincidence?

https://travis-ci.com/socketry/falcon/jobs/163846367 - scroll to the end.

dentarg:
@ioquatix I think you are seeing https://bugs.ruby-lang.org/issues/15499

We see it for our app using Puma cluster mode, it does not exit on SIGTERM. On Heroku, we are getting [R12 - Exit timeout](https://devcenter.heroku.com/articles/error-codes#r12-exit-timeout) errors.

MSP-Greg:
I believe this issue may have also been causing failing tests on Travis when ruby-head was 2.6.0dev, the tests are skipped on Windows/Appveyor due to lack of `fork`...

guilleiguaran:
I think this is fixed by https://github.com/puma/puma/pull/1738 also

evanphx:
This sounds like a bug in ruby 2.6.

ioquatix:
Let me see if I can reproduce the issue in 2.6.1

MSP-Greg:
I believe this issue has been around since 2018-Aug, between r64316 and r64376.

Some travis build info:

Failed builds
[...]

PR #1741, opened 2019-03-12
cluster - worker shutdown - use WNOHANG with nil return tests

MSP-Greg:
Ruby 2.6 introduced a change that affects worker shutdown.  I believe the change occurred between 'commits' r64316 and r64376 of ruby trunk.

1. Added code using `Process::WNOHANG` along with needed logic.  Adds worker status (via `$?`) and total shutdown time to log.

2. Logged status matches status returned by current code (exit 0) for Ruby 2.5.x and lower.

Co-authored-by: MSP-Greg <greg.mpls@gmail.com>
Co-authored-by: guilleiguaran <guilleiguaran@gmail.com>

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -37,7 +37,25 @@ module Puma
       @workers.each { |x| x.term }
 
       begin
-        @workers.each { |w| Process.waitpid(w.pid) }
+        if RUBY_VERSION < '2.6'
+          @workers.each { |w| Process.waitpid(w.pid) }
+        else
+          # below code is for a bug in Ruby 2.6+, above waitpid call hangs
+          t_st = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+          pids = @workers.map(&:pid)
+          loop do
+            pids.reject! do |w_pid|
+              if Process.waitpid(w_pid, Process::WNOHANG)
+                log "    worker status: #{$?}"
+                true
+              end
+            end
+            break if pids.empty?
+            sleep 0.5
+          end
+          t_end = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+          log format("    worker shutdown time: %6.2f", t_end - t_st)
+        end
       rescue Interrupt
         log "! Cancelled waiting for workers"
       end
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,8 +24,8 @@ rvm:
   - 2.2.10
   - 2.3.8
   - 2.4.5
-  - 2.5.3
-  - 2.6
+  - 2.5.5
+  - 2.6.2
   - ruby-head
 
 matrix:
@@ -35,13 +35,12 @@ matrix:
       env: RUBYOPT="--jit"
     - rvm: 2.3.8
       os: osx
-    - rvm: 2.5.3
+    - rvm: 2.5.5
       os: osx
     - rvm: jruby-9.2.6.0
     - rvm: jruby-head
 
   allow_failures:
-    - rvm: 2.6
     - rvm: ruby-head
     - rvm: ruby-head
       env: RUBYOPT="--jit"


MSP-Greg:
force-pushed change - cleaned up sleep logic

guilleiguaran:
Thanks for this!

I prefer this version over #1738 since I appreciate the extra logs but I'll keep #1738 open just in case if the Puma authors don't want the extra logging.

MSP-Greg:
@evanphx

See https://github.com/puma/puma/issues/1674#issuecomment-472416252

As to the code I added, it's replacing line 40 of:
https://github.com/puma/puma/blob/ca03c520757e4ed1caab28bf4f60ec2ffe5b2585/lib/puma/cluster.rb#L35-L44

1. Line 40 did not change the `@workers` array, so my code does not. 
2. Since `Process.waitpid(w_pid, Process::WNOHANG)` is essentially non-blocking on a running process, I wanted the code to loop thru the whole `@workers` array, as the original blocked on a running process.  Also, I added code to not add a 'sleep' if the all processes were shutdown.
3. I can remove the [logging](https://travis-ci.org/puma/puma/jobs/505283278#L612), but it might be helpful to see the 'shutdown' time?

lib/puma/cluster.rb
```diff
@@ -37,7 +37,23 @@ def stop_workers
...
+            if Process.waitpid(w_pid, Process::WNOHANG)
+              log "    worker status: #{$?}"
+              pids.delete w_pid
+              break
+            else
+              wait = true
+            end
+          end
+          sleep 0.5 if wait
+        end
```
AlexWayfer: There are two cycles instead of one.

You can use `redo`.

See https://github.com/puma/puma/pull/1738#discussion_r264748021

MSP-Greg: @AlexWayfer

Thanks.  As mentioned, I did it this way so it would loop thru all the workers on first pass, then go back for the nil returns.  I'd rather use `Process.waitpid(-1, Process::WNOHANG)`, as that's a bit cleaner, but it seems to repeatedly fail on trunk JIT, I never tried with 2.6.x JIT...

AlexWayfer: > As mentioned, I did it this way so it would loop thru all the workers on first pass, then go back for the nil returns.

It's strange algorithm.

Option 1, with `redo`:

1. Go through all workers
2. Sleep and redo on the first fail (wait every first fail)

Option 2:

1. Go through all workers and attempt to stop them
2. Delete from array in success, wait after all if any failed
3. Retry if there are any remained

Option 3, your choice, as I see:

1. Go through all workers
2. Delete in success, and retry (go further)
3. Don't do anything in fail, just sleep in the end
4. Retry until there are any

I think, the first option is simplest (but may be longest), the second is most optimal, and the third is strange.

MSP-Greg: I changed the logic to use reject!

AlexWayfer: > I changed the logic to use reject!

You can use `loop do` instead of `while true do`. 😉

MSP-Greg: I often look here:
https://msp-greg.github.io/ruby_trunk/file.control_expressions.html

`loop do` isn't mentioned there.  Time to review kernel methods again...

evanphx:
I don't see why we need this. Is `Process.wait` without WNOHANG fundementally broken on ruby now? This PR just implements a busy polling loop which is unnecessary if a blocking `Process.wait` works the way it should, which is that it waits until a child is finished.

MSP-Greg:
> Is Process.wait without WNOHANG fundamentally broken on ruby now?

As I think I've mentioned elsewhere, is it a bug or a breaking change?  I tried to create a small repo of it, but being a windows type, I can't `fork` locally.  Couldn't find a way to repo in Windows.

There are a few Ruby tests with waitpid & waitpid2, and also a few specs.  I believe these have continued to pass even when Puma starting having issues last year with trunk.

> busy polling loop

I tried to make it as 'light' as possible, with the minimum number of waitpid calls.

guilleiguaran:
@evanphx indeed, it seems to be caused by https://github.com/ruby/ruby/commit/48b6bd74e2febde095ac85d818e94c0e58677647 but doesn't look like it will be fixed anytime soon since the Ruby team itself isn't sure about how to fix it 😞

evanphx:
This is absolutely a bug in ruby, not a behavior change. Looks like they attempted to fix it here: https://github.com/ruby/ruby/commit/9e66910b3bd85de32e95cf019ed285a36aecfd9e

MSP-Greg:
I could add a conditional on `RUBY_VERSION` and/or comments about the code being needed for a bug in Ruby 2.6+.

See https://github.com/puma/puma/issues/1674#issuecomment-474142356

evanphx:
I'm pretty annoyed that ruby-core has broken a fundamental and simple unix function and hasn't fixed it.

Conditionalizing it on ruby version is fine.

MSP-Greg:
Okay with a string (`RUBY_VERSION < '2.6`), which will fail on Ruby 2.10?

MSP-Greg:
Done

PR #1748, opened 2019-03-19
Cluster#check_workers - waitpid logic

MSP-Greg:
Simplify deleting workers responding to waitpid.  One loop instead of several.

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -218,14 +218,9 @@ module Puma
       # during this loop by giving the kernel time to kill them.
       sleep 1 if any
 
-      while @workers.any?
-        pid = Process.waitpid(-1, Process::WNOHANG)
-        break unless pid
+      @workers.reject! { |w| Process.waitpid(w.pid, Process::WNOHANG) }
 
-        @workers.delete_if { |w| w.pid == pid }
-      end
-
-      @workers.delete_if(&:dead?)
+      @workers.reject!(&:dead?)
 
       cull_workers
       spawn_workers


nateberkopec:
Hm, any comments on how this removes `break`?

MSP-Greg:
@nateberkopec

Sorry, confused.  You mean this or ##1741?

lib/puma/cluster.rb
```diff
@@ -218,14 +218,9 @@ def check_workers(force=false)
       # during this loop by giving the kernel time to kill them.
       sleep 1 if any
 
-      while @workers.any?
-        pid = Process.waitpid(-1, Process::WNOHANG)
-        break unless pid
```
nateberkopec: @MSP-Greg talking about this break here. also the waitpid is slightly different. wondering what effect the differences will have.

MSP-Greg: 1. `Process.waitpid(-1, Process::WNOHANG)` will check **all** child processes, not just workers.  Not sure if that is an issue or not.

2. With the '-1' parameter, I believe the existing code will return pid's for all child processes that are shut down, and then stop from the break.  I believe that the following will do the same, just limited to workers (and a bit cleaned up):
```ruby
@workers.reject! { |w| Process.waitpid(w.pid, Process::WNOHANG) }
```

Locally, I'm just windows, so the only way I can check code using `fork` is thru Travis.  That also leaves me somewhat unfamiliar with it, so I'll consider test results, which are green.  I'm also aware that workers may be shutdown due to a misbehaving application, and that can be difficult to test.

Hence, I'll defer to your judgement...

nateberkopec: > Process.waitpid(-1, Process::WNOHANG) will check all child processes, not just workers. Not sure if that is an issue or not.

I think this will be fine unless anyone is doing really weird stuff with `fork`.

nateberkopec:
This is a kinda subtle change so I'm gonna have to defer to @evanphx

evanphx:
Looks good, I prefer to not use -1 and use an explicit pid when we can.

PR #1750, opened 2019-03-19
Autoload cluster.rb & single.rb

MSP-Greg:
See issue 1731.


evanphx:
I'm going to fix this by just moving the const in question.

PR #1763, opened 2019-04-07
[RFC] Add a --plugin option. #1761

### LINKED ISSUE #1761: Impossible to load plugins with prune_bundler and workers > 0
### Steps to reproduce

1) Create a gem with a plugin inside `puma/plugin/test.rb`

2) Create a `config/puma.rb` with

```
plugin 'test'

workers 2

prune_bundler
```

3) start with `./bin/puma`

```
/home/ylecuyer/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/puma-3.12.1/lib/puma/plugin.rb:50:in `rescue in find': Unable to find plugin: test (Puma::UnknownPlugin)
```

### Expected behavior

It should boot the worker with the plugin loaded

### Actual behavior

It fails to find the plugin when starting the workers because bundler as been pruned and when requiring the plugin it is not in `$LOAD_PATH` for the plugin to work in this condition it has to be in the core puma gem. kind of useless

### System configuration

**Ruby version**: ruby 2.5.3p105
**Rails version**: Rails 4.2.11.1
**Puma version**: puma 3.12.1

ylecuyer:
This is a fix that requires less changes to the current base code to fix #1761 

It adds a --plugin option that takes the name of the gems loaded in the configuration with the plugin dsl.
When puma-wild is launched, the require paths of the plugin are added to the dependency to load into $LOAD_PATH when puma-wild starts up.

The code is not beautiful, I'll do a cleanup if the overall idea is approved.

### DIFF
--- a/lib/puma/cli.rb
+++ b/lib/puma/cli.rb
@@ -161,6 +161,10 @@ module Puma
             user_config.prune_bundler
           end
 
+          o.on "--plugins PLUGINS", "Add plugins to the $LOAD_PATH" do |arg|
+            user_config.plugins arg.split(':')
+          end
+
           o.on "-q", "--quiet", "Do not log requests internally (default true)" do
             user_config.quiet
           end
--- a/lib/puma/configuration.rb
+++ b/lib/puma/configuration.rb
@@ -133,6 +133,7 @@ module Puma
 
       @options     = UserFileDefaultOptions.new(user_options, default_options)
       @plugins     = PluginLoader.new
+      @plugins_to_load = []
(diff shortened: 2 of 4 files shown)


nateberkopec:
> The code is not beautiful, I'll do a cleanup if the overall idea is approved.

Great work! I think it just definitely needs a test for your original issue report.

nateberkopec:
Same here, needs test for inclusion into 4.1

nateberkopec:
Closing as stale, please reopen if you can add a test.

PR #1767, opened 2019-04-10
1756 pid monitoring

jessebs:
Pull request to add status reporting while using pids.

I had some tests that I ended up reverting.  TestPumaControlCli#test_control_url interfered with any other threaded tests and the fact that signals are used to kill the puma process means the tests were self destructing.

Testing was doable using forked processes but they aren't supported by windows.

### DIFF
--- a/lib/puma/control_cli.rb
+++ b/lib/puma/control_cli.rb
@@ -121,9 +121,11 @@ module Puma
         @control_url = sf.control_url
         @control_auth_token = sf.control_auth_token
         @pid = sf.pid
-      elsif @pidfile
+      elsif @pidfile && File.readable?(@pidfile)
         # get pid from pid_file
         @pid = File.open(@pidfile).gets.to_i
+      else
+        @pid = nil
       end
     end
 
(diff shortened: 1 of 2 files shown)


lib/puma/control_cli.rb
```diff
@@ -121,7 +121,7 @@ def prepare_configuration
         @control_url = sf.control_url
         @control_auth_token = sf.control_auth_token
         @pid = sf.pid
-      elsif @pidfile
+      elsif @pidfile && File.readable?(@pidfile)
```
nateberkopec: Why was this added?

nateberkopec:
>TestPumaControlCli#test_control_url interfered with any other threaded tests and the fact that signals are used to kill the puma process means the tests were self destructing.

Hm :(

nateberkopec:
@jessebs 👋

lib/puma/control_cli.rb
spk: o/ this was added to avoid `No such file or directory @ rb_sysopen - tmp/pids/server.pid` when defined in `config/puma.rb` and have instead `Neither pid nor control url available` from https://github.com/puma/puma/blob/v3.12.1/lib/puma/control_cli.rb#L183

nateberkopec: I think this could be a separate PR-plus-tests, then.

spk: > I think this could be a separate PR-plus-tests, then.

First PR done in https://github.com/puma/puma/pull/1824

nateberkopec:
Closing for #1824

Issue #1769, opened 2019-04-11
ability to turn off backtraces in pumactl


jessebs:
### Steps to reproduce

1) run `pumactl badcommand`

### Expected behavior

An error message that the command is invalid

### Actual behavior

An error message that the command is invalid, as well as a stacktrace that is useless to end users

### System configuration

**Ruby version**: 2.5.1
**Puma version**: 3.12.1

nateberkopec:
Great PR as well. I wonder if we should show the stacktrace at all though? Maybe we should just remove the stacktrace entirely?

jessebs:
I fully agree with removing the trace.

nateberkopec:
It looks like it was added by @schneems in #1239 as a sort of temporary debugging measure, but it was committed. I think it's really only useful for us maintainers, and we can add it back in locally if we need it.

Please close your old PR and open a new one removing the line and I will merge.

PR #1770, opened 2019-04-11
Ability to silence pumactl backtraces

jessebs:
Addresses #1769

Adds a new flag (-N, --no-backtrace) to pumactl to prevent displaying of backtraces.

I would think this should be default behavior (and add a --verbose flag), but I didn't want to break backward compatibility.


nateberkopec:
Great code but unsure about approach - see comment on issue.

nateberkopec:
Closing as abandoned - please reopen when you are able respond.

Issue #1499, opened 2018-01-03
systemd socket activation Errno::EBADF: Bad file descriptor - not a socket file descriptor


olegantonyan:
### Steps to reproduce

1) Follow the instructions here https://github.com/puma/puma/blob/master/docs/systemd.md and create `puma.socket` and `puma.service` files for systemd socket activation
2) After starting the service check its status `systemctl status puma.socket puma.service`

### Expected behavior

Successful start

### Actual behavior

```
● puma.socket - puma accept sockets
   Loaded: loaded (/etc/systemd/system/puma.socket; enabled; vendor preset: enabled)
   Active: failed (Result: service-start-limit-hit) since Wed 2018-01-03 08:59:03 UTC; 3s ago
   Listen: 127.0.0.1:3000 (Stream)

Jan 03 08:58:59  systemd[1]: Listening on puma accept sockets.
Jan 03 08:59:03  systemd[1]: puma.socket: Unit entered failed sta

● puma.service - puma server
   Loaded: loaded (/etc/systemd/system/puma.service; enabled; vendor preset: enabled)
   Active: failed (Result: exit-code) since Wed 2018-01-03 08:59:03 UTC; 3s ago
  Process: 11775 ExecStart=/bin/bash -lci bundle exec puma --config ./config/puma.rb -b tcp:/
 Main PID: 11775 (code=exited, status=1/FAILURE)
      CPU: 581ms
```

Puma's log:
```
Errno::EBADF: Bad file descriptor - not a socket file descriptor
shared/bundle/ruby/2.4.0/gems/puma-3.11.0/lib/puma/binder.rb:65:in `for_fd'
shared/bundle/ruby/2.4.0/gems/puma-3.11.0/lib/puma/binder.rb:65:in `block (2 levels) in import_from_env'
shared/bundle/ruby/2.4.0/gems/puma-3.11.0/lib/puma/binder.rb:63:in `times'
shared/bundle/ruby/2.4.0/gems/puma-3.11.0/lib/puma/binder.rb:63:in `block in import_from_env'
shared/bundle/ruby/2.4.0/gems/puma-3.11.0/lib/puma/binder.rb:57:in `each'
shared/bundle/ruby/2.4.0/gems/puma-3.11.0/lib/puma/binder.rb:57:in `import_from_env'
shared/bundle/ruby/2.4.0/gems/puma-3.11.0/lib/puma/launcher.rb:52:in `initialize'
[...]

PR #1777, opened 2019-04-18
Add a note about bundle and socket activation

tagliala:
When running Puma with bundle and socket activation, there could be issues
related to shell interactivity.

Ref: https://github.com/puma/puma/issues/1499#issuecomment-481772886


docs/systemd.md
```diff
@@ -74,6 +74,10 @@ clustered mode and application preload.
 **Note:** Socket activation doesn't currently work on jruby. This is
 tracked in [#1367].
 
+**Note:** Do not run Puma with `bundle` when using with socket
+activation. This may cause issues with shell interactivity (Ref:
+[1499#issuecomment-481772886](https://github.com/puma/puma/issues/1499#issuecomment-481772886)
+
```
dekellum: So I wonder if the mentioned incompatibility with ruby 2.6 is really about the version of bundler it includes? 

As mentioned in the original issue https://github.com/puma/puma/issues/1499#issuecomment-358867492, does `bundle exec` with the `--keep-file-descriptors` flag also avoid the problem you were seeing? If so then you might change this to say something like: 

_Any wrapper scripts which `exec`, or other indirections in `ExecStart`, may result in activated socket file descriptors being closed before they reach the puma master process.  For example, if using `bundle exec`, pass the `--keep-file-descriptors` flag._

dekellum:
Cc: @itolosa  Suggestions for better describing the problem?

docs/systemd.md
tagliala: Hi @dekellum

> So I wonder if the mentioned incompatibility with ruby 2.6 is really about the version of bundler it includes?

I don't know that. I'm using bundler 1.16.6 with ruby 2.5 and 1.17.2 with ruby 2.6

> As mentioned in the original issue #1499 (comment), does bundle exec with the --keep-file-descriptors flag also avoid the problem you were seeing? 

Yes, it does. Didn't try that before, I didn't even read that comment 🤦‍♂️ 

> Any wrapper scripts which exec, or other indirections in ExecStart, may result in activated socket file descriptors being closed before they reach the puma master process. For example, if using bundle exec, pass the --keep-file-descriptors flag.

Since keep-file-descriptors work, I will force-push in a moment

itolosa:
> Cc: @itolosa Suggestions for better describing the problem?

I think that would be a good idea to also include the suggestion of using binstubs when no --keep-file-descriptor flag is available

nateberkopec:
@itolosa I'm unsure - do you need more time to add things or is this ready to merge?

itolosa:
RTM! the reference should be enough

tagliala:
@dekellum I would like to change the reference from the comment to the whole thread

I mean, from

https://github.com/puma/puma/pull/1777/files#diff-38e9e6fc9e4f94c189480d539dbbcda9R78

```
[1499#issuecomment-481772886](https://github.com/puma/puma/issues/1499#issuecomment-481772886)
```

to `[#1499]`

https://github.com/puma/puma/pull/1777/files#diff-38e9e6fc9e4f94c189480d539dbbcda9R78

Also, the suggestion to provide a binstub could be useful for some use cases.

dekellum:
The sample config mentions a binstub alternative, but sure, adding a sentence could make it more clear, e.g. `bundle exec` can be avoided by using a `puma` executable generated by `bundle binstubs puma`.

tagliala:
@dekellum I've just force-pushed again. Please take a look at the new commit message and sentence

dekellum:
LGTM, thanks.

PR #1784, opened 2019-04-24
small fixes - binder.rb client.rb test_integration.rb

MSP-Greg:
Fixes errors/warning shown in CI.

`binder.rb` fix removes error in CI when using a tempfile (which also rescues `Errno::ENOENT`), but not sure about rescue in general use...

### DIFF
--- a/lib/puma/binder.rb
+++ b/lib/puma/binder.rb
@@ -50,7 +50,13 @@ module Puma
 
     def close
       @ios.each { |i| i.close }
-      @unix_paths.each { |i| File.unlink i }
+      @unix_paths.each do |i|
+        # Errno::ENOENT is intermittently raised
+        begin
+          File.unlink i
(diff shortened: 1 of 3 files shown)


lib/puma/binder.rb
```diff
@@ -50,7 +50,8 @@ def env(sock)
 
     def close
       @ios.each { |i| i.close }
-      @unix_paths.each { |i| File.unlink i }
+      # Errno::ENOENT is intermittently raised
+      @unix_paths.each { |i| File.unlink(i) rescue Errno::ENOENT }
```
nateberkopec: please turn this into a multi-liner. doing this one line is equivalent to `rescue nil`.

MSP-Greg: That was stupid of me.  Apologies.
Thanks for pointing it out...

PR #1120, opened 2016-10-23
Fixed the issue of @listeners getting nil io

prathamesh-sonpatki:
- `add_tcp_listener` and `add_ssl_listener` return `nil` now for
  localhost addresses and it was getting added to the `@listeners`.
- Later on the file descriptor `nil` was getting converted to 0 and
  resulting into `BadFileDescriptor` error as per this gist -
  https://gist.github.com/prathamesh-sonpatki/bc0cac8929dbf17316e8b4b7dda93e20.
- This issue is now fixed and added test case.
- This fixes `rails restart` command on Rails 5 + Puma 3.5 and above.

PR #1786, opened 2019-04-25
Report tcp 0 port properly. Fixes #1679

### LINKED ISSUE #1679: Print corret port when -p 0
# Summary
For some reason, I struggle to run web applications in default ports 4000, 3000 (the one used by puma), 4200, etc.
So I passed the switch -p 0 to rails to start in a random server,
`rails s -p 0`
It is ok, it finds a free port, but in the output it says:

>  Listening on tcp://localhost:0
Can you fix that so it outputs not the input but the actual port where the server is listening?

**Ruby version**: ruby 2.3.3-p222
**Rails version**: 5.2.1
**Puma version**: 3.12.0 (through puma --version)

evanphx:


nateberkopec:
LGTM

test/test_binder.rb
```diff
@@ -54,7 +56,8 @@ def setup
   def test_localhost_addresses_dont_alter_listeners_for_ssl_addresses
     @binder.parse(["ssl://localhost:10002?key=#{@key}&cert=#{@cert}"], @events)
 
-    assert_equal [], @binder.listeners
+    assert_equal "ssl://localhost:10002?key=#{@key}&cert=#{@cert}",
```
nateberkopec: I don't understand why this PR adds a listener, and why that's "ok".

nateberkopec:
Oh interesting. When I don't add these to the listeners array, `test_sigterm_closes_listeners_on_forked_servers` breaks. So that's weird.

nateberkopec:
#1120 feels relevant.

PR #1788, opened 2019-04-28
Add constants for availability of SSLv3 & TLSv1, adjust tests, add no_tlsv1 test

MSP-Greg:

PR #1799, opened 2019-05-19
Add server test for no_tlsv1, updates to mini_ssl.c

MSP-Greg:
Commit 1 - `test/test_puma_server_ssl.rb`:

1. remove `setup`, add `start_server` to allow ctx to be changed in tests
2. add `test_tls_v1_rejection`

Commit 2 - `mini_ssl.c`:

1. use TLS_method in SSL_CTX_new
2. start using SSL_CTX_set_min_proto_version for protocol selection

### DIFF
--- a/ext/puma_http11/mini_ssl.c
+++ b/ext/puma_http11/mini_ssl.c
@@ -142,7 +142,7 @@ VALUE engine_init_server(VALUE self, VALUE mini_ssl_ctx) {
   VALUE obj;
   SSL_CTX* ctx;
   SSL* ssl;
-  int ssl_options;
+  int min, ssl_options;
 
   ms_conn* conn = engine_alloc(self, &obj);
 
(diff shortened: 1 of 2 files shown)


MSP-Greg:
@nateberkopec

Ok, maybe I bounce around a little too much.  I just noticed that the test added here I also added in PR #1788.  Also, I'm not sure if the mini_ssl.c changes will merge with it...

I can redo this with only the mini_ssl.c changes.

nateberkopec:
@MSP-Greg I've marked 3 of your PRs for inclusion in 4.0.1: https://github.com/puma/puma/milestone/4 Do you foresee any issues/changes that need to be made?

MSP-Greg:
@nateberkopec

> Do you foresee any issues/changes that need to be made?

Yes.  Some of the code in this & #1788 was using version checks, etc for 'compile' function checks.  With all the different builds of OpenSSL, I think using `have_func` / `have_macro` in `extconf.rb` is a lot better solution (and also visible in build logs), as below from #1832:

https://github.com/puma/puma/blob/224a387fad36996987f616059d8a22c6a8d4a710/ext/puma_http11/extconf.rb#L13-L19

Hence, if possible, could #1832 be merged, then I'll combine this and #1788 in a new PR?

I've been thinking about how to move forward with setting minimum SSL/TLS protocol versions.  We've already got [`Puma::MiniSSL::Context#no_tlsv1=`](https://msp-greg.github.io/puma/Puma/MiniSSL/Context.html#no_tlsv1=-instance_method), but, in the future, cotinuing to add things like `#no_tlsv1_1=` gets messy.

It might make sense to add something identical to [`OpenSSL::SSL::SSLContext#min_version=`](https://msp-greg.github.io/ruby_trunk/openssl/OpenSSL/SSL/SSLContext.html#min_version=-instance_method) to `Puma::MiniSSL::Context`?  But then, the issue is what happends with `#no_tlsv1`?  Does `#min_version=` override it?

Regardless, adding `#min_version=` might need to wait until after 4.0.1...

nateberkopec:
> It might make sense to add something identical to OpenSSL::SSL::SSLContext#min_version= to Puma::MiniSSL::Context? But then, the issue is what happends with #no_tlsv1? Does #min_version= override it?

I think we can punt this to 4.1.

MSP-Greg:
Closing in favor of #1836.

PR #1803, opened 2019-05-21
Fixes the bug that wouldn't allow no_token be set to true

### LINKED ISSUE #1193: Setting no_token option causes crash

Jesus:
This recovers PR #1321 by @manishval. Closes #1193 

It's the same code as the original PR but it also includes a test to cover the reported bug.

### DIFF
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -105,7 +105,7 @@ module Puma
       end
 
       if opts[:no_token]
-        auth_token = :none
+        auth_token = ''
       else
         auth_token = opts[:auth_token]
         auth_token ||= Configuration.random_token
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -105,7 +105,7 @@ module Puma
       end
 
       if opts[:no_token]
-        auth_token = ''
+        auth_token = 'none'
       else
         auth_token = opts[:auth_token]
         auth_token ||= Configuration.random_token
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -105,6 +105,11 @@ module Puma
       end
 
       if opts[:no_token]
+        # We need to use 'none' rather than :none because this value will be
+        # passed on to an instance of OptionParser, which doesn't support
+        # symbols as option values.
+        #
+        # See: https://github.com/puma/puma/issues/1193#issuecomment-305995488
         auth_token = 'none'
       else
         auth_token = opts[:auth_token]
--- a/test/test_pumactl.rb
+++ b/test/test_pumactl.rb
@@ -29,6 +29,23 @@ class TestPumaControlCli < Minitest::Test
     assert_equal "t3-pid", control_cli.instance_variable_get("@pidfile")
   end
 
+  def test_control_no_token
+    opts = [
+      "--config-file", "test/config/control_no_token.rb",
+      "start"
+    ]
+
+    control_cli = Puma::ControlCLI.new opts, @ready, @ready
+    t = Thread.new do
+      Thread.current.abort_on_exception = true
+      control_cli.run
+    end
+
+    assert_equal 'none', control_cli.instance_variable_get("@control_auth_token")
+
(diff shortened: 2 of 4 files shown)


lib/puma/dsl.rb
```diff
@@ -105,7 +105,7 @@ def activate_control_app(url="auto", opts={})
       end
 
       if opts[:no_token]
-        auth_token = :none
+        auth_token = 'none'
```
nateberkopec: Please add a comment here about why it's important we use a string rather than a symbol. I could see someone trying to "optimize" this later :D

nateberkopec:
Thank you @Jesus! What a resurrection!

Needs one thing added and I think it's good to go.

lib/puma/dsl.rb
dentarg: Shouldn't there be tests that fail if changed?

Jesus: @dentarg the test already checks if the value is `'none'` https://github.com/puma/puma/pull/1803/files#diff-5199cb8753ebe54bb452866fea92a0d7R44

If you think there're more cases we should cover I'm willing to know.

nateberkopec: I see what you mean 'dentarg' - yes I know the tests will fail, but still I'd like a record of *why* this is important for the future.

test/test_pumactl.rb
```diff
@@ -29,6 +29,23 @@ def test_config_file
...
   end
 
+  def test_control_no_token
+    opts = [
+      "--config-file", "test/config/control_no_token.rb",
+      "start"
+    ]
+
+    control_cli = Puma::ControlCLI.new opts, @ready, @ready
+    t = Thread.new do
```
MSP-Greg: The assert passes with the thread being created.

So, should it be removed, or an assert/refute added for it?

Jesus: Sorry I don't understand what you mean. Would you mind elaborating?

Also I think there's a better way to write this test which shouldn't require a thread: I think checking that the value is a String rather than checking that the server boots successfully should be enough.

Jesus:
Thanks Nate for your effort viewing my PRs, I appreciate it a lot!

Please let me sort out what @MSP-Greg is pointing out and this should be ok to go.

test/test_pumactl.rb
nateberkopec: not sure what @MSP-Greg means either but this test is also nondeterministic - control_cli.run may not finish before we run the assertion

nateberkopec: I think you could just fix that with a .join though

MSP-Greg: I think the assert statement could be moved above the `Thread.new` statement, or, the instance variable is set with `Puma::ControlCLI.new`, not `#run`.  Sorry, gone for most of the morning...

Jesus: I think it's been a mistake trying to test two things in a single test: It was explicitly testing that the value was being set to 'none' and also implicitly testing that the server could boot with `no_token: true`.

I've removed the part about booting the server.

Jesus:
Sorry for the back and forth, I'll use `[WIP]` next time I commit incomplete work on a PR.

This should be ok to go unless you want to add anything further.

nateberkopec:
> Sorry for the back and forth, I'll use [WIP] next time I commit incomplete work on a PR.

Not a problem! I'll merge when it's green.

PR #1806, opened 2019-05-24
Updates documentation for Server#handle_request

Jesus:
It was referencing some non-existing params and some behavior that I
don't think is present anymore in this method.


nateberkopec:
👏

schneems:
Thanks for the PR. If you're interested in more docs stuff, the Reactor class changed from `select` to using nio4r (uses epoll/kqueue).

PR #1808, opened 2019-05-30
[close #1802] Close listeners on SIGTERM

### LINKED ISSUE #1802: H13 (connection closed with no response) from Puma on Heroku
For full description see reproduction repository https://github.com/schneems/puma_connection_closed_reproduction.

> Note: The hanging issue noted in there is due to a Ruby 2.6 bug and has been fixed in master.

## Possible cause and fix

It looks like durring SIGTERM the socket connection is never explicitly closed. The only way it is closed is when the process calls `exit(0)`. Instead, we can manually close the socket which will prevent additional incoming requests but will still allow us to respond to requests that have already been accepted.

schneems:
Currently when a SIGTERM is sent to a puma cluster, the signal is trapped, then sent to all children, it then waits for children to exit and then the parent process exits. The socket that accepts connections is only closed when the parent process calls `exit 0`. The problem with this flow is there is a period of time where there are no child processes to work on an incoming connection, however the socket is still open so clients can connect to it. When this happens, the client will connect, but the connection will be closed with no response. Instead, the desired behavior is for the connection from the client to be rejected. This allows the client to re-connect, or if there is a load balance between the client and the puma server, it allows the request to be routed to another node.

This PR fixes the existing behavior by manually closing the socket when SIGTERM is received before shutting down the workers/children processes. When the socket is closed, any incoming requests will fail to connect and they will be rejected, this is our desired behavior. Existing requests that are in-flight can still respond.

 ## Test

This behavior is quite difficult to test, you'll notice that the test is far longer than the code change. In this test we send an initial request to an endpoint that sleeps for 1 second. We then signal to other threads that they can continue. We send the parent process a SIGTERM, while simultaneously sending other requests. Some of these will happen after the SIGTERM is received by the server. When that happens we want none of the requests to get a `ECONNRESET` error, this would indicate the request was accepted but then closed. Instead we want `ECONNREFUSED`.

I ran this test in a loop for a few hours and it passes with my patch, it fails immediately if you remove the call to close the listeners.

```
$ while m test/test_integration.rb:235; do :; done
```

 ## Considerations

This PR only fixes the problem for "cluster" (i.e. multi-worker) mode. When trying to reproduce the test with single mode, on (removing the `-w 2` config) it already passes. This leads us to believe that either the bug does not exist in single threaded mode, or at the very least reproducing the bug via a test in the single threaded mode requires a different approach.

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -401,6 +401,8 @@ module Puma
           log "Early termination of worker"
           exit! 0
         else
+          @launcher.send(:close_binder_listeners)
+
           stop_workers
           stop
 
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -36,16 +36,26 @@ module Puma
 
       begin
         if RUBY_VERSION < '2.6'
-          @workers.each { |w| Process.waitpid(w.pid) }
+          @workers.each do |w|
+            begin
+              Process.waitpid(w.pid)
+            rescue Errno::ECHILD
+              # child is already terminated
+            end
+          end
         else
           # below code is for a bug in Ruby 2.6+, above waitpid call hangs
           t_st = Process.clock_gettime(Process::CLOCK_MONOTONIC)
           pids = @workers.map(&:pid)
           loop do
             pids.reject! do |w_pid|
-              if Process.waitpid(w_pid, Process::WNOHANG)
-                log "    worker status: #{$?}"
-                true
+              begin
+                if Process.waitpid(w_pid, Process::WNOHANG)
+                  log "    worker status: #{$?}"
+                  true
+                end
+              rescue Errno::ECHILD
+                true # child is already terminated
               end
             end
             break if pids.empty?
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -411,7 +411,7 @@ module Puma
           log "Early termination of worker"
           exit! 0
         else
-          @launcher.send(:close_binder_listeners)
+          @launcher.close_binder_listeners
 
           stop_workers
           stop
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -214,6 +214,15 @@ module Puma
       end
     end
 
+    def close_binder_listeners
+      @binder.listeners.each do |l, io|
+        io.close
+        uri = URI.parse(l)
+        next unless uri.scheme == 'unix'
+        File.unlink("#{uri.host}#{uri.path}")
+      end
+    end
+
     private
 
     def reload_worker_directory
@@ -318,16 +327,6 @@ module Puma
     def prune_bundler?
       @options[:prune_bundler] && clustered? && !@options[:preload_app]
     end
-
-    def close_binder_listeners
-      @binder.listeners.each do |l, io|
-        io.close
-        uri = URI.parse(l)
-        next unless uri.scheme == 'unix'
-        File.unlink("#{uri.host}#{uri.path}")
-      end
-    end
-
 
     def generate_restart_data
       if dir = @options[:directory]
--- a/test/rackup/1second.ru
+++ b/test/rackup/1second.ru
@@ -1,0 +1,4 @@
+run lambda { |env|
+  sleep 1
+  [200, {}, ["Hello World"]]
+}
--- a/test/test_integration.rb
+++ b/test/test_integration.rb
@@ -232,6 +232,58 @@ class TestIntegration < Minitest::Test
     assert_equal "Hello World", new_reply
   end
 
+  def test_sigterm_closes_listeners_on_forked_servers
+    skip NO_FORK_MSG unless HAS_FORK
+    pid = start_forked_server("-w 2 -q test/rackup/1second.ru")
+    threads = []
+    initial_reply = nil
+    next_replies = []
+    condition_variable = ConditionVariable.new
+    mutex = Mutex.new
+
+    threads << Thread.new do
+      s = connect
+      mutex.synchronize { condition_variable.broadcast }
+      initial_reply = read_body(s)
+    end
+
+    threads << Thread.new do
+      mutex.synchronize {
+        condition_variable.wait(mutex, 1)
+        Process.kill("SIGTERM", pid)
+      }
+    end
+
+    10.times.each do |i|
+      threads << Thread.new do
+        mutex.synchronize { condition_variable.wait(mutex, 1.5) }
+
+        begin
+          s = connect
+          read_body(s)
+          next_replies << :success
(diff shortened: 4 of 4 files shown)


nateberkopec:
The maintainers think this looks good, but could use feedback. If you're experiencing issues with closed connections during shutdowns please give this a shot ❤️

electron0zero:
We had this issue, our workaround was to send SIGTERM to nginx(we run puma behind nginx) and then sleep for 15 seconds and send SIGTERM to puma.

adamlogic:
I see H13 errors all the time when downscaling with Rails Autoscale. I've deployed this patch, and so far so good.

Screen Shot 2019-06-04 at 1 54 53 PM

It's only been a couple hours, I'll post another update tomorrow.

schneems:
Awesome, thanks for the info @adamlogic!

adamlogic:
After giving it about 24 hours, I'm definitely seeing _different_ behavior, although I can't quite make sense of it. I typically get 1-2 H13 errors during a downscale or restart, spread throughout the day. Yesterday I only saw one instance of H13 errors, but it was a burst of 37 errors. Quite unusual.

Screen Shot 2019-06-05 at 11 58 30 AM

I'm hoping that was an anomaly. Going to keep this patch running for now.

schneems:
There was also an incident at that time, so it might be related? Thanks for running this, keep me updated.

schneems:
How did we do on the next 24 hours? 

One theory on where this might be going south is that the default behavior for Puma is not to drain the socket backlog. There's a setting for it though. Just a theory, haven't tested.

adamlogic:
No H13's in the last 36+ hours. 👍 

Screen Shot 2019-06-06 at 12 24 49 PM

lib/puma/cluster.rb
```diff
@@ -401,6 +401,8 @@ def setup_signals
           log "Early termination of worker"
           exit! 0
         else
+          @launcher.send(:close_binder_listeners)
```
evanphx: Should we just expose this as a normal method?

adamlogic:
Spoke too soon. Got 4 H13 errors last night.

Screen Shot 2019-06-07 at 8 56 32 AM

Dug into my logs and found this error. This is the only instance of this error in my logs, even though I've had many downscale events.

```
2019-06-06T21:00:59.137527+00:00 heroku web.2 - - State changed from up to down
2019-06-06T21:01:00.403377+00:00 heroku web.2 - - Stopping all processes with SIGTERM
2019-06-06T21:01:00.428773+00:00 heroku web.2 - - Stopping all processes with SIGTERM
2019-06-06T21:01:01.404286+00:00 app web.2 - - [4] - Gracefully shutting down workers...
2019-06-06T21:01:01.408456+00:00 app web.2 - - bundler: failed to load command: puma (/app/vendor/bundle/ruby/2.5.0/bin/puma)
2019-06-06T21:01:01.408661+00:00 app web.2 - - Errno::ECHILD: No child processes
2019-06-06T21:01:01.408666+00:00 app web.2 - -   /app/vendor/bundle/ruby/2.5.0/bundler/gems/puma-184e1510a97c/lib/puma/cluster.rb:39:in `waitpid'
2019-06-06T21:01:01.408669+00:00 app web.2 - -   /app/vendor/bundle/ruby/2.5.0/bundler/gems/puma-184e1510a97c/lib/puma/cluster.rb:39:in `block in stop_workers'
2019-06-06T21:01:01.408672+00:00 app web.2 - -   /app/vendor/bundle/ruby/2.5.0/bundler/gems/puma-184e1510a97c/lib/puma/cluster.rb:39:in `each'
2019-06-06T21:01:01.408674+00:00 app web.2 - -   /app/vendor/bundle/ruby/2.5.0/bundler/gems/puma-184e1510a97c/lib/puma/cluster.rb:39:in `stop_workers'
2019-06-06T21:01:01.408676+00:00 app web.2 - -   /app/vendor/bundle/ruby/2.5.0/bundler/gems/puma-184e1510a97c/lib/puma/cluster.rb:406:in `block in setup_signals'
2019-06-06T21:01:01.408678+00:00 app web.2 - -   /app/vendor/bundle/ruby/2.5.0/bundler/gems/puma-184e1510a97c/lib/puma/cluster.rb:531:in `rescue in run'
2019-06-06T21:01:01.408680+00:00 app web.2 - -   /app/vendor/bundle/ruby/2.5.0/bundler/gems/puma-184e1510a97c/lib/puma/cluster.rb:494:in `run'
2019-06-06T21:01:01.408682+00:00 app web.2 - -   /app/vendor/bundle/ruby/2.5.0/bundler/gems/puma-184e1510a97c/lib/puma/launcher.rb:186:in `run'
2019-06-06T21:01:01.408684+00:00 app web.2 - -   /app/vendor/bundle/ruby/2.5.0/bundler/gems/puma-184e1510a97c/lib/puma/cli.rb:80:in `run'
2019-06-06T21:01:01.408686+00:00 app web.2 - -   /app/vendor/bundle/ruby/2.5.0/bundler/gems/puma-184e1510a97c/bin/puma:10:in `<top (required)>'
2019-06-06T21:01:01.408689+00:00 app web.2 - -   /app/vendor/bundle/ruby/2.5.0/bin/puma:23:in `load'
2019-06-06T21:01:01.408691+00:00 app web.2 - -   /app/vendor/bundle/ruby/2.5.0/bin/puma:23:in `<top (required)>'
2019-06-06T21:01:01.975425+00:00 heroku web.2 - - Process exited with status 1
```

LMK if there's any more info that would be helpful.

schneems:
I’ve seen that error before, it happens because the child got closed before the parent tried to wait on it. When there is no child we should rescue that exception and keep going. 

Heroku sends TERM to all processes not just the one it spawned. So sometimes the children close before the parent can send a TERM to them.

I’ll update later today with a fix for that and let you know.

schneems:
@adamlogic I just added https://github.com/puma/puma/pull/1808/commits/8c78ee201e0ab934745d8b9f63f96d0dad75bb5b which should resolve that error. I don't know if that's the cause of the H13 or just covering up some other behavior.

If fixing that error doesn't resolve the H13 then I would suggest setting `drain_on_shutdown` in your `config/puma.rb` to see if it helps (but please don't set this until after we verify that my above patch works, I would like to rule out one thing at a time and that's easier to do if we only change one thing).

adamlogic:
👍 Thanks! Just deployed with 84ce04d.

I'll post another update on Monday.

adamlogic:
No H13 errors or `Errno::ECHILD` errors in the past 72 hours. 🎉 

Screen Shot 2019-06-10 at 6 45 46 AM

nateberkopec:
That's the cleanest looking Heroku metrics timeline I've ever seen

PR #1809, opened 2019-06-04
[WIP] Native websocket support

Jesus:
This is Evan's original work from 3 years ago. I've rebased it and made the minimum updates required to make it work for a basic chat application.

Still a long way to go for supporting this spec https://github.com/boazsegev/iodine/blob/master/SPEC-WebSocket-Draft.md & making this stable enough to merge.


evanphx:
I'm happy for us to collectively move in this direction. Is it the direction we want to take? Or do we want to give enough hooks to provide the websocket functionality itself as a gem?

Jesus:
Thanks Evan for taking the time to write here.

I was thinking the web server would just provide what the spec indicates, hoping that the spec is good enough for the upper layers to do their job. I'd take care of proving this correct at a later stage.

You're definitely the best one to decide way we want to approach this. Where would you set the boundaries of Puma's responsibility?

evanphx:
The spec spoken to is about supporting Websockets within the webserver. It's fine to do that if we want to. I'm just throwing another option out there, that we expose the lowest bits so they can be used by other tooling if need be.

Jesus:
What use case do you have in mind?

evanphx:
Anything that would take over the data stream for a request and want to wait on IO, mostly.

Jesus:
Again, thanks for your feedback. I've had this PR on hold not because I'm not willing to change the approach, but because I want to explore your suggestion but I'm not sure I understood correctly.

Let me see if I'm getting right what you're saying:

Most of the code of this PR would be moved to its own gem, let's call it puma-websockets, then Puma would provide hooks (providing callbacks and exposing methods) so puma-websockets can do its own thing.

Then the discussion would be what are those hooks that Puma needs to provide, and I'm guessing basically Puma would notify when there's data to read in a Websocket and it also would allow puma-websockets to invoke the Rack app when the websocket needs.

Is that the sort of thing you're describing?

PR #1607, opened 2018-07-05
Fix chunked ending check

harmdewit:
As stated in #1480 according [to the spec](https://tools.ietf.org/html/rfc7230#section-3.3) a valid chunked body is formatted like this:

```
4.1.  Chunked Transfer Coding

   The chunked transfer coding wraps the payload body in order to
   transfer it as a series of chunks, each with its own size indicator,
   followed by an OPTIONAL trailer containing header fields.  Chunked
   enables content streams of unknown size to be transferred as a
   sequence of length-delimited buffers, which enables the sender to
   retain connection persistence and the recipient to know when it has
   received the entire message.

     chunked-body   = *chunk
                      last-chunk
                      trailer-part
                      CRLF

     chunk          = chunk-size [ chunk-ext ] CRLF
                      chunk-data CRLF
     chunk-size     = 1*HEXDIG
     last-chunk     = 1*("0") [ chunk-ext ] CRLF

     chunk-data     = 1*OCTET ; a sequence of chunk-size octets

   The chunk-size field is a string of hex digits indicating the size of
   the chunk-data in octets.  The chunked transfer coding is complete
   when a chunk with a chunk-size of zero is received, possibly followed
   by a trailer, and finally terminated by an empty line.

   A recipient MUST be able to parse and decode the chunked transfer
   coding.
```

An example would be something like this:

```
GET / HTTP/1.1\r\nConnection: Keep-Alive\r\nTransfer-Encoding: chunked\r\n\r\n1\r\nh\r\n4\r\nello\r\n0\r\n\r\n
```

Notice the 2 CLRF's at the ending of the request. One is of the last last-chunk and the other one is the trailing CRLF of the chunked-body. Currently it Puma only accounts for a single CLRF and incorrectly sees the second CLRF as the start of a following request, which is malformed.

This PR checks for a second CLRF and skips it if present.

Fixes #1456 
Fixes #1480 

As a sidenode, does Puma parse trailer-parts?

PR #1812, opened 2019-06-08
Fix a bug that the last CRLF of chunked body may be used in the next request

kou:
The last CRLF of chunked body is checked by #1607. But it's
incomplete. If a client sends the last CRLF (or just LF) after Puma
processes "0\r\n" line, the last CRLF (or just LF) isn't dropped in
the "0\r\n" process:

https://github.com/puma/puma/blob/675344e8609509b0d767ae7680436b3b382d8394/lib/puma/client.rb#L183-L192

    if line.end_with?("\r\n")
      len = line.strip.to_i(16)
      if len == 0
        @body.rewind
        rest = io.read
        # rest is "" with no the last CRLF case and
        # "\r" with no last LF case.
        # rest.start_with?("\r\n") returns false for
        # Both of these cases.
        rest = rest[2..-1] if rest.start_with?("\r\n")
        @buffer = rest.empty? ? nil : rest
        set_ready
        return true
      end

The unprocessed last CRLF (or LF) is used as the first data in the
next request. Because Puma::Client#reset sets `@parsed_bytes`
to 0.

https://github.com/puma/puma/blob/675344e8609509b0d767ae7680436b3b382d8394/lib/puma/client.rb#L100-L109

    def reset(fast_check=true)
      @parsed_bytes = 0

It means that data in `@buffer` (it's "\r" in no the last LF case) and
unread data in input socket (it's "\r\n" in no the last CRLF case and
"\n" in no the last LF case) are used used as the first data in the
next request.

This change fixes these cases by the followings:

  * Ensures reading the last CRLF by setting `@partial_part_left` when
    CRLF isn't read in processing "0\r\n" line.

  * Introduces a `@in_last_chunk` new state to detect whether the last
    CRLF is waiting or not. It's reset in Puma::Client#reset.

### DIFF
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -232,9 +232,9 @@ module Puma
 
       if @in_last_chunk
         set_ready
-        return true
+        true
       else
-        return false
+        false
       end
     end
 
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -68,6 +68,8 @@ module Puma
       @remote_addr_header = nil
 
       @body_remain = 0
+
+      @in_last_chunk = false
     end
 
     attr_reader :env, :to_io, :body, :io, :timeout_at, :ready, :hijacked,
@@ -108,6 +110,7 @@ module Puma
       @ready = false
       @body_remain = 0
       @peerip = nil
+      @in_last_chunk = false
 
       if @buffer
         @parsed_bytes = @parser.execute(@env, @buffer, @parsed_bytes)
@@ -166,7 +169,7 @@ module Puma
           chunk = chunk[@partial_part_left..-1]
           @partial_part_left = 0
         else
-          @body << chunk
+          @body << chunk if @partial_part_left > 2 # don't include the last \r\n
           @partial_part_left -= chunk.size
           return false
         end
@@ -184,12 +187,20 @@ module Puma
         if line.end_with?("\r\n")
           len = line.strip.to_i(16)
           if len == 0
+            @in_last_chunk = true
             @body.rewind
             rest = io.read
-            rest = rest[2..-1] if rest.start_with?("\r\n")
-            @buffer = rest.empty? ? nil : rest
-            set_ready
-            return true
+            last_crlf_size = "\r\n".bytesize
+            if rest.bytesize < last_crlf_size
+              @buffer = nil
+              @partial_part_left = last_crlf_size - rest.bytesize
+              return false
+            else
+              @buffer = rest[last_crlf_size..-1]
+              @buffer = nil if @buffer.empty?
+              set_ready
+              return true
+            end
           end
 
           len += 2
@@ -219,7 +230,12 @@ module Puma
         end
       end
 
-      return false
+      if @in_last_chunk
+        set_ready
+        return true
+      else
+        return false
+      end
     end
 
     def read_chunked_body
--- a/test/test_puma_server.rb
+++ b/test/test_puma_server.rb
@@ -564,7 +564,7 @@ EOF
     @server.run
 
     sock = TCPSocket.new @host, @server.connected_port
-    sock << "GET / HTTP/1.1\r\nConnection: close\r\nTransfer-Encoding: chunked\r\n\r\n1\r\nh\r\n4\r\nello\r\n0\r\n"
+    sock << "GET / HTTP/1.1\r\nConnection: close\r\nTransfer-Encoding: chunked\r\n\r\n1\r\nh\r\n4\r\nello\r\n0\r\n\r\n"
 
     data = sock.read
 
@@ -586,7 +586,7 @@ EOF
     sock << "GET / HTTP/1.1\r\nConnection: close\r\nTransfer-Encoding: chunked\r\n\r\n1\r\n"
     sleep 1
 
-    sock << "h\r\n4\r\nello\r\n0\r\n"
+    sock << "h\r\n4\r\nello\r\n0\r\n\r\n"
 
     data = sock.read
 
@@ -608,7 +608,7 @@ EOF
     sock << "GET / HTTP/1.1\r\nConnection: close\r\nTransfer-Encoding: chunked\r\n\r\n1\r\nh\r\n"
     sleep 1
 
-    sock << "4\r\nello\r\n0\r\n"
+    sock << "4\r\nello\r\n0\r\n\r\n"
 
     data = sock.read
 
@@ -630,7 +630,7 @@ EOF
     sock << "GET / HTTP/1.1\r\nConnection: close\r\nTransfer-Encoding: chunked\r\n\r\n1\r"
     sleep 1
 
-    sock << "\nh\r\n4\r\nello\r\n0\r\n"
+    sock << "\nh\r\n4\r\nello\r\n0\r\n\r\n"
 
     data = sock.read
 
@@ -652,7 +652,7 @@ EOF
     sock << "GET / HTTP/1.1\r\nConnection: close\r\nTransfer-Encoding: chunked\r\n\r\n1"
     sleep 1
 
-    sock << "\r\nh\r\n4\r\nello\r\n0\r\n"
+    sock << "\r\nh\r\n4\r\nello\r\n0\r\n\r\n"
 
     data = sock.read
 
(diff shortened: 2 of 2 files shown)


test/test_puma_server.rb
```diff
@@ -786,11 +786,23 @@ def test_chunked_keep_alive_two_back_to_back
     @server.run
 
     sock = TCPSocket.new @host, @server.connected_port
-    sock << "GET / HTTP/1.1\r\nConnection: Keep-Alive\r\nTransfer-Encoding: chunked\r\n\r\n1\r\nh\r\n4\r\nello\r\n0\r\n\r\n"
+    sock << "GET / HTTP/1.1\r\nConnection: Keep-Alive\r\nTransfer-Encoding: chunked\r\n\r\n1\r\nh\r\n4\r\nello\r\n0\r\n"
```
nateberkopec: I'm not a huge fan of this test. I think a test for this PR would ideally test for the bug that you found. The test should process 2 requests in a row, and the 2nd request should not start with a CRLF.

lib/puma/client.rb
```diff
@@ -165,7 +168,7 @@ def decode_chunk(chunk)
           chunk = chunk[@partial_part_left..-1]
           @partial_part_left = 0
         else
-          @body << chunk
+          @body << chunk if @partial_part_left > 2 # don't include the last \r\n
```
nateberkopec: this feels brittle, because we're not actually certain that the last 2 characters are \r\n.

lib/puma/client.rb
```diff
@@ -183,12 +186,20 @@ def decode_chunk(chunk)
...
           if len == 0
+            @in_last_chunk = true
             @body.rewind
             rest = io.read
-            rest = rest[2..-1] if rest.start_with?("\r\n")
-            @buffer = rest.empty? ? nil : rest
-            set_ready
-            return true
+            last_crlf_size = "\r\n".bytesize
+            if rest.bytesize < last_crlf_size
```
nateberkopec: this feels pretty brittle too. aren't there edge cases where this would be `true`?

lib/puma/client.rb
```diff
@@ -218,7 +229,12 @@ def decode_chunk(chunk)
         end
       end
 
-      return false
+      if @in_last_chunk
+        set_ready
+        return true
```
nateberkopec: unneccessary `return`

nateberkopec:
Thanks for catching this.

I think this PR needs a clearer test that actually describes and tests the bug that you laid out in your very thorough and complete PR description. From there, we'll be able to write an implementation that catches it.

test/test_puma_server.rb
kou: Do you imagine a test like the following?

```ruby
sock = TCPSocket.new @host, @server.connected_port
sock << "GET / HTTP/1.1\r\nConnection: Keep-Alive\r\nTransfer-Encoding: chunked\r\n\r\n1\r\nh\r\n4\r\nello\r\n0\r\n\r\n"
sock << "GET / HTTP/1.1\r\nConnection: Keep-Alive\r\nTransfer-Encoding: chunked\r\n\r\n4\r\ngood\r\n3\r\nbye\r\n0\r\n\r\n"

h = header(sock)
assert_equal ["HTTP/1.1 200 OK", "Content-Length: 0"], h
assert_equal "hello", body

h = header(sock)
# Not "HTTP/1.1 400 Bad Request" (Request started with "\r\n" is bad request)
assert_equal ["HTTP/1.1 200 OK", "Content-Length: 0"], h
assert_equal "goodbye", body
```

This doesn't reproduce this case. Because this case should delay "\r\n" in the last chunk.

The following test reproduces this case but this expects that we don't require the last empty chunk to process chunked body. This is the current Puma behavior. Should we keep this behavior?

```ruby
def test_chunked_last_empty_chunk_delay
  # ...
  sock = TCPSocket.new @host, @server.connected_port
  sock << "GET / HTTP/1.1\r\nConnection: Keep-Alive\r\nTransfer-Encoding: chunked\r\n\r\n1\r\nh\r\n4\r\nello\r\n0\r\n"

  h = header(sock)
  assert_equal ["HTTP/1.1 200 OK", "Content-Length: 0"], h
  assert_equal "hello", body # Processed without the last "\r\n"

  sock << "\r\n"
  sock << "GET / HTTP/1.1\r\nConnection: Keep-Alive\r\nTransfer-Encoding: chunked\r\n\r\n4\r\ngood\r\n3\r\nbye\r\n0\r\n\r\n"

  h = header(sock)
  assert_equal ["HTTP/1.1 200 OK", "Content-Length: 0"], h
  assert_equal "goodbye", body

  sock.close
end
```

lib/puma/client.rb
kou: I think so too but the current implementation doesn't check content: https://github.com/puma/puma/blob/675344e8609509b0d767ae7680436b3b382d8394/lib/puma/client.rb#L163

So I followed this style.

I think that robust content check should be done in a separated pull request.

lib/puma/client.rb
kou: There are edge cases. For example, the first request omits the last empty chunk:

```ruby
sock = TCPSocket.new @host, @server.connected_port
# No last empty chunk
sock << "GET / HTTP/1.1\r\nConnection: Keep-Alive\r\nTransfer-Encoding: chunked\r\n\r\n1\r\nh\r\n4\r\nello\r\n0\r\n"
# The leading "GE" is consumed as the last empty chunk (\r\n)
sock << "GET / HTTP/1.1\r\nConnection: Keep-Alive\r\nTransfer-Encoding: chunked\r\n\r\n4\r\ngood\r\n3\r\nbye\r\n0\r\n\r\n"

h = header(sock)
assert_equal ["HTTP/1.1 200 OK", "Content-Length: 0"], h
assert_equal "hello", body

h = header(sock)
# The next request starts with "T / HTTP/..." (No "GE")
assert_equal ["HTTP/1.1 200 OK", "Content-Length: 0"], h
assert_equal "goodbye", body

sock.close
```

I omitted robust content check to follow the current style.

I think that robust content check should be done in a separated pull request.

lib/puma/client.rb
kou: I think so too but I follow the current style:

  * https://github.com/puma/puma/blob/675344e8609509b0d767ae7680436b3b382d8394/lib/puma/client.rb#L156
  * https://github.com/puma/puma/blob/675344e8609509b0d767ae7680436b3b382d8394/lib/puma/client.rb#L300

There are no `return` cases too:

  * https://github.com/puma/puma/blob/675344e8609509b0d767ae7680436b3b382d8394/lib/puma/client.rb#L336
  * https://github.com/puma/puma/blob/675344e8609509b0d767ae7680436b3b382d8394/lib/puma/client.rb#L372

Which style is Puma style?

kou:
I've left a comment about test at https://github.com/puma/puma/pull/1812#discussion_r291850895 .

lib/puma/client.rb
nateberkopec: From here on out (I make the rules now!) no more unnecessary returns.

lib/puma/client.rb
nateberkopec: Agreed.

lib/puma/client.rb
nateberkopec: OK

test/test_puma_server.rb
```diff
@@ -786,11 +786,23 @@ def test_chunked_keep_alive_two_back_to_back
...
+      sock << "\n"
+      last_crlf_written = true
+    end
 
     h = header(sock)
     assert_equal ["HTTP/1.1 200 OK", "Content-Length: 0"], h
     assert_equal "hello", body
+    assert_equal true, last_crlf_written
+
+    last_crlf_writer.join
```
nateberkopec: Doesn't this line have to be *before* `assert last_crlf_written`?

lib/puma/client.rb
kou: OK. I've removed unnecessary returns from this pull request.

test/test_puma_server.rb
kou: We need to put this here.
This tests that Puma must wait the last CRLF to complete HTTP body processing.
If we put this before `assert_equal true, last_crlf_written`, the assertion is passed even if Puma doesn't wait the last CRLF. (This is the current behavior. And this causes this problem.)

nateberkopec:
Tests failing now.

kou:
I've added missing last CRLF for chunked request into existing tests.
Now CI is green.

nateberkopec:
Thank you for your contribution!

PR #1816, opened 2019-06-10
Allow to define a custom formatter for logs

ylecuyer:
This PR allows to have a custom formatter for logs through a `log_formatter` dsl inside the configuration file

Example with a new rails app and no `log_formatter`:

```
=> Booting Puma
=> Rails 5.2.3 application starting in development 
* Version 3.12.1 (ruby 2.5.5-p157), codename: Llamas in Pajamas
* Environment: development
* Listening on tcp://localhost:3000
Use Ctrl-C to stop
```

With the config :

```
log_formatter do |str|
  "[#{Process.pid}] [#{Socket.gethostname}] #{Time.now}: #{str}"
end
```

```
=> Booting Puma
=> Rails 5.2.3 application starting in development 
[25473] [ylecuyer-U32U] 2019-06-10 22:29:43 +0200: Puma starting in single mode...
[25473] [ylecuyer-U32U] 2019-06-10 22:29:43 +0200: * Version 3.12.1 (ruby 2.5.5-p157), codename: Llamas in Pajamas
[25473] [ylecuyer-U32U] 2019-06-10 22:29:43 +0200: * Min threads: 5, max threads: 5
[25473] [ylecuyer-U32U] 2019-06-10 22:29:43 +0200: * Environment: development
[25473] [ylecuyer-U32U] 2019-06-10 22:29:43 +0200: * Listening on tcp://localhost:3000
[25473] [ylecuyer-U32U] 2019-06-10 22:29:43 +0200: Use Ctrl-C to stop
```

### DIFF
--- a/History.md
+++ b/History.md
@@ -8,15 +8,19 @@
   * Min worker timeout (#1716)
   * Add option to suppress SignalException on SIGTERM (#1690)
   * Allow mutual TLS CA to be set using `ssl_bind` DSL (#1689)
+  * Reactor now uses nio4r instead of `select` (#1728)
+  * Minimum Ruby version now >= 2.5 (#1813)
   * Add log_formatter configuration
 * x bugfixes
-  * Do not accept new requests on shutdown (#1685)
+  * Do not accept new requests on shutdown (#1685, #1808)
   * Fix 3 corner cases when request body is chunked (#1508)
   * Change pid existence check's condition branches (#1650)
   * Don't call .stop on a server that doesn't exist (#1655)
   * Implemented NID_X9_62_prime256v1 (P-256) curve over P-521 (#1671)
   * Fix @notify.close can't modify frozen IOError (RuntimeError) (#1583)
-
+  * Fix Java 8 support (#1773)
+  * Fix error `uninitialized constant Puma::Cluster` (#1731)
+  * Fix `not_token` being able to be set to true (#1803)
 
 ## 3.12.1 / 2019-03-19
 
--- a/History.md
+++ b/History.md
@@ -1,12 +1,16 @@
 ## Master
 
-x features
+* Features
+  * Add log_formatter configuration (#1816)
+
+* Bugfixes
+  * Your bugfix goes here (#Github Number)
+
+## 4.0.1 / 2019-07-11
 
 * 2 bugfixes
-  * Socket removed after reload (#1829)
-  * Add extconf tests for DTLS_method & TLS_server_method, use in minissl.rb.  (#1832)
-* 1 feature
-  * Add log_formatter configuration
+  * Fix socket removed after reload - should fix problems with systemd socket activation. (#1829)
+  * Add extconf tests for DTLS_method & TLS_server_method, use in minissl.rb. Should fix "undefined symbol: DTLS_method" when compiling against old OpenSSL versions. (#1832)
 
 
 ## 4.0.0 / 2019-06-25
@@ -1439,3 +1443,12 @@ be added back in a future date when a java Puma::MiniSSL is added.
 ## 1.0.0 / 2012-03-29
 
 * Released!
+
+## Ignore - this is for maintainers to copy-paste during release
+## Master
+
+* Features
+  * Your feature goes here (#Github Number)
+
+* Bugfixes
+  * Your bugfix goes here (#Github Number)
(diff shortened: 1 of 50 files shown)


ylecuyer:
@nateberkopec can't this PR be included in the incoming 4.0.1 release ?

nateberkopec:
That wouldn't be very SemVer of me, would it? 😆

nateberkopec:
Thanks for keeping this updated, I think we can get it in w/4.1.

ylecuyer:
OK great to hear thanks

nateberkopec:
I'm cool with this. Needs a test.

nateberkopec:
Still needs a test here if it will get into 4.1

ylecuyer:
I have added some tests

lib/puma/launcher.rb
```diff
@@ -81,7 +83,7 @@ def initialize(conf, launcher_args={})
       set_rack_environment
 
       if clustered?
-        @events.formatter = Events::PidFormatter.new
+        @events.formatter = Events::PidFormatter.new unless @options[:log_formatter]
```
nateberkopec: Actually, let's just make this `@events.formatter ||= Events::PidFormatter.new`. I don't like have two conditionals looking at `options[:log_formatter]`.

History.md
```diff
@@ -11,8 +11,7 @@
 * 2 bugfixes
   * Fix socket removed after reload - should fix problems with systemd socket activation. (#1829)
   * Add extconf tests for DTLS_method & TLS_server_method, use in minissl.rb. Should fix "undefined symbol: DTLS_method" when compiling against old OpenSSL versions. (#1832)
-* 1 other
```
nateberkopec: ? Why remove?

test/test_events.rb
```diff
@@ -152,7 +152,7 @@ def test_pid_formatter
   end
 
   def test_custom_log_formatter
```
nateberkopec: Oh, I didn't even realize this is a thing.

So what does this PR give us? I'm not sure the CustomFormatter class does anything, as it has the same interface as a Proc.

However, I think the config option is useful. I'm not sure how one would have added a custom formatter before 

I'm not sure there's any benefit to the CustomFormatter class added here.

test/test_cli.rb
```diff
@@ -253,6 +253,30 @@ def test_state_file_callback_filtering
...
+    assert_instance_of Puma::Events::PidFormatter, cli.launcher.events.formatter
+  end
+
+  def test_log_formatter_custom_single
+    cli = Puma::CLI.new [ "--config", "test/config/custom_log_formatter.rb" ]
+    assert_instance_of Puma::Events::CustomFormatter, cli.launcher.events.formatter
+  end
+
+  def test_log_formatter_custom_clustered
+    skip_on :jruby, :windows
```
nateberkopec: Would prefer @MSP-Greg's approach to the skips here.

lib/puma/launcher.rb
```diff
@@ -63,6 +60,11 @@ def initialize(conf, launcher_args={})
       @options = @config.options
       @config.clamp
 
+      @events.formatter = Events::CustomFormatter.new(@options[:log_formatter]) if @options[:log_formatter]
+
+      @binder        = Binder.new(@events)
```
nateberkopec: Pretty sure moving this is unnecessary, let's keep it where it is.

lib/puma/launcher.rb
nateberkopec: Oh right, duh, there's already a formatter.

Instead, let's just do it all in one conditional then. Keep our impact in as much of one place as possible.

```ruby
@events.formatter = if options[:log_formatter]
  options[:log_formatter]
elsif clustered?
  Events::PidFormatter
end
```

History.md
ylecuyer: This is an error I'll revert

test/test_cli.rb
ylecuyer: done :+1:

test/test_events.rb
ylecuyer: i remove the CustomFormatter class and used a proc

lib/puma/launcher.rb
ylecuyer: Did something a little different so I don't have to add an else case with the default formatter

PR #1818, opened 2019-06-11
Updates docs, removes refs. to IO.select()

Jesus:
Hi @schneems,

I'm sending this PR as per your suggestion in https://github.com/puma/puma/pull/1806#issuecomment-500509845

By the way, personally I'm annoyed by long lines in source code.  Would you be interested in another PR trimming down these lines to a max of 80? I've seen some parts of Puma respect this limit while others just don't. What's the preference of the maintainers?

And one last thought, your long texts in the source code have helped me a lot recently while researching this codebase, so MANY THANKS!

### DIFF
--- a/lib/puma/reactor.rb
+++ b/lib/puma/reactor.rb
@@ -24,7 +24,7 @@ module Puma
   # which stores it in an array and waits for any of the connections to be ready for reading.
   #
   # The waiting/wake up is performed with nio4r, which will use the apropriate backend (libev, Java NIO or
-  # just plain IO#select). The call to `NIO::Selector#select` (hereinafter `select()`) will "wake up" and
+  # just plain IO#select). The call to `NIO::Selector#select` will "wake up" and
   # return the references to any objects that caused it to "wake". The reactor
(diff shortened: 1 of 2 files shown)


lib/puma/reactor.rb
```diff
@@ -69,19 +70,18 @@ def initialize(server, app_pool)
 
     private
 
-
     # Until a request is added via the `add` method this method will internally
     # loop, waiting on the `sockets` array objects. The only object in this
     # array at first is the `@ready` IO object, which is the read end of a pipe
     # connected to `@trigger` object. When `@trigger` is written to, then the loop
-    # will break on `IO.select` and return an array.
+    # will break on `select()` and return an array.
```
schneems: I think all these `select()`'s are technically accurate but a bit confusing since `IO.select` and `NIO::Selector#select` are two different things. Can we be more accurate and say `NIO::Selector#select` instead?

Jesus: I don't see why not. Thanks for your review.

schneems:
Thanks!

PR #1822, opened 2019-06-18
Swallow connection errors when sending early hints

### LINKED ISSUE #1640: Early hints + javascript_include_tag = Puma:ConnectionError
### Steps to reproduce

1) Procfile enables early hints eg. web: bundle exec puma -C config/puma.rb --early-hints

2) Make use of `javascript_include_tag` helper

3) Production environment on Heroku

### Expected behavior

Testing out --early-hints on Heroku, I would expect the hints to be generated as described in [Implement H2 Early Hints for Rails](https://github.com/rails/rails/pull/30744).

### Actual behavior

I'm seeing some intermittent socket timeout errors when I have --early-hints enabled on my app, which is hosted on Heroku. It's only showing up about once a day, and only in production.

Some lucky visitor's request is throwing one Puma:ConnectionError for each `javascript_include_tag` in my application layout. If one call fails, they all fail in the order they appear in the template. I've never experienced the end-result personally, leading me to guess that the visitor probably won't notice as their script tags would just load normally.

```Aug 16 12:34:37 #<Puma::ConnectionError: Socket timeout writing data> 
Aug 16 12:34:37 /app/vendor/bundle/ruby/2.4.0/gems/puma-3.11.4/lib/puma/server.rb:972:in 'rescue in fast_write' 
Aug 16 12:34:37 /app/vendor/bundle/ruby/2.4.0/gems/puma-3.11.4/lib/puma/server.rb:963:in 'fast_write' 
Aug 16 12:34:37 /app/vendor/bundle/ruby/2.4.0/gems/puma-3.11.4/lib/puma/server.rb:609:in 'block in handle_request' 
Aug 16 12:34:37 /app/vendor/bundle/ruby/2.4.0/gems/actionpack-5.2.0/lib/action_dispatch/http/request.rb:217:in 'send_early_hints' 
Aug 16 12:34:37 /app/vendor/bundle/ruby/2.4.0/gems/actionview-5.2.0/lib/action_view/helpers/asset_tag_helper.rb:96:in 'javascript_include_tag' 
Aug 16 12:34:37 /app/vendor/bundle/ruby/2.4.0/gems/sprockets-rails-3.2.1/lib/sprockets/rails/helper.rb:156:in 'block in javascript_include_tag' 
Aug 16 12:34:37 /app/vendor/bundle/ruby/2.4.0/gems/sprockets-rails-3.2.1/lib/sprockets/rails/helper.rb:154:in 'map' 
Aug 16 12:34:37 /app/vendor/bundle/ruby/2.4.0/gems/sprockets-rails-3.2.1/lib/sprockets/rails/helper.rb:154:in 'javascript_include_tag' 
Aug 16 12:34:37 /app/app/views/layouts/application.html.erb:97:in 'block in _app_views_layouts_application_html_erb__4118899528439397404_70152019751760' 
Aug 16 12:34:37 /app/app/views/layouts/application.html.erb:95:in 'each' 
Aug 16 12:34:37 /app/app/views/layouts/application.html.erb:95:in '_app_views_layouts_application_html_erb__4118899528439397404_70152019751760'
```

### System configuration

**Ruby version**: 2.4.1
**Rails version**: 5.2.0
**Puma version**: 3.11.4

Jesus:
I've been thinking on how Puma should behave when the early hints
can't be sent because the connection has been lost.

I first thought that it should behave the same way it does when the
connection is lost before main response is sent, I've seen that it
swallows those errors (https://git.io/fjVtg) so perhaps we should do
the same here.

There's one consideration though. Because the early hints are sent at
the very beginning of the app execution, we may want to abort this
request altogether and save resources by throwing an exception and
causing Rails to interrupt execution.

Finally I decided to overlook that case just because I believe it's
least surprising from the Rack app point of view. Also
I haven't seen any other example of the web server interrupting
the apps execution due to an error on the lower layer.

Fixes #1640.

### DIFF
--- a/test/test_puma_server.rb
+++ b/test/test_puma_server.rb
@@ -190,29 +190,25 @@ EOF
   end
 
   def test_early_hints_are_ignored_if_connection_lost
-    # Have a long early hints header so we have time to close connection
-    # before the server sends the whole content.
-    links = Array.new(100) { |n| "</script_#{n}.js>; rel=preload" }.join("\n")
-
     app = proc { |env|
-      env['rack.early_hints'].call("Link" => links)
+      env['rack.early_hints'].call("Link" => "</script.js>; rel=preload")
       [200, { "X-Hello" => "World" }, ["Hello world!"]]
     }
 
     events = Puma::Events.strings
     server = Puma::Server.new app, events
-
+    def server.fast_write(*args)
+      raise Puma::ConnectionError
+    end
     server.add_tcp_listener @host, @port
     server.early_hints = true
     server.run
 
-    # Make a request and close the socket immediately, we expect the
-    # connection to be closed while the server is sending the early hints
+    # This request will cause the server to try and send early hints
     sock = TCPSocket.new @host, server.connected_port
     sock << "HEAD / HTTP/1.0\r\n\r\n"
-    sock.close
 
-    # Let the server fail...
+    # Give the server some time to try to write (and fail)
     sleep 0.1
 
     # Expect no errors in stderr
--- a/test/test_puma_server.rb
+++ b/test/test_puma_server.rb
@@ -189,6 +189,36 @@ EOF
     assert_equal expected_data, data
   end
 
+  def test_early_hints_are_ignored_if_connection_lost
+    # Have a long early hints header so we have time to close connection
+    # before the server sends the whole content.
+    links = Array.new(100) { |n| "</script_#{n}.js>; rel=preload" }.join("\n")
+
+    app = proc { |env|
+      env['rack.early_hints'].call("Link" => links)
+      [200, { "X-Hello" => "World" }, ["Hello world!"]]
+    }
+
+    events = Puma::Events.strings
+    server = Puma::Server.new app, events
+
+    server.add_tcp_listener @host, @port
+    server.early_hints = true
+    server.run
+
+    # Make a request and close the socket immediately, we expect the
+    # connection to be closed while the server is sending the early hints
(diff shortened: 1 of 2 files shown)


dentarg:
> I've seen that it swallows those errors (https://git.io/fjVL8)

Here's the [future proof version of that GitHub link](https://github.com/puma/puma/blob/c242e76f4d14dd7582ce27062c6b0d26ff4abaf5/lib/puma/server.rb#L489-L490).

Jesus:
I've updated both the Github message & the commit message with your link. Thank you!

nateberkopec:
Works for me. I think you could write a test that just has fast_write blow up/raise.

Jesus:
Test added and I think this is ready to merge 🚢

test/test_puma_server.rb
```diff
@@ -189,6 +189,36 @@ def test_early_hints_works
     assert_equal expected_data, data
   end
 
+  def test_early_hints_are_ignored_if_connection_lost
+    # Have a long early hints header so we have time to close connection
+    # before the server sends the whole content.
+    links = Array.new(100) { |n| "</script_#{n}.js>; rel=preload" }.join("\n")
```
nateberkopec: Rather than do this, which is a nondeterministic test, let's just do something like:

```ruby
server = Puma::Server.new app, events
def server.fast_write(*args)
  raise ConnectionError
end
```

Jesus: Ah, great! I wanted to do that but I couldn't figure out how to modify `#fast_write` without affecting the whole test suite.

Thanks!

Issue #1765, opened 2019-04-10
pumactl does not respond to status while using PIDs


jessebs:
### Steps to reproduce

1) in puma.rb:

```
daemonize true
pidfile '/tmp/app.pid'
```

2) pumactl start

3) pumactl status

### Expected behavior

I expect some type of status indicating whether or not puma is running

### Actual behavior

Nothing

### System configuration

**Ruby version**: 2.5.1
**Sinatra version**: 2.0.5
**Puma version**: 3.12.1

jessebs:
See pull request #1767

PR #1824, opened 2019-06-23
Add status to pumactl with pidfile

spk:
ref https://github.com/puma/puma/pull/1767
fix https://github.com/puma/puma/issues/1765

First PR adding status output when using pidfile option

cc @nateberkopec


nateberkopec:
🚢

PR #1825, opened 2019-06-25
v4.0.0

schneems:


ylecuyer:
Btw @schneems why did you do a major upgrade as there are no breaking changes in the changelog ?

nateberkopec:
Because we're not "officially" SemVer and we can do what we want?

We swapped out one of the most important internal parts of Puma with a new dependency. It doesn't break the API, but it's a "major" change in our opinion.

Issue #1768, opened 2019-04-11
More debug to QUERY_STRING max length


ylecuyer:
### Steps to reproduce

1) Make an HTTP query with a query string bigger than 1024*10

2) read the output of puma stderr

```
2019-04-11 11:37:02 +0200: HTTP parse error, malformed request (127.0.0.1): #<Puma::HttpParserError: HTTP element QUERY_STRING is longer than the (1024 * 10) allowed length (was 15839)>
```

### Expected behavior

The actual HTTP uri should be logged to know which query triggered the error

### Actual behavior

The error message only shows the error and not the associated URI

### System configuration

**Ruby version**: ruby 2.5.3p105
**Rails version**: Rails 4.2.11.1
**Puma version**: puma-3.12.0

nateberkopec:
I wonder if we should log something other than the URL too.

nateberkopec:
Merged on master.

ylecuyer:
Thanks

PR #1831, opened 2019-07-01
Add REQUEST_PATH on parse error message

spk:
Request get `/plop` and query string bigger than 1024*10

```
2019-07-01 21:31:08 +0200: HTTP parse error, malformed request (127.0.0.1/plop): #<Puma::HttpParserError: HTTP element QUERY_STRING is longer than the (1024 * 10) allowed length (was 20481)>
```

ref https://github.com/puma/puma/issues/1768

### DIFF
--- a/lib/puma/events.rb
+++ b/lib/puma/events.rb
@@ -93,7 +93,10 @@ module Puma
     # parsing exception.
     #
     def parse_error(server, env, error)
-      @stderr.puts "#{Time.now}: HTTP parse error, malformed request (#{env[HTTP_X_FORWARDED_FOR] || env[REMOTE_ADDR]}): #{error.inspect}\n---\n"
+      @stderr.puts "#{Time.now}: HTTP parse error, malformed request " \
+        "(#{env[HTTP_X_FORWARDED_FOR] || env[REMOTE_ADDR]}#{env[REQUEST_PATH]}): " \
+        "#{error.inspect}" \
+        "\n---\n"
     end
 
     # An SSL error has occurred.
--- a/test/test_events.rb
+++ b/test/test_events.rb
@@ -156,4 +156,25 @@ class TestEvents < Minitest::Test
 
     assert_equal "-> ready", out
   end
+
+  def test_parse_error
+    port = 0
+    host = "127.0.0.1"
+    app = proc { |env| [200, {"Content-Type" => "plain/text"}, ["hello\n"]] }
+    events = Puma::Events.strings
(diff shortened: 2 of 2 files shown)


nateberkopec:
Works for me. Test?

spk:
> Works for me. Test?

Added in https://github.com/puma/puma/pull/1831/commits/ea555ef9d617782c03ca0554ca8069d92f109838

ylecuyer:
Would it be possible to log the full url? Instead of just the path

nateberkopec:
Awesome, thanks for adding a test ❤️

spk:
> Would it be possible to log the full url? Instead of just the path

If you mean with the full url adding the query string in this case it wouldn't be possible since we cannot parse it, the QUERY_STRING is going to be empty.

For the IP its going to be present when puma is behind an http roxy forwarding the env x-forwarded-for or remote-addr cf https://github.com/spk/puma/blob/ea555ef9d617782c03ca0554ca8069d92f109838/lib/puma/events.rb#L97

I hope I answer your question and of course I let Nate correct me if I'm wrong !

nateberkopec:
I appreciate the effort you put in to the test here, but I'm trying to avoid adding more integration tests like this unless absolutely necessary.

I'd prefer a simpler [test like the one here.](https://github.com/puma/puma/blob/master/test/test_http11.rb#L84), or an even simpler one that simply calls parse_error and checks stdout.

spk:
> I appreciate the effort you put in to the test here, but I'm trying to avoid adding more integration tests like this unless absolutely necessary.
> 
> I'd prefer a simpler [test like the one here.](https://github.com/puma/puma/blob/master/test/test_http11.rb#L84), or an even simpler one that simply calls parse_error and checks stdout.

I've looked up I don't think its possible to check the stderr log here, we already have a test for [max params](https://github.com/spk/puma/blob/master/test/test_http11.rb#L141) and this raise directly... Maybe I'm wrong thanks for your help!

nateberkopec:
@spk Since the method you're changing is actually in Puma::Events, you'll need to start a new Puma::Server with a custom IO object (don't use global stderr/stdout), and then you can cause a parser error and check the IO object. Maybe the tests for Server or Events would be a better inspiration. I would put this test in with the event tests because that's where this method actually lives.

spk:
Thanks Nate for your help hope its ok with this https://github.com/spk/puma/commit/55334b2ff1ab3f7eb8c7e3a593be332f74e2f2f5

PR #1832, opened 2019-07-01
Use OpenSSL DTLS_method & TLS_server_method when available

MSP-Greg:
2nd commit changes a require in Rakefile to require_relative...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,9 @@ rvm:
 matrix:
   fast_finish: true
   include:
+    - rvm: 2.2 
+      dist: trusty 
+      env: NOTES="Trusty OpenSSL 1.0.1"
     - rvm: ruby-head
       env: RUBYOPT="--jit"
     - rvm: 2.4.6
--- a/History.md
+++ b/History.md
@@ -1,7 +1,9 @@
 ## Master
 
 x features
-x bugfixes
+
+* ? bugfixes
+  * Add extconf tests for DTLS_method & TLS_server_method, use in minissl.rb.  (#1832)
 
 ## 4.0.0 / 2019-06-25
 
--- a/Rakefile
+++ b/Rakefile
@@ -3,7 +3,7 @@ require "rake/testtask"
 require "rake/extensiontask"
 require "rake/javaextensiontask"
 require "rubocop/rake_task"
-require 'puma/detect'
+require_relative 'lib/puma/detect'
 require 'rubygems/package_task'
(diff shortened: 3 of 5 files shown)


MSP-Greg:
Looking over mini_ssl, the line:

https://github.com/puma/puma/blob/a2cc0537a40c1b1c5dfca831a4e93d0967940b62/ext/puma_http11/mini_ssl.c#L172

can probably have a similar PR for `TLS_server_method`...

This would replace some of the code in #1799, and I will review the code in #1788 & #1799, as the two PR's share some of the same code (my mistake, I forgot about the first PR).

nateberkopec:
* Needs a changelog entry
* I'm going to say the trusty CI build covers this testing-wise

MSP-Greg:
> Needs a changelog entry

Done.

> I'm going to say the trusty CI build covers this testing-wise

Agreed.  Fixes issue #1833

MSP-Greg:
@nateberkopec

Thanks.  Since there's some overlap (my mistake), I'll now combine #1788 & #1799 with another PR.  Hopefully, SSL tests will then pass for any OpenSSL build/configuration and be a good indicator of functionality.

I agree that a `MiniSSL::Context#min_version=` method/attr is best left for later...

MSP-Greg:
@nateberkopec, et al

I saw the Friday release mention.

**Question:** Don't know future plans re releases, but I think one thing people might want is to also have a `no_tlsv1_1=` method to lock Puma to TLSv1.2 or higher.

Make it so `#no_tlsv1_1 = true` overrides `#no_tlsv1 = false`.

Ideally,  `MiniSSL::Context#min_version=` is probably a better solution, but given that new TLS versions are really uncommon, having two 'no_' methods isn't cumbersome.  Maybe when TLSv1.4 or TLSv2.0 is released five years from now...

Thoughts?  Mention because it is much easier than `MiniSSL::Context#min_version=`...

nateberkopec:
> I think one thing people might want is to also have a no_tlsv1_1= method to lock Puma to TLSv1.2 or higher.

That's fine in the near-term I think. Probably for 4.1.

MSP-Greg:
Ok, so **not for the next release**?

It's relatively easy to add, so I could add it when I combine #1788 & #1799 

IOW, generate the new PR (and include `no_tlsv1_1=`) within the next few days...

nateberkopec:
> not for the next release?

Correct. I'd like to keep new methods out of 4.0.1.

PR #1836, opened 2019-07-09
Openssl updates

MSP-Greg:
**1st commit - OpenSSL 1.1.1 updates, add `#no_tlsv1_1`**
1.1 Adds three constants to MiniSSL, as many OpenSSL builds cannot/will not connect with older protocols:
```
   OPENSSL_NO_SSL3
   OPENSSL_NO_TLS1
   OPENSSL_NO_TLS1_1
```
1.2 Uses `SSL_CTX_set_min_proto_version` when available
1.3 Add `#no_tlsv1_1` method to `Puma::MiniSSL::Context`

**2nd commit - travis.yml - add bionic 18.04 job**
2.1 Add Ubuntu bionic 18.04 job with Ruby 2.6.3.  This tests with OpenSSL 1.1.1

**3rd commit - test_puma_server_ssl.rb - add tests for `#no_tlsv1`, `#no_tlsv1_1`**
3.1 Adds tests for `#no_tlsv1` and `#no_tlsv1_1`
3.2 Changes setup (via adding a `start_server` method), allowing the context to be passed to a block

**4th commit - add no_tlsv1_1 to binder, config, etc**
4.1 Adds use of `#no_tlsv1_1` with binder.rb and dsl.rb
4.2 Adds tests to test_binder.rb and test_config.rb for above
4.3 Adds feature to MiniSSL.java

### DIFF
--- a/test/test_puma_server_ssl.rb
+++ b/test/test_puma_server_ssl.rb
@@ -1,8 +1,6 @@
 require_relative "helper"
 require "puma/minissl"
 require "puma/puma_http11"
-# net/http (loaded in helper) does not necessarily load OpenSSL
-require "openssl" unless Object.const_defined? :OpenSSL
 
 #———————————————————————————————————————————————————————————————————————————————
 #             NOTE: ALL TESTS BYPASSED IF DISABLE_SSL IS TRUE
@@ -27,13 +25,25 @@ DISABLE_SSL = begin
             rescue
               true
             else
+              # net/http (loaded in helper) does not necessarily load OpenSSL
+              require "openssl" unless Object.const_defined? :OpenSSL
               false
             end
 
 class TestPumaServerSSL < Minitest::Test
 
   def setup
-    return if DISABLE_SSL
+    @http = nil
+    @server = nil
+  end
+
+  def teardown
+    @http.finish if @http && @http.started?
+    @server.stop(true) if @server
+  end
+
+  # yields ctx to block, use for ctx setup & configuration
+  def start_server
     @port = UniquePort.call
     @host = "127.0.0.1"
 
@@ -51,6 +61,8 @@ class TestPumaServerSSL < Minitest::Test
 
     ctx.verify_mode = Puma::MiniSSL::VERIFY_NONE
 
+    yield ctx if block_given?
+
     @events = SSLEventsHelper.new STDOUT, STDERR
     @server = Puma::Server.new app, @events
     @ssl_listener = @server.add_ssl_listener @host, @port, ctx
@@ -61,13 +73,8 @@ class TestPumaServerSSL < Minitest::Test
     @http.verify_mode = OpenSSL::SSL::VERIFY_NONE
   end
 
-  def teardown
-    return if DISABLE_SSL
-    @http.finish if @http.started?
-    @server.stop(true)
-  end
-
   def test_url_scheme_for_https
+    start_server
     body = nil
     @http.start do
       req = Net::HTTP::Get.new "/", {}
@@ -81,6 +88,7 @@ class TestPumaServerSSL < Minitest::Test
   end
 
   def test_request_wont_block_thread
+    start_server
     # Open a connection and give enough data to trigger a read, then wait
     ctx = OpenSSL::SSL::SSLContext.new
     ctx.verify_mode = OpenSSL::SSL::VERIFY_NONE
@@ -100,6 +108,7 @@ class TestPumaServerSSL < Minitest::Test
   end
 
   def test_very_large_return
+    start_server
     giant = "x" * 2056610
 
     @server.app = proc do
@@ -118,6 +127,7 @@ class TestPumaServerSSL < Minitest::Test
   end
 
   def test_form_submit
+    start_server
     body = nil
     @http.start do
       req = Net::HTTP::Post.new '/'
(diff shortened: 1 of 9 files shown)


MSP-Greg:
A few notes:

1. I mentioned the new `#no_tlsv1_1` method in another thread, immediately after that, I googled TLSv1.1.  There are several major sites (including GitHub?) that have disabled TLSv1.1.  It's time, and I had already started on the PR.

2. Travis/rvm seems to have Ruby master/trunk/head messed up again.  If it isn't fixed soon, I'll patch then.  I think `RUBYOPT=--jit` needs to be defined after `bundle install`, which fixed the issue in another repo...

3. Ruby 2.6.3 is currently tested in Xenial (1.0.2), and on Appveyor, Bionic, & OSX, all with 1.1.1.  Since OSX jobs take quite a while, I'd suggest removing the OSX 2.6.3 job.  On OSX, both the 2.4 & 2.5 jobs test on 1.1.1.

4. Lastly, I think with the additions/changes in this PR, all OpenSSL issues are resolved, and nothing will need to be added/changed in the foreseeable future.  After all, we're testing on 1.0.1 thru 1.1.1, all is passing, and protocols can be restricted if needed...

nateberkopec:
🙏 Really solid work on all of this @MSP-Greg. The way a PR should be done!

Will review soon, looks like there was an unrelated build failure.

MSP-Greg:
@nateberkopec 

Thank you, really.  Sorry for adding the method, but I needed a break from a Ruby/JS mess, and I've done a few other OpenSSL PR's this week, similar issues.  So, I just wanted to get it done...

> looks like there was an unrelated build failure.

That pissed me off.

1. Passed in my fork.
2. Been involved with testing enought that I generally don't write unstable tests.  Checked other jobs, and some ran all the SSL tests before the tests that failed.
3. I think the failed tests involve `fork`, and being a windows guy, I can't test them locally.

Long story short, I guess we'll all keep an eye on it.  Re Travis, two people generate all the windows Ruby builds used in CI, but Travis/rvm can't keep their trunk/master/head build working.  Shaking my head going AFK...

test/test_puma_server_ssl.rb
```diff
@@ -33,7 +33,17 @@ def ssl_error(server, peeraddr, peercert, error)
 class TestPumaServerSSL < Minitest::Test
 
   def setup
-    return if DISABLE_SSL
+    @http = nil
+    @server = nil
+  end
+
+  def teardown
+    @http.finish if !@http.nil? && @http.started?
```
nateberkopec: style nitpick but I hate unnecessary predicates. I think `if @http` is sufficient here, as is `unless @server` in the next line.

test/test_puma_server_ssl.rb
```diff
@@ -127,6 +138,47 @@ def test_ssl_v3_rejection
...
+  def test_tls_v1_rejection
+    skip("TLSv1 protocol is unavailable") if Puma::MiniSSL::OPENSSL_NO_TLS1
+    start_server { |ctx| ctx.no_tlsv1 = true }
+
+    if @http.respond_to? :max_version=
+      @http.max_version = :TLS1
+    else
+      @http.ssl_version = :TLSv1
+    end
+    # Ruby 2.4.5 on Travis raises ArgumentError
```
nateberkopec: lmao

nateberkopec:
The Ruby here is good with me, needs a review from someone else on the C.

nateberkopec:
> I think RUBYOPT=--jit needs to be defined after bundle install
> I'd suggest removing the OSX 2.6.3 job. On OSX, both the 2.4 & 2.5 jobs test on 1.1.1.

make-it-so-captain-102843

test/test_puma_server_ssl.rb
MSP-Greg: @nateberkopec

I just pushed this change, and it's breaking things.

Although I can't recall what/why, there are times when connections aren't 'made' due to testing for restricted protocol versions.  Are you ok with returning these to the original code in the PR?

MSP-Greg:
Should be ok if the 'review from someone else on the C' is good.

Ruby master will fail on Travis until a bundler issue specific to 'nested bundle commands used when bundler is a default gem' is fixed.  PR submitted to bundler...

test/test_puma_server_ssl.rb
nateberkopec: Yup! if the predicate is necessary, then it's necessary. just thought it wasn't.

MSP-Greg:
@nateberkopec

1. rebased
2. added the code to use `no_tlxv1_1` to binder, config, tests, etc.  Previously, the PR just added it to `MiniSSL::Context`, which isn't very useful...

MSP-Greg:
2019-08-03 15:30 UTC, I selfishly rebased to trigger a CI run to test ruby-head.

The fix for 'nested bundle commands when bundle is a default gem' has been pushed from Bundler to ruby-head/trunk, and ruby-head is now passing.

A small patch in this PR additionally fixes ruby-head with `--jit`, and it is also passing.

See:
https://travis-ci.org/puma/puma/builds/567321441

Note that this wasn't an issue on Appveyor, as all the tests that have bundle commands also use `fork`, which is unavailable on Windows & JRuby...

PR #1837, opened 2019-07-10
Avoid mutating global STDOUT & STDERR

montanalow:
The current behavior prevents apps from benefitting from `STDOUT.sync = false` during configuration phases, since the initial call to `Event.stdio` is relatively late in the boot cycle.

### DIFF
--- a/lib/puma/events.rb
+++ b/lib/puma/events.rb
@@ -29,8 +29,8 @@ module Puma
     #
     def initialize(stdout, stderr)
       @formatter = DefaultFormatter.new
-      @stdout = stdout
-      @stderr = stderr
+      @stdout = stdout.dup
+      @stderr = stderr.dup
 
       @stdout.sync = true
       @stderr.sync = true
--- a/test/test_events.rb
+++ b/test/test_events.rb
@@ -6,7 +6,6 @@ class TestEvents < Minitest::Test
 
     assert_instance_of Puma::NullIO, events.stdout
     assert_instance_of Puma::NullIO, events.stderr
-    assert_equal events.stdout, events.stderr
   end
 
   def test_strings
@@ -19,8 +18,17 @@ class TestEvents < Minitest::Test
   def test_stdio
     events = Puma::Events.stdio
 
-    assert_equal STDOUT, events.stdout
-    assert_equal STDERR, events.stderr
+    # events.stdout is a dup, so same file handle, different ruby object, but inspect should show the same file handle
+    assert_equal STDOUT.inspect, events.stdout.inspect
+    assert_equal STDERR.inspect, events.stderr.inspect
+  end
+
+  def test_stdio_respects_sync
+    STDOUT.sync = false
+    events = Puma::Events.stdio
+
+    assert !STDOUT.sync
+    assert events.stdout.sync
   end
 
   def test_register_callback_with_block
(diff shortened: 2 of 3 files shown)


nateberkopec:
I'm afraid I don't understand the use case.

This PR mostly changes how our internal logs interact with flush. This seems weird, first of all, because the whole point of `sync = false` is, IMO, to let Ruby do the `flush`ing for you.

I also don't understand the PR title - how was global stdout and stderr being modified before this PR?

montanalow:
`$stdout = STDOUT` by default.

This call results in setting `$stdout.sync = true`. 
https://github.com/puma/puma/blob/master/lib/puma/events.rb#L149

In the current code path, whenever `Puma::Events.stdio` is called, `STDOUT` & `STDERR` will be changed to `sync`. `sync` is not the default behavior for `STDOUT` & `STDERR` so this is a side effect on a global variable that will have application wide results.

I would like to prevent this side effect, so Puma applications can run with `STDOUT.sync # => false` 

I'm unfamiliar with the reasoning behind the current code setting `sync = true`, but my patch preserves the behavior of synchronized output for `Events`, by flushing `stdout` and `stderr` any time output is written.

nateberkopec:
Why not just change the default IO to a new STDOUT/STDERR object, then?

E.g. 

```ruby
def self.stdio
    Events.new $stdout.dup, $stderr.dup
end
```

nateberkopec:
That leaves the separate issue of us forcing you to use sync = false, but at least we avoid the global side effect.

nateberkopec:
Actually, even better:

```ruby 
def self.stdio 
    stdout = $stdout.dup
    stdout.sync = false 
    stderr = $stderr.dup 
    stderr.sync = false
    Events.new stdout, stderr
end

then rm the sync code from Events#initialize 
```

montanalow:
@nateberkopec To catch `DEFAULT = new(STDOUT, STDERR)` as well, I moved the dup inside initialize, which I think aligns with your suggestions.

nateberkopec:
I think that's OK, but you'll need to make sure that [the rack handler](https://github.com/puma/puma/blob/master/lib/rack/handler/puma.rb#L69) changes it's arguments so that the correct item is used.

montanalow:
Not sure I follow, are you worried about the resulting `StringIO.new.dup` from `Rack::Events.strings`?

nateberkopec:
On that line, `::Puma::Events.stdio` (which is the code path we'll take if the the silent option is NOT present) will return the global stdout/stderr. That is then passed into Puma::Launcher#new, avoiding the change that you've made here b/c the event is passed in explicitly.

montanalow:
Sorry if I'm still missing something, but I thought that's what my change would address. By `dup`ing the args from all callers inside `Events#intitialize`, before `sync` is set, there shouldn't be any way to get update the globals, even if you pass them in.

`Events.stdio` will be `dup`ped, so won't impact `$stdout`, and `Events::DEFAULT` will also be `dup`ped to avoid mutating STDOUT, when called from `Puma::Launcher#initialize` here https://github.com/puma/puma/blob/master/lib/puma/launcher.rb#L48

nateberkopec:
Well, write a test and we'll find out then 😆 You'll have to fix the tests that are looking for output at $stdout/$stderr too unfortunately.

nateberkopec:
Looking at this again and I think you're probably write, but it does need new tests + fixes of the old ones for a merge.

nateberkopec:
Thanks for all your work @montanalow !

nateberkopec:
Who could have guessed the entire Puma logging ecosystem somehow depended on us modifying STDOUT.sync to true... a good lesson for software maintainership - modifying a global is probably worth a major version bump!

PR #1844, opened 2019-07-14
Instrument start time

ylecuyer:
Instrument start time for workers/launcher and expose them through Puma.stats


lib/puma/cluster.rb
```diff
@@ -95,12 +95,13 @@ def initialize(idx, pid, phase, options)
         @signal = "TERM"
         @options = options
         @first_term_sent = nil
+        @started_at = Time.now
```
nateberkopec: cluster inherits from runner, so I believe this is already initialized, no?

nateberkopec:
Code looks fine, but what's the use case here?

Isn't this info already available in `ps aux`, for example?

nateberkopec:
Also wondering how this interacts with restarts and child process kills/restarts

ylecuyer:
> Code looks fine, but what's the use case here?
> 
> Isn't this info already available in `ps aux`, for example?

This is the use case https://github.com/ylecuyer/puma-status

> Also wondering how this interacts with restarts and child process kills/restarts

When I tested, the `started_at` field is reset at restart

lib/puma/cluster.rb
ylecuyer: But this is for the `Worker` class as you can see L89

SuperQ:
Nice, thanks, this will also be useful for our Prometheus metrics endpoint.

PR #1850, opened 2019-07-18
More test fixups

nateberkopec:


MSP-Greg:
@nateberkopec

https://github.com/puma/puma/blob/30fbd8442e844b950b59b00c65850c526e0312e1/win_gem_test/puma.ps1#L40

Needs to change to:
```
Update-Gems minitest, minitest-retry, minitest-proveit, rack, rake
```

I guess I could have used bundler, but it has the same issues as on Travis...

test/helper.rb
```diff
@@ -10,18 +11,19 @@
...
-rescue LoadError
-  warn "Failed to load bundler ... this should only happen during package building"
-end
+# begin
+#   require "bundler/setup"
+#   # bundler/setup may not load bundler
+#   require "bundler" unless Bundler.const_defined?(:ORIGINAL_ENV)
+# rescue LoadError
+#   warn "Failed to load bundler ... this should only happen during package building"
+# end
```
olleolleolle: Now that CI has passed, also without this section, can it be omitted altogether?

test/test_web_server.rb
```diff
@@ -79,10 +59,30 @@ def test_header_is_too_long
...
     long = "GET /test HTTP/1.1\r\nContent-length: #{body.length}\r\n\r\n" + body
-    do_test(long, (Puma::Const::CHUNK_SIZE * 2) - 400)
+    socket = do_test(long, (Puma::Const::CHUNK_SIZE * 2) - 400)
+    assert_match "hello", socket.read
   end
 
+  private
+
+  def do_test(string, chunk, close_after=nil)
+    # Do not use instance variables here, because it needs to be thread safe
```
nateberkopec: Not sure this is true.

test/helper.rb
nateberkopec: This was the only thing blocking my merge :D

test/test_web_server.rb
MSP-Greg: I think a new instance is used in every thread?  If so, instance variables should be ok.  JFYI, I'm running `test_puma_server_ssl.rb` right now locally, and it passed running parallel...

Issue #1775, opened 2019-04-16
Socket removed after reload


j15e:
### Steps to reproduce

I followed instruction on the wiki to setup puma with systemd on Ubuntu 18.04 (https://github.com/puma/puma/blob/master/docs/systemd.md#socket-activation) but puma reload is not working as expected with the latest version of puma.

During a reload, I see these logs which highlight the problem (with `PUMA_DEBUG=1`):

> Apr 16 03:38:22 fizz bundle[1184]: % Registered unix:/home/app/web/shared/tmp/sockets/puma.sock for activation from LISTEN_FDS
> Apr 16 03:38:22 fizz bundle[1184]: [1184] Puma starting in cluster mode...
> Apr 16 03:38:22 fizz bundle[1184]: [1184] * Version 3.12.1 (ruby 2.5.5-p157), codename: Llamas in Pajamas
> Apr 16 03:38:22 fizz bundle[1184]: [1184] * Min threads: 8, max threads: 16
> Apr 16 03:38:22 fizz bundle[1184]: [1184] * Environment: production
> Apr 16 03:38:22 fizz bundle[1184]: [1184] * Process workers: 2
> Apr 16 03:38:22 fizz bundle[1184]: [1184] * Phased restart available
> Apr 16 03:38:22 fizz bundle[1184]: [1184] * Activated unix:///home/app/web/shared/tmp/sockets/puma.sock
> Apr 16 03:38:22 fizz bundle[1184]: [1184] Use Ctrl-C to stop
> Apr 16 03:38:25 fizz bundle[1184]: [1184] - Worker 1 (pid: 1225) booted, phase: 0
> Apr 16 03:38:25 fizz bundle[1184]: [1184] - Worker 0 (pid: 1221) booted, phase: 0
> Apr 16 03:38:43 fizz systemd[1]: Reloading Puma HTTP Server.
> Apr 16 03:38:43 fizz bundle[1184]: [1184] - Starting phased worker restart, phase: 1
> Apr 16 03:38:43 fizz bundle[1184]: [1184] + Changing to /home/app/web/current
> Apr 16 03:38:43 fizz systemd[1]: Reloaded Puma HTTP Server.
> Apr 16 03:38:45 fizz bundle[1184]: [1184] - Stopping 1221 for phased upgrade...
> Apr 16 03:38:45 fizz bundle[1184]: [1184] - TERM sent to 1221...
[...]

PR #1851, opened 2019-07-18
Update binder, test_cli, test_integration - fix unix sockets

MSP-Greg:
@nateberkopec

Since I'm UnixSocket challenged, I looked at some of the Core test code.  Possible fix may revolve around [`File.socket?`](https://msp-greg.github.io/ruby_trunk/Core/File.html#socket?-class_method).  Used it in the `Binder#close` method.  The rest is just cleanup, etc.

Don't know about the issue with deleted files (#1775)

Passed in my fork without all the errors...

### DIFF
--- a/lib/puma/binder.rb
+++ b/lib/puma/binder.rb
@@ -395,7 +395,7 @@ module Puma
     def inherit_unix_listener(path, fd)
       @unix_paths << path
 
-      if fd.kind_of? UNIXServer
+      if fd.kind_of? TCPServer
         s = fd
       else
         s = UNIXServer.for_fd fd
--- a/lib/puma/binder.rb
+++ b/lib/puma/binder.rb
@@ -50,12 +50,9 @@ module Puma
 
     def close
       @ios.each { |i| i.close }
-      @unix_paths.each do |i|
-        # Errno::ENOENT is intermittently raised
-        begin
-          unix_socket = UNIXSocket.new i
-          unix_socket.close
-        rescue Errno::ENOENT
+      @unix_paths.uniq.each do |i|
+        if i && !File.socket?(i) && File.exist?(i)
+          File.unlink(i)
         end
       end
     end
@@ -398,7 +395,7 @@ module Puma
     def inherit_unix_listener(path, fd)
       @unix_paths << path
 
-      if fd.kind_of? TCPServer
+      if fd.kind_of? UNIXServer
         s = fd
       else
         s = UNIXServer.for_fd fd
--- a/test/test_cli.rb
+++ b/test/test_cli.rb
@@ -12,7 +12,7 @@ class TestCLI < Minitest::Test
 
     @tmp_path2 = "#{@tmp_path}2"
 
-    File.unlink @tmp_path if File.exist? @tmp_path
+    File.unlink @tmp_path  if File.exist? @tmp_path
     File.unlink @tmp_path2 if File.exist? @tmp_path2
 
     @wait, @ready = IO.pipe
@@ -26,9 +26,8 @@ class TestCLI < Minitest::Test
   end
 
   def teardown
-    File.unlink @tmp_path if File.exist? @tmp_path
+    File.unlink @tmp_path  if File.exist? @tmp_path
     File.unlink @tmp_path2 if File.exist? @tmp_path2
-
     @wait.close
     @ready.close
   end
@@ -107,6 +106,7 @@ class TestCLI < Minitest::Test
 
     cli.launcher.stop
     t.join
+    s.close
   end
 
   def test_control
@@ -131,6 +131,7 @@ class TestCLI < Minitest::Test
 
     cli.launcher.stop
     t.join
+    s.close
   end
 
   def test_control_stop
@@ -154,6 +155,7 @@ class TestCLI < Minitest::Test
     assert_equal '{ "status": "ok" }', body.split("\r\n").last
 
     t.join
+    s.close
   end
 
   def control_gc_stats(uri, cntl)
--- a/test/test_integration.rb
+++ b/test/test_integration.rb
@@ -12,8 +12,8 @@ require "open3"
 class TestIntegration < Minitest::Test
 
   def setup
-    @state_path = "test/test_puma.state"
-    @bind_path = "test/test_server.sock"
+    @state_path   = "test/test_puma.state"
+    @bind_path    = "test/test_server.sock"
     @control_path = "test/test_control.sock"
     @token = "xxyyzz"
 
@@ -26,10 +26,6 @@ class TestIntegration < Minitest::Test
   end
 
   def teardown
-    File.unlink @state_path rescue nil
-    File.unlink @bind_path rescue nil
-    File.unlink @control_path rescue nil
-
     @wait.close
     @ready.close
 
@@ -42,6 +38,10 @@ class TestIntegration < Minitest::Test
 
       @server.close
     end
+
+    File.unlink @state_path   if File.exist? @state_path
+    File.unlink @bind_path    if File.exist? @bind_path
+    File.unlink @control_path if File.exist? @control_path
   end
 
   def server_cmd(argv)
@@ -198,6 +198,7 @@ class TestIntegration < Minitest::Test
         done = true
       end
     end
+
     # Stop
     ccli = Puma::ControlCLI.new ["-S", @state_path, "stop"], sout
     ccli.run


test/test_cli.rb
```diff
@@ -26,9 +26,8 @@ def wait_booted
   end
 
   def teardown
-    File.unlink @tmp_path if File.exist? @tmp_path
-    File.unlink @tmp_path2 if File.exist? @tmp_path2
-
+    File.unlink @tmp_path  if @tmp_path && File.socket?(@tmp_path)
+    File.unlink @tmp_path2 if @tmp_path && File.socket?(@tmp_path2)
```
Jesus: Did you mean...
```ruby
File.unlink @tmp_path2 if @tmp_path2 && File.socket?(@tmp_path2)
```
?

lib/puma/binder.rb
```diff
@@ -50,12 +50,9 @@ def env(sock)
...
     def close
       @ios.each { |i| i.close }
-      @unix_paths.each do |i|
-        # Errno::ENOENT is intermittently raised
-        begin
-          unix_socket = UNIXSocket.new i
-          unix_socket.close
-        rescue Errno::ENOENT
+      @unix_paths.uniq.each do |i|
+        if i && !File.socket?(i) && File.exist?(i)
```
nateberkopec: Why not check existence before checking if it's a socket?

test/test_cli.rb
```diff
@@ -107,6 +106,7 @@ def test_control_clustered
 
     cli.launcher.stop
     t.join
+    s.close
```
nateberkopec: Hmmmm... why is this necessary? Shouldn't Puma do this by itself?

test/test_integration.rb
```diff
@@ -198,12 +198,13 @@ def test_phased_restart_via_pumactl
         done = true
       end
     end
+    assert File.exist? @bind_path
```
nateberkopec: Why did moving this before the CLI invocation matter? I assume the state would be the same.

test/test_cli.rb
MSP-Greg: Thanks. Long day, stupid mistake.  I'll fix with a push.

test/test_integration.rb
MSP-Greg: I ran thru several iterations in my fork.  At one point, the file did not exist after stop.  An, since the issue was that sockets were disappearing after a restart, I moved it.

MSP-Greg:
@nateberkopec & @Jesus

After I submitted the PR, I was bothered by some of the test teardown code.  I think that all the unlink commands should be of the of the form:

```ruby
File.unlink @state_path if File.exist? @state_path
```

Thoughts?

lib/puma/binder.rb
MSP-Greg: I believe File.socket?(i) is the check for whether it should be left in place due to a restart.

I left `File.exist?(i)` in just so an error wouldn't be generated if something like a graceful shutdown was run and the 'other side' removed the file?

nateberkopec: Ohhhh I get it. So on a restart, the file descriptor is not a socket anymore (from Ruby's perspective)?

nateberkopec:
BTW @MSP-Greg and @Jesus I've really appreciated both of your contributions and responsiveness over the last few weeks.

IMO Greg it bothers me that we're even unsure that the file exists or not.

lib/puma/binder.rb
MSP-Greg: Umm, backwards maybe?  On restart, we want to leave connections intact, so we don't want to unlink the file?

So, on restart, since the 'other side' still has the socket open, `File.socket?(i)` is true, but it's negated in the logic to unlink the file...

test/test_cli.rb
MSP-Greg: As mentioned, I can't just play around with Unix Sockets locally.  But:

1. The test creates `s`, not Puma. Hence, the test should close it.  Think 'not quite right' PR's.
2. `teardown` doesn't have access to `s`., but it should unlink the file if Puma doesn't.

I'll remove if you'd like.  I was just trying to make the tests bomb-proof.  If we need more asserts or tests, that's another thing...

MSP-Greg:
@nateberkopec 

> IMO Greg it bothers me that we're even unsure that the file exists or not.

Are you speaking of within tests, or Puma?

We've got something communicating with Puma, we'll call it VOG.  Maybe on a shutdown, VOG issues the command, assumes Puma receives it, waits a brief amount of time, then closes the socket and unlinks the file.  Maybe Puma takes a bit more time to shutdown before it hits its own close/unlink logic.  I don't know, but I'd prefer it to be bomb-proof...

MSP-Greg:
Just pushed a change that aligns the test setup/teardown unlink code to always be of the form:

```ruby
File.unlink @tmp_path2 if File.exist? @tmp_path2
```

MSP-Greg:
Please review errors in Travis:

https://travis-ci.org/puma/puma/jobs/560543589#L851

This seems to be a consistent error with Ruby 2.2.

https://travis-ci.org/puma/puma/jobs/560543589#L1061

This maybe intermittent, seems to only be Ruby 2.2.

I checked builds 3567 and 3570.

nateberkopec:
https://travis-ci.org/puma/puma/jobs/560543589#L851 is my fault due to that test being parallelized

MSP-Greg:
Thanks.  Changing to parallel testing is always fun (Ruby Core on Windows).  Re the first issue, I dug around, and found something I forgot about.

See https://bugs.ruby-lang.org/issues/13239

Maybe we should skip the test on Ruby 2.2?

MSP-Greg:
I ran some 'debug' builds, and through all the tests, every item in binder's @unix_paths returns true for `File.socket?`, so no files are unlinked...

rromanchuk:
@MSP-Greg i'm now running your branch and i can report my deploy is now fixed, thanks!

MSP-Greg:
@rromanchuk

Thank you for testing it, and glad to hear it works.

test/test_integration.rb
MSP-Greg: I moved this back, I must have moved before I got Binder working correctly.

Jesus:
> BTW @MSP-Greg and @Jesus I've really appreciated both of your contributions and responsiveness over the last few weeks.

Absolutely, **thank you** for all the feedback and support.

lib/puma/binder.rb
evanphx: I don't see how this logic makes sense at all. Why would the file exist and be a regular file? File.socket? will return true even if there is nothing connected to it I believe.

MSP-Greg: > File.socket? will return true even if there is nothing connected to it I believe

Quick check on Travis showed that `File.socket? __FILE__` returned false.

>  Why would the file exist and be a regular file? 

Normally it wouldn't.  But if `File.socket?` does work, and hence, nothing is connected, shouldn't it be unlinked?

MSP-Greg:
@evanphx

Regarding your comment above, I should have mentioned that the code needed for Puma's handling of UNIXSockets was committed in 2013 at https://github.com/puma/puma/commit/f124bc21cd6556ad0ad2779e5a71a02a6a67fc95 in [`Binder#add_unix_listener`](https://msp-greg.github.io/puma/Puma/Binder.html#add_unix_listener-instance_method).

The code in [`Binder#close`](https://msp-greg.github.io/puma/Puma/Binder.html#close-instance_method) only needs to close all the members of @ios.

Hence, I could remove all the code associated with `unlink`, or I could add a comment like 'Unlink files only if the client has closed their UNIXSocket.'

Thoughts?

rromanchuk:
@MSP-Greg weird, for some reason this started regressing, investigating now

MSP-Greg:
@rromanchuk

I pushed one commit since your first test, I think around July 19, 2019 10:32 CDT.

I can revert it, it was some odd code.  Do you know when things started regressing?

rromanchuk:
@MSP-Greg let me do a sanity check, i just deployed this morning when i saw it regressed, i have it locked to your branch, but not commit, seems likely due to that latest commit, but let me actually verify first, dont want to send you on a goose chase

MSP-Greg:
@rromanchuk

Not really a goose chase, it would take me at least 30 seconds...

rromanchuk:
@MSP-Greg  If you revert so i can deploy one more time for quick sanity test, less variables for me to mess up on redeploy, like setting the wrong `ref`

MSP-Greg:
@rromanchuk

Done.

rromanchuk:
`sudo systemctl restart puma` is wiping the socket and causing nginx to explode again,  let me slow down for a second and make sure I know which code i've got running. I quadruple checked this when when it was working, so this is strange, i'm going to make sure my vendor and gem env is actually running the code I want running.

MSP-Greg:
Thanks for checking it. I'm working in another repo, so I'll be at the keyboard for quite a while...

rromanchuk:
OK, i reverted to `3.12.1` so i could confirm the regression wasn't form my own configuration between upgrades.  Things are working fine after downgrading, but i'm going to give this another go, this time a little more methodically, I may have been conflating two separate issues, and i'm not convinced my lockfile was pulling down the correct commit. 

Deployment setup for context:
https://gist.github.com/rromanchuk/25f5d123e266c7b776a2a8ed60f707aa

Heavily borrowed from https://github.com/puma/puma/issues/1499

lib/puma/binder.rb
evanphx: So you're understanding is that `File.socket?` will return false if nothing is connected? I've never heard of that behavior and kind of doubt it because the entry type is backed into the directory entry from the beginning.

MSP-Greg: As I stated above, I ran it thru several os's and Ruby versions on Travis
```ruby
puts "File.socket? __FILE__  #{File.socket? __FILE__}", RUBY_DESCRIPTION
```

all print the first line as:
```
File.socket? __FILE__  false
```

See https://travis-ci.org/MSP-Greg/testing-travis/builds/561254063

I'm not sure if your implying that once something is connected, and then removed or closed, `File.socket?` will return true?  If so, I didn't check for that.

evanphx: I'm unsure what you're testing with that code because of course a `.rb` file is not a socket. My point is that once a unix domain socket is created, it will return true from `File.socket?` no matter if it's used or not.

So that would mean your code is assuming that maybe the path specified to be a socket is for some reason not actually a domain socket but does exist? I don't see how that would be the case.

MSP-Greg: Ok.  As mentioned above, until a recent commit (but before this PR), the code included
```ruby
@unix_paths.each { |i| File.unlink i }
```

So every file contained in @unix_paths was unlinked.  Also note that there have been and continue to be a number of `Errno::ECONNREFUSED` errors raised in Travis testing.  Some of the issues related to UNIXSockets were related to connection issues on restart.

Hence, I wasn't sure if all the items in @unix_paths should be affected, which resulted in the conditional.  Since I'm a windows guy, I also checked the Ruby UNIXSocket test files, and they were using `File.socket?` in places.

After those changes, all the tests pass, and the `Errno::ECONNREFUSED` errors disappeared.   The other changes were just things I thought stabilized testing.

Also, note the last commit where I returned a line in `inherit_unix_listener` to the original code:
```
if fd.kind_of? TCPServer
```

This seems very odd.

I don't recall every detail of this, I can review it more tomorrow.

MSP-Greg:
@evanphx

I spent some more time on Travis, if a file is created by a UNIXSocket/Server, and everything is disconnected/closed, `File.socket?` will still return true.  Hence, I believe you were correct, and with that, there are never any paths unlinked.  Sorry for the trouble.

I'm closing this, although I may revisit the test changes, as some of them may help with test isolation...

nateberkopec:
> Sorry for the trouble.

No trouble Greg! Sometimes you're right, sometimes you're wrong! 😆

MSP-Greg:
@nateberkopec

Quick question.  If an easier answer. when is ENV[`LISTEN_FDS`] used?

nateberkopec:
Got me on that one chief, you'd have to back through the git blame. Unfamiliar with that feature.

MSP-Greg:
Thanks.  Might have found something, the code came from PR #231 the first line of the description is:
"Added a systemd compatibile socket activation"

which I think is a topic that has come up in an issue...

PR #1858, opened 2019-07-21
Do not override RUBYOPT on travis

spk:
Ref https://travis-ci.org/puma/puma/jobs/561782375#L242

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ before_script:
   - bundle exec rake compile
 
 script:
-  - bundle exec rake
+  - RUBYOPT=$TEST_RUBYOPT bundle exec rake
 
 rvm:
   - 2.2.10
@@ -39,7 +39,7 @@ matrix:
       dist: trusty 
       env: NOTES="Trusty OpenSSL 1.0.1"
     - rvm: ruby-head
-      env: RUBYOPT="--jit"
+      env: TEST_RUBYOPT="--jit"
     - rvm: 2.4.6
       os: osx
       osx_image: xcode10.2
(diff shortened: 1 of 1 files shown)


MSP-Greg:
JFYI, did essentially the same thing in #1836 since I was tired of seeing the message.

Regardless, ruby-head is broken for any CI's that have nested bundle commands (like Puma), and although the patches are committed in bundler, they still have to be ported over to ruby-head...

nateberkopec:
What's the message you're referring to @MSP-Greg?

MSP-Greg:
You mean 'what am I talking about?'  Sorry, I was 'brief'.

Not sure if it's temporary or permanent, but the `--jit` option cannot be in place when installing/compiling an extension gem.  I assume it's buried inside the mkmf std-lib, but I haven't checked.

So, if one wants to use `--jit` for CI testing, it must be set after 'rake compile' or the equivalent.  This PR addresses that, I wanted to point out that I also addressed it in PR 1836.

Also, ruby-head is the only place bundler is currently a default gem.  Bundler (only as a default gem) had a bug that didn't allow nested 'bundle' commands to execute.  The bug has been fixed in bundler master.  Nested commands are used here, but only on Travis, as the tests are skipped on Appveyor.

So, even if this PR or mine is accepted, ruby-head Travis CI will still fail tests until an update of bundler is pushed to ruby-head...

spk:
I did not know about your patch @MSP-Greg your fix looks better could you maybe extract it to another PR so I can close this one, thanks cheers !

MSP-Greg:
> your fix looks better

Thanks. Matter of opinion, they seem pretty equivalent.  Thanks for taking the time to submit the PR.

I may have given a complete explanation, but the that's from a self-imposed requirement that I stay aware of ruby-head issues/problems, both with Travis & Appveyor.

My Windows/MinGW ruby-head build has been used for a few years by several repos for CI.  If I'm not aware of the issues/problems, then my build may break CI for others.  I don't like anyone's CI to break...

Anyway, Thanks.

PR #1861, opened 2019-07-24
WIP: Better: Restart workers in batches / clustered mode.

ylecuyer:
This PR adds the option `restart_workers_in_batch_of` so that not all workers boot at the same time.

Before:

```
=> Booting Puma
=> Rails 5.2.3 application starting in development 
[11938] Puma starting in cluster mode...
[11938] * Version 4.0.1 (ruby 2.5.5-p157), codename: 4 Fast 4 Furious
[11938] * Min threads: 4, max threads: 4
[11938] * Environment: development
[11938] * Process workers: 17
[11938] * Phased restart available
[11938] * Listening on tcp://localhost:3000
[11938] Use Ctrl-C to stop
[11938] * Starting control server on unix:///tmp/puma-status-1563997129438-11938
[11938] - Worker 0 (pid: 11964) booted, phase: 0
[11938] - Worker 1 (pid: 11976) booted, phase: 0
[11938] - Worker 2 (pid: 11978) booted, phase: 0
[11938] - Worker 3 (pid: 11997) booted, phase: 0
[11938] - Worker 4 (pid: 12012) booted, phase: 0
[11938] - Worker 5 (pid: 12024) booted, phase: 0
[11938] - Worker 6 (pid: 12036) booted, phase: 0
[11938] - Worker 7 (pid: 12048) booted, phase: 0
[11938] - Worker 8 (pid: 12061) booted, phase: 0
[11938] - Worker 9 (pid: 12074) booted, phase: 0
[11938] - Worker 10 (pid: 12086) booted, phase: 0
[11938] - Worker 11 (pid: 12096) booted, phase: 0
[11938] - Worker 12 (pid: 12111) booted, phase: 0
[11938] - Worker 13 (pid: 12124) booted, phase: 0
[11938] - Worker 14 (pid: 12138) booted, phase: 0
[11938] - Worker 15 (pid: 12150) booted, phase: 0
[11938] - Worker 16 (pid: 12163) booted, phase: 0
```

With the option set to 3:

```
=> Booting Puma
=> Rails 5.2.3 application starting in development 
[11556] Puma starting in cluster mode...
[11556] * Version 4.0.1 (ruby 2.5.5-p157), codename: 4 Fast 4 Furious
[11556] * Min threads: 4, max threads: 4
[11556] * Environment: development
[11556] * Process workers: 17
[11556] * Phased restart available
[11556] * Listening on tcp://localhost:3000
[11556] Use Ctrl-C to stop
[11556] * Starting control server on unix:///tmp/puma-status-1563997022501-11556
[11556] Booting 3 workers
[11556] - Worker 0 (pid: 11583) booted, phase: 0
[11556] - Worker 1 (pid: 11585) booted, phase: 0
[11556] - Worker 2 (pid: 11606) booted, phase: 0
[11556] Booting 3 workers
[11556] - Worker 3 (pid: 11622) booted, phase: 0
[11556] - Worker 4 (pid: 11626) booted, phase: 0
[11556] - Worker 5 (pid: 11648) booted, phase: 0
[11556] Booting 3 workers
[11556] - Worker 6 (pid: 11661) booted, phase: 0
[11556] - Worker 7 (pid: 11674) booted, phase: 0
[11556] - Worker 8 (pid: 11687) booted, phase: 0
[11556] Booting 3 workers
[11556] - Worker 9 (pid: 11691) booted, phase: 0
[11556] - Worker 10 (pid: 11713) booted, phase: 0
[11556] - Worker 11 (pid: 11726) booted, phase: 0
[11556] Booting 3 workers
[11556] - Worker 12 (pid: 11739) booted, phase: 0
[11556] - Worker 13 (pid: 11743) booted, phase: 0
[11556] - Worker 14 (pid: 11746) booted, phase: 0
[11556] Booting 2 workers
[11556] - Worker 15 (pid: 11778) booted, phase: 0
[11556] - Worker 16 (pid: 11791) booted, phase: 0
```

Test TBD

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -156,7 +156,14 @@ module Puma
       diff = @options[:workers] - @workers.size
       return if diff < 1
 
+      if @options[:restart_workers_in_batch_of]
+        return unless @workers.all?(&:booted?) 
+        diff = [diff, @options[:restart_workers_in_batch_of]].min
+      end
+
       master = Process.pid
+
+      log "Booting #{diff} workers"
 
       diff.times do
         idx = next_worker_index
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -157,7 +157,7 @@ module Puma
       return if diff < 1
 
       if @options[:restart_workers_in_batch_of]
-        return unless @workers.all?(&:booted?) 
+        return unless @workers.all?(&:booted?)
         diff = [diff, @options[:restart_workers_in_batch_of]].min
       end
 
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -236,6 +236,13 @@ module Puma
       @options[:restart_cmd] = cmd.to_s
     end
 
+    # In clustered mode, instead of restarting all workers at once,
+    # restart workers in batch of `count`
+    # Usually count is set to the number of processors available
+    def restart_workers_in_batch_of(count)
+      @options[:restart_workers_in_batch_of] = count
+    end
+
     # Store the pid of the server in the file at +path+.
     def pidfile(path)
       @options[:pidfile] = path.to_s


nateberkopec:
Use case?

ylecuyer:
With a server with X core and Y workers of one thread (because code isn't thread safe) and no preload

When you start the server with X > Y the load on the server will be high, and some workers will timeout because all the app couldn't be loaded. They will try to start again all at the sametime, failling again and again.

It is better to start them X by X

nateberkopec:
Hm, why not just set a more appropriate level of workers or, better yet, do a rolling restart or heroku-style "preboot" swap when deploying so avoid a high-load situation during deploy?

ylecuyer:
The issue is with cold boot here or when doing a rollback and a quick restart is needed, so rolling isn't an option.

The level of workers is OK, they work well when they all achieve to boot.

Preboot is a nice solution when you are able to modify the system which isn't the case

nateberkopec:
We shouldn't try to fix every possible problem that can happen in a server deployment in Puma itself. I don't think the solution for stampeding-herd problem you're describing belongs in Puma, but in the environment to which it's deployed. There's far better ways of solving this problem.

jarthod:
Hi @nateberkopec,

How about preload + rolling restart then? will it be supported at some point? (like in passenger) because the current version of puma is kind of impossible to cold start at high volume (machines with 20+ workers) without a minute of downtime. And if we use the preload we loose the ability to do rolling restart ☹

> Hm, why not just set a more appropriate level of workers

Not sure what you mean by that? People can't just decide to have fewer workers if they need this much to handle the traffic? The problem is mostly that when you start 30 workers, instead of starting 1 or 2 after 10 sec ans start serving traffic, puma prefers to start the 30 at the same time, killing the machine and blocking all requests for 1m+ plus (until it timeouts and reboot indefinitely in the worst case) sounds like this cluster mode wasn't made for small numbers only?

nateberkopec:
> How about preload + rolling restart then? 

Sure, I would definitely support that PR.

> Hm, why not just set a more appropriate level of workers Not sure what you mean by that? 

In @ylecuyer's example they said they were using a number of workers less than the number of cores. That's not the optimal setting. Ideally you would be running at cores == workers or cores * 1.5 == workers. With less traffic per worker, they'd be less likely to timeout and die.

> puma prefers to start the 30 at the same time, killing the machine and blocking all requests for 1m+ plus

I understand this, my point is I think that there's better ways of solving that than starting your workers in smaller batches (which this PR does, and doesn't actually solve the problem you're describing). Adding a reverse proxy, running two servers at once and swapping at the load balancer level like Heroku, or rolling restarts are much better options. @ylecuyer's original example was with preloading off, so rolling restarts are an option.

jarthod:
> Sure, I would definitely support that PR.

The readme says "Preloading can’t be used with phased restart, since phased restart kills and restarts workers one-by-one, and preload_app copies the code of master into the workers." so I guess it's not easy but I'm not sure to understand the real issue here? I guess the problem is that if you don't reload the main process you'll fork the same code again? so you probably need some kind of preload process like passenger does to achieve this?

> In @ylecuyer's example they said they were using a number of workers less than the number of cores. That's not the optimal setting. Ideally you would be running at cores == workers or cores * 1.5 == workers. With less traffic per worker, they'd be less likely to timeout and die.

Good point he actually made a typo ^^ what he meant was more workers than cores of course, in our case it's like 50 workers for 20 cores / 40 threads CPU. The timeout I was talking about is not related to traffic, I'm talking about the boot timeout after which if the process is not ready puma kills it and start again (which makes the problem even worse).

> my point is I think that there's better ways of solving that than starting your workers in smaller batches (which this PR does, and doesn't actually solve the problem you're describing).

Why doesn't it solve the problem? if we start only 10 workers instead of 50 at the same time, they'll be available ~3 times more quickly to start serving requests (reducing downtime and risk of timeout on boot). No? 

> Adding a reverse proxy, running two servers at once and swapping at the load balancer level like Heroku

Sure if we had a couple weeks to spend writing a routing layer that could work (at least for the downtime part), but it would also add more complexity and would not actually fix the issue because starting the puma process would still kill the machine (CPU load)

> or rolling restarts are much better options. @ylecuyer's original example was with preloading off, so rolling restarts are an option.

Yes we're using this of course (and it's great) but talking about when we can't use rolling restart here, which are case of quick rollback for example. Rolling restarts takes about 10 minutes with 50 processes (preload would also reduce this I guess)

Overall I'm not saying this change is the best idea of course ^^ it's kind of a small improvement but I think it makes puma a bit nicer to the machine in case of high number of worker and can't really hurt can it?

nateberkopec:
> Sure if we had a couple weeks to spend writing a routing layer that could work 

The core problem here is that instead, if this is merged, *I* have to spend a few *years* maintaining this feature (as we do every feature added to Puma). It "can't hurt" except for the fact that every single feature we had makes Puma harder to understand and maintain. I want to set a high bar for what gets added into Puma, and if there's other good ways _outside_ of Puma for you to solve a particular problem, I'm going to push you that way rather than add another doohickey onto Puma.

jarthod:
Yep ok, totally makes sense for an open source project. What about the preloading + phased restart feature then? if added to puma it would probably add substantial complexity, would you be ok for this one?

nateberkopec:
> What about the preloading + phased restart feature then?

I'll ask Evan about if this is possible. I feel like it could be. In any case, I think it would be worth it, as preloading should be compatible with any/all other features. There's a good reason to preload, and there's a good reason to phased restart, therefore I would like to support the combination of those features.

It should also help the startup time issue here, as preloading takes care of a lot of that.

jarthod:
Yes in our case preloading + phased restart would totally solve the problem. And even more, as it would also make phase restart way faster and cheaper. But of course it's way more complicated to write and I'm not sure we can afford this amount of time. So even if we can afford it, we don't want to waste time if the feature is not wanted.

Also about my questions earlier:
> The readme says "Preloading can’t be used with phased restart, since phased restart kills and restarts workers one-by-one, and preload_app copies the code of master into the workers." so I guess it's not easy but I'm not sure to understand the real issue here? I guess the problem is that if you don't reload the main process you'll fork the same code again? so you probably need some kind of preload process like passenger does to achieve this?

Do you have some feedback about this?

nateberkopec:
> if you don't reload the main process
exactly, the master process needs to re-exec. or we could use a separate process that the new workers would fork from.

nateberkopec:
from @evanphx: 

> no, it would have to be in puma itself
the idea because the phased restart is that it’s methodical
and if you do it with preload, you have to have the master process loaded with the new code before you start
that makes it less methodical because the master might crash trying to perform it’s pre-spawn restart
either that or you structure the code such that the master doesn’t load the code at all
and instead there is a sort of stem-cell process that forks to create the workers
in which case, to do phased restart you start a new stem-cell with the new code
and kill off workers from the old stem-cell
and create ones from the new stem-cell

jarthod:
Ok thanks, so this last option is the one I suggested above which has been chosen by Passenger (screenshot from one of my personal project):
image
A lighweight core process starts a "preloader" process holding the app (highlighted in this screenshot) and then forks from this preloader. When a restart is requested a new preloader is launched and starts forking new workers while the previous one stays until all of his children workers are stopped.

So that should work fine but is a pretty big refactoring

Issue #1675, opened 2018-11-08
Puma does not shutdown gracefully


Adithya-copart:
Using `Ctrl + C` or `pumactl stop` does not shutdown puma gracefully.

**Gemfile:**
```source 'https://rubygems.org'

ruby '2.5.0', engine: 'jruby', engine_version: '9.2.0.0'

gem 'sinatra'
gem 'sucker_punch'

gem 'pry'
gem 'puma'
```

**Gemfile.lock:**
```GEM
  remote: https://rubygems.org/
  specs:
    coderay (1.1.2)
    concurrent-ruby (1.0.5-java)
    ffi (1.9.25-java)
    method_source (0.8.2)
    mustermann (1.0.3)
    pry (0.10.4-java)
      coderay (~> 1.1.0)
      method_source (~> 0.8.1)
      slop (~> 3.4)
      spoon (~> 0.0)
    puma (3.12.0-java)
    rack (2.0.5)
    rack-protection (2.0.0)
      rack
    sinatra (2.0.0)
      mustermann (~> 1.0)
      rack (~> 2.0)
      rack-protection (= 2.0.0)
      tilt (~> 2.0)
    slop (3.6.0)
    spoon (0.0.6)
      ffi
    sucker_punch (2.1.1)
      concurrent-ruby (~> 1.0)
    tilt (2.0.8)

PLATFORMS
  java

DEPENDENCIES
  pry
  puma
  sinatra
  sucker_punch

RUBY VERSION
   ruby 2.5.0p0 (jruby 9.2.0.0)

BUNDLED WITH
   1.16.5
```

**test_app.rb:**
```
class TestApp < Sinatra::Base

  class TestJob
    include SuckerPunch::Job

    def perform
      puts "The job #{self} is starting to sleep 5sec in #{Thread.current}"
      hsh = Thread.list.reduce({}) {|hsh,t| hsh[t] = t.status; hsh}
      # puts "Thread list in #{Thread.current}: #{hsh}"
      sleep 5
      puts "The job #{self} woke up from sleep in #{Thread.current}"
    end
  end

  get '/' do
    3.times {TestJob.perform_async}
  end

  get '/pry' do
    binding.pry
  end

end
```
**config.ru**
```
require 'sinatra/base'
require 'sucker_punch'
require 'pry'
require './test_app.rb'

run TestApp
```

### Steps to reproduce

1) `bundle install` and then `bundle exec puma`.

2) Hit the URL `get '/'` to trigger the jobs.

[...]

Issue #73, opened 2012-04-14
Socket 'already in use'


kyledrake:
If there is a socket present, the server fails to start:

```
bundle exec puma -b unix:///var/run/website.sock
Puma 0.9.5 starting...
* Min threads: 0, max threads: 16
* Listening on unix:///var/run/website.sock
Address already in use - /var/run/website.sock
```

My first thought was to have Puma remove this socket file before it quits, but I'm not sure what the best-practices for UNIX socket file management are so I wanted to file a ticket before sending any pull requests so the bearded unix wizards could chime in.

This is being connected to an nginx proxy, so it's possible that nginx is holding this file open and that's what's causing the problem.

PR #1868, opened 2019-07-28
Using at_exit for graceful stop upon receiving a SIGINT in JRuby

Adithya-copart:
This resolves https://github.com/puma/puma/issues/1675.
Sorry about the noise in that issue but you can find a sample app [here](https://github.com/puma/puma/issues/1675#issuecomment-515584846) demonstrating the problem.

The `graceful_stop` is not using the same thread as the `at_exit` blocks and is being run in the JVM signal handling thread in JRuby.

So, the `thread.join` [here](https://github.com/puma/puma/blob/31d91c351a2bdd4d1aad3ec5434ae3da8306a8ae/lib/puma/server.rb#L980) is joining `@thread` on the signal handler thread and I'm not sure if that is the intention.

This fix will execute `graceful_stop` within the `at_exit` block on the main thread.

Link: https://github.com/jruby/jruby/issues/5437

### DIFF
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -406,12 +406,6 @@ module Puma
 
       begin
         Signal.trap "SIGINT" do
-          if Puma.jruby?
-            @status = :exit
-            at_exit { graceful_stop }
-            exit
-          end
-
           stop
         end
       rescue Exception
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -408,7 +408,7 @@ module Puma
         Signal.trap "SIGINT" do
           if Puma.jruby?
             @status = :exit
-            graceful_stop
+            at_exit { graceful_stop }
             exit
           end
 
--- a/test/rackup/jruby-sigint.ru
+++ b/test/rackup/jruby-sigint.ru
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+
+java_import java.util.concurrent.ThreadPoolExecutor
+java_import java.lang.Runnable
+
+# From https://github.com/ruby-concurrency/concurrent-ruby/blob/d11b29c37b81320ca7126cc9cd85f4f3d17a78a3/lib/concurrent/executor/java_thread_pool_executor.rb#L111-L117
+pool = java.util.concurrent.ThreadPoolExecutor.new(
+    2,
+    2,
+    60,
+    java.util.concurrent.TimeUnit::SECONDS,
+    java.util.concurrent.LinkedBlockingQueue.new,
+    java.util.concurrent.ThreadPoolExecutor::AbortPolicy.new
+  )
+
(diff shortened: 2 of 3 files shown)


nateberkopec:
Needs a test. Probably just an adaptation/simplification of your reproduction app.

Adithya-copart:
@nateberkopec I'm not sure how to come up with a working test.
I'm running into an issue after adding a test to `test_integration.rb`.

The problem is executing the following code in a irb session will block at ` Process.wait(server.pid)` when running on `master` but not in this branch.

However, if I add the same code to a file and run it, the puma process is shutting down properly on both the branches. I'm not sure what to make out from this.

```
require 'open3'

server = IO.popen("rackup test/rackup/jruby-sigint.ru")

# Wait for bootup
true while server.gets !~ /Ctrl-C/

t1 = Time.now
Process.kill(:INT, server.pid)
Process.wait(server.pid)
p "Time elapsed: #{Time.now - t1}"

```

Adithya-copart:
```
pid = spawn("rackup test/rackup/jruby-sigint.ru")

# Wait until bootup
sleep 6
t1 = Time.now
Process.kill(:INT, pid)
Process.wait(pid)
p "Time elapsed: #{Time.now - t1}"
```

This works inside a file, I'll see if I can integrate this into a test instead.
I might need to make sure the `teardown` works if the server fails to shutdown.

Adithya-copart:
Faced https://github.com/jruby/jruby/issues/1050#issuecomment-521021792 with an approach using `Timeout.timeout` and `Process.wait`.

So, I came-up with a test that fails on master and passes in this branch but it does use `sleep`.

Adithya-copart:
@nateberkopec I opened https://github.com/jruby/jruby/issues/5844 to understand https://github.com/puma/puma/pull/1868#issuecomment-520677982.

@headius investigated the issue and added https://github.com/jruby/jruby/issues/5844#issuecomment-524203618.

This piece of code may not longer be needed in Puma. It was initially added in `cli.rb` via https://github.com/puma/puma/commit/1dbe28e3b906f332733ad54bbdb9a4763ba24052 a long time ago and was later moved to `launcher.rb`.

#1675 no longer happens when I remove this code. The puma server shutdown as expected.
https://github.com/puma/puma/blob/9fb12283dbadf6e2bad04d5e67fe3fc9c29eede5/lib/puma/launcher.rb#L408-L413

nateberkopec:
@Adithya-copart It looks like Evan added that code in response to this comment: https://github.com/puma/puma/issues/73#issuecomment-10452101

What do you think? I guess Ctrl-C still shuts down the JVM right? Is that what you're seeing?

Adithya-copart:
Yep, Ctrl+C is shutting down the JVM without that piece of code.

Tested on my Mac and a JRuby docker container.

nateberkopec:
Does it remove the socket still, though? I guess it must if the tests are passing. Originally that's why this code was added, ctrl-c just blew up the JVM without properly removing the socket.

nateberkopec:
@Adithya-copart maybe the test should check to make sure the socket is removed? 

Adithya-copart:
Updated the test to check if the port is available after shutdown.
Let me know if this is not what you had in mind.

nateberkopec:
Closing in favor of #1961

PR #1870, opened 2019-07-29
fix test/test_unix_socket.rb

MSP-Greg:
Fix to work reliably on platforms without UNIXSockets


MSP-Greg:
Don't know if this is the result in a change to Mintest, but in my fork I had two different jobs erroring as follows:
```
Error:
TestPumaUnixSocket#test_server:
NoMethodError: undefined method `stop' for nil:NilClass
        test/test_unix_socket.rb:19:in `teardown'
```

PR fixes the error by not running any code unless `UNIX_SKT_EXIST`, but shows the skip.

PR #1871, opened 2019-07-29
Appveyor - change lower Ruby version to 2.3 [skip travis]

MSP-Greg:
Current nio4r gem (2.4.0) requires Ruby 2.3 or greater.  The Appveyor build/test system uses 'gem install puma'.

At present, RubyGems ignores ruby version constraints when installing dependencies.

### DIFF
--- a/win_gem_test/puma.ps1
+++ b/win_gem_test/puma.ps1
@@ -13,7 +13,7 @@ Make-Const repo_name 'puma'
 Make-Const url_repo  'https://github.com/puma/puma.git'
 
 #———————————————————————————————————————————————————————————————— lowest ruby version
-Make-Const ruby_vers_low 22
+Make-Const ruby_vers_low 23
 # null = don't compile; false = compile, ignore test (allow failure);
 # true = compile & test
 Make-Const trunk     $false ; Make-Const trunk_x64     $false


MSP-Greg:
The last Windows release of Ruby 2.2 was 2.2.6, dated 2016-11-15, while the latest release was 2.2.10, dated 2018-03-28.

The last Windows release of Ruby 2.3 was 2.3.3, dated 2016-11-21, while the latest release was 2.3.8, dated 2018-10-18.

These were built with the older, unmaintained RubyInstaller system, which was based on MSYS.  Since Ruby 2.4, builds have been done using RubyInstaller2 and MSYS2.  MSYS2 is maintained and is currently using some libraries that are newer than the libraries included with 'bionic 18.04'.

Long story short, I'm not sure I see a need for Windows testing on Ruby 2.2, or even Ruby 2.3.

Regardless, I need to address this issue with the build system, so I'll be adding code.  I'll either allow install via bundler (which has caused issues with Travis in the past, especially on ruby-head), or I'll set it to allow gem installation with code for version selection.

At present, a code block exists for after the gem is installed.  I may add one for before the gem is installed, which would allow Ruby version dependent gem installation...  Sorry.

MSP-Greg:
Closing in favor off #1872, which fixes the problem.  Above still holds, but testing is now running again with Ruby 2.2 on Windows/Appveyor.

nateberkopec:
> Long story short, I'm not sure I see a need for Windows testing on Ruby 2.2, or even Ruby 2.3.

I'm OK with this as I don't think there will be issues with older versions of Ruby + windows that we won't find out about from testing on latest (or later) Ruby + Windows.

MSP-Greg:
@nateberkopec

Given that Appveyor CI time is currently less than Travis, and given that I fixed the issue, I think leaving 2.2 in AV testing is best for now.

Note that Puma can no longer be installed with RubyGems on Ruby 2.2 (on any platform), as the nio4r dependency will fail.  Ok with bundler install though...

PR #1877, opened 2019-08-02
Merge configuration examples

### LINKED ISSUE #1876: Merge examples/config.rb -> dsl.rb
Currently, docs are duplicated in both locations. Let's just remove the examples file (or delete it all and just leave a comment saying "go read dsl.rb") and merge all that stuff into lib/dsl.rb.

We should probably alphabetize dsl.rb at the same time for easier lookup and reference.

Jesus:
3 commits:

1. Merges documentation from both files.
2. Sorts methods alphabetically.
3. Changes in here should have been part of the first commit, it's just some fixes like typos etc.

Closes #1876. You want it, you got it 🙃.

Just one thing, perhaps alphabetic sorting isn't the best and we should have different sections so all hook methods are grouped together. Lmk if you agree and I'll update the PR.


nateberkopec:
Kaboom!

> and we should have different sections so all hook methods are grouped together.

Which methods are you referring to specifically?

Jesus:
`on_worker_boot`, `after_worker_fork`, etc.

nateberkopec:
Hmmm, you're probably right. Most people use CTRL/CMD-F anyway right so alphabetizing doesn't really matter. Sort by subject as you see fit, maybe just leave the rest in alphabetical order at the end.

nateberkopec:
Just a suggestion though. Organize as you see best.

Jesus:
Ok, if you think that sorting alphabetically doesn't really matter I'd discard this PR and merge a different one where no sorting is done at all.

As you said, people usually use the editor search tool anyway. Also it gets sorted automatically once you build the documentation, ex.: https://www.rubydoc.info/gems/puma/Puma/DSL

Imo, traceability of code is preferable.

nateberkopec:
> Imo, traceability of code is preferable.

Great point.

PR #1882, opened 2019-08-05
Remove confusing `private` call

okuramasafumi:
`private` method doesn't make class methods private.
(To make class methods private, use `private_class_method`.)
That was confusing so it's removed.

### DIFF
--- a/lib/rack/handler/puma.rb
+++ b/lib/rack/handler/puma.rb
@@ -88,7 +88,7 @@ module Rack
           "Verbose"         => "Don't report each request (default: false)"
         }
       end
-    private
+
       def self.set_host_port_to_config(host, port, config)
         config.clear_binds! if host || port
 


nateberkopec:
Hmmm... it looks like we were depending on it being public internally too!

nateberkopec:
What do you think - should we just remove the private and just say that method is public (which it has been...possibly for ever), or re-architect the code to deal with this somehow and make it actually private?

okuramasafumi:
Oh, then I think we should remove the confusing `private` call now and re-architecting should be done in another PR.

nateberkopec:
Works for me @okuramasafumi

okuramasafumi:
@nateberkopec OK, I modified this PR to just remove `private` call.

nateberkopec:
merge

Thank you!

PR #1864, opened 2019-07-28
Detach dead workers to prevent zombie processes

dzunk:
Fixes #1855. When a worker is found dead, detach it from the master process with [`Process.detach`](https://ruby-doc.org/core-2.6.3/Process.html#method-c-detach). This will reap the dead process in a separate thread and prevent zombie children lingering in the process table.

Before:
```
bx puma --debug -w 4 test/rackup/hello-bind.ru
* snip *
[66227] - Worker 0 (pid: 66230) booted, phase: 0
[66227] - Worker 1 (pid: 66250) booted, phase: 0
[66227] - Worker 2 (pid: 66253) booted, phase: 0
[66227] - Worker 3 (pid: 66255) booted, phase: 0
```

```
$ kill 66230
$ kill 66250
$ kill 66253
$ ps aux | grep "[Z]+"
matt             66253   0.0  0.0        0      0 s004  Z+    8:58PM   0:00.00 (ruby)
matt             66250   0.0  0.0        0      0 s004  Z+    8:58PM   0:00.00 (ruby)
matt             66230   0.0  0.0        0      0 s004  Z+    8:58PM   0:00.00 (ruby)
```

After:
```
bx puma --debug -w 4 test/rackup/hello-bind.ru
* snip *
[67422] - Worker 0 (pid: 67468) booted, phase: 0
[67422] - Worker 1 (pid: 67469) booted, phase: 0
[67422] - Worker 2 (pid: 67470) booted, phase: 0
[67422] - Worker 3 (pid: 67471) booted, phase: 0
```
```
$ kill 67468
$ kill 67469
$ kill 67470
$ ps aux | grep "[Z]+"
# nothing
```

PR #1884, opened 2019-08-05
Track and wait on dead workers to prevent zombies

evanphx:
This fixes a bug introduced in #1748 and supersedes the work in PR #1864. 

The bug was the the old logic used `-1` to `Process.waitpid` which would catch any pids, including ones that we accidentally already removed from `@workers` before waiting on them.

This logic tracks the dead workers separately so we're sure to wait on them.

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -24,6 +24,7 @@ module Puma
 
       @phase = 0
       @workers = []
+      @dead_workers = []
       @next_check = nil
 
       @phased_state = :idle
@@ -36,14 +37,11 @@ module Puma
 
       begin
         if RUBY_VERSION < '2.6'
-          @workers.each do |w|
-            begin
-              Process.waitpid(w.pid)
-            rescue Errno::ECHILD
-              # child is already terminated
-            end
-          end
+          @workers.each { |w| w.check_pid }
         else
+          # Clean up the list to make sure we only consider running workers
+          @workers.reject! { |w| w.check_pid }
+
           # below code is for a bug in Ruby 2.6+, above waitpid call hangs
           t_st = Process.clock_gettime(Process::CLOCK_MONOTONIC)
           pids = @workers.map(&:pid)
@@ -130,6 +128,7 @@ module Puma
       end
 
       def term
+        return unless @pid
         begin
           if @first_term_sent && (Time.now - @first_term_sent) > @options[:worker_shutdown_timeout]
             @signal = "KILL"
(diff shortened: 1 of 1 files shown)


dentarg:
Should the test from #1864 be added too?

MSP-Greg:
I looked at this in both 3.12.1 and master with the test from #1864 added.  Some interesting things, along with the issues that have appeared with Ruby 2.6+.

Anyway, if the code in `check_workers` that is causing these issues is replaced with:
```ruby
pids = []
while pid = Process.waitpid(-1, Process::WNOHANG) do
  pids << pid
end
@workers.reject! { |w| w.dead? || pids.include?(w.pid) }
```

Everything passes (with a few other minor changes), including the mentioned test, but Ruby 2.2 fails intermittently on the new zombie test.  It's essentially a simplification of the code that existed in 3.12.1.  but it does use '-1', which you've mentioned you'd prefer not to use...

Top two commits at https://github.com/MSP-Greg/puma/commits/zombies.  I deleted one that may make 2.2 failures less common.

Travis test is: https://travis-ci.org/MSP-Greg/puma/builds/568144883

evanphx:
@MSP-Greg Oh great! Please open a PR against that branch and we'll use it.

MSP-Greg:
@evanphx - see PR #1887.  Testing right now.  Sorry for jumping the gun, but I've been looking at this stuff too much...  Now watch, Appveyor will fail...

MSP-Greg:
@evanphx

When you have a minute, a look at #1836 would be appreciated.

The only real feature added is `no_tlsv1_1`, which was mostly about duplicating the code from `#no_tlsv1`.  A `minimum_protocol_version` or something similar would be better, but I don't think we'll need to add a `no_tlsv1_2` parameter/option anytime soon...

Otherwise, it's a few things that make SSL testing better, as there are some false positives, etc happening.  Obviously, false positives are not good.  Also, I've got a patch for the 'http & OpenSSL 1.1.1 (TLSv1.3)' problem, but I really would rather do it on top of #1836...  Thanks, Greg

evanphx:
@MSP-Greg No problem, merged #1836.

MSP-Greg:
@evanphx 

Thanks.  I'm fixing Travis in #1887 'cluster.rb - no zombies' right now

nateberkopec:
just pasting this comment here for my own review later: https://github.com/puma/puma/pull/1864#issuecomment-518533458

nateberkopec:
Closing in favor of #1887

Issue #1445, opened 2017-11-01
pumactl doesn't default to config/puma/<env>.rb as puma does


x-yuri:
I believe these two links describe the issue in detail, but feel free to ask questions:

https://github.com/puma/puma/blob/v3.10.0/lib/puma/control_cli.rb#L76
https://github.com/puma/puma/blob/v3.10.0/lib/puma/configuration.rb#L195

As a result, I can start `puma` with `bundle exec puma`, or `bundle exec pumactl start`, but can't stop with `bundle exec pumactl stop`. I've got to:

`bundle exec pumactl -F config/puma/production.rb stop`

Or else it asks for pid or control url.

PR #1885, opened 2019-08-05
Check puma config environment file on control_cli

spk:
Ref https://github.com/puma/puma/issues/1445

Allow to pass env for pumactl and mimic config loading with first `config/puma/<env>.rb` then `config/puma.rb` when no config provided from command line


nateberkopec:
:clap: 🚢

PR #1829, opened 2019-07-01
fix #1775

Fudoshiki:
#1775

PR #1886, opened 2019-08-06
fix Binder#close

MSP-Greg:
Previous to #1829, `Binder#close` looped thru @unix_paths and unlinked all of them.  This seems to cause issues with UNIXSockets on some restarts (phased?).

#1829 removed the `unlink`, but also added two lines that create a UNIXSocket and immediately close it.  I've seen the code often, and I'm not sure what effect it has.  Hence, removed it.

Travis test: https://travis-ci.org/MSP-Greg/puma/builds/568149578

macOS Ruby 2.4 seems to behaving issues, seen that more than once today.  Maybe time to jump to xcode11 from xcode10.2?

### DIFF
--- a/lib/puma/binder.rb
+++ b/lib/puma/binder.rb
@@ -50,14 +50,6 @@ module Puma
 
     def close
       @ios.each { |i| i.close }
-      @unix_paths.each do |i|
-        # Errno::ENOENT is intermittently raised
-        begin
-          unix_socket = UNIXSocket.new i
-          unix_socket.close
-        rescue Errno::ENOENT
-        end
-      end
     end
 
     def import_from_env


MSP-Greg:
Forgot to mention that by removing the code, the `Errno::ECONNREFUSED` errors no longer appear in the logs.  The errors aren't causing failures (??), but they do clutter things up...

nateberkopec:
>  The errors aren't causing failures

AFAICT this is because the errors are raised in setup/teardown, not the test itself.

nateberkopec:
nvm that is incorrect, minitest does raise correctly in setup/teardown, we must be doing something gross to rescue it

MSP-Greg:
> doing something gross

I hate it when that happens...

here's the current code I'm speaking of:
https://github.com/puma/puma/blob/b32de94ee689705d4d5009c63c17b73a5060ef7a/lib/puma/binder.rb#L53-L60
That's generating all the output in the test logs, but all the tests are passing with Puma 'handling' the exception.  I thought two possibilities, given that the PR eliminated the `unlink`:
1. Maybe it wasn't needed.
2. Maybe `Errno::ECONNREFUSED` should be added to line 58

???

MSP-Greg:
@Fudoshiki

Sorry for the ping.  If you have a minute...

As mentioned above, #1829 is causing quite a bit of output in the test logs due to a `Errno::ECONNREFUSED` being raised.  One option is this PR, another is to add `Errno::ECONNREFUSED` to the rescue, as shown in PR #1899.  Wondering what your thoughts are?

Fudoshiki:
Need to test without part
```ruby
 @unix_paths.each do |i| 
   # Errno::ENOENT is intermittently raised 
   begin 
     unix_socket = UNIXSocket.new i 
     unix_socket.close 
   rescue Errno::ENOENT 
   end 
 end 
```

If it doesn't affect anything, why should keep this code

nateberkopec:
I'm going to say "no" to this one, as we simply can't work on Puma today on the basis of "the tests are green, therefore this is good." The test suite a) isn't up to that standard and b) it's difficult to write tests for Puma so test coverage will always never be complete.

MSP-Greg:
@nateberkopec

This isn't about CI being 'green' or anything related.  This is about whether the error should be trapped here in `Server#handle_servers`:

https://github.com/puma/puma/blob/8520fddc14c2907b81758574e4ec73539af942e6/lib/puma/server.rb#L419-L426

and hence, causing the conditional at lines 420 - 423 to not execute.  If it should propagate from `Binder#close` to `Server#graceful_shutdown` to `Server#handle_servers`, then it should stay.

If not, then this PR removes the lines causing the error, PR #1899 catches the error.

nateberkopec:
I was taking issue with your reasoning here @MSP-Greg:

> all the tests are passing with Puma 'handling' the exception

and 

>  I've seen the code often, and I'm not sure what effect it has. Hence, removed it.

We simply can't remove code based on this reasoning alone, for the reasons I outlined above.

But your recent comment has explained the change a bit more, so I will re-read and re-consider.

MSP-Greg:
The above concerns the error being raised (which is cluttering up the Travis CI job logs).  UNIXServers created by Puma are shown below.  Note that both are added to `@ios`.

https://github.com/puma/puma/blob/1046165a9c8766ce5c1db8db4d894e200f4c981c/lib/puma/binder.rb#L381-L383

and

https://github.com/puma/puma/blob/1046165a9c8766ce5c1db8db4d894e200f4c981c/lib/puma/binder.rb#L405-L407

The code in question in this PR:

https://github.com/puma/puma/blob/1046165a9c8766ce5c1db8db4d894e200f4c981c/lib/puma/binder.rb#L51-L61

Note that all items in the `@ios` array are closed, and then the code tries to open a UNIXSocket on the just closed io on line 56, which generates an error.

Just saying.  gentle reminder about this or PR #1899.  I think this is probably better...

Fudoshiki:
I agree with @MSP-Greg 

```ruby
@unix_paths.each do |i|
  # Errno::ENOENT is intermittently raised
  begin
    unix_socket = UNIXSocket.new i
    unix_socket.close
  rescue Errno::ENOENT
  end
end
```
This is useless part now

Before #1829 it was just

```ruby
@unix_paths.each do |i|
  # Errno::ENOENT is intermittently raised
  begin
    File.unlink i
  rescue Errno::ENOENT
  end
end
```
We should have removed only this `File.unlink i` to fix socket file missing
IMHO final method should look that

```ruby
def close 
  @ios.each { |i| i.close } 
end 
```

MSP-Greg:
Another gentle reminder about this and the issue with the error propagating as mentioned above in https://github.com/puma/puma/pull/1886#issuecomment-521845940...

MSP-Greg:
38 commits behind master, rebased...

nateberkopec:
Closing in favor of #1945

PR #1887, opened 2019-08-06
cluster.rb - no zombies

MSP-Greg:
Commits:

1. **'cluster.rb - fixup for Cluster#check_workers, removes zombies?'** - reverts problematic code in `Cluster#check_workers` to a simplified version of what was used for several years, thru 3.12.1.  Current code has been revised from previous due to issues with newer Ruby versions, etc.  Simplified version appears to be working correctly.

2. **'test_integration.rb - add zombie test'** - uses test from PR #1864 (not merged, closed)

3. **'travis.yml - misc updates'** - updates a few items to make use of current/newer releases.  Hoping to fix some issues with intermittent failures.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,25 +37,27 @@ rvm:
 matrix:
   fast_finish: true
   include:
-    - rvm: 2.2 
-      dist: trusty 
+    - rvm: 2.2.10
+      dist: trusty
       env: OS="Trusty 14.04 OpenSSL 1.0.1"
-    - rvm: 2.6
+    - rvm: 2.6.3
       dist: bionic
       env: OS="Bionic 18.04 OpenSSL 1.1.1"
     - rvm: ruby-head
       env: jit=yes
     - rvm: 2.4.6
       os: osx
-      osx_image: xcode10.2
-      env: OS="osx xcode10.2"
+      osx_image: xcode11
+      env: OS="osx xcode11"
     - rvm: 2.5.5
       os: osx
-      osx_image: xcode10.2
-      env: OS="osx xcode10.2"
+      osx_image: xcode11
+      env: OS="osx xcode11"
     - rvm: jruby-9.2.7.0
       env: JRUBY_OPTS="--debug" JAVA_OPTS="--add-opens java.base/sun.nio.ch=org.jruby.dist --add-opens java.base/java.io=org.jruby.dist --add-opens java.base/java.util.zip=ALL-UNNAMED --add-opens java.base/java.util=ALL-UNNAMED --add-opens java.base/java.security.cert=ALL-UNNAMED --add-opens java.base/java.security=ALL-UNNAMED --add-opens java.base/java.io=ALL-UNNAMED"
     - rvm: jruby-head
+      dist: bionic
+      env: OS="Bionic 18.04"
 
   allow_failures:
     - rvm: ruby-head
@@ -63,6 +65,8 @@ matrix:
       env: jit=yes
     - rvm: jruby-9.2.7.0
     - rvm: jruby-head
+      dist: bionic
+      env: OS="Bionic 18.04"
 
 env:
   global:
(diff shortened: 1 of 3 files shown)


MSP-Greg:
Forgot to mention - test added here failed on current master...

MSP-Greg:
Re CI, there are two Ruby 2.2.10 jobs, one failed, which seems to be an intermittent failure.  See https://bugs.ruby-lang.org/issues/13239.  There was a recent patch for it, but it was only backported to Ruby 2.3.x.  The failure may involve `#close` calls where the io is already closed.  Adding `closed?` might help...

As to the failure in 2.3.8, it passed several times in my fork.  Not sure what to think of it...

evanphx:
So a better test, rather than relying on `Process.kill` to search, is to just perform a `Process.waitpid(-1, Process::WNOHANG)` and verify that nothing is returned, which will indicate that the server waited on them properly.

lib/puma/cluster.rb
```diff
@@ -227,9 +227,11 @@ def check_workers(force=false)
...
       sleep 1 if any
 
-      @workers.reject! { |w| Process.waitpid(w.pid, Process::WNOHANG) }
-
-      @workers.reject!(&:dead?)
+      pids = []
+      while pid = Process.waitpid(-1, Process::WNOHANG) do
+        pids << pid
+      end
+      @workers.reject! { |w| w.dead? || pids.include?(w.pid) }
```
dzunk: Could move these lines into a `reap_workers` method to match `cull_workers` and `spawn_workers`.

MSP-Greg:
@evanphx 

> So a better test

Updated.  Thanks.

lib/puma/cluster.rb
MSP-Greg: Good suggestion, I'll leave coding style up to otheers.  It's four line, probably only called by the method it's contained in, and any test would involve that method...

MSP-Greg:
Sometime ago [`Binder#stop_workers`](https://github.com/puma/puma/blob/3ea247174e47bd23a7b792b760ac1aaf9a8cb035/lib/puma/cluster.rb#L33-L70) was revised for any issue with Ruby 2.6 and later.  In it is the following loop:
https://github.com/puma/puma/blob/3ea247174e47bd23a7b792b760ac1aaf9a8cb035/lib/puma/cluster.rb#L38-L45

If the first worker in the array is the one needing the longest time to shut down, the code will not check the others until it's waited for the first to complete.  Maybe I've got something wrong, but this seems incorrect.

So, should the code simply use a revised version of the code currently used for Ruby 2.6+ (removing the code for Ruby < 2.6)?

If so, a commit here, or another PR?

.travis.yml
```diff
@@ -37,32 +37,36 @@ rvm:
 matrix:
   fast_finish: true
   include:
-    - rvm: 2.2 
-      dist: trusty 
+    - rvm: 2.2.10
```
nateberkopec: In the future @MSP-Greg, I'd prefer you make changes to travis.yml in separate PRs. That way, they won't get blocked by the corresponding feature/bugfix.

MSP-Greg: I was thinking the same.

While I was testing/debugging, Travis seemed to be acting up, and I wasn't sure if it was affecting the CI results I was seeing.. IOW, didn't just decide to throw it in, I was having failures...

How about I remove it right now & add another PR?

nateberkopec:
@MSP-Greg definitely a separate PR for anything regarding #stop_workers

MSP-Greg:
@nateberkopec

Does the paragraph after the code snippet make sense?

JFYI, I'm looking at some tests where multiple requests that take a varying amount of time to complete are used, and maybe better testing as far as graceful shutdown/phased restart can be done.  Also, externally killing workers and checking respawn.  Not today...

nateberkopec:
> Does the paragraph after the code snippet make sense?

IDK, ask me later on that PR after I release 4.1 😆

PR #1883, opened 2019-08-05
Merge files documenting configuration options (take 2)

Jesus:
- 0 lines of executable code is changed in this PR (only comments).
- Documentation from both files are now combined in `dsl.rb`.
- Fixed some typos and made some small corrections.

Closes #1876

PR #1888, opened 2019-08-06
Remove example config reference from readme

ClikeX:
https://github.com/puma/puma/pull/1883 removed the config example file, but the readme still refered to it. I figure the broken link can be removed.


nateberkopec:
merge

this shows why this change was necessary - why should we have 2 different locations to learn about our configuration DSL?!

PR #1892, opened 2019-08-07
cluster.rb - fixup worker wait logic in #check_workers and #stop_workers

MSP-Greg:
After #1887 was accepted, I was wondering about the wait code in stop_workers.  While thinking about that, I was also bothered by:

1. `Process.waitpid(-1, Process::WNOHANG)` can raise `Errno::ECHILD`, but isn't caught in the current code used in `check_workers`.  This may be an issue with extenal use of `SIGTERM`.

2. @evanphx expressed a preference for never using `wait(-1)`, which I would agree with.

3. Intermittent failures have happened in CI, but always on Ruby versions < 2.6.

4. Code in `stop_workers` is Ruby version specific, but not because of a new feature.  I've never liked that.

JFYI, every job passed in my fork on this.  I may run it a few more times.

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -35,34 +35,19 @@ module Puma
       @workers.each { |x| x.term }
 
       begin
-        if RUBY_VERSION < '2.6'
-          @workers.each do |w|
+        pids = @workers.map(&:pid)
+        loop do
+          pids.reject! do |w_pid|
             begin
-              Process.waitpid(w.pid)
-            rescue Errno::ECHILD
-              # child is already terminated
-            end
-          end
-        else
-          # below code is for a bug in Ruby 2.6+, above waitpid call hangs
-          t_st = Process.clock_gettime(Process::CLOCK_MONOTONIC)
-          pids = @workers.map(&:pid)
-          loop do
-            pids.reject! do |w_pid|
-              begin
-                if Process.waitpid(w_pid, Process::WNOHANG)
-                  log "    worker status: #{$?}"
-                  true
-                end
-              rescue Errno::ECHILD
-                true # child is already terminated
+              if Process.wait(w_pid, Process::WNOHANG)
+                true
               end
+            rescue Errno::ECHILD
+              true # child is already terminated
             end
-            break if pids.empty?
-            sleep 0.5
           end
-          t_end = Process.clock_gettime(Process::CLOCK_MONOTONIC)
-          log format("    worker shutdown time: %6.2f", t_end - t_st)
+          break if pids.empty?
+          sleep 0.2
         end
       rescue Interrupt
         log "! Cancelled waiting for workers"
@@ -227,11 +212,13 @@ module Puma
       # during this loop by giving the kernel time to kill them.
       sleep 1 if any
 
-      pids = []
-      while pid = Process.waitpid(-1, Process::WNOHANG) do
-        pids << pid
+      @workers.reject! do |w|
+        begin
+          Process.wait(w.pid, Process::WNOHANG)
+        rescue Errno::ECHILD
+          true # child is already terminated
+        end
       end
-      @workers.reject! { |w| w.dead? || pids.include?(w.pid) }
 
       cull_workers
       spawn_workers


lib/puma/cluster.rb
```diff
@@ -227,11 +212,13 @@ def check_workers(force=false)
       # during this loop by giving the kernel time to kill them.
       sleep 1 if any
 
-      pids = []
-      while pid = Process.waitpid(-1, Process::WNOHANG) do
-        pids << pid
+      @workers.reject! do |w|
```
nateberkopec: this code now seems similar enough to be extracted into a new method, yes?

lib/puma/cluster.rb
```diff
@@ -35,34 +35,19 @@ def stop_workers
...
+            rescue Errno::ECHILD
+              true # child is already terminated
             end
-            break if pids.empty?
-            sleep 0.5
           end
-          t_end = Process.clock_gettime(Process::CLOCK_MONOTONIC)
-          log format("    worker shutdown time: %6.2f", t_end - t_st)
+          break if pids.empty?
+          sleep 0.2
```
nateberkopec: Why do we sleep/wait for everyone to die here but not on line 215?

lib/puma/cluster.rb
```diff
@@ -35,34 +35,19 @@ def stop_workers
...
+              if Process.wait(w_pid, Process::WNOHANG)
+                true
               end
+            rescue Errno::ECHILD
+              true # child is already terminated
             end
-            break if pids.empty?
-            sleep 0.5
           end
-          t_end = Process.clock_gettime(Process::CLOCK_MONOTONIC)
```
nateberkopec: not a fan of logging? 😆

lib/puma/cluster.rb
MSP-Greg: I did think about that, but see below.

lib/puma/cluster.rb
MSP-Greg: `stop_workers` is just that, it needs to wait for **all the workers to stop**.

`check_workers` is just that, it checks for workers (possibly stopped externally)  that have been `SIGTERM`'d, and removes then from the array, then moves on.  It's not meant to pause, as it essentially runs continuously.

lib/puma/cluster.rb
MSP-Greg: I don't know.  I think I added that recently.  I can log/debug lines if you'd like...

lib/puma/cluster.rb
MSP-Greg: One thing I meant to check into is that if a worker is stopped externally, I thought that `check_workers` might possibly wait up to `Const::WORKER_CHECK_INTERVAL` (5 seconds) before it runs `wait` and then respawns?  Is that terrible, or ok?  Or is it running before then by something setting the `force` parameter to true?

Make sense?

A little distracted right now...

lib/puma/cluster.rb
evanphx: This logic seems like the same one that causing zombies because we will delete from `@workers` if a process is dead due to signal but the process is still running. When it actually does exit, we won't call wait on it, so this reintroduces the bug.

MSP-Greg: > if a process is dead due to signal

I did 'turn around' things so that wait was called before `dead?` was checked, but I've kind of left the `dead?` check in the because I thought it was previously used for a long time.  But, `wait` may not have been using `WNOHANG` then.  Also, I need to review how & when `@dead` is set, as I'm not clear on that.

Regardless, what you're saying makes sense to me, should I remove the ` || w.dead?` code on line 220?

MSP-Greg: @evanphx

As above, I removed the `|| w.dead?` code in my fork, and it passed.  Would you prefer that?  I think I would...

I don't think there are any tests that create a situation where a worker needs a 'graceful' shutdown and then externally closes it, and the waits to see if a zombie exists.  Maybe a bit messy...

MSP-Greg: I removed `|| w.dead?`

evanphx: This changes the logic so we'll need to valid it: namely should workers be respawned if they are marked as dead but the process still exists? Prior to this change, it did that. But with this change, we don't respawn dead workers until the process actually goes away. We need to decide which is right.

MSP-Greg: IDN.  If this does cause an issue, maybe there needs to be a  'pid array', which is just used for `wait` calls?  Also remove from workers if `wait` returns the pid?

MSP-Greg: I haven't had time to look at `@dead`.  If a worker pid is killed (SIGTERM) outside of Puma, will `@dead` be true?

evanphx: Using a separate array is what I started to do in that other PR. I don't think it's probably necessary if we remove `w.dead?` from here, but I'm mentioning it just so that we've at least thought about it.

evanphx: I looked through the code and it's fine. The main effect will be that there will be a very small delay perhaps before spawning a replacement worker and that's ok. The dead status code is sent right before the worker exits anyway.

MSP-Greg: Thanks for looking at it.  Next major computer purchase is a system for Ubuntu/*nix something...

nateberkopec: @MSP-Greg or just rent an AWS instance by the hour!

MSP-Greg: @evanphx 

After finally noticing the `reject!` typo (funny how `reject` doesn't work), I think I've got the code for a 'separate array' working.  If you'd prefer it, let me know.

Patch: https://github.com/MSP-Greg/puma/commit/f141b027f14421df0d45c29af43b6af17c38560d
Travis: https://travis-ci.org/MSP-Greg/puma/builds/569122723

@nateberkopec you're on macOS I assume.  You're 90% of the way there already.  I might as well be in the next solar system...

MSP-Greg:
@evanphx, @nateberkopec

This PR an some of the issues involving worker 'proper exit' do need more tests.  One thing I wonder is if there's a need for a `Worker#term?` method to indicate whether `Worker#term` has been called.

For worker 'w', we have logic like the following used a `Worker.each` loop:
```ruby
begin
  Process.wait(w.pid, Process::WNOHANG)
rescue Errno::ECHILD
  true # child is already terminated
end
```

With the addition of the `#term?` flag/attribute/method
```ruby
begin
  if Process.wait(w.pid, Process::WNOHANG)
    true
  else
    w.term if w.term?
    nil
  end
rescue Errno::ECHILD
  true # child is already terminated
end
```

MSP-Greg:
Closing in favor of #1908

PR #1794, opened 2019-05-09
Bundler: switch from deprecated with_clean_env to with_original_env

bensie:
`with_clean_env` blows away all "BUNDLE_*" environment variables without consideration for the ones that you may have wanted to retain. It's also deprecated. For example, starting your server with `BUNDLE_GEMFILE=Gemfile.rails6` would delete this environment variable on app restart when using `prune_bundler`, when all you really want is to restore the environment to what it was before Bundler was activated.

The [environment preserver](https://github.com/bundler/bundler/blob/master/lib/bundler/environment_preserver.rb) provided by Bundler goes to greater lengths to avoid stomping on things that are intended to be preserved and `with_original_env` is a beautiful thing.

PR #1893, opened 2019-08-08
Preserve BUNDLE_GEMFILE and add a test for it

seven1m:
Starting your server with BUNDLE_GEMFILE=Gemfile.rails6 would delete this environment variable on app restart when using `--prune-bundler`, when all you really want is to restore the variable to what it was before Bundler was activated.

I worked with @bensie on this and this PR should supersede #1794.


test/test_integration.rb
```diff
@@ -239,6 +239,18 @@ def test_restart_closes_keepalive_sockets_workers
     assert_equal "Hello World", new_reply
   end
 
+  def test_restart_with_prune_bundler_keeps_bundle_gemfile_env
+    gemfile_path = File.expand_path("../Gemfile", __dir__)
+    duplicate_gemfile_path = gemfile_path + ".duplicate"
+    FileUtils.cp(gemfile_path, gemfile_path + ".duplicate")
```
seven1m: I'm not in love with the idea of doing a file copy in a test like this, but I thought this approach was preferable to having a duplicate Gemfile that is checked into git. I'm open to other suggestions though!

nateberkopec: I'm not sure I agree. I don't like having tests create/destroy files, especially in the root dir, as they tend to be created and then I ctrl-c out of my tests and the file is never deleted.

Why not just put a Gemfile somewhere appropriate in the test/ directory, like the other fixtures we have for Rackup files?

nateberkopec: It would remove ~3 lines from this test, at least.

seven1m: The Gemfile has to be a copy of the real Gemfile. If it is missing something or gets outdated, then Puma won’t start in the test. Also, because the Gemfile contains the directive `gemspec`, it must live in the root of the repo next to the puma.gemspec file.

seven1m: I do understand and agree with your sentiment about Ctrl-C, for sure!

seven1m: I have another idea for writing this test... I'll try to push another commit shortly. Thanks for looking!

seven1m: @nateberkopec please take another look at the PR. @bensie and I were able to make the test more generic by using PATH and `Bundler.with_original_env` rather than focusing solely on BUNDLER_GEMFILE. We believe this makes the intention clearer and the code simpler!

nateberkopec:
Hm, I wonder if we'll need to major version bump for this as people may have been relying on the old behavior of Bundle env variables getting wiped clean.

bensie:
It's hard to imagine the scenario where you'd depend on that behavior, since there's no way to inject the Bundler variables you _do_ want during the Puma reload and you're left with inconsistent state depending on whether you `reload` or `restart`.

But yeah it's up to you, I guess sometimes a bug fix is technically breaking 🤷‍♂

nateberkopec:
I think this is good to go, but I'm going to wait on merging it until it becomes more clear to me what the next 2-3 months of release schedule looks like.

If required immediately, I suggest using a fork until that time.

nateberkopec:
Actually it looks like you have a small issue to fix - I think your test needs to be skipped on Windows.

test/test_integration.rb
```diff
@@ -240,6 +240,7 @@ def test_restart_closes_keepalive_sockets_workers
   end
 
   def test_restart_with_prune_bundler_keeps_bundler_env
+    skip NO_FORK_MSG unless HAS_FORK
```
nateberkopec: Do this test actually need to use cluster mode/forking? If we don't need to, we should remove it and allow the test to run on JRuby.

test/test_integration.rb
```diff
@@ -240,6 +240,7 @@ def test_restart_closes_keepalive_sockets_workers
...
     initial_reply, new_reply = restart_server_and_listen("-q -w 2 --prune-bundler test/rackup/hello-bundler-env.ru", "PATH" => "hello-bundler:#{ENV["PATH"]}", "BUNDLER_ORIG_PATH" => nil)
```
nateberkopec: as in do you actually need "-w 2" here or not.

seven1m: The `-w 2` should not be needed, but I cannot get the build to succeed on AppVeyor. The errors seem to be compile errors? Any tips here would be appreciated! 🙏

MSP-Greg: At present, Puma isn't installed with bundler on Appveyor.  Sometime soon, but probably after a determination about using GitHub Actions for CI...

zw963:
thank you a lot, i will test this when merge this PR merge into master.

nateberkopec:
OK, problem: for me with Bundler 1.17.2, this test passes with the original implementation.

I tried a couple of other env variables, everything is passing.

seven1m:
Ahhh! You're completely right. I don't even know what happened, but I really thought I did this right. I *know* this one-line change fixes issues in our applications -- we use this patch in production to allow us to reload our apps while keeping our environment variables from getting wiped out.

I really struggled to write a good test to prove this, and I guess I didn't even succeed in writing a proper test. 🤦‍♂ 

I'm open to suggestions!

If you're ok with it, can we leave this open and I'll work on fixing the test...?

nateberkopec:
Oh of course we can leave it open. If it fixes _something_ then the state is changed somehow, we just need to document what that is with a test.

dentarg:
@nateberkopec @seven1m I think the test need to test with an `BUNDLE_` environment variable, it is only those that are wiped, see the the `clean_env` implementation in Bundler (it is the same in 1.17.2 and 2.0.2)

* https://github.com/bundler/bundler/blob/v1.17.2/lib/bundler.rb#L283-L306
* https://github.com/bundler/bundler/blob/v2.0.2/lib/bundler.rb#L283-L306

(In Bundler 2.1.0 that method has been renamed to [`unbundled_env`](https://github.com/bundler/bundler/blob/v2.1.0.pre.3/lib/bundler.rb#L340-L361))

nateberkopec:
> I think the test need to test with an BUNDLE_ environment variable

Pretty sure I tried that and it still passed with original implementation, but now I can't remember.

nateberkopec:
@seven1m Did you see @dentarg's comment? It's possible that would help get this test working correctly. 

We need a test that's not passing on master for this to work. Maybe #2018 will help us to figure out what needs to be tested here?

seven1m:
Thanks for the tip, I'll take a look at this and see what I can do today.

seven1m:
@nateberkopec I found a way to test this finally. Please let me know what you think!

test/test_integration_cluster.rb
```diff
@@ -133,6 +133,32 @@ def test_stuck_phased_restart
     worker_respawn { |phase0_worker_pids| Process.kill :USR1, @pid }
   end
 
+  # It does not wipe out BUNDLE_GEMFILE et al
+  def test_usr2_restart_preserves_bundler_environment
+    skip_unless_signal_exist? :USR2
```
nateberkopec: I don't think this skip is necessary. `preload_bundler` can be triggered by starting any Puma cluster that doesnt `preload_app!`.

nateberkopec: Then again, I'm not sure how you test it without a USR2 restart, so maybe ignore me?

nateberkopec:
It looks like the 2.2 test run is hanging somewhere inside your test, probably waiting on a response from the server or the server to boot. I would run the test by itself using 2.2 locally and see if you can figure out what line of the test it's hanging on.

seven1m:
I am apparently unable to reproduce the issue seen on the Travis build. 😭 

I was able to create an environment fairly close to Travis using Docker, here is the `Dockerfile` for posterity:

```
FROM ubuntu:14.04

RUN apt-get update
RUN apt-get install -y -q build-essential libreadline-dev libssl-dev wget curl git gnupg2
RUN gpg --keyserver hkp://pool.sks-keyservers.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3 7D2BAF1CF37B13E2069D6956105BD0E739499BDB
RUN curl -sSL https://get.rvm.io | bash -s stable
RUN /bin/bash -l -c ". /etc/profile.d/rvm.sh && rvm install 2.2.10 && rvm --default use 2.2.10"
RUN /bin/bash -l -c ". /etc/profile.d/rvm.sh && gem update --system 2.7.10 --no-document && gem install bundler -v '< 2'"
WORKDIR /puma
COPY . /puma
RUN /bin/bash -l -c ". /etc/profile.d/rvm.sh && bundle install"
```

Moving the test into its own file fixed the issue in my environment, but it seems that did not fix the problem on Travis.

Unfortunately, I've just about reached the limit of my time to work on this project. I hate to give up on it, but I'm really out of ideas and time.

If anyone has any brilliant ideas for fixing the Travis build, let me know!

seven1m:
Update: I got all the builds to pass, but I'm nervous it's the wrong answer. 😬 

I'm still working on it...

bensie:
🍿 

@nateberkopec if the outstanding issue is really just Ruby 2.2-specific and you're thinking this PR can only be merged for Puma 5.x, any shot of removing support for 2.2 then? As of March 2018, 2.2 is EOL.

nateberkopec:
We base Ruby version support on usage, not on EOL, so there’s a bit of lag there.

lib/puma/launcher.rb
```diff
@@ -297,8 +297,10 @@ def prune_bundler
 
       log '* Pruning Bundler environment'
       home = ENV['GEM_HOME']
+      bundle_gemfile = ENV['BUNDLE_GEMFILE']
```
nateberkopec: Oh, what happened?

test/test_preserve_bundler_env.rb
```diff
@@ -0,0 +1,33 @@
...
+require_relative "helpers/integration"
+
+class TestPreserveBundlerEnv < TestIntegration
+  def setup
+    skip NO_FORK_MSG unless HAS_FORK
+    super
+  end
+
+  # It does not wipe out BUNDLE_GEMFILE et al
+  def test_usr2_restart_preserves_bundler_environment
```
nateberkopec: I thought about the skip here more and I guess it makes sense that while the code technically _runs_ on non-usr2 restarts it doesn't actually have a testable effect unless you're doing a usr2 restart. This test should be skipped if platform support isn't present (the way you had it before).

lib/puma/launcher.rb
seven1m: Something about the environment in the Travis build was causing `Bundler.with_original_env` to make Puma load in an endless loop. Puma would load, it would call `Bundler.with_original_env`, then call `Kernel.exec('puma-wild', ...)`, then the loop would start over again.

You can see that loop in [this failed build](https://travis-ci.org/puma/puma/jobs/650170319?utm_medium=notification&utm_source=github_status).

I could not reproduce this issue locally, even with the same version of Ubuntu, same version of Ruby, and same version of RVM. There are probably other pieces in the environment on the Travis build that I am not aware of and did not fully replicate.

Testing this required, as you can see, pushing many commits to track down the problem.

The `prune_bundler` method is recursive, as you know, and it relies on the line `return unless defined?(Bundler)` as its terminal case. Something about the Travis environment (I suspect it's RVM) was causing Bundler to be loaded every time the method recursed.

As I am out of time to work on this, I went back to the hacky solution that copies the needed env var (`BUNDLE_GEMFILE`) back into the environment after we run `Bundler.with_clean_env`.

It's not my favorite solution, at all. But it meets our needs and should not break anything in other people's environments.

nateberkopec:
Gonna test this again locally against master, but from the code perspective it LGTM.

History.md
```diff
@@ -10,6 +10,7 @@
 * Bugfixes
   * Your bugfix goes here (#Github Number)
   * Windows update extconf.rb for use with ssp and varied Ruby/MSYS2 combinations (#2069)
+  * Preserve `BUNDLE_GEMFILE` env var on restart (#1893)
```
nateberkopec: OK last request! Let's change this to "... env var when using prune_bundler".

nateberkopec:
Fails on master! One more change to the changelog and then we hit the big green button.

lib/puma/launcher.rb
AlexWayfer: > I could not reproduce this issue locally, even with the same version of Ubuntu, same version of Ruby, same version of Bundler, and same version of RVM. There are probably other pieces in the environment on the Travis build that I am not aware of and did not fully replicate.

Travis is pretty dirty in my experience, so I've switched my repos to other CIs. You can consider something similar too.

nateberkopec: We've moved almost everything to Github Actions, but Ruby 2.2 can't be moved (yet?).

AlexWayfer: Ruby 2.2 is no more maintained, even security, but I understand that you want to support its users. But I personally consider its usage dangerous and prefer to drop support from gems.

dentarg: Since https://github.com/ruby/setup-ruby/releases/tag/v1.14.0 it is possible to use Ruby 2.2 on GitHub Actions... but it could be that there's something else missing for the Puma test setup (e.g. that version wasn't built with the OpenSSL version we want to test Puma with, I'm sure @MSP-Greg knows about this and maybe even have tried it, I don't have the time to double check it now)

seven1m: I should note that I believe Travis' environment actually saved us from shipping a bug here. *Something* in that particular environment was causing the Bundler constant to be loaded earlier than anticipated. I suspect it is some RVM magic.

dentarg: One difference between Travis and GitHub Actions is (was) the `bundle install` command

Travis does the following (https://travis-ci.org/puma/puma/jobs/656438353#L885)

    bundle install --jobs=3 --retry=3 --path=${BUNDLE_PATH:-vendor/bundle}

Before #2126 (5 days ago), `--path` wasn't used on GitHub Actions, now it is:

https://github.com/puma/puma/blob/8f282206dcdd07d765553179fe31d61acfd08dda/.github/workflows/ruby.yml#L38-L39

Haven't checked if it matters for this particular discussion but thought it was worth mentioning here

PR #1894, opened 2019-08-08
cluster.rb - add @worker_pids, fixup Process.wait logic in #check_workers and #stop_workers

MSP-Greg:
Adds @worker_pids array, which is used for Process.wait loops, allowing @workers to be purged of dead Workers

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -23,7 +23,18 @@ module Puma
       super cli, events
 
       @phase = 0
+
+      # @workers - array of Workers
+      #   Used to determine when new workers should be spawned.
+      #
+      # @worker_pids - array of Worker pids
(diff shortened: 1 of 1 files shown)


MSP-Greg:
This PR is comparable to #1892, but allows Puma to spawn a new worker when a current worker is considered dead, but hasn't yet exited.  Based on discussion in #1892.

MSP-Greg:
Do not merge this.  @workers is used by much of the code, so I believe it should never have a 'dead' worker removed, which the current PR does.  But, some of the code (possibly `#cull_workers`), should be taking into account dead workers in one form or another.

It's complicated.  I need to think about it more...  But, if a worker is only spawned when an 'old' worker has exited, PR #1892 is fine.  If we want a worker to be spawned when an 'old' worker is SIGTERM'd, that will require more work...

PR #1900, opened 2019-08-09
test_integration.rb - fix intermittent test_no_zombie_children

MSP-Greg:
When run with other parallel tests, previous code would inteermittently return pid's not related to the test.

Skipped on Ruby 2.2.  See:
https://bugs.ruby-lang.org/issues/13239 can't modify frozen IOError (RuntimeError)

I think a fix was backported to 2.2, but it introduced another bug (the one above).  It was then to late to backport to 2.2.

### DIFF
--- a/test/test_integration.rb
+++ b/test/test_integration.rb
@@ -377,6 +377,7 @@ class TestIntegration < Minitest::Test
 
   def test_no_zombie_children
     skip NO_FORK_MSG unless HAS_FORK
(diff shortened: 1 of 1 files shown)


nateberkopec:
Makes sense. Should we call it something other than zombies tho, since they're not actually zombie process entries?

MSP-Greg:
Sure.  What would you prefer?

nateberkopec:
Added some stuff

Issue #1902, opened 2019-08-10
testing - rackup folder - sleep


MSP-Greg:
There are several files in the test/rackup folder, and a few make use of `sleep`, with varying times.

The same thing may be done when 'app' lambdas are declared in a test method.  I've worked locally and in my fork with a file similar to:

```ruby
run lambda { |env|
  dly = (env['REQUEST_PATH'][/\/(d+)sec/,1] || '0').to_i
  sleep dly
  [200, {}, ["Delayed #{dly}"]]
}
```
as a replacement, and also used a constant lambda defined in `helper.rb'.

This came about when thinking of tests for external KILL/TERM, graceful shutdown, phased-restarts, etc.  A way to clearly create requests with specific time delays seems needed.

The above needs a GET of '\/<x\>sec', where \<x\> is an integer, and replies with 'Delayed \<x>\'.

Hence, two questions:

1. Stupid idea or otherwise?
2. If a 'good' idea', what to name the file, the constant, the GET and the reply?  Kind of sensitive to ESL issues, as contributors are worldwide...

nateberkopec:
Makes sense. 

I think we could add it a new "helpers/apps.rb" file that's required by test/helper. Namespace something like `ExampleRackApps::SLEEP`. I think you should just use the world `sleep` as it describes what the app actually does, which is call sleep.

dentarg:
Can this issue be closed now that #1903 was merged?

MSP-Greg:
Ok with me...

PR #1903, opened 2019-08-11
tests - clean up - three commits

MSP-Greg:
1. Commit 'tests - cleanup rackup folder, add sleep.ru, add helpers/app.rb' - Cleans up rackup folder, adds sleep.ru, which allows setting request 'sleep' time via the URI.  Add companion 'app' in `TestApps::SLEEP`.  See Issue #1902.

2. Commit 'tests - change nextport to UniquePort.call, remove nextport, mutex' - helper.rb currently has two methods to select an unused port.  Removed the `nextport` method, and added a mutex to `UniquePort.call` for parallel testing.

3. Commit  'test_integration.rb - make UNIXSocket file names unique` - for parallel testing

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -16,6 +16,7 @@ require "timeout"
 require "minitest/autorun"
 require "minitest/pride"
 require "minitest/proveit"
+require_relative "helpers/apps"
 
 $LOAD_PATH << File.expand_path("../../lib", __FILE__)
 Thread.abort_on_exception = true
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -44,10 +44,10 @@ end
 
 module UniquePort
   @port  = 3211
+  @mutex = Mutex.new
 
   def self.call
-    @port += 1
-    @port
+    @mutex.synchronize { @port += 1 }
   end
 end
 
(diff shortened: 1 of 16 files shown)


MSP-Greg:
Travis failed on Xcode11 Ruby 2.4.6.  Most of the intermittent failures seem to involve this particular job.

Anyone know what the standard Ruby version is with newer macOS versions?  Maybe we should change it to 2.6.3?

MSP-Greg:
All three Travis builds passed.  There was a single ruby 2.3.3p222 (32 bit) job that failed on AppVeyor, which is an old version of 2.3 (most recent is 2.3.8).  I have seen the method fail before.  Human readable error text:

```
Program: C:\Ruby23\bin\ruby.exe File: ext/puma_http11/http11_parser.rl, Line 121
Expression: parser->nread <= len && "nread longer than length"

This application has requested the Runtime to terminate it in an unusual way.
Please contact the application's support team for more information.

rake aborted!

Command failed with status (3)

Tasks: TOP => default => win_test
(See full trace by running task with --trace)
Http11ParserTest#test_horrible_queries =
```

nateberkopec:
Its-So-Beautiful

Issue #1904, opened 2019-08-11
[Bug] TestIntegration#test_phased_restart_via_pumactl, proper Worker shutdown


MSP-Greg:
While creating PR #1903, which is directed towards test stability, I also moved some timers around in the following test method:

https://github.com/puma/puma/blob/da6a9f9e57575af6601f21e214cfc32a56161247/test/test_integration.rb#L159-L207

This test method currently tests whether at least one worker has been rebooted via ‘phased-restart’.

If one inspects the test logs (see https://travis-ci.org/puma/puma/pull_requests, builds 3706 – 3708), there are several jobs where the above test takes a bit over 40 seconds.  I believe this would be considered a bug.

In the test, `delay` is set to 40, which is the time the single http request will take to complete.  The test also sets `DSL#worker_shutdown_timeout` to 2 seconds (default is 30), and `Const::WORKER_CHECK_INTERVAL` is the normal 5 seconds.  So a test time of 40 seconds indicates that Puma is not sending a ‘SIGKILL’ to the worker when it should be.

If this is the case, Puma may also not be sending ‘SIGKILL’ to workers shutdown externally, which is sometimes done when external indicators show that the worker (probably actually the ‘app’) is misbehaving/stuck/frozen.

In general, one could define four states for a worker:

A.	Booted and running
B.	Sent a SIGTERM
C.	Exited
D.	Process is removed from the ‘process table’ via `Process.wait`

Some issues have been specific to D, others indicate that the worker was in state B, but may have never proceeded to C (or D).

I believe the bug involves the current code in `Cluster#check_workers`.  It removes workers in the `@workers` array which are `dead?`, which should correspond to state B, but by removing them, they may not move to states C or D.  This was brought up by @evanphx, and after reviewing the code, I agree.

The code in PR #1892 fixes the above issue, but in a [comment](https://github.com/puma/puma/pull/1892#issuecomment-519943855), I mentioned that I’m not sure 1892 works correctly with externally SIGTERM’d workers, as they may need the fix mentioned, especially if they are misbehaving/stuck/frozen.  I have a branch in my fork with a fix for that.

I very possibly have this all wrong.  If not, merging PR #1903 would be helpful as request timers are needed for additional tests...

MSP-Greg:
One other thought, regarding docs.  If I have things correct, if a Worker is 'misbehaving/stuck/frozen', whether it is shut down by Puma or externally, it will take approx

[...]

PR #1908, opened 2019-08-13
Fixes Cluster worker shutdown/restart

MSP-Greg:
Hopefully fixes issues with correct handling of workers when:

1. Using phased-restart
2. externally SIGTERM'd workers

'Misbehaving' workers may not have been previously handled correctly.

Commit 'cluster.rb - fixup worker wait in Cluster, add Worker#term?' - relatively simple code that changes the handling of workers, making sure to trigger SIGKILL if `@options[:worker_shutdown_timeout]` has been exceeded, correct `wait` logic, and does not remove `dead?` workers from Cluster `@workers` array

Commit 'Adds two tests for worker SIGTERM/respawn and phased-restart' - adds two tests for the above.  They're somewhat brittle, and may fail, but not often.  They take about 30 seconds each.

I added the tests to current master and on top of PR #1892, and both of the new tests failed in every job.

See PR #1892 and Issue #1904 for additional discussion.

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -35,34 +35,10 @@ module Puma
       @workers.each { |x| x.term }
 
       begin
-        if RUBY_VERSION < '2.6'
-          @workers.each do |w|
-            begin
-              Process.waitpid(w.pid)
-            rescue Errno::ECHILD
-              # child is already terminated
-            end
-          end
-        else
-          # below code is for a bug in Ruby 2.6+, above waitpid call hangs
-          t_st = Process.clock_gettime(Process::CLOCK_MONOTONIC)
-          pids = @workers.map(&:pid)
-          loop do
-            pids.reject! do |w_pid|
-              begin
-                if Process.waitpid(w_pid, Process::WNOHANG)
-                  log "    worker status: #{$?}"
-                  true
-                end
-              rescue Errno::ECHILD
-                true # child is already terminated
-              end
-            end
-            break if pids.empty?
-            sleep 0.5
-          end
-          t_end = Process.clock_gettime(Process::CLOCK_MONOTONIC)
-          log format("    worker shutdown time: %6.2f", t_end - t_st)
+        loop do
+          wait_workers
+          break if @workers.empty?
+          sleep 0.2
         end
       rescue Interrupt
         log "! Cancelled waiting for workers"
@@ -99,6 +75,7 @@ module Puma
         @last_checkin = Time.now
         @last_status = '{}'
         @dead = false
+        @term = false
       end
 
       attr_reader :index, :pid, :phase, :signal, :last_checkin, :last_status, :started_at
@@ -120,6 +97,10 @@ module Puma
         @dead = true
       end
 
+      def term?
+        @term
+      end
+
       def ping!(status)
         @last_checkin = Time.now
         @last_status = status
@@ -134,9 +115,9 @@ module Puma
           if @first_term_sent && (Time.now - @first_term_sent) > @options[:worker_shutdown_timeout]
             @signal = "KILL"
           else
+            @term ||= true
             @first_term_sent ||= Time.now
           end
-
           Process.kill @signal, @pid
         rescue Errno::ESRCH
         end
@@ -227,12 +208,7 @@ module Puma
       # during this loop by giving the kernel time to kill them.
       sleep 1 if any
 
-      pids = []
-      while pid = Process.waitpid(-1, Process::WNOHANG) do
-        pids << pid
-      end
-      @workers.reject! { |w| w.dead? || pids.include?(w.pid) }
-
+      wait_workers
       cull_workers
       spawn_workers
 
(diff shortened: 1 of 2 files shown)


test/test_integration.rb
```diff
@@ -72,8 +72,12 @@ def start_forked_server(argv)
 
   def stop_forked_server(pid)
     Process.kill(:TERM, pid)
-    sleep 1
-    Process.wait2(pid)
+    sleep 2
+    begin
+      Process.wait2(pid)
+    rescue Errno::ECHILD
+      15
```
dentarg: `15`?

MSP-Greg: @dentarg

Being a windows type, I can't test this locally.

The change above in `stop_forked_server` was done because I did see infrequent (and false) Travis failures from it.  But, it isn't really part of this PR, so I reverted it.

It is wrong (should be `[nil, 15]`), but given the 'green' builds, it shows how infrequent the failures were...  Thanks.

MSP-Greg:
Tests updated to make use of `after_worker_fork` hook to track worker creation.  Faster and more reliable...

test/test_integration.rb
```diff
@@ -389,7 +389,207 @@ def test_no_zombie_children
...
+    port = UniquePort.call
+    workers_booted = 0
+
+    conf = Puma::Configuration.new do |c|
+      c.bind "tcp://#{host}:#{port}"
+      c.threads 1, 1
+      c.workers 2
+      c.worker_shutdown_timeout 2
+      c.app TestApps::SLEEP
+      c.after_worker_fork { |idx| workers_booted += 1 }
```
nateberkopec: Why do we need a new mechanism for determining if the process has finished booting? We already have two methods.

test/test_integration.rb
```diff
@@ -389,7 +389,207 @@ def test_no_zombie_children
...
+    end
+
+    l = Puma::Launcher.new conf, :events => @events
+
+    t = Thread.new do
+      Thread.current.abort_on_exception = true
```
nateberkopec: Don't need to set, this is already set on all thread in helper.rb

test/test_integration.rb
```diff
@@ -389,7 +389,207 @@ def test_no_zombie_children
...
+        Process.wait(pid, Process::WNOHANG)
+        pid
+      rescue Errno::ECHILD
+        nil # child is already terminated
+      end
+    }.compact
+
+    Thread.kill worker0
+    Thread.kill worker1
+    http0   = nil
```
nateberkopec: what's the point of setting all of these to nil?

test/test_integration.rb
```diff
@@ -389,7 +389,207 @@ def test_no_zombie_children
...
+    assert_operator (Time.now.to_f - start_time).round(2), :<, 32
+
+    msg = "old_pids #{old_pids.inspect}  new_pids #{new_pids.inspect}  old_waited #{old_waited.inspect}"
+    assert_equal 2, new_pids.length, msg
+    assert_equal 2, old_pids.length, msg
+    assert_equal new_pids, (new_pids - old_pids), "#{msg}\nBoth workers should be replaced"
+    assert_empty old_waited, msg
+  end
+
+  def test_worker_phased_restart
```
nateberkopec: These tests are exactly the same except for 1 line, could we clean that up a bit?

test/test_integration.rb
```diff
@@ -389,7 +389,207 @@ def test_no_zombie_children
...
+    Thread.kill worker0
+    Thread.kill worker1
+    http0   = nil
+    http1   = nil
+    cluster = nil
+
+    l.stop
+    assert_kind_of Thread, t.join, "server didn't stop"
+
+    assert_operator (Time.now.to_f - start_time).round(2), :<, 32
```
nateberkopec: maybe add together what you're looking for explicitly here: 

```ruby
assert_operator (Time.now.to_f - start_time).round(2), :<, (conf.worker_shutdown_timeout + 30)
```

I forgot, where does the 30 come from?

test/test_integration.rb
```diff
@@ -389,7 +389,207 @@ def test_no_zombie_children
...
+
+    msg = "old_pids #{old_pids.inspect}  new_pids #{new_pids.inspect}  old_waited #{old_waited.inspect}"
+    assert_equal 2, new_pids.length, msg
+    assert_equal 2, old_pids.length, msg
+    assert_equal new_pids, (new_pids - old_pids), "#{msg}\nBoth workers should be replaced"
```
nateberkopec: imo this would be slightly clearer:

```ruby 
assert_empty new_pids & old_pids
```

lib/puma/cluster.rb
```diff
@@ -227,11 +220,18 @@ def check_workers(force=false)
       # during this loop by giving the kernel time to kill them.
       sleep 1 if any
 
-      pids = []
-      while pid = Process.waitpid(-1, Process::WNOHANG) do
-        pids << pid
+      @workers.reject! do |w|
```
nateberkopec: This looks the same as the looped one, so we should extract a new method.

MSP-Greg:
@nateberkopec

I rebased and updated most of the items you mentioned.  Some of the work in the test file could be shared by other tests, saving that for another PR...

PR #1911, opened 2019-08-15
Gemspec Changelog URI

sj26:
Add changelog uri to spec metadata so that there is a "Changelog" link on https://rubygems.org/gems/puma. I went looking for this when comparing an update recently.


nateberkopec:
Nice, thanks.

PR #1915, opened 2019-08-17
Msp greg zombie test fix intermittent

nateberkopec:
+ remove unused signal helper

Replaces #1900


MSP-Greg:
@nateberkopec

Thanks.

JFYI, my ISP is having issues, started Wednesday, will continue until at least next Friday, 23-Aug.  In a cellular data dead zone unless I stand on the roof...

PR #1920, opened 2019-08-20
Even accept

nateberkopec:
Attempts to spread out work more appropriately. More loaded workers will listen less.


nateberkopec:
So, next steps here will be to find a benchmark and a benchmarking app that reproduces the original case described in #1646. I tried (https://github.com/puma/puma/pull/1646#issuecomment-519981602) but couldn't find anything yet.

nateberkopec:
I think that the way a typical load generating tool (such as wrk) generates load means that this benchmarking this PR is a little more difficult than usual. I think the benchmarking app I laid out previously will work, but I need to work on the load generation to make it work in a more production-like way.

nateberkopec:
I’ve been trying this with Siege as well (which emits requests after a set delay rather than after the last request has been responded to) and still not seeing a result. Next step will be for me to reproduce using TechEmpower, which is the only benchmark we can claim has gotten better so far.

ayufan:
I opened this issue https://github.com/puma/puma/issues/2078, that seems to talk almost about the same problem as this one.

However, this cares about a busy workers, but not rather idle ones.

@nateberkopec @dentarg What do you think about the other issue and the proposed solution (not a sleep in that form, but something close that looks at thread pool), but by injecting latency earlier. I would think about making this change as a non-changing behavior (disabled by default), and this could be then enabled to improve the performance.

I tested the changes introduced by this branch: https://docs.google.com/spreadsheets/d/1y2YqrPPgZ-RtjKiCGplJ7YkwjMXkZx6vEPq7prFbUn4/edit#gid=0.
It seems to not bring any major improvements. I did also
re-run my branch again to ensure that this was no fluke,
and it still brings substantial improvements.

lib/puma/thread_pool.rb
```diff
@@ -192,24 +201,39 @@ def <<(work)
...
+            # check for work again after a short timeout, which is a function
+            # of the available threads.
+            # Otherwise if threads are busy, block without timeout
+            # until a thread finishes its work.
+            timeout = if threads_available
+                        (busy_threads.to_f / @max) * WORK_AVAILABLE_TIMEOUT_FACTOR
+                      end
+          end
+
           return busy_threads if @max > busy_threads
```
wjordan: This line was added by #1648, and makes the behavior of this PR quite different from the one in #1646 (which is based on a commit prior to #1648). Here, the call to `@not_full.wait` will _never_ be reached unless `busy_threads >= @max` (ie, the thread pool is maxxed out with requests), which would be an extremely rare case.

I believe this is the reason this PR has no observable effect in benchmarks.

wjordan:
Re-ran the TechEmpower benchmark to confirm this PR is currently not working, see https://github.com/puma/puma/pull/2079#issuecomment-564809228.

nateberkopec:
Closing in favor of the 2 open and active PRs addressing this issue: #2079 and #2092

PR #1925, opened 2019-08-23
test_integration.rb - cleanup, try removing log output on unneeded 'forked' servers

MSP-Greg:
Commit 'test_integration.rb - minor code clean up' - cleans up file, no real changes

Commit 'start_forked_server => server.pid ?' - stop forking servers, trying to eliminate log output.  There are some retries that may indicate issues.

And yes, `#test_sigterm_closes_listeners_on_forked_servers` still intermittently fails, usually on macOS jobs.

Regarding instance variables in setup, Minitest runs every test in a new instance.  Hence, creating strings, etc really shouldn't be considered bad.  Removed all the past object code, so objects are only created as needed.

### DIFF
--- a/test/test_integration.rb
+++ b/test/test_integration.rb
@@ -5,150 +5,55 @@ require "puma/cli"
 require "puma/control_cli"
 require "open3"
 
-# TODO: Remove over-utilization of @instance variables
-# TODO: remove stdout logging, get everything out of my rainbow dots
-
 class TestIntegration < Minitest::Test
   HOST  = "127.0.0.1"
   TOKEN = "xxyyzz"
 
+  BASE = defined?(Bundler) ?
+    "bundle exec #{Gem.ruby} -Ilib bin/puma" :
+    "#{Gem.ruby} -Ilib bin/puma"
+
   def setup
-    @state_path   = "test/test_#{name}_puma.state"
-    @bind_path    = "test/test_#{name}_server.sock"
-    @control_path = "test/test_#{name}_control.sock"
-
-    @server = nil
-
-    @wait, @ready = IO.pipe
-
-    @events = Puma::Events.strings
-    @events.on_booted { @ready << "!" }
+    @ios_to_close = []
+    @state_path   = "test/#{name}_puma.state"
+    @bind_path    = "test/#{name}_server.sock"
+    @control_path = "test/#{name}_control.sock"
   end
 
   def teardown
-    File.unlink @state_path   rescue nil
-    File.unlink @bind_path    rescue nil
-    File.unlink @control_path rescue nil
-
-    @wait.close
-    @ready.close
-
-    if @server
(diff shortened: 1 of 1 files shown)


nateberkopec:
> Minitest runs every test in a new instance. Hence, creating strings, etc really shouldn't be considered bad.

Yeah it's more just the ugliness than the actual correctness that inspired that TODO

nateberkopec:
This is a little too parallel with my existing integration cleanup PR. I'm going to prioritize that one. I just have to figure out why it's not working on Linux.

MSP-Greg:
@nateberkopec

> This is a little too parallel with my existing integration cleanup PR

I just pushed what I can do with this.

I saw your [integration_cleanup](https://github.com/puma/puma/tree/integration_cleanup) branch, and tried working with it, but starting with something that's broken can be nonproductive.  Also, the differences between Ubuntu & macOS are beyond my pay grade...

When you asked for changes in test_integration.rb on PR #1908,  I decided to add some of things I had done with your branch, and this is the result of that.

1. i tried creating a 'test_phased_restart_via_pumactl' test that used 'server' instead of 'run_launcher', and it kept failing in Puma control_cli.rb code.  For now, I give up.

2. You renamed a few tests, and I agree that the names are cluttered.  At present, we've got suffixes of 'workers', 'on_forked_servers', 'clustered_mode', and a prefix of 'test_workers', etc.  Maybe if there are two companion tests or a specific test, simply a 'single' or 'cluster' suffix might be used?

3. [`test_phased_restart_via_pumactl`](https://github.com/puma/puma/blob/d6c017b05b5af20e50aef119919837100b3004f4/test/test_integration.rb#L59-L94) is similar to [`test_worker_phased_restart`](https://github.com/puma/puma/blob/d6c017b05b5af20e50aef119919837100b3004f4/test/test_integration.rb#L315-L318).  I added a time based assert to the first, as it's timing should be governed by 'worker_shutdown_timeout'.

Hence, feel free to use whatever.  Some of the code I added (like `@ios_to_close`) is just 'good practices', and allows objects to be created without doing so as instance variables.

I'm pissed about the macOS intermittent failures.  I think we're all tired of the need to look at test logs due to test stability issues...

MSP-Greg:
@nateberkopec

While AFK, had a few thoughts, and I think that with the 2nd commit, this is similar to what you're looking for.  Many of the tests here are end-to-end, maybe a few should move elsewhere, like the 'stuck_workers' tests.  Some of the common methods may be applicable to other test files.

I'll probably jinx it by mentioning it, but the macOS jobs have been passing & stable in my fork while working on the 2nd commit...

nateberkopec:
Cool! There's some things you were doing in here that I prefer, I just need to sit down and "merge" the two PRs so they're just the way I like.

nateberkopec:
Merged w/ 9efee4e97

PR #1927, opened 2019-08-24
cluster.rb - remove worker #dead?, #dead!, @dead

MSP-Greg:
While working on PR #1908, I added a Worker method `#term?` to indicate whether `Worker#term` had been called.  Also, previous code that rejected workers from the `@workers` array where `dead?` was true was removed.

At the time, I wondered what `dead`'s purpose was.  It was also bothersome that workers contain threads, and `alive?` is a thread method, and most would consider `dead` to be the inverse.  Also, it's confusing, especially since code and comments in `thread_pool.rb` refer to `dead_workers`...

Upon reviewing the code, I cannot see anywhere where the state of `Worker#dead?` or `@dead` is used.  It is set here on line 510:

https://github.com/puma/puma/blob/9fb12283dbadf6e2bad04d5e67fe3fc9c29eede5/lib/puma/cluster.rb#L503-L512

This PR changes line 510 to 'w.term' and removes the two 'dead' methods.  If this seems improper, the PR could be changed to call `#term` in `Cluster#check_workers` for any dead workers...

**EDIT:** This may also help intermittent  test_integration.rb failures in macOS builds on Travis, which I've seen in my fork, as I can't test cluster locally..

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -74,7 +74,6 @@ module Puma
         @started_at = Time.now
         @last_checkin = Time.now
         @last_status = '{}'
-        @dead = false
         @term = false
       end
 
@@ -87,14 +86,6 @@ module Puma
       def boot!
         @last_checkin = Time.now
         @stage = :booted
-      end
-
-      def dead?
-        @dead
-      end
-
-      def dead!
-        @dead = true
       end
 
       def term?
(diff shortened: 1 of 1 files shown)


nateberkopec:
> This may also help intermittent test_integration.rb failures in macOS builds on Travis, which I've seen in my fork, as I can't test cluster locally..

welp, guess that didn't work 

Screen Shot 2019-08-24 at 7 02 43 AM

I've notice the 2.4 one fails more often than the 2.5 one

nateberkopec:
Great catch, agreed that this is dead code.

nateberkopec:
Looks like it was added here https://github.com/puma/puma/commit/3fdf0c968153afc696a1cb237c1bec68e9d0eb5e and the delete_if? dead has been removed at some point in the intervening years.

PR #1934, opened 2019-08-27
Allow semicolon in url-path

### LINKED ISSUE #1122: Semicolon in url-path
I use puma with redmine.
If redmine attachment filename contains ";" (semicolon) puma drop it and symbols which follows after, so filename does not match, and we get error.

on puma:
In browser we go to "http://localhost:3000/attachments/download/301/attachment-with-;-semicolon.txt"
and in logs we have:

```
Started GET "/attachments/download/301/attachment-with-" for ::1 at 2016-10-25 16:37:02 +0300
Processing by AttachmentsController#download as HTML
  Parameters: {"id"=>"301", "filename"=>"attachment-with-"}
```

on webrick:
url in browser: "http://localhost:3000/attachments/download/301/attachment-with-;-semicolon.txt"
logs:

```
Started GET "/attachments/download/301/attachment-with-;-semicolon.txt" for ::1 at 2016-10-25 16:41:39 +0300
Processing by AttachmentsController#download as HTML
  Parameters: {"id"=>"301", "filename"=>"attachment-with-;-semicolon.txt"}
```

I also did test with "thin" and "passenger" but problem confirmed only with puma.

zarelit:
Allow unescaped semicolons in the URL, improves compatibility with applications that use it as a parameter separator such as Angular 2 or Redmine.

Fixes #1122

### DIFF
--- a/ext/puma_http11/http11_parser.c
+++ b/ext/puma_http11/http11_parser.c
@@ -38,7 +38,7 @@ static void snake_upcase_char(char *c)
 
 #line 40 "ext/puma_http11/http11_parser.c"
 static const int puma_parser_start = 1;
-static const int puma_parser_first_final = 47;
+static const int puma_parser_first_final = 46;
 static const int puma_parser_error = 0;
 
 static const int puma_parser_en_main = 1;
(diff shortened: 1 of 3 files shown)


zarelit:
The tests on MacOSX are flaky, here are the same tests on the same commit, passing: https://travis-ci.org/zarelit/puma/builds/577223164

nateberkopec:
😮 Thanks for pushing this up so quickly! Will review soon.

zarelit:
Thank you for the review and merge @evanphx @nateberkopec

PR #1943, opened 2019-09-02
Using RbConfig for host os check

Adithya-copart:
I'm browsing through the code and came across this piece of code which is initially added in https://github.com/puma/puma/commit/6da831aec3eb4bd237008a30a539e0885475dc8a

https://github.com/puma/puma/blob/818aa4a12e643606a1fd8a6cd3557c51a1521d80/lib/puma/server.rb#L108-L109

This condition will not return a truthy value for JRuby even on Linux and I wanted to see the results from travis with this change to understand if it's intentional.

```
$ docker run -it jruby /bin/bash
root@097863b86e82:/# irb
irb(main):001:0> RUBY_PLATFORM =~ /linux/
=> nil
irb(main):002:0> RUBY_PLATFORM
=> "java"
irb(main):003:0> RbConfig::CONFIG['host_os'] =~ /linux/
=> 0
```

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -105,7 +105,7 @@ module Puma
     # On Linux, use TCP_CORK to better control how the TCP stack
     # packetizes our stream. This improves both latency and throughput.
     #
-    if RUBY_PLATFORM =~ /linux/
+    if RbConfig::CONFIG['host_os'] =~ /linux/
       UNPACK_TCP_STATE_FROM_TCP_INFO = "C".freeze
 
       # 6 == Socket::IPPROTO_TCP


Adithya-copart:
Looks like Java doesn't support `TCP_CORK` as this is platform specific.

```
root@0392432c7bb7:/# irb -r socket
irb(main):001:0> sock = TCPSocket.new 'example.com', 80
=> #<TCPSocket:fd 5>
irb(main):002:0> sock.setsockopt(6, 3, 1)
        7: from /opt/jruby/bin/irb:13:in `<main>'
        6: from org/jruby/RubyKernel.java:1193:in `catch'
        5: from org/jruby/RubyKernel.java:1193:in `catch'
        4: from org/jruby/RubyKernel.java:1425:in `loop'
        3: from org/jruby/RubyKernel.java:1061:in `eval'
        2: from (irb):2:in `evaluate'
        1: from org/jruby/ext/socket/RubyBasicSocket.java:329:in `setsockopt'
Errno::ENOPROTOOPT (Protocol not available - Protocol not available)
irb(main):003:0> 
```

PR #1945, opened 2019-09-02
Fix #1842

### LINKED ISSUE #1842: Socket removed after reload, part II
Continuing the discussion from #1775, it seems like the fix in 4.0.1 may not have done the trick for everyone.

I made a mistake [here](https://github.com/puma/puma/issues/1775#issuecomment-505855372) by not forking the discussion into two parts, because some people were reporting problems on 3.12 and some were not. We may have two separate issues here. I'd like to reset the discussion to here with what people are experiencing on 4.0.1.

nateberkopec:
* [x] Consider #1886

### DIFF
--- a/test/test_integration.rb
+++ b/test/test_integration.rb
@@ -198,6 +198,7 @@ class TestIntegration < Minitest::Test
     ccli.run
 
     assert_kind_of Thread, t.join, "server didn't stop"
+    assert File.exist? @bind_path
   end
 
   def test_kill_unknown_via_pumactl
--- a/test/test_integration.rb
+++ b/test/test_integration.rb
@@ -188,19 +188,20 @@ class TestIntegration < Minitest::Test
     until done
       @events.stdout.rewind
       log = @events.stdout.readlines.join("")
-      if log =~ /- Worker \d \(pid: \d+\) booted, phase: 1/
+      if log =~ /- Worker 1 \(pid: \d+\) booted, phase: 1/
         assert_match(/TERM sent/, log)
-        assert_match(/- Worker \d \(pid: \d+\) booted, phase: 1/, log)
+        # TODO: This test sends term twice to each worker and then kills them
+        # Why are there two terms to each worker?
         done = true
       end
     end
 
-    assert File.exist? @bind_path
+    assert File.exist?(@bind_path), "Bind path must exist after phased restart"
     ccli = Puma::ControlCLI.new ["-S", @state_path, "stop"], sout
     ccli.run
 
     assert_kind_of Thread, t.join, "server didn't stop"
-    refute File.exist? @bind_path
+    refute File.exist?(@bind_path), "Bind path must be removed after stop"
   end
 
   def test_kill_unknown_via_pumactl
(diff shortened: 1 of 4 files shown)


nateberkopec:
This is _almost_ there for me locally. It's now only failing on the phased restart test, because something appears to be closing the listener unnecessarily (no idea what that is yet but shouldn't be too hard to figure out.)

test/test_integration.rb
```diff
@@ -193,12 +193,12 @@ def test_phased_restart_via_pumactl
         done = true
       end
     end
-    # Stop
+
+    assert File.exist? @bind_path
```
nateberkopec: @Fudoshiki Why did you originally insert this assertion _after_ the stop command is issued? Shouldn't stopping the server remove the binder?

Fudoshiki: We don't need this `assert` now

nateberkopec: Don't we? The original issue was that this file was removed after a phased restart was begun.

nateberkopec:
Possible clue in the output
```
[94299] - TERM sent to 94301...
[94299] - TERM sent to 94301...
[94299] - KILL sent to 94301...
```
This is intentional, as the test is trying to trigger a long response (sleep 40 seconds) and then have the phased restart mechanism timeout the worker trying to restart it, but it's possible that this triggers binding.close somewhere.

test/test_integration.rb
Fudoshiki: - If we call `phased-restart` - the file must exist, connections from old workers should close
- If we stopping the server, the file must be deleted

Maybe we need move this in `until` block above

nateberkopec:
Okay, I think I've got it?

Basically we're calling `binder.close` all over the place in server.rb, always have been. However, when phased restarting, each worker will eventually call binder.close. That's fine for the first worker but the second will fail.

Previously the @own_binder logic in #1685 prevented phased restarting worker processes from calling binding.close, because it wasn't their "own" binder. Makes sense.

So, we just have to figure out how to add that back in, in a way that _also_ closes the socket when we are shutting down in a non-phased-restart scenario.

nateberkopec:
Things I know:

* The tests are correct now and seem to be describing the correct behavior.
* #1685 made a big mistakey by removing the own_binder logic.
* Just randomly removing code and hoping the tests pass is a Bad Idea without a complete understanding of the situation with Puma.
* Binder closing is surprisingly complicated, as it needs to work in tons of different scenarios - phased restart, regular restart, shutdown in a cluster, shutdown in single mode, etc.

What I'm not sure of:
* The code path for a phased restart/shutdown is different when using signals than when using control-cli over a socket, is this bug only present on one codepath or the other?
* own_binder logic isnt really working because now no-one is closing the socket correctly, as every cluster runner inherits its listener from the launcher.

MSP-Greg:
Similar to what you're thinking, in the past I've wondered why Binder has `@ios` and `@unix_paths`, with `@listeners` containing very similar information.

Kind of goes hand in hand with the similarity between `Binder#close` and `Launcher#close_binder_listeners`...

PR #1947, opened 2019-09-03
Move MacOS to allow_fail until we fix flaky test

nateberkopec:


dentarg:
Which test is it that is flaky? 

(I know it has come up a number of times that some tests are flaky, but I haven't looked closely. I thought it might be good to document explicitly what needs to be fixed for this to change be removed in the future. Maybe there's already some issue about.)

nateberkopec:
Good point. I went back and looked and actually master doesn't fail that often on Mac, and over the last 4 weeks hasn't failed on a consistent test. I know @MSP-Greg will have an opinion on this but I think for now actually it's fine and isn't failing often enough to be a problem.

PR #1956, opened 2019-09-10
[PR 1952] 2st verification - updated tests & GitHub Actions

MSP-Greg:
PR #1955 shows that the tests in this PR do not pass on current master.

Below is a summary of the changes here.  The last commit adds two tests that use UNIXSockets for the bind address, verifying that unix files seem to be handled correctly during shutdowns and restarts.

### lib files - binder.rb, cluster.rb, launcher.rb

1. binder.rb - Add PR #1886

2. cluster.rb - Add PR #1952, additional minor changes

3. launcher.rb - #close_binder_listeners - close io's as fast as possible.  Unlink files in separate iteration.

#### Request handling during server TERM - two tests

`#test_term_closes_listeners_cluster_tcp`
`#test_term_closes_listeners_cluster_unix`

using `#term_closes_listeners_cluster`

Send requests 10 per second.  Send 10, then :TERM server, then send another 30.
No more than 10 should throw Errno::ECONNRESET.

#### Request handling during phased restart - two tests

`#test_usr1_all_respond_cluster_tcp`
`#test_usr1_all_respond_cluster_unix`

using `#usr1_all_respond_cluster`

Send requests 1 per second.  Send 1, then :USR1 server, then send another 24.
All should be responded to, and at least three workers should be used

#### Stuck worker tests - two tests

`#test_stuck_external_term_spawn_cluster`
Tests whether externally TERM'd 'stuck' workers are proper re-spawned.

`#test_stuck_phased_restart_cluster`
Tests whether 'stuck' workers are properly shutdown during phased-restart.

### GitHub Actions - workflow.yml

1. Install ragel for Ubuntu and macOS.  Both intermittently failed compile without doing so.

2. Windows - Add step to update MSYS2 gcc, install ragel and proper OpenSSL package.

3. Actions CI seems stable, and it finishes well before either Travis or AppVeyor.

### Misc test files changes

Added `parallelize_me!` to a few files, misc changes

### DIFF
--- a/lib/puma/binder.rb
+++ b/lib/puma/binder.rb
@@ -50,14 +50,6 @@ module Puma
 
     def close
       @ios.each { |i| i.close }
-      @unix_paths.each do |i|
-        # Errno::ENOENT is intermittently raised
-        begin
-          unix_socket = UNIXSocket.new i
-          unix_socket.close
-        rescue Errno::ENOENT
-        end
-      end
     end
 
     def import_from_env
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -217,11 +217,12 @@ module Puma
     end
 
     def close_binder_listeners
-      @binder.listeners.each do |l, io|
-        io.close
+      # close binder listeners as fast as possible, so separate loop
+      @binder.listeners.each { |_, io| io.close }
+      @binder.listeners.each do |l, _|
         uri = URI.parse(l)
         next unless uri.scheme == 'unix'
-        File.unlink("#{uri.host}#{uri.path}")
+        File.unlink "#{uri.host}#{uri.path}"
       end
     end
 
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -47,7 +47,11 @@ module UniquePort
   @mutex = Mutex.new
 
   def self.call
-    @mutex.synchronize { @port += 1 }
+    @mutex.synchronize {
+      @port += 1
+      @port = 3307 if @port == 3306  # MySQL on Actions
+      @port
+    }
   end
 end
 
@@ -70,6 +74,34 @@ end
 
 module TestSkips
 
+  # finds bad signals, value is an array of symbols
+  BAD_SIGNAL_LIST = begin
+    code = <<-HEREDOC
+      data = (%w[HUP INT TTIN TTOU USR1 USR2] - Signal.list.keys).join(' ')
+
+      begin
+        Signal.trap('TERM') { }
+      rescue ArgumentError
+        data << 'TERM'
+      end
+
+      puts data
+    HEREDOC
+    pipe = IO.popen RbConfig.ruby, 'r+'
+    pid = pipe.pid
+    pipe.puts code
+    pipe.close_write
+
+    begin
+      Process.kill :TERM, pid
+      pipe.read.strip.split(' ').sort.map(&:to_sym)
+    rescue Errno::EINVAL
+      "#{pipe.read.strip} TERM".split(' ').sort.map(&:to_sym)
+    ensure
+      Process.wait pid
+    end
+  end
+
   # usage: skip NO_FORK_MSG unless HAS_FORK
   # windows >= 2.6 fork is not defined, < 2.6 fork raises NotImplementedError
   HAS_FORK = ::Process.respond_to? :fork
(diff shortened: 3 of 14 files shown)


test/test_puma_server.rb
```diff
@@ -1,9 +1,10 @@
 require_relative "helper"
 
 class TestPumaServer < Minitest::Test
+  parallelize_me!
 
   def setup
-    @port = 0
+    @port = UniquePort.call
```
nateberkopec: Why can't we just use port = 0 everywhere, I wonder?

MSP-Greg: Good question.  Only thing is that all the code would have to be changed for the 'clients', but we could set that in `wait_for_server_to_boot`?

I'm !AFK now, so as soon as the CI finishes, I'll have a look, as it's pretty 'fresh' in my mind...

MSP-Greg:
@nateberkopec

One typo I didn't explicitly mention:

https://github.com/puma/puma/blob/8e2c461de47d04459ea10c5136fdcfd5ee9ec031/test/helpers/integration.rb#L20-L25

Line 25 should be combined with line 24.  In case you've got another commit waiting.

lib/puma/binder.rb
```diff
@@ -50,14 +50,6 @@ def env(sock)
 
     def close
       @ios.each { |i| i.close }
-      @unix_paths.each do |i|
```
nateberkopec: This doesn't fix the problem, as bind paths are still not removed after program stop.

lib/puma/launcher.rb
```diff
@@ -217,11 +217,12 @@ def restart_args
     end
 
     def close_binder_listeners
-      @binder.listeners.each do |l, io|
-        io.close
+      # close binder listeners as fast as possible, so separate loop
```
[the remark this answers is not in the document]
MSP-Greg: We want the listener sockets closed quickly once shutdown starts.  Anything that requires checking the disk can slow down.  Small issue.

lib/puma/binder.rb
MSP-Greg: A few days ago, I was really clear on what was where in terms of Binder `@ios`, `@unix_paths`, and `@listeners`.  I should have added some comments...

Paths are closed in `Binder#close_listeners`, They aren't in `Binder#close`

test/helper.rb
```diff
@@ -70,6 +74,34 @@ def run(*)
 
 module TestSkips
 
+  # finds bad signals, value is an array of symbols
+  BAD_SIGNAL_LIST = begin
```
nateberkopec: Not a huge fan of this, what's wrong with Signal.list?

nateberkopec:
OK, let's try again with this:

* 1 commit that adds the tests described in the OP
* 1 commit each for the 3 changes described in the lib section of the OP (so, 3 more commits)
* Save everything else, e.g. Github Actions, the signal test helper changes, etc for other PRs.

test/helper.rb
nateberkopec: I appreciate the idea, but let's just remove :TERM from the list if on Windows, this spawning-a-process thing feels like the nuclear option.

nateberkopec: Also, if :TERM doesnt work on windows, why do tests that use not fail? e.g. `test_term_closes_listeners` or `test_term_suppress`

lib/puma/binder.rb
nateberkopec: [See this assertion I've added to master.](https://github.com/puma/puma/commit/2ae66ce61c6ab0ef4e4df15b324eaf953667979b#diff-e405df74eaf050d8fc90276c05c35b39R17) Anything that purports to fix this problem needs to pass this test (which is failing on master).

test/helper.rb
nateberkopec: Of of course, because you can't fork on windows. D'oh!

MSP-Greg: > let's just remove :TERM from the list if on Windows, this spawning-a-process thing feels like the nuclear option

Understandable.

I really hate defining by OS, as opposed to actually checking for functionality.  Maybe it (`:TERM`) could change in a year or something.  Probably unlikely...

MSP-Greg:
Of the above three items, '1 commit that adds the tests described in the OP' is the messiest.  I'm happy to look at it or do the PR, but you may want to do it.

I was told I'd be receiving info about an ongoing project early this week, which will keep me busy for several days.  Hence, the PR.  I've now been told it will be later this week or next...

PR #1935, opened 2019-08-28
CI: Use Ruby 2.4.7, 2.5.6, 2.6.4

olleolleolle:
This PR updates the build matrix.

PR #1965, opened 2019-09-13
Update test_integration files per PR #1956

MSP-Greg:
#### Request handling during server TERM - two tests

`#test_term_closes_listeners_tcp`
`#test_term_closes_listeners_unix`

using `#term_closes_listeners`

Send requests 10 per second.  Send 10, then :TERM server, then send another 30.
No more than 10 should throw Errno::ECONNRESET.

#### Request handling during phased restart - two tests

`#test_usr1_all_respond_tcp`
`#test_usr1_all_respond_unix`

using `#usr1_all_respond`

Send requests 1 per second.  Send 1, then :USR1 server, then send another 24.
All should be responded to, and at least three workers should be used

#### Stuck worker tests - two tests

`#test_stuck_external_term_spawn`
Tests whether externally TERM'd 'stuck' workers are proper re-spawned.

`#test_stuck_phased_restart`
Tests whether 'stuck' workers are properly shutdown during phased-restart.

### helper files/methods changes

1. Changes to allow binding to TCP or UNIX, see kwarg unix:
2. Skip on Windows for signal TERM

### DIFF
--- a/test/helpers/integration.rb
+++ b/test/helpers/integration.rb
@@ -20,24 +20,21 @@ class TestIntegration < Minitest::Test
 
   def teardown
     if defined?(@server) && @server
-      begin
-        Process.kill "INT", @server.pid
-      rescue
-        Errno::ESRCH
-      end
-      begin
-        Process.wait @server.pid
-      rescue Errno::ECHILD
-      end
-      @server.close unless @server.closed?
-      @server = nil
+      stop_server @pid, signal: :INT
     end
 
     @ios_to_close.each do |io|
       io.close if io.is_a?(IO) && !io.closed?
       io = nil
     end
+    refute File.exist?(@bind_path), "Bind path must be removed after stop"
     File.unlink(@bind_path) rescue nil
+
+    # wait until the end for OS buffering?
+    if defined?(@server) && @server
+      @server.close unless @server.closed?
+      @server = nil
+    end
   end
 
   private
@@ -56,8 +53,10 @@ class TestIntegration < Minitest::Test
   end
 
   def stop_server(pid = @pid, signal: :TERM)
-    Process.kill signal, pid
-    sleep 1
+    begin
+      Process.kill signal, pid
+    rescue Errno::ESRCH
+    end
     begin
       Process.wait2 pid
     rescue Errno::ECHILD
@@ -65,16 +64,16 @@ class TestIntegration < Minitest::Test
   end
 
   def restart_server_and_listen(argv)
-    cli_server(argv)
+    cli_server argv
     connection = connect
     initial_reply = read_body(connection)
-    restart_server(connection)
+    restart_server connection
     [initial_reply, read_body(connect)]
   end
 
   # reuses an existing connection to make sure that works
   def restart_server(connection)
-    Process.kill :USR2, @server.pid
+    Process.kill :USR2, @pid
     connection.write "GET / HTTP/1.1\r\n\r\n" # trigger it to start by sending a new request
     wait_for_server_to_boot
   end
--- a/test/helpers/integration.rb
+++ b/test/helpers/integration.rb
@@ -15,6 +15,7 @@ class TestIntegration < Minitest::Test
 
   def setup
     @ios_to_close = []
+    @bind_path    = "test/#{name}_server.sock"
   end
 
   def teardown
@@ -36,26 +37,31 @@ class TestIntegration < Minitest::Test
       io.close if io.is_a?(IO) && !io.closed?
       io = nil
     end
+    File.unlink(@bind_path) rescue nil
   end
 
   private
 
-  def cli_server(argv, bind = nil)
-    if bind
-      cmd = "#{BASE} bin/puma -b #{bind} #{argv}"
+  def cli_server(argv, unix: false)
+    if unix
+      cmd = "#{BASE} bin/puma -b unix://#{@bind_path} #{argv}"
     else
       @tcp_port = UniquePort.call
       cmd = "#{BASE} bin/puma -b tcp://#{HOST}:#{@tcp_port} #{argv}"
     end
     @server = IO.popen(cmd, "r")
     wait_for_server_to_boot
+    @pid = @server.pid
     @server
   end
 
-  def send_term_to_server(pid)
-    Process.kill(:TERM, pid)
+  def stop_server(pid = @pid, signal: :TERM)
+    Process.kill signal, pid
     sleep 1
-    Process.wait2(pid)
+    begin
+      Process.wait2 pid
+    rescue Errno::ECHILD
+    end
   end
 
   def restart_server_and_listen(argv)
@@ -77,8 +83,8 @@ class TestIntegration < Minitest::Test
     true while @server.gets !~ /Ctrl-C/ # wait for server to say it booted
   end
 
-  def connect(path = nil)
-    s = TCPSocket.new HOST, @tcp_port
+  def connect(path = nil, unix: false)
+    s = unix ? UNIXSocket.new(@bind_path) : TCPSocket.new(HOST, @tcp_port)
     @ios_to_close << s
     s << "GET /#{path} HTTP/1.1\r\n\r\n"
     true until s.gets == "\r\n"
--- a/test/helpers/integration.rb
+++ b/test/helpers/integration.rb
@@ -52,6 +52,9 @@ class TestIntegration < Minitest::Test
     @server
   end
 
+  # rescue statements are just in case method is called with a server
+  # that is already stopped/killed, especially since Process.wait2 is
+  # blocking
   def stop_server(pid = @pid, signal: :TERM)
     begin
       Process.kill signal, pid
--- a/test/test_integration_cluster.rb
+++ b/test/test_integration_cluster.rb
@@ -122,7 +122,7 @@ class TestIntegrationCluster < TestIntegration
   def term_closes_listeners(unix: false)
     skip_unless_signal_exist? :TERM
 
-    cli_server "-w #{WORKERS} -t 0:6 -q test/rackup/sleep_pid.ru", unix: unix
+    cli_server "-w #{WORKERS} -t 0:6 -q test/rackup/sleep_step.ru", unix: unix
     threads = []
     replies = []
     mutex = Mutex.new
@@ -135,29 +135,52 @@ class TestIntegrationCluster < TestIntegration
         threads << Thread.new do
           sleep i.to_f/div
           Process.kill :TERM, @pid
-          mutex.synchronize { replies << :term_sent }
+          mutex.synchronize { replies[i] = :term_sent }
         end
       else
         threads << Thread.new do
-          thread_run replies, i.to_f/div, 1, mutex, refused, unix: unix
+          thread_run_step replies, i.to_f/div, 1, i, mutex, refused, unix: unix
         end
       end
     end
 
     threads.each(&:join)
 
-    responses = replies.count { |r| r[/\ASlept 1/] }
-    resets    = replies.count { |r| r == :reset    }
-    refused   = replies.count { |r| r == :refused  }
-    msg = "#{responses} responses, #{resets} resets, #{refused} refused"
+    failures  = replies.count(:failure)
+    successes = replies.count(:success)
+    resets    = replies.count(:reset)
+    refused   = replies.count(:refused)
 
-    $debugging_info << "#{full_name}\n    #{msg}\n"
+    r_success = replies.rindex(:success)
+    l_reset   = replies.index(:reset)
+    r_reset   = replies.rindex(:reset)
+    l_refused = replies.index(:refused)
 
-    assert_operator 9,  :<=, responses, msg
+    msg = "#{successes} successes, #{resets} resets, #{refused} refused, failures #{failures}"
+
+    assert_equal 0, failures, msg
+
+    assert_operator 9,  :<=, successes, msg
 
     assert_operator 10, :>=, resets   , msg
 
     assert_operator 20, :<=, refused  , msg
+
+    # Interleaved asserts
+    # UNIX binders do not generate :reset items
+    if l_reset
+      assert_operator r_success, :<, l_reset  , "Interleaved success and reset"
+      assert_operator r_reset  , :<, l_refused, "Interleaved reset and refused"
+    else
+      assert_operator r_success, :<, l_refused, "Interleaved success and refused"
+    end
+
+  ensure
+    if passed?
+      $debugging_info << "#{full_name}\n    #{msg}\n"
+    else
+      $debugging_info << "#{full_name}\n    #{msg}\n#{replies.inspect}\n"
+    end
   end
 
   # Send requests 1 per second.  Send 1, then :USR1 server, then send another 24.
@@ -177,7 +200,7 @@ class TestIntegrationCluster < TestIntegration
 
     24.times do |delay|
       threads << Thread.new do
-        thread_run replies, delay, 1, mutex, refused, unix: unix
+        thread_run_pid replies, delay, 1, mutex, refused, unix: unix
       end
     end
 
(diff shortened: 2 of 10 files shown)


test/helpers/integration.rb
```diff
@@ -36,26 +37,31 @@ def teardown
       io.close if io.is_a?(IO) && !io.closed?
       io = nil
     end
+    File.unlink(@bind_path) rescue nil
```
nateberkopec: Let's fail the test if this path exists.

test/helpers/integration.rb
```diff
@@ -36,26 +37,31 @@ def teardown
...
 
-  def send_term_to_server(pid)
-    Process.kill(:TERM, pid)
+  def stop_server(pid = @pid, signal: :TERM)
+    Process.kill signal, pid
     sleep 1
-    Process.wait2(pid)
+    begin
+      Process.wait2 pid
+    rescue Errno::ECHILD
```
nateberkopec: Shouldn't we just wait2 without sleeping, then we don't need to rescue ECHILD?

test/test_integration_cluster.rb
```diff
@@ -112,117 +91,166 @@ def test_term_worker_clean_exit
...
+    assert_operator 9,  :<=, responses, msg
 
-    worker1 = Thread.new do
-      begin
-        req1 = Net::HTTP::Get.new "/sleep40", {}
-        http1.start.request(req1) { |rep1| body1 = rep1.body }
-      rescue
+    assert_operator 10, :>=, resets   , msg
+
+    assert_operator 20, :<=, refused  , msg
```
nateberkopec: This feels like a pretty big change. Wasn't the original test that _all_ responses must be refused?

test/test_integration_cluster.rb
```diff
@@ -232,19 +260,29 @@ def clean_exit_pids(pids)
...
-    events.on_booted { ready << "!" }
-
-    launcher = Puma::Launcher.new conf, :events => events
-
-    thr = Thread.new { launcher.run }
-
-    # wait for boot from `events.on_booted`
-    wait.sysread 1
+  # used with thread_run to define correct 'refused' errors
+  def thread_run_refused(unix: false)
```
nateberkopec: This feels like the normalization of deviance to me.

test/test_integration_pumactl.rb
```diff
@@ -16,15 +15,15 @@ def teardown
     begin
       # refute File.exist?(@bind_path), "Bind path must be removed after stop"
     ensure
-      [@bind_path, @state_path, @control_path].each { |p| File.unlink(p) rescue nil }
+      [@state_path, @control_path].each { |p| File.unlink(p) rescue nil }
```
nateberkopec: We should still unlink bind_path it if it exists.

nateberkopec:
The changes to #term_closes_listeners feel weird to me. It feels like we're just normalizing deviance of the test, which was originally that all connections were refused.

Thanks for this, I realize this PR probably took a lot of effort! The new integration test suite is a joy, it's so much nicer than it used to be.

test/helpers/integration.rb
MSP-Greg: At least one test is checking an exit code:

https://github.com/puma/puma/blob/17c1d2997e7d78dcda3f25e72baa2bc29fb6d036/test/test_integration_cluster.rb#L60-L65.  Shouldn't the wait be in place for that?

test/test_integration_cluster.rb
MSP-Greg: I may be incorrect, but I think the updated test is probably more accurate.  GIven that the system may be loaded down, resource starved, parallel tests, etc, it will take some finite amount of time to shut down.

I figured this would define a baseline for that time, it may be able to be improved, but it certainly will show if the code if changed and the time is affected for the worse.

test/test_integration_cluster.rb
MSP-Greg: It bothers me also.  Given that it seems to only apply to macOS and it's intermittent...

MSP-Greg:
> The changes to #term_closes_listeners feel weird to me. It feels like we're just normalizing deviance of the test, which was originally that all connections were refused.

1. Since you're macOS (?), have you checked any of these tests locally?  I wonder if you see the intermittent failures?  Anyway, see below...

2. I wanted to get the integration tests stable.  At some point, you gotta 'ship'.

3. It's not the best comparison, but when building ruby-loco (Windows Ruby trunk), the total times on AppVeyor can run anywhere from about 30 minutes to 50 minutes.  Both the build time and the test time vary.  I haven't seen that much deviation on Travis or GH Actions.  Also, there have been plenty of times where I could successfully run a test (for days) locally, but running it on AppVeyor either always failed or failed intermittently.

4. I think CI is probably better than local.  Maybe some 'TODO" notes need to be added to the tests.

test/helpers/integration.rb
nateberkopec: wait2 waits for the pid to exit, then returns the exit code. I don't see why a sleep or rescue is necessary.

nateberkopec:
> I wonder if you see the intermittent failures?

No, never 😞

test/test_integration_cluster.rb
nateberkopec: Then let's at least only allow the deviance there. I'd prefer we just skip the test entirely on Darwin with a note, rather than allow all platforms to slide backward.

test/helpers/integration.rb
MSP-Greg: Ok.  The sleep should be able to be removed, not sure if the rescue can be.  I'll remove both.

MSP-Greg:
> No, never 😞

Which kind of gets back to what's a better indicator of Puma use, testing locally or testing on CI?  And if CI testing has intermittent issues, how to move forward...

Anyway, any interest in merging #1886 and/or #1952 so this is passing, then start making the changes discussed?

nateberkopec:
Pull in #1952 for sure.

For #1886, you can try it, but [I would want to see this assertion uncommented and passing.](https://github.com/puma/puma/blob/e96e7980681c41f9ade0dac940553f0f2bd8f205/test/test_integration_pumactl.rb#L17)

MSP-Greg:
> For #1886, you can try it, but I would want to see this assertion uncommented and passing.

I try it in a branch based on this.

MSP-Greg:
@nateberkopec 

I think I've gotten the tests to a 'reasonable' state.  Several builds passed on both Actions and Travis. 
 See:

https://github.com/MSP-Greg/puma/commits/update-test-integration-bind-path

Three commits:

1. [Update test_integration files per PR puma#1956](https://github.com/MSP-Greg/puma/commit/88ec9de18583f3f593a1ec0ba35c32f3d1e9b8a2) - updates only the split up test files to changes similar to PR #1935 

2. [Add Binder#close_unix_paths, test_integration_* parallelize_me!](https://github.com/MSP-Greg/puma/commit/5243d2abf38f79eb6c5669e9c8765cbd4eda4057) - both lib files and test files.

    2.1 Adds a `Binder#close_unix_paths` method, and a matching `Launcher#close_binder_unix_paths` method.  I tried to add this to the existing methods in Launcher, but decided that it was easiest to add it to the end of the `run` method in `Single` and `Cluster`, since that is certainly where a clear 'finish/done/quit' is processed.

    2.2 lib files - changed most/all of the signal references from strings to symbols.

    2.3  test files - test/test_cli.rb - test_control_clustered - this was affecting Minitest's debug output, some additions that may not be needed, but it's clear what they do.

    2.4 test files - misc cleanup

3. [Add full_name for logging](https://github.com/MSP-Greg/puma/commit/1ed75c559ef31169fc502d0aca1d8b6555d5d752) - some of this is in the previous commit.  Since we may want to look at some test data over several builds/OS's, etc, I added an output section for the tests.  Travis example is [here](https://travis-ci.org/MSP-Greg/puma/jobs/584999595#L1301).

I'm sure I've forgotten some, but the main questions are:

A. If you have time for a quick review, is this ok?

B. If B, how would you like it divided up?

C. I think I've adjusted for most of the items in your review comments.

test/helpers/integration.rb
nateberkopec: @MSP-Greg ^^^

test/helpers/integration.rb
nateberkopec: @MSP-Greg ^^^^

test/test_integration_cluster.rb
```diff
@@ -112,117 +91,166 @@ def test_term_worker_clean_exit
...
-    worker_respawn { |l, phase0_worker_pids|
-      phase0_worker_pids.each { |p| Process.kill :TERM, p }
-    }
+    skip_unless_signal_exist? :TERM
+
+    worker_respawn(0) do |phase0_worker_pids|
+      last = phase0_worker_pids.last
+      phase0_worker_pids.each do |pid|
+        Process.kill :TERM, pid
+        sleep 4 unless pid == last
```
nateberkopec: Why 4? Please leave a comment.

nateberkopec: I say this because ocasionally in integration tests we have to sleep specific times for specific reasons, so if there are reasons, they should be documented.

test/test_integration_cluster.rb
nateberkopec: I still can't quite get past this. This test wasn't failing intermittently on Linux/Windows before, so I don't like that it is (?) now and we need this "<" stuff.

test/helpers/integration.rb
MSP-Greg: Just pushed an update.

MSP-Greg:
I think I fixed everything (except a previously duplicated 'term?' method in cluster.rb.

See:
https://github.com/MSP-Greg/puma/runs/225512262
and
https://travis-ci.org/MSP-Greg/puma/builds/586123815

Commits on the branch that I used above are at:
https://github.com/MSP-Greg/puma/commits/update-test-integration-lib

I updated the tests, created a new branch from it, then added PR's #1952, #1970, & #1973, along with an Actions yml file.  All passed, importantly, I believe the tests are stable...

MSP-Greg:
Question - should the :INFO signal be tested against on JRuby?

MSP-Greg:
> I still can't quite get past this.

Yeah, I know.

> This test wasn't failing intermittently on Linux/Windows before, so I don't like that it is (?) now and we need this "<" stuff.

  I created the new test to clearly spread out the workers a bit more.  I was debugging the output of the old test, and felt something different was needed.  The new tests clearly show that there is a small amount of time where request handling is indeterminate, which is reasonable.  The original PR mentions shutting down in relation to load balancers, which is odd, as I would think that one would shut down a server after it was removed from the load balancer, not before.

I kept the tests because it is a metric that shouldn't change.  If code is introduced that increases the number of resets, something may be wrong.

MSP-Greg:
Just pushed with added comments.  Sorry, should have added [skip CI]...

nateberkopec:
> The new tests clearly show that there is a small amount of time where request handling is indeterminate, which is reasonable.

How about we do this instead: block until the connection closes (received 1 conn closed), then send 30 requests, expect all to be rejected (I think I might have gotten the states wrong here, but basically keep sending up to X_LARGE_NUMBER requests until the conn gets into the state we want, then send like 30 more and expect it to stay there).

MSP-Greg:
> How about we do this instead

Maybe we could try this for a while and see how well it works?  I'm real leery of of tests that wait for state changes.

Obviously, we do have to wait for a booting server to become stable, but past that, nothing in the real world is waiting for the server, our tests also should not.   Doing so often leads to false positives, and we've had enough problems with those...

**Off topic:**

I've contributed to several gems/repos testing, but I did a lot of work getting the two main Ruby test suites stable and passing on Windows,  There were also stability issues with parallel testing on all platforms.  It was common for particular tests to fail on cloud CI but pass locally or non-parallel.  Total test count also jumped around in parallel testing.

Re Puma, it's often running with an app that is using far more resources than Puma itself (disk IO, memory, OS calls, etc).  Hence, running tests locally is really about as far from a good test environment as one could get.  We can't duplicate the 'resource hungry app' in CI (or locally), but running parallel is a good start.

In some respects, this is kind of a 'point of view' thing, but one particular issue is creating a passing test locally, then finding out it fails or isn't stable in CI.  Modifying it for CI may create a false positive test.  In some respects, I'm lucky that I can't run the tests locally...

nateberkopec:
>  nothing in the real world is waiting for the server, our tests also should not. Doing so often leads to false positives, and we've had enough problems with those...

I see this as being similar to booting. You're moving from state A to state B. It's important that we don't go to state A again or even to state C, we're just in State B forever.

I've seen the Mac test fail in such crazy ways (like reset to refusted back to reset and then to success) and we need to capture those fails, not say that going from A-> B -> A || C is ok if you only do it a few times. My proposal captures this.

MSP-Greg:
> I've seen the Mac test fail in such crazy ways (like reset to refusted back to reset and then to success)

I've seen that with the current test, and that's the reason I changed how the request threading works.

Rather than try and determine what state the server is in, should we see if the array divides into three arrays, the first containing only 'responds', the second containing only `:resets`, and the third containing only `:refused`?  IOW, the array has no 'interleaving' of the three?

nateberkopec:
> should we see if the array divides into three arrays, the first containing only 'responds', the second containing only :resets, and the third containing only :refused? IOW, the array has no 'interleaving' of the three?

That works for me!

MSP-Greg:
Don't know if you saw it, it was an edit, I added:
I'll make a branch from this and add the actual array to the debug output to check if it's interleaved...

For now, I'll rebase since all the 'lib' PR's needed for this to work are merged.  I'll start on the interleaved issue on my fork after that.

test/test_integration_single.rb
```diff
@@ -71,28 +73,28 @@ def test_term_not_accepts_new_connections
   def test_int_signal_with_background_thread_in_jruby
```
Adithya-copart: This method name could be updated to `test_int_shutdown` or `test_int_exit` while this file is being modified.
Also, `skip_unless :jruby` can be removed.

MSP-Greg: Done, but not yet pushed.

MSP-Greg:
@nateberkopec

Took a few more breaks than normal and added interleave.  But, the interleave can only be tested between :reset & :refused.  Since the requests have a delay of one second in the 'app', they continue into the :refused sections.  Removing the delay will remove all :reset entries, as I believe they're raised when the accepted request is read?

I added the array in debug output, see:

https://github.com/MSP-Greg/puma/commit/44fc57b5e3b21de6846dfded01b02817abbb7462/checks

It's at the bottom of the test step in all jobs but windows.  Also, #1952 is needed.  I believe this will pass and be stable when that is added.  Lastly, :INT signal doesn't work on Windows...

MSP-Greg:
The main issue with the above is that I kept with the 'original' concept for loading the 'replies' array with request information.  That concept loaded the array in the order that requests were handled, whether that be an actual response or raising an error.  Given that the 'app' was sleeping, the array order did not match the 'sent' order.  That array is helpful, but not optimal for our needs.

Light bulb went off, and I wrote an ru file that allows loading the array in the order that the requests are sent.

Much better.  Using that concept for the array, the array (tcp only) shows three non-interleaved sets, the first being responses, second is resets, and third is refused.  Unix sockets never seem to generate  any :reset members...

Haven't moved the code back into this branch yet.

nateberkopec:
Needs a rebase.

MSP-Greg:
> Needs a rebase.

Done.  Passed except for JRuby 9.2.8.0...

nateberkopec:
Congrats, that was a lot of work @MSP-Greg 👏

MSP-Greg:
Thanks.  The intermittent tests were driving me crazy, probably the same for you.  IDK about JRuby 9.2.8.0, seems to need more work.

Now we'll move on to the next phase.  I saw #1976, and wanted to write some tests for how Puma started when a UNIX control and/or bind socket existed already, or if the file existed, but no sockets.

Couldn't help but think how much easier it would be if #1971 was in place...

BTW, the gist is totally wrong, I've already got an update that I'll start working with.  Rather than setting config via a call to `cli_server`, we need a method like `setup_puma` which would be the first thing done in a test.  That allows some skip statements to move into it, and also allows for tests that require something created before starting Puma, like in #1976...

PR #1973, opened 2019-09-16
PluginRegistry#fire_background - fix up per issue 1972

MSP-Greg:
1. PluginRegistry#fire_background - fix

2. Add test_plugin.rb with a simple test

May be the first test with `Puma::ControlCLI` that's using TCP, so it runs on Windows...

### DIFF
--- a/lib/puma/plugin.rb
+++ b/lib/puma/plugin.rb
@@ -62,9 +62,9 @@ module Puma
     end
 
     def fire_background
-      @background.each do |b|
+      @background.each_with_index do |b, i|
         Thread.new do
-          set_thread_name "plugin background"
+          Puma.set_thread_name "plugin background #{i}"
           b.call
         end
       end
--- a/test/test_plugin.rb
+++ b/test/test_plugin.rb
@@ -1,0 +1,39 @@
+require_relative "helper"
+require_relative "helpers/integration"
+
+class TestPlugin < TestIntegration
+  def test_plugin
+    skip "Skipped on Windows Ruby < 2.5.0, Ruby bug" if windows? && RUBY_VERSION < '2.5.0'
+    @tcp_bind = UniquePort.call
+    @tcp_ctrl = UniquePort.call
+
+    Dir.mkdir("tmp") unless Dir.exist?("tmp")
+
+    cli_server "-b tcp://#{HOST}:#{@tcp_bind} --control-url tcp://#{HOST}:#{@tcp_ctrl} --control-token #{TOKEN} -C test/config/plugin1.rb test/rackup/hello.ru"
+    File.open('tmp/restart.txt', mode: 'wb') { |f| f.puts "Restart #{Time.now}" }
+
+    true while (l = @server.gets) !~ /Restarting\.\.\./
+    assert_match(/Restarting\.\.\./, l)
+
+    true while (l = @server.gets) !~ /Ctrl-C/
+    assert_match(/Ctrl-C/, l)
+
+    out = StringIO.new
+
+    cli_pumactl "-C tcp://#{HOST}:#{@tcp_ctrl} -T #{TOKEN} stop"
+    true while (l = @server.gets) !~ /Goodbye/
+
+    @server.close
+    @server = nil
+    out.close
+  end
+
+  private
+
+  def cli_pumactl(argv)
+    pumactl = IO.popen("#{BASE} bin/pumactl #{argv}", "r")
+    @ios_to_close << pumactl
+    Process.wait pumactl.pid
+    pumactl
+  end
+end
--- a/test/config/plugin1.rb
+++ b/test/config/plugin1.rb
@@ -1,0 +1,1 @@
+plugin 'tmp_restart'


MSP-Greg:
I didn't add/fix it here, but some of the IO operations in events.rb intermittently cause `Errno::EPIPE` errors on shutdown...

MSP-Greg:
For quite a while I've been working on worker/UNIXSocket issues/tests, and all the work I had to push to Travis/Actions.

The issue here I could finally work on locally.  So the test file individually, and the whole test suite **runs fine locally**, but in the AppVeyor build here and a build I just ran in my fork, both locked up.

JFYI, also ran a Travis build, all passed but one 'standard' failure on one macOS job.  See:
https://travis-ci.org/MSP-Greg/puma/builds/585830442

I hate this.  Please hold until I figure out what's happening on AppVeyor...

MSP-Greg:
The current Appveyor build system constructs a pre-compiled gem, installs it, then runs the tests.

The single test in test_plugin.rb loads the 'sample' plugin, lib/puma/plugin/tmp_restart.rb, and sees if it restarts Puma.  Note that the test failed on current master.

Anyway, the gem does not have a 'tmp' directory.  I did locally and Travis does from compiling.  The sample plugin requires it.  Added a line to it to create the directory if it doesn;'t exist.

Should pass now.  It's already completed tests on two versions of Ruby...

MSP-Greg:
I've worked on this enough, don't know if there are better names for the plugin threads.

Final result is use `cli_pumactl`, and Window Ruby 2.4 (and maybe earlier) have a bug that I may or may not try to track down.  Also, Actions now has Ruby 2.3.x on Ubuntu & macOS.

So, before I skipped Windows 2.4.x and earlier, Actions is here:
https://github.com/MSP-Greg/puma/commit/9b9791a0340b43e76a1fa2809257bc6e6391e638/checks

Actions after I added the skip:
https://github.com/MSP-Greg/puma/commit/70d9bafc85daaa0c8fd063ef343694ecf98441b6/checks

Tests are still wonderfully intermittent.

lib/puma/plugin.rb
```diff
@@ -62,10 +62,16 @@ def add_background(blk)
...
 
     def fire_background
-      @background.each do |b|
-        Thread.new do
-          set_thread_name "plugin background"
-          b.call
+      # remove conditional when Ruby 2.2 support is dropped,
+      # no Thread#name method
+      if Thread.current.respond_to?(:name=)
+        @background.each_with_index do |b, i|
```
nateberkopec: set_thread_name already has code that deals with this, you don't need to do it here.

lib/puma/plugin/tmp_restart.rb
```diff
@@ -10,6 +10,7 @@ def start(launcher)
 
     # If we can't write to the path, then just don't bother with this plugin
     begin
+      Dir.mkdir("tmp") unless Dir.exist?("tmp")
```
nateberkopec: Are we adding this purely so that the tests work?

test/test_plugin.rb
```diff
@@ -0,0 +1,40 @@
+require_relative "helper"
+require_relative "helpers/integration"
+
+class TestPlugin < TestIntegration
+  def test_plugin
+    skip "Skipped on Windows & Ruby 2.4.x" if windows? && RUBY_VERSION < '2.5.0'
```
nateberkopec: Should leave a comment here as to why

test/test_plugin.rb
```diff
@@ -0,0 +1,40 @@
...
+  private
+
+  def cli_pumactl(argv)
+    pumactl = IO.popen("#{BASE} bin/pumactl #{argv}", "r")
+    @ios_to_close << pumactl
+    Process.wait pumactl.pid
+    pumactl
+  end
+
+
```
nateberkopec: rm extra whitespace

lib/puma/plugin.rb
nateberkopec: Isn't the fix purely that we should be calling `Puma.set_thread_name`?

lib/puma/plugin/tmp_restart.rb
nateberkopec: point being, can't we do this in the test?

lib/puma/plugin.rb
MSP-Greg: The only reason `Puma.set_thread_name` exists is to support Ruby 2.2.x.  Come Christmas time...

test/test_plugin.rb
MSP-Greg: IDC.  It's Windows specific, and 2.4 is 'Currently in security maintenance phase.'  On Windows, I suspect there are very few people using 2.4.

test/test_plugin.rb
MSP-Greg: I would have thought my favorite software (RuboCop) would have picked that...

test/test_plugin.rb
MSP-Greg: Fixed.

test/test_plugin.rb
MSP-Greg: Fixed.

lib/puma/plugin/tmp_restart.rb
MSP-Greg: No.  Or, maybe on Windows as it's currently tested in Appveyor.  We're doing it because compiling creates the directory.  It won't work without compiling.  If it's a 'sample', it shouldn't require that the `tmp` folder exist.

lib/puma/plugin.rb
nateberkopec: We're going to be supprting 2.2 for the forseeable future so I'd prefer we stick to the method that already exists.

lib/puma/plugin/tmp_restart.rb
nateberkopec: I really don't want to modify the behavior of a years-old plugin just we can get our appveyor tests to pass (which already have an expiration date due to Github CI)

MSP-Greg: I'm think more of someone wanting use it after installing the gem.  Does the `tmp` folder exists then?

nateberkopec: When it comes to touching files on people's deployments, I really don't want to do it. Most people's deploys are an absolute house of cards. If we don't absolutely need to do this and no one has requested it in issues, I don't want to do it.

MSP-Greg: To even use the plugin, it will touch files on people's deployments.

Maybe we should add a rescue with a message that the folder must exist?

lib/puma/plugin.rb
MSP-Greg: Done.

MSP-Greg:
I fixed whatever I broke earlier today...

lib/puma/plugin/tmp_restart.rb
nateberkopec: what if Puma doesn't have permissions to create the folder? now this plugin (silently) dies where before it would continue to work. People may have deploy scripts out there that work around this and a different user or privileged script may create the folder after Puma boots.

My point is that this is actually a major API change that's unneccessary for this PR is trying to accomplish, regardless of where or not it's a good idea to do this.

nateberkopec: Like I feel like we can almost guarantee someone will create an issue about this if we deploy this in 4.2.

MSP-Greg: > Like I feel like we can almost guarantee someone will create an issue

Bring it on!

> a major API change

In an 'example' plugin file?

Anyway, this seems to be a 'hot button' issue, so I moved the statement in question to the test....

MSP-Greg: @nateberkopec @dentarg

I just did a fresh install of puma in 2.6.4, and set up an existing server script (for viewing docs locally) to use the plugin.

With no 'tmp' folder, there was no message, but it didn't create `restart.txt` either, so the plugin wasn't loaded.

When I added the `tmp` folder, it works.

Hence, the current plugin will not load without an existing 'tmp' folder.

Maybe one of your could try it?

dentarg: I think @nateberkopec is right here @MSP-Greg 

The [docs are pretty clear](https://github.com/puma/puma/blob/v4.1.1/docs/plugins.md)

> tmp_restart: Restarts the server if the file `tmp/restart.txt` is touched

If anyone wants to use this plugin, they will create the directory. So I think the tests should do that. 

(I can also add that `tmp/` exists by default on Heroku, as their build system puts stuff there.)

> In an 'example' plugin file?

`lib/puma/plugin/tmp_restart.rb` doesn't look like an example to me, it is the plugin, isn't it?

MSP-Greg: @dentarg

Thanks.

> `lib/puma/plugin/tmp_restart.rb` doesn't look like an example to me, it is the plugin, isn't it?

Maybe we'll say a simple example of the hooks/methods needed to create a simple plugin?

To use it, one needs to create a config file with the line `plugin 'tmp_restart'` and add it to the puma command arguments, along with `-C`.

The current plugin is using the equivalent of `"#{Dir.pwd}/tmp"`, which may or may not exist...

PR #1974, opened 2019-09-18
4.2.0

nateberkopec:
See also the open milestone, which contains everything remaining to be done before this release.

### DIFF
--- a/lib/puma/const.rb
+++ b/lib/puma/const.rb
@@ -100,8 +100,8 @@ module Puma
   # too taxing on performance.
   module Const
 
-    PUMA_VERSION = VERSION = "4.1.1".freeze
-    CODE_NAME = "Fourth and One".freeze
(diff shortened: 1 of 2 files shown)


lib/puma/const.rb
```diff
@@ -100,8 +100,8 @@ class UnsupportedOption < RuntimeError
   # too taxing on performance.
   module Const
 
-    PUMA_VERSION = VERSION = "4.1.1".freeze
-    CODE_NAME = "Fourth and One".freeze
+    PUMA_VERSION = VERSION = "4.2.0".freeze
+    CODE_NAME = "Distant Airhorns".freeze
```
dentarg: How do you pick these? 🤔

nateberkopec:
Gonna let the dust settle a bit here and then release monday morning.

Issue #1986, opened 2019-09-24
Error after binding to socket and port in one config


flaminestone:
Puma can not to be configuring for work with socket and port in one time.
In previous version (4.1.1) all work fine.
**Error**:
```
Traceback (most recent call last):
	12: from /usr/local/bundle/bin/puma:23:in `<main>'
	11: from /usr/local/bundle/bin/puma:23:in `load'
	10: from /usr/local/bundle/gems/puma-4.2.0/bin/puma:10:in `<top (required)>'
	 9: from /usr/local/bundle/gems/puma-4.2.0/lib/puma/cli.rb:80:in `run'
	 8: from /usr/local/bundle/gems/puma-4.2.0/lib/puma/launcher.rb:172:in `run'
	 7: from /usr/local/bundle/gems/puma-4.2.0/lib/puma/single.rb:98:in `run'
	 6: from /usr/local/bundle/gems/puma-4.2.0/lib/puma/runner.rb:154:in `load_and_bind'
	 5: from /usr/local/bundle/gems/puma-4.2.0/lib/puma/binder.rb:89:in `parse'
	 4: from /usr/local/bundle/gems/puma-4.2.0/lib/puma/binder.rb:89:in `each'
	 3: from /usr/local/bundle/gems/puma-4.2.0/lib/puma/binder.rb:107:in `block in parse'
	 2: from /usr/local/bundle/gems/puma-4.2.0/lib/puma/binder.rb:107:in `each'
	 1: from /usr/local/bundle/gems/puma-4.2.0/lib/puma/binder.rb:111:in `block (2 levels) in parse'
/usr/local/bundle/gems/puma-4.2.0/lib/puma/binder.rb:111:in `ip_unpack': need IPv4 or IPv6 address (SocketError)
``` 

**Puma config:**
```
bind "unix:///puma.sock"
bind 'tcp://0.0.0.0:3000'
```

**To Reproduce**
File with application (app.rb):

```ruby
require 'sinatra'
class App < Sinatra::Base
  get '/' do
    render 'ok!'
  end
end
```

File with puma config (config.puma.rb)
```
bind "unix:///puma.sock"
bind 'tcp://0.0.0.0:3000'
```
sinatra config (config.ru):
```
require_relative 'app'

run Rack::URLMap.new('/' => App)

configure do
  set :server, :puma
end
```

Gemfile:
```
source 'https://rubygems.org'
gem 'puma', '4.2.0'
# gem 'puma', '4.1.1'
[...]

PR #1987, opened 2019-09-24
binder.rb - fix using both UNIX and TCP listeners

MSP-Greg:
See issue #1986 

1st Commit - 'binder.rb - fix double binds and misc'

1. Fix mult binds when unix is first, followed by ip bind.  See issue #1986.

2. When localhost is used for ssl or tcp, output actual ip's bound to.

3. Fix issues with IPv6 bind logging (brackets).

4. Move ssl context initialization code from Binder#parse to new method, #ssl_ctx_init.

2nd Commit - 'test_binder.rb - add double bind tests, misc'

1. Add test combinations of multiple binds, localhost, port 0, etc.

2. Allow most tests using ssl binders to run on JRuby, including `MiniSSL::Context#no_tlsv*` tests, which all were previously only run on MRI Ruby.

3. Close sockets created by Binder#parse.

4. Refactor code.

5. Some redundant asserts.  Could use cleanup to make more orthogonal.

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,10 +1,10 @@
 ## Master
 
 * Features
-  * Your feature goes here (#Github Number)
+  * binder.rb - cleanup handling of port 0, localhost, etc.  Add tests (#1987)
 
 * Bugfixes
-  * Your bugfix goes here (#Github Number)
+  * binder.rb - handle 'double binds' when unix, then ip.  Add tests (#1987)
 
 ## 4.2.0 / 2019-09-23
 
--- a/lib/puma/binder.rb
+++ b/lib/puma/binder.rb
@@ -86,6 +86,7 @@ module Puma
     end
 
     def parse(binds, logger)
+      @logger = logger
       binds.each do |str|
         uri = URI.parse str
         case uri.scheme
@@ -103,16 +104,6 @@ module Puma
             bak = params.fetch('backlog', 1024).to_i
 
             io = add_tcp_listener uri.host, uri.port, opt, bak
-
-            @ios.each do |i|
-              addr = if i.local_address.ipv6?
-                "[#{i.local_address.ip_unpack[0]}]:#{i.local_address.ip_unpack[1]}"
-              else
-                i.local_address.ip_unpack.join(':')
-              end
-
-              logger.log "* Listening on tcp://#{addr}"
-            end
           end
 
           @listeners << [str, io] if io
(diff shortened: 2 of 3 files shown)


MSP-Greg:
re forced push, I think `skipped?` may only work in `teardown`?

nateberkopec:
Closing in favor of #1989

nateberkopec:
that one got too complicated, coming back to this

MSP-Greg:
@nateberkopec

I spent some more time looking at this, and decided to add tests for activated & inherited sockets. 
 Turned into a rabbit hole kind of thing, but I think I've got it working.  Also, added needed cleanup for sockets (Binder @ios, etc).

See the files changed in https://github.com/MSP-Greg/puma/tree/unix-and-tcp-listeners-t.  Last four commits, except for 'local test rakefile, Action' (`Rakefile_test` lets me run locally tests on a branch other than 'master' in my fork, and the changes to `ruby.yml` allow me to run Actions given the same).

I could separate it into another PR, but that may have merge issues is anything is changed in the first.  Thoughts?

Question - `Binder#import_from_env` loads `@inherited_fds` and `@activated_sockets`.

Inherited sockets (`PUMA_INHERIT_\d`) have an exact fd passed via individual ENV settings, and the ENV settings are created by Puma.

Activated sockets ( systemd `LISTEN_FDS`) have a single number passed in, and fd's are looped looking for the matching uri.  But, binder fd(s) may not be sequential above STDERR (3, 4, etc).

The code (in `Binder#parse`) below closes/deletes any unused fd's in :

https://github.com/puma/puma/blob/079b284d13e2d9920b29ce234a0fa3e0deae569e/lib/puma/binder.rb#L250-L260

Aren't we removing fd's in `@activated_sockets` that we possibly shouldn't?

nateberkopec:
@MSP-Greg re: those questions, please contact me via the email address on my profile

dentarg:
Why discuss offline instead of GitHub?

MSP-Greg:
Patrik,

> Why discuss offline instead of GitHub?

**OFF TOPIC**

I think I can state that recently both Nate & I said something similar to "let's focus on a few existing, fixable issues and get a point release out", referring to this and the systemd issues.  I think we both also started on things that morphed into larger, more involved changes, which went beyond what is needed for the point release.

My post above was partially in frustration, as some of the work I did in my fork stabilized test_binder.rb, but I'd need to cherry pick those changes back into this PR, and take the additional test work (and its changes to the lib files) for another PR.

Said another way, sometimes it's very frustrating to work on A (this PR), then work on B (my branch), and realize that B needs to be divided, one part moving back into A, and the balance saved for later.  Been, there, done that, and it requires a lot of concentration, as A and B are stable, but dividing them may make A unstable until the 'division' is 'perfect'.

Summing up, we're doing a 'bug fix' push.  Next push is for refactoring, better test coverage, improvements, etc.

So, I'm going to start working on that division.  Nate & I have not communicated, so I'm not speaking for him in any way.

Sorry for being off-topic, Greg

MSP-Greg:
Updated.

MSP-Greg:
With the last commit, most tests using ssl binders run with JRuby, which is a change from the previous version, where very few ran.

See:

https://travis-ci.org/puma/puma/jobs/590983851#L411
and
https://travis-ci.org/puma/puma/jobs/590983851#L538

MSP-Greg:
Luckily, last Travis CI build allowed both JRuby jobs to run TestBinder before they locked up.

All passed, including the `MiniSSL::Context#no_tlsv*` tests, which were previously only run on MRI Ruby.

9.2.8.0 - https://travis-ci.org/puma/puma/jobs/591284029#L352-L365
head - https://travis-ci.org/puma/puma/jobs/591284032#L414-L427

nateberkopec:
Ok, now THIS one's too complicated. We need just tests + a simple hack fix for 4.2.1, not a full refactor (which is also what I attempted too). I know it's tempting because it's obviously going to help a lot of the bugs we're seeing around activation. The problem with a full refactor here is that the test coverage on binders is only _ok_ and it's certainly nonexistent for inheriting/activating.

MSP-Greg:
The 'double path' bug has been fixed

My current version of test/test_binder.rb is compatible with master, but I'd like to add more test coverage.  So, I can either update this, or close this and create another PR.

Thoughts?

@nateberkopec - your call

nateberkopec:
Closing in favor of #1989

PR #1989, opened 2019-09-25
Binder refactor

nateberkopec:
* [x] Bindings inheritance model? SSLBinder, UnixBinder, etc

Doing an inheritance model could probably help me take a lot of the complexity out of Binder#parse and the various inherit_ and add_ listener methods...

### DIFF
--- a/lib/puma/binder.rb
+++ b/lib/puma/binder.rb
@@ -149,60 +149,7 @@ module Puma
 
           MiniSSL.check
 
-          ctx = MiniSSL::Context.new
-
-          if defined?(JRUBY_VERSION)
-            unless params['keystore']
-              @events.error "Please specify the Java keystore via 'keystore='"
-            end
-
-            ctx.keystore = params['keystore']
-
-            unless params['keystore-pass']
-              @events.error "Please specify the Java keystore password  via 'keystore-pass='"
-            end
-
-            ctx.keystore_pass = params['keystore-pass']
-            ctx.ssl_cipher_list = params['ssl_cipher_list'] if params['ssl_cipher_list']
-          else
-            unless params['key']
-              @events.error "Please specify the SSL key via 'key='"
-            end
-
-            ctx.key = params['key']
-
-            unless params['cert']
-              @events.error "Please specify the SSL cert via 'cert='"
-            end
-
-            ctx.cert = params['cert']
-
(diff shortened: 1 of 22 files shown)


lib/puma/binder.rb
```diff
@@ -229,6 +217,18 @@ def parse(binds, logger)
...
       end
 
+      @ios.each do |i|
+        protocol, addrinfo = case i
+        when TCPServer
+          ["tcp", i.local_address]
+        when UNIXServer
+          ["unix", i.local_address]
+        when Puma::MiniSSL::Server
+          ["ssl", i.to_io.local_address]
```
nateberkopec: This has the advantage of logging port 0 binds with SSL correctly too.

nateberkopec:
@MSP-Greg thoughts?

nateberkopec:
Oh, on windows, UNIXServer isn't even defined?

nateberkopec:
I need to think about how this affects inherited listeners, and whether or not I should be iterating through `@listeners` or `@ios`

nateberkopec:
I don't think it matters whether or not I'm iterating through listeners or ios. Unfortunately, those are _supposed_ to be the same data AFAICT. I think we should probably just have one collection in binders (merging ios and listeners), a collection of a new value object (call it Binding?). That would make a lot of stuff a lot nicer.

I don't really want to do that refactoring right now but I think I have to in order to log inherited sockets once rather than twice (which this will now do).

nateberkopec:
This is ending up a lot more intense than something that can just fix that bug and go into a patch release. Once I finish this I'll walk it back to something that can, maybe even just Greg's original fix + test.

nateberkopec:
I think it's time for me to stop the spike for now and pull back a few of these things into master.

MSP-Greg:
@nateberkopec

As to the emoji I added, it reminded me of how "I want to fix a few issues" often turns into a lot more.  At least that's been the case with me...

> I don't think it matters whether or not I'm iterating through listeners or ios. Unfortunately, those are supposed to be the same data AFAICT.

**I think** the current code only loads listeners in the parse code.  ios is loaded in the add/inherit code. So, control url's aren't in listeners, and any sockets created via re-entrant calls in the add/inherit code are also not.  **I think**...

nateberkopec:
Closing as unmergeable, but I do want to revisit the ideas in here very soon.

Issue #1971, opened 2019-09-16
Test Reorganization


MSP-Greg:
1. Cluster Integration Tests - Two recent concerns are correct handling of unlinking UNIXSocket files and requests being handled correctly during 'stop'.  Regarding controlling Puma, the code paths are different dependent on whether control is via a socket or via a signal.  So, a 'full matrix' would have the following:<br/><br/>Server Bind: TCP, UNIX<br/>Server Control: TCP, UNIX, Signal<br/><br/>I'm not suggesting all tests be performed on all six combinations, but there certainly some critical tests that should be tested on more than the current two.  This would require changing the arguments for `cli_server` and `cli_pumactl`, allowing code to be shared between tests.<br/><br/>Maybe two kwargs - `bind:` & `ctrl:` for `cli_server`, and `ctrl` would set an instance variable so `cli_pumactl` would sense it?

2. `TestIntegrationPumactl#cli_pumactl` (test_integration_pumactl.rb)  - Rather than spinning up `ruby  pumactl`, should we just use `Puma::ControlCLI.new(<args>).run`?  Any IO issues?

3. test_cli.rb - Given that every call to `cli_server` is using Puma::CLI, I don't see a reason to start Puma in its tests.  But, since breakage in it will cascade to other tests, testing using its instance variables, etc is certainly a good idea.  May need to run up a server to test Puma::Events.  Maybe add a test_events.rb file?

Thoughts?

nateberkopec:
Re: 1, great point. two kwargs makes sense.

2. I think we need a few test for "full up integration" of the ControlCLI. If removing that one would mean there are no tests that run a real control process, then we should not remove it. Otherwise, we can.

3. More tests is always better!

MSP-Greg:
Cool.  Given the rebasing and dicing/chopping I've done with current code changes, I'd like to get a few PR's merged before I start on these changes.

MSP-Greg:
Re point 2, I did say 'Any IO issues?'

After wrestling with #1973, yes, there are IO issues.  I started with a simple test and `Puma::ControlCLI.new(<args>).run`.  No, just no.  There's some IO mapping in Puma::ControlCLI, and tests just got weird.  So, we should stick with `cli_pumactl`...

MSP-Greg:
Once we get the current set of PR's merged, I will start working on this.  After some more thought, the above:

Server Bind (bind: kwarg): TCP, UNIX
Server Control (ctrl: kwarg): TCP, UNIX, Signal

[...]

PR #1990, opened 2019-09-25
Test timeout gets [changelog skip]

nateberkopec:
Discovered this working locally and wondering why my tests were hanging.

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -45,6 +45,22 @@ def hit(uris)
   end
 end
 
+module WaitForServerLogs
+  def wait_until_server_logs(msg, server: @server)
+    regex = Regexp === msg ? msg : Regexp.new(Regexp.escape msg)
+    l = ""
+
+    Timeout.timeout(ENV["CI"] ? 30 : 5) do
+      while true
+        l = server.gets
+        l[regex]
+      end
+    end
+
+    assert l[regex], "Server timeout waiting for '#{msg}' in:\n    #{caller(1..2).join "\n    "}"
+  end
+end
+
 module UniquePort
   @port  = 3211
   @mutex = Mutex.new
--- a/test/helpers/integration.rb
+++ b/test/helpers/integration.rb
@@ -6,6 +6,8 @@ require "open3"
 # Only single mode tests go here. Cluster and pumactl tests
 # have their own files, use those instead
 class TestIntegration < Minitest::Test
+  include WaitForServerLogs
+
   HOST  = "127.0.0.1"
   TOKEN = "xxyyzz"
   WORKERS = 2
@@ -82,7 +84,7 @@ class TestIntegration < Minitest::Test
   end
 
   def wait_for_server_to_boot
-    true while @server.gets !~ /Ctrl-C/ # wait for server to say it booted
+    wait_until_server_logs(/Ctrl-C/) # wait for server to say it booted
   end
 
   def connect(path = nil, unix: false)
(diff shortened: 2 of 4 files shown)


MSP-Greg:
Good idea.

> Discovered this working locally

Had the same problem.

> `def wait_until_server_logs(server, msg)`

Maybe something like `def wait_until_server_logs(msg, server: @server)`?

I guess I like kwargs for optional parameters...

I'm working on this issue in a branch, incorporating the ideas of the 'setup_puma' concept (https://github.com/puma/puma/issues/1971).

The issue with needing a pid, (and hence, another process) to test signalling leaves one without much info if something crashes.  I changed to using spawn, grabbing stderr, which (sometimes) has info re the crash.  Also fixes an issue on Windows (and maybe other OS's?) when using `Kernel.exec` in the 'other' process.

Adding a method like this makes it easier to throw all the output to the console, and if it times out, check the stderr output...

nateberkopec:
The failure for Windows on 2.5/2.6 is really weird: https://github.com/puma/puma/pull/1990/checks?check_run_id=241355633#step:9:104

MSP-Greg:
My equivalent to `wait_until_server_logs` (`wait_server`) uses `IO.select` instead of `Timeout.timeout`, which is Ruby and adds another thread.  It also has a `debug:` kwarg, which dumps everything read/gotten to STDOUT as it's looking for the match.

It seems to be working well, and JFYI, when I changed `parallelize_me!` so that it was disabled for JRuby in a few files, JRuby CI became a lot more stable and even passes sometimes.

In general, we've got parallel testing, which adds a thread, `TimeoutEveryTestCase` adds another thread, etc, etc.  Given that many of the JRuby CI logs show a lot of 'waiting on join' kinds of timeouts,  I figured switching to `IO.select` might be helpful.  It may even be possible to eliminate `TimeoutEveryTestCase`...

> The failure for Windows on 2.5/2.6 is really weird

I think Windows Ruby 2.4 is skipped.  I've got all versions passing in my 'enhanced test framework'...

MSP-Greg:
@nateberkopec

Here's the code I was using.  I added quite a bit for string/regex, debugging, etc.
```ruby
def wait_server(re, server: @server, debug: false, timeout: 10)
  regex = Regexp === re ? re : Regexp.new(Regexp.escape re)
  STDOUT.puts('',
    "#{' ' * 20}*** #{full_name} wait_server #{re} ***") if debug
  l = ''

  while IO.select([server], nil, nil, timeout) do
    l = server.gets
    STDOUT.puts "#{l}" if debug
    break if l[regex]
  end
  # TODO add debugging output to use server STDERR if available
  assert l[regex], "Server timeout waiting for '#{re}' in:\n    #{caller(1..2)}"
  l
end
```

Re the TODO comment, I switched to spawn from IO.popen, so stderr is available, but often doesn't have any helpful info.

nateberkopec:
Thanks for the code. Windows 2.5./2.6 still failing on `TestLauncher#test_dependencies_and_files_to_require_after_prune_is_correctly_built_with_extra_deps = 0.01 s = .
^CThe process tried to write to a nonexistent pipe.`

I think this is coming from the Plugin test class.

nateberkopec:
Hm. Still failing on https://github.com/puma/puma/pull/1990/checks?check_run_id=246697262#step:9:75

MSP-Greg:
I think :INT below needs to change to :KILL or Windows may have issues:

https://github.com/puma/puma/blob/60d2ab5898eb0ab7a38d2f0b4b9c920fd5f90312/test/helpers/integration.rb#L23-L26

Also, I know I revised the below, something like `#{caller(1..2).join "\n    "}`, and I may have changed the indexes (four spaces after \n)...
```ruby
assert l[regex], "Server timeout waiting for '#{re}' in:\n    #{caller(1..2)}"
```

PR #1991, opened 2019-09-25
[ci skip] Require Changelog at PR time

schneems:
We frequently forget to ask for a changelog entry, and then when we want to release the release maintainer has to go back and manually add them all. It would be better if the changelog was required at PR time so the original submitter can think of the best wording. If a PR is trivial people can add `[ci skip]` or `[changelog skip]`.

This implementation uses github actions.


nateberkopec:
I think in a perfect world the action would also comment on failure, but that seems a lot more complicated to do with the actions kit rn, so lets 🚢

nateberkopec:
whoops luckily i merged that 10 seconds after you pushed that fix 😆

PR #1992, opened 2019-09-25
Revert close unix paths

MSP-Greg:
For testing with systemd


MSP-Greg:
Closing.  See https://github.com/puma/puma/pull/2007

PR #1993, opened 2019-09-25
[changelog skip] Document the changelog check in README

dentarg:
Follow-up to https://github.com/puma/puma/pull/1991


dentarg:
Should probably have used `[ci skip]` 🤦‍♂ oh well

PR #2002, opened 2019-09-27
Increase max size of querystring and url in parser

JohnnyTing:


JohnnyTing:
feat: update the QUERY_STRING  default value

nateberkopec:
I'm not sure how to evaluate a change like this. We get this request every once in a while ("please just make the buffers bigger in the parser"), but I don't know how to evaluate the pros/cons of such a request.

Issue #1997, opened 2019-09-26
One test is logging a warning about SSL session not being started yet


nateberkopec:
`/Users/nateberkopec/.rubies/ruby-2.6.4/lib/ruby/2.6.0/openssl/buffering.rb:322: warning: SSL session is not started yet.`

The last remaining shit in our green dots, after this, Operation No Shit In Our Green Dots is complete

Confusion:
The test logging this is `test/test_puma_server_ssl#test_request_wont_block_thread`. The reason is that it writes something to an SSLSocket without having set up an SSL session. The warning originates from MRI's `ext/openssl/ossl_ssl.c`and is unconditional. If you add a `socket.connect` you'll get a 
```
HTTP parse error, malformed request (): #<Puma::HttpParserError: Invalid HTTP format, parsing fails.>
```
so then the thread isn't blocked by definition and the test doesn't achieve its goal. 

So I think there are possible solutions:
1) Suppress Ruby warnings around this test
2) Start the SSL session to such an extent that the warning isn't emitted, but the connection still expects data. 

I'm not sure if the second is possible and I'm not sure if the first would be acceptable to you.

MSP-Greg:
See PR https://github.com/puma/puma/pull/2003.  Choose option 1, $VERBOSE

nateberkopec:
Good digging @Confusion. First isn't acceptable to me (let's just fix it!).

AndrewSpeed:
:wave: I'm willing to take a look at this one, though I'll admit I'm pretty unfamiliar with the `OpenSSL` module

From what I can tell so far, we'll need to call `OpenSSL::SSL::SSLSocket#connect` in order to start the SSL session.

After doing this, the next hurdle I've run into is `OpenSSL::SSL::SSLSocket#verify_result` returning `20`, which [maps to `X509_V_ERR_UNABLE_TO_GET_ISSUER_CERT_LOCALLY`](https://linux.die.net/man/1/verify), so I'm assuming we'll need to have an SSL certificate generated for the tests in order to get past this issue.

Does this :point_up: sound like I'm going down the right path so far?

nateberkopec:
> Does this ☝️ sound like I'm going down the right path so far?

Yup!

PR #2003, opened 2019-09-30
Fix test server ssl.rb warning

MSP-Greg:
Error is generated in 'client code', swap $VERBOSE for statement only, ensure in case...

See https://github.com/puma/puma/issues/1997

### DIFF
--- a/History.md
+++ b/History.md
@@ -5,6 +5,7 @@
 
 * Bugfixes
   * Your bugfix goes here (#Github Number)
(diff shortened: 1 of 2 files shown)


nateberkopec:
> Error is generated in 'client code',

What do you mean?

I'd rather just fix the issue than disable the warning. Also this test is parallelized so we can't just be modifying globals in it.

nateberkopec:
Glad you figured out which test it was though!

PR #2010, opened 2019-10-04
Strip header whitespace. Fix #1890. Code by @matthewd

### LINKED ISSUE #1890: Trailing whitespace in headers is not stripped, violating RFC 2616/7230
### Steps to reproduce

With config.ru:

```ruby
run(proc{|env| [200, {}, [env['HTTP_FOO'].inspect, "\n"]]})
```

Start server:

```
rackup -o 127.0.0.1
```

Test with header with trailing whitespace:

```
curl -HFoo:"  Bar  Blah  " http://127.0.0.1:9292
```

### Expected behavior

Output should be:

```
"Bar  Blah"
```

### Actual behavior

Output is:

```
"Bar  Blah  "
```

### System configuration

Not dependent on system configuration.

I believe this is a violation of RFC 2616 section 4.2 :

```
   The field-content does not include any leading or trailing LWS:
   linear white space occurring before the first non-whitespace
   character of the field-value or after the last non-whitespace
   character of the field-value. Such leading or trailing LWS MAY be
   removed without changing the semantics of the field value.
```

and RFC 7230 section 3.2.4:

```
   A field value might be preceded and/or followed by optional
   whitespace (OWS); a single SP preceding the field-value is preferred
   for consistent readability by humans.  The field value does not
   include any leading or trailing whitespace: OWS occurring before the
   first non-whitespace octet of the field value or after the last
   non-whitespace octet of the field value ought to be excluded by
   parsers when extracting the field value from a header field.
```

Unicorn and Webrick both correctly strip the trailing whitespace.  Thin has the same bug as Puma, but appears dead upstream.

nateberkopec:
Needs a java equivalent.

### DIFF
--- a/ext/puma_http11/puma_http11.c
+++ b/ext/puma_http11/puma_http11.c
@@ -200,6 +200,8 @@ void http_field(puma_parser* hp, const char *field, size_t flen,
     f = rb_str_new(hp->buf, new_size);
   }
 
+  while (vlen > 0 && isspace(value[vlen - 1])) vlen--;
+
   /* check for duplicate header */
   v = rb_hash_aref(hp->request, f);
 
--- a/History.md
+++ b/History.md
@@ -1,5 +1,5 @@
 * Features
-  * Your feature goes here (#Github Number)
+  * Strip whitespace at end of HTTP headers (#2010)
 
 * Bugfixes
   * Your bugfix goes here (#Github Number)
--- a/ext/puma_http11/org/jruby/puma/Http11.java
+++ b/ext/puma_http11/org/jruby/puma/Http11.java
@@ -87,7 +87,9 @@ public class Http11 extends RubyObject {
                 validateMaxLength(flen, MAX_FIELD_NAME_LENGTH, MAX_FIELD_NAME_LENGTH_ERR);
                 validateMaxLength(vlen, MAX_FIELD_VALUE_LENGTH, MAX_FIELD_VALUE_LENGTH_ERR);
 
-                ByteList b = new ByteList(Http11.this.hp.parser.buffer,field,flen);
+                ByteList buffer = Http11.this.hp.parser.buffer;
+
+                ByteList b = new ByteList(buffer,field,flen);
                 for(int i = 0,j = b.length();i<j;i++) {
                     if((b.get(i) & 0xFF) == '-') {
                         b.set(i, (byte)'_');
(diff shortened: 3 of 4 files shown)


ext/puma_http11/puma_http11.c
```diff
@@ -200,6 +200,8 @@ void http_field(puma_parser* hp, const char *field, size_t flen,
     f = rb_str_new(hp->buf, new_size);
   }
 
+	while (vlen > 0 && isspace(value[vlen - 1])) vlen--;
```
olleolleolle: Is the whitespace wonky here?

Jesus:
You might know this already but just in case, I want to advertise a little known feature of git. You can create a commit with several authors, just add this to your commit message:

    Co-authored-by: Matthew Draper <matthew@trebex.net>

More at: https://help.github.com/en/articles/creating-a-commit-with-multiple-authors

ext/puma_http11/puma_http11.c
nateberkopec: ya I think it is

headius:
@nateberkopec This patch should do it. It's larger than yours because there were duplicate loads of the parse buffer that I didn't want to duplicate a third time.

```diff
diff --git a/ext/puma_http11/org/jruby/puma/Http11.java b/ext/puma_http11/org/jruby/puma/Http11.java
index 59dde372..7e38c2b2 100644
--- a/ext/puma_http11/org/jruby/puma/Http11.java
+++ b/ext/puma_http11/org/jruby/puma/Http11.java
@@ -87,7 +87,9 @@ public class Http11 extends RubyObject {
                 validateMaxLength(flen, MAX_FIELD_NAME_LENGTH, MAX_FIELD_NAME_LENGTH_ERR);
                 validateMaxLength(vlen, MAX_FIELD_VALUE_LENGTH, MAX_FIELD_VALUE_LENGTH_ERR);
 
-                ByteList b = new ByteList(Http11.this.hp.parser.buffer,field,flen);
+                ByteList buffer = Http11.this.hp.parser.buffer;
+
+                ByteList b = new ByteList(buffer,field,flen);
                 for(int i = 0,j = b.length();i<j;i++) {
                     if((b.get(i) & 0xFF) == '-') {
                         b.set(i, (byte)'_');
@@ -105,7 +107,9 @@ public class Http11 extends RubyObject {
                   f.cat(b);
                 }
 
-                b = new ByteList(Http11.this.hp.parser.buffer, value, vlen);
+                while (vlen > 0 && Character.isWhitespace(buffer.get(value + vlen - 1))) vlen--;
+
+                b = new ByteList(buffer, value, vlen);
                 v = req.op_aref(req.getRuntime().getCurrentContext(), f);
                 if (v.isNil()) {
                     req.op_aset(req.getRuntime().getCurrentContext(), f, RubyString.newString(runtime, b));
```

Looking at this code has made me realize how much it can be improved. The ByteList constructions here are doing full copies of the parse buffer each time, for example, and going out to java.lang.String just for a comparison is surprisingly expensive.

PR #2011, opened 2019-10-04
Allow for symbols to be used in verify_mode

### LINKED ISSUE #1222: Puma does not load a CA when verification mode is set to :peer
### Steps to reproduce

1) Configure Puma to accept SSL connections:
```
key =  File.expand_path "./certs/my_key.key"
cert = File.expand_path "./certs/my_cert.crt"
ca = File.expand_path "./certs/my_ca.crt"

ssl_bind "0.0.0.0", 9292, :ca => ca, :cert => cert, :key => key, :verify_mode => :peer
```

2) Start Puma, and get the following:
```
 starting in single mode...
* Version 3.6.2 (ruby 2.3.1-p112), codename: Sleepy Sunday Serenity
* Environment: production
ERROR: Please specify the SSL ca via 'ca='
```

### Expected behavior

Puma should just start and accept TLS connections.

### Actual behavior

Puma exits 1, requesting a CA be specified.

### Why it's happening

If mode is set to verify a peer, it requests a CA:
https://github.com/puma/puma/blob/0912b0cea1576e96ee4e84a84cfff2a80750ecf6/lib/puma/binder.rb#L178

However, the ssl bind method will not attach a CA param to the URL it generates:
https://github.com/puma/puma/blob/master/lib/puma/dsl.rb#L273

If this is a bug and not just me being dumb/wrong, I have no problem submitting a PR to fix it. 

### System configuration

**Ruby version**: ruby 2.3.1p112 (2016-04-26 revision 54768)
**Rails version**: rails (5.0.1)

iovis:
Fixes #1222 

`:verify_mode` was expecting strings to be used so this line would fail: https://github.com/puma/puma/blob/master/lib/puma/dsl.rb#L402

### DIFF
--- a/History.md
+++ b/History.md
@@ -6,6 +6,7 @@
 * Bugfixes
   * Fix socket activation of systemd (pre-existing) unix binder files (#1842, #1988)
   * Deal with multiple calls to bind correctly (#1986, #1994, #2006)
+  * Accepts symbols for `verify_mode` (#1222)
 
 ## 4.2.0 / 2019-09-23
 
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -396,7 +396,7 @@ module Puma
     #     keystore_pass: password
     #   }
     def ssl_bind(host, port, opts)
-      verify = opts.fetch(:verify_mode, 'none')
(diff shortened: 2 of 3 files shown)


nateberkopec:
EBPyFlZUIAEmT9-

PR #2012, opened 2019-10-05
Small JRuby cleanups and optimizations

headius:
This PR contains a quick pass over the Http11 class and the generated parser to reduce allocation and other antipatterns.

* Fast path hash access that does not need ThreadContext.
* Reduced transcoding of Java strings into Ruby strings by caching ByteList versions statically.
* Reduced allocation of call back objects by making them static.
* Reduced loads of `runtime` and `parser` fields by passing them into callbacks.

The callbacks could also be made into Java lambdas, removing the anonymous classes from the jar file, but that would tie this code to Java 8 or higher. I believe there may still be some JRuby folks with legacy apps on JRuby 1.7 + Java 7. Lambdas versus anon classes would not make any performance difference here.

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,7 +1,7 @@
 ## Master
 
 * Features
-  * Your feature goes here (#Github Number)
+  * Optimize HTTP parser for JRuby (#2012)
 
 * Bugfixes
   * Fix socket activation of systemd (pre-existing) unix binder files (#1842, #1988)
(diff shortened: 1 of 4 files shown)


headius:
See my comment here for reference: https://github.com/puma/puma/pull/2010#issuecomment-538661575

headius:
Note: This PR does not include the changes from #2010. That patch may not apply cleanly over this one but it's a trivial change.

Issue #1564, opened 2018-04-18
Errno::EINVAL when SSL is enabled and Firefox does not trust self-signed cert


perlun:
### Steps to reproduce

1) Use a self-signed certificate with Puma.
2) Browse to the https://localhost:12345 port.
3) Firefox will show a security warning, asking you to add an exception.

### Expected behavior

The request should be processed without errors.

### Actual behavior

```
Error in reactor loop escaped: Invalid argument - getpeername(2) (Errno::EINVAL)
/Users/plundberg/.rvm/gems/ruby-2.5.1/gems/puma-3.11.3/lib/puma/minissl.rb:155:in `peeraddr'
/Users/plundberg/.rvm/gems/ruby-2.5.1/gems/puma-3.11.3/lib/puma/minissl.rb:155:in `peeraddr'
/Users/plundberg/.rvm/gems/ruby-2.5.1/gems/puma-3.11.3/lib/puma/reactor.rb:93:in `rescue in block in run_internal'
/Users/plundberg/.rvm/gems/ruby-2.5.1/gems/puma-3.11.3/lib/puma/reactor.rb:73:in `block in run_internal'
/Users/plundberg/.rvm/gems/ruby-2.5.1/gems/puma-3.11.3/lib/puma/reactor.rb:43:in `each'
/Users/plundberg/.rvm/gems/ruby-2.5.1/gems/puma-3.11.3/lib/puma/reactor.rb:43:in `run_internal'
/Users/plundberg/.rvm/gems/ruby-2.5.1/gems/puma-3.11.3/lib/puma/reactor.rb:154:in `block in run_in_thread'
```

This is mostly a minor annoyance; after adding the exception it seems to work fine.

### System configuration

**Ruby version**: 2.5.1
**Rails version**: N/A
**Puma version**: 3.11.3

sevenseacat:
I added the security exception in Firefox and I'm still seeing the error in my logs.

Ruby version: 2.5.1
Rails version: 5.1
Puma version: 3.12.0

sevenseacat:
False alarm - there were a few other domains I needed to add security exceptions for as well!

PR #2014, opened 2019-10-06
Fix peeraddr raises Errno::EINVAL inside reactor

hahmed:
Hey @nateberkopec

I had a look into the issue #1564 and could not easily mock the object that returns the `peeraddr` as its inside the reactor class. 

Possible reasons why this error could occur? https://github.com/ruby/ruby/blob/ddf5020e4fcae5ed28a064af10124a032590452f/ext/socket/socket.c#L366

- Errno::EINVAL - the address length used for the _sockaddr_ is not a valid
- Errno::EINVAL - the +socket+ is already bound to an address, and the
-  protocol does not support binding to the new _sockaddr_ or the socket has been shut down.
- Errno::EINVAL - the address length is not a valid length for the address family

My idea was to create a test that has an expired SSL, that means we hit this part of the code https://github.com/puma/puma/blob/181777381574173fdc29f873c04cc266037c1cda/lib/puma/reactor.rb#L234 then have the `peeraddr` throw from a mocked object, but could not seem to do that.

Any idea on how I can do this, or is there another way I can go about writing a test for this?

### DIFF
--- a/lib/puma/reactor.rb
+++ b/lib/puma/reactor.rb
@@ -237,7 +237,7 @@ module Puma
                 ssl_socket = c.io
                 begin
                   addr = ssl_socket.peeraddr.last
-                rescue IOError
+                rescue IOError, Errno::EINVAL
                   addr = "<unknown>"
                 end
 
--- a/test/test_puma_server_ssl.rb
+++ b/test/test_puma_server_ssl.rb
@@ -199,6 +199,18 @@ class TestPumaServerSSL < Minitest::Test
       assert_match(msg, @events.error.message) if @events.error
     end
   end
+
+  def test_peeraddr_raises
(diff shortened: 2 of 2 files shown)


nateberkopec:
Yikes. I got about neck-deep into this one before I realized we would need a ground-up redesign of Reactor and Client to really stub this correctly, or give up and use any-instance mocks.

Merging without a test, under great pain.

nateberkopec:
Closed w 70e381d853f09520b198d17523676409010a9a49

Thanks for your 👀 @hahmed, this one turned out to be a lot more complicated to test than almost anything we've looked at yet...

hahmed:
Thanks, yeah I tried everything to find a way to test this, at least I learned a lot about how the reactor works 😂

PR #2016, opened 2019-10-07
this line gets overwritten when the YAML is parsed so we should removed [changelog skip]

zreisman:
…e it or line 47

Line is being ignore when YAML is parsed. One or the other should be removed or they should be merged. 
@olleolleolle @nateberkopec

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ matrix:
       env: jit=yes
     - rvm: jruby-9.2.8.0
       dist: bionic
-      env: OS="Bionic 18.04"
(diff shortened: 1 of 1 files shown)


olleolleolle:
Hello! Thanks for locating that, and yes, merging them, by placing `OS="Bionic 18.04"` first on the line would be helpful.

olleolleolle:
@zreisman Oh, one more thing I forgot to mention. Now that the `env` is a single long line, that must be used to match in the `allow_failures` section. The easiest thing is to duplicate, verbatim, the `env` line into the allowed_failures.

Issue #1949, opened 2019-09-05
Make "puma-heroku" the default config


nateberkopec:
The gem is a bit misnamed, it's just "good practices in config". I think we could mainline most of the settings in there for 5.0 and sunset the gem.

olleolleolle:
The items in [puma/puma-heroku](https://github.com/puma/puma-heroku/blob/master/lib/puma/plugin/heroku.rb) to be moved into [lib/puma/configuration.rb](https://github.com/puma/puma/blob/master/lib/puma/configuration.rb) could include:

- [ ] Worker Count via ENV variable where supported: query the environment variable `WEB_CONCURRENCY` (default: 1) for workers count (unless JRuby or Windows, then do not set it)
- [ ] if using workers, preload the app
- [ ] Set min and max threads via ENV variable(s): min_threads, max_threads set to `Integer(ENV['RAILS_MAX_THREADS'] || ENV['MAX_THREADS'] || 5)` (perhaps w/ a RAILS_MIN_THREADS, too?)
- [ ] Listen to RAILS_ENV for environment: `:environment => -> { ENV['RACK_ENV'] || "development" },` to also include `ENV['RAILS_ENV']`
- [ ] Rails 5.2 and lower: pre-fork & on_worker_boot database connecting

Any these items that have higher priority?

olleolleolle:
@nateberkopec From memory: I've only used it because the option setting was there, and then used the same values for "max". Merging them into one useful setting: a good thing.

nateberkopec:
Also re: Set min and max threads via ENV variable(s): min_threads, max_threads set to Integer(ENV['RAILS_MAX_THREADS'] || ENV['MAX_THREADS'] || 5) (perhaps w/ a RAILS_MIN_THREADS, too?)

We should keep the default of 16 on non-MRI platforms - so we should use a new default of 5 on MRI only.

jalevin:
Was having issues getting all tests to run locally. need to add to history / docs #2143

ukolovda:
Hello!
Also it will be useful in Docker containers. 
I need setup default parameters from environment variables, not from command-line parameters.

PR #2022, opened 2019-10-10
Configuration: Get environment from RAILS_ENV, too

olleolleolle:
This PR changes configuration defaults.

The **default value for environment** is also read from `RAILS_ENV`, if `RACK_ENV` can't be found.

## Overview

See #1949

The goal of that Issue is that configuration defaults in [puma/puma-heroku](https://github.com/puma/puma-heroku/blob/master/lib/puma/plugin/heroku.rb) move into [lib/puma/configuration.rb](https://github.com/puma/puma/blob/master/lib/puma/configuration.rb).

### DIFF
--- a/lib/puma/control_cli.rb
+++ b/lib/puma/control_cli.rb
@@ -22,7 +22,7 @@ module Puma
       @control_auth_token = nil
       @config_file = nil
       @command = nil
-      @environment = ENV['RACK_ENV']
+      @environment = ENV['RACK_ENV'] || ENV['RAILS_ENV']
 
       @argv = argv.dup
       @stdout = stdout
--- a/test/test_cli.rb
+++ b/test/test_cli.rb
@@ -360,11 +360,22 @@ class TestCLI < Minitest::Test
     assert_equal %w[a b], extra_dependencies
   end
 
-  def test_environment
+  def test_environment_rack_env
     ENV.delete 'RACK_ENV'
 
     Puma::CLI.new ["--environment", @environment]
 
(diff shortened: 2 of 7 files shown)


test/test_cli.rb
```diff
@@ -313,11 +313,20 @@ def test_load_path
...
   end
 
-  def test_environment
+  def test_environment_rack_env
     ENV.delete 'RACK_ENV'
 
     Puma::CLI.new ["--environment", @environment]
 
-    assert_equal ENV['RACK_ENV'], @environment
+    assert_equal @environment, ENV['RACK_ENV']
```
olleolleolle: This improves the output on failing  - the order of arguments is: `assert_equal exp, actual`

nateberkopec:
@olleolleolle friendly nudge/reminder to fix the failing tests

olleolleolle:
Huh, it's just like I forgot to push one of the files. I'll get to my computer and investigate!

lib/puma/control_cli.rb
```diff
@@ -22,7 +22,7 @@ def initialize(argv, stdout=STDOUT, stderr=STDERR)
       @control_auth_token = nil
       @config_file = nil
       @command = nil
-      @environment = ENV['RACK_ENV']
+      @environment = ENV['RACK_ENV'] || ENV['RAILS_ENV'] || 'development'
```
olleolleolle: This one! This feels dubious!

olleolleolle:
@nateberkopec There, I think this is ready for a Squash now.

PR #2024, opened 2019-10-10
Fix-up Markdown in History.md [ci skip]

dentarg:
Stuff between `<` and `>` wasn't rendered.


nateberkopec:
Nice catch, I always make this mistake too.

PR #2025, opened 2019-10-11
Use public ios method in test [changelog skip]

composerinteralia:
ios is an `attr_reader`, so this effectively the same thing, but testing
a public method instead of reaching into the internal state of the class
being tested seems like a win to me.


composerinteralia:
I will rebase once the CI failures mentioned in https://github.com/puma/puma/commit/d41cb68364e48869abf5ba22e79990336b975164 are resolved

nateberkopec:
👍

PR #2032, opened 2019-10-13
Fix loading plugins with prune_bundler

hahmed:
Fixes https://github.com/puma/puma/issues/1761

This PR allows allows one to add their own non bundled plugins into the default directory `./lib/puma/plugin/something.rb` or specify another root directory inside the configuration where we will find the plugin.

The current behaviour in puma fails to find non bundled plugins.

The code I wrote around how we load the default root path for plugin lookup and how its assigned via the dsl may need to be improved (please let me know?), by adding the ability to override the root plugin lookup path we can make it testable.

To test the plugin, I created a plugin at `test/puma/plugin/hello.rb` and used `STDERR.puts` to output a message. I then used `capture_subprocess_io` in `test/test_plugin.rb` to check the output in the server, is there a better way? This was the easiest plugin I could think off...

cc/ @nateberkopec

### DIFF
--- a/test/test_plugin.rb
+++ b/test/test_plugin.rb
@@ -28,6 +28,24 @@ class TestPlugin < TestIntegration
     out.close
   end
 
+  def test_non_bundled_plugin
+    skip "Skipped on Windows Ruby < 2.5.0, Ruby bug" if windows? && RUBY_VERSION < '2.5.0'
+
+    @tcp_bind = UniquePort.call
+    @tcp_ctrl = UniquePort.call
+
+    out, err = capture_subprocess_io do
+      cli_server "-C test/config/custom_plugin.rb test/rackup/hello.ru"
+    end
+
+    _, status = stop_server
+
+    assert_equal "hello world!!!", err.strip
+
+    @server.close
+    @server = nil
+  end
+
   private
 
   def cli_pumactl(argv)
--- a/test/test_plugin.rb
+++ b/test/test_plugin.rb
@@ -28,24 +28,6 @@ class TestPlugin < TestIntegration
     out.close
   end
 
-  def test_non_bundled_plugin
-    skip "Skipped on Windows Ruby < 2.5.0, Ruby bug" if windows? && RUBY_VERSION < '2.5.0'
-
-    @tcp_bind = UniquePort.call
-    @tcp_ctrl = UniquePort.call
-
-    out, err = capture_subprocess_io do
-      cli_server "-C test/config/custom_plugin.rb test/rackup/hello.ru"
-    end
-
-    _, status = stop_server
-
-    assert_equal "hello world!!!", err.strip
-
-    @server.close
-    @server = nil
-  end
-
   private
 
   def cli_pumactl(argv)
--- a/History.md
+++ b/History.md
@@ -10,6 +10,7 @@
   * Fix pumactl defaulting puma to development if an environment was not specified (#2035)
   * Fix closing file stream when reading pid from pidfile (#2048)
   * Fix a typo in configuration option `--extra_runtime_dependencies` (#2050)
+  * Fix loading plugins with prune_bundler (#2032)
 
 ## 4.2.1 / 2019-10-07
 
--- a/lib/puma/configuration.rb
+++ b/lib/puma/configuration.rb
@@ -133,6 +133,7 @@ module Puma
 
       @options     = UserFileDefaultOptions.new(user_options, default_options)
       @plugins     = PluginLoader.new
+      @all_plugins = []
       @user_dsl    = DSL.new(@options.user_options, self)
       @file_dsl    = DSL.new(@options.file_options, self)
       @default_dsl = DSL.new(@options.default_options, self)
@@ -142,7 +143,7 @@ module Puma
       end
     end
 
-    attr_reader :options, :plugins
+    attr_reader :options, :plugins, :all_plugins
 
     def configure
       yield @user_dsl, @file_dsl, @default_dsl
@@ -272,6 +273,7 @@ module Puma
     end
 
     def load_plugin(name)
+      @all_plugins << name
       @plugins.create name
     end
 
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -253,7 +253,9 @@ module Puma
         "#{d.name}:#{spec_for_gem(d.name).version}"
       end
 
-      [deps, require_paths_for_gem(puma) + extra_runtime_deps_directories]
+      plugin_paths = @config.all_plugins.map { |plugin| "puma/plugin/#{name}" }
+
+      [deps, require_paths_for_gem(puma) + extra_runtime_deps_directories + plugin_paths]
     end
 
     def extra_runtime_deps_directories
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -253,7 +253,7 @@ module Puma
         "#{d.name}:#{spec_for_gem(d.name).version}"
       end
 
-      plugin_paths = @config.all_plugins.map { |plugin| "puma/plugin/#{name}" }
+      plugin_paths = @config.all_plugins.map { |name| "puma/plugin/#{name}" }
 
       [deps, require_paths_for_gem(puma) + extra_runtime_deps_directories + plugin_paths]
     end
--- a/test/config/custom_plugin.rb
+++ b/test/config/custom_plugin.rb
@@ -1,2 +1,0 @@
-plugin 'hello'
-prune_bundler
--- a/test/config/custom_plugin.rb
+++ b/test/config/custom_plugin.rb
@@ -1,0 +1,2 @@
+plugin 'hello'
+prune_bundler
--- a/test/puma/plugin/hello.rb
+++ b/test/puma/plugin/hello.rb
@@ -1,7 +1,0 @@
-require 'puma/plugin'
-
-Puma::Plugin.create do
-  def start(launcher)
-    STDERR.puts "hello world!!!"
-  end
-end
--- a/test/puma/plugin/hello.rb
+++ b/test/puma/plugin/hello.rb
@@ -1,0 +1,7 @@
+require 'puma/plugin'
+
+Puma::Plugin.create do
+  def start(launcher)
+    STDERR.puts "hello world!!!"
+  end
+end
--- a/test/test_launcher.rb
+++ b/test/test_launcher.rb
@@ -50,7 +50,7 @@ class TestLauncher < Minitest::Test
   def test_prune_bundler_loads_plugins
     skip_on :no_bundler
     conf = Puma::Configuration.new do |c|
-      c.plugin 'hello'
+      c.plugin 'tmp_restart'
     end
     _, dirs = launcher(conf).send(:dependencies_and_files_to_require_after_prune)
     assert_equal(3, dirs.length)
--- a/test/test_launcher.rb
+++ b/test/test_launcher.rb
@@ -43,16 +43,16 @@ class TestLauncher < Minitest::Test
     end
     dep_dirs = launcher(conf).send(:extra_runtime_deps_directories)
 
-    assert_equal(3, dep_dirs.length)
+    assert_equal(1, dep_dirs.length)
     assert_match(%r{gems/rdoc-[\d.]+/lib$}, dep_dirs.first)
   end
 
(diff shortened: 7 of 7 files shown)


test/test_plugin.rb
```diff
@@ -28,6 +28,23 @@ def test_plugin
     out.close
   end
 
+  def test_non_bundled_plugin
+    skip "Skipped on Windows Ruby < 2.5.0, Ruby bug" if windows? && RUBY_VERSION < '2.5.0'
```
dentarg: I assume this is just copied from above, but I wonder if it is still true for the test below?

dentarg:
> To test the plugin, I created a plugin at `test/puma/plugin/hello.rb` and used `STDERR.puts` to output a message. I then used `capture_subprocess_io` in `test/test_plugin.rb` to check the output in the server, is there a better way?

What about having the plugin write to a file? Maybe a simpler test can be written then.

test/test_plugin.rb
nateberkopec: yeah try leaving that skip out

hahmed: Done

nateberkopec: Looks like that's a "no" lol https://github.com/puma/puma/pull/2032/checks?check_run_id=260440680#step:9:121

nateberkopec:
@hahmed Can you talk a bit about why you chose this approach over the way @ylecuyer did it in #1763?

test/test_plugin.rb
dentarg: Good to know... I wonder what bug this is, I guess @MSP-Greg must have added this skip initially

hahmed:
> What about having the plugin write to a file? Maybe a simpler test can be written then.

This is a relatively simple test -- was hoping there was an easier way to write it. 

For example we have the launcher in the plugin so we can call `launcher.events.write message` and we can perhaps pass in the `@events` into the `@server` and read it from there.

test/test_plugin.rb
hahmed: lol -- reverted, btw I have a windows machine somewhere, I would test this out on windows but it hurts so much using windows 😂

hahmed:
> @hahmed Can you talk a bit about why you chose this approach over the way @ylecuyer did it in #1763?

@nateberkopec the PR by @ylecuyer loads the path inside the launcher (`lib/puma/launcher.rb`), whereas the require for bundled plugins happens inside `lib/puma/plugin.rb` 

https://github.com/puma/puma/pull/1763/files#diff-cdee464b1408f50d2df33316479cd064R259 -- here we are using bundler to pull the gems, but all we really need is to require the files?

I'm happy to move that solution but I did not understand what the need for bundler to load non bundled plugins was...

test/test_plugin.rb
MSP-Greg: Cute.  I'm 60+, I can say that to someone your age.  Anyway, Windows should work now.  BTW, running from just rake needs a `pathname` require.

hahmed: Nice work on that PR @MSP-Greg I can see that windows works 👍✨

nateberkopec:
@hahmed https://github.com/puma/puma/issues/1761#issuecomment-544793748 

What do you think? It feels to me like adding new DSL options is probably unnecessary and we can fix this without any input from the user.

nateberkopec:
Accidentally close, my b.

nateberkopec:
> The current behaviour in puma fails to find non bundled plugins.

FWIW the current behavior in Puma fails to find ANY plugins whose
plugin name doesn't exactly match its gem name.

nateberkopec:
I guess we could just document my workaround in #1761, and if someone _really wants_ to add arbitrary paths to their prune_bundler we can add that later.

hahmed:
>  It feels to me like adding new DSL options is probably unnecessary
Totally agree, will have a read of your comments and respond.

hahmed:
Looks like there is a few issues 
1. FWIW the current behavior in Puma fails to find ANY plugins whose
plugin name doesn't exactly match its gem name.
2. someone really wants to add arbitrary paths to their prune_bundler we can add that later.
3. finding custom plugins (this we can document by saying to the user, add your require statement, I'm not sure if this is super clear anywhere?)

With this PR we only want to address? https://github.com/puma/puma/issues/1761#issuecomment-544793748

hahmed:
@nateberkopec I managed to get the plugin names re: your comment https://github.com/puma/puma/pull/2032#issuecomment-544793891 just need to find way to write a test for bundled plugins

nateberkopec:
Does this work with, say, the puma-heroku plugin?

hahmed:
With the settings and config;

```
threads 3, 5
plugin :heroku
workers 2
prune_bundler
```
With the gems;
```
source "https://rubygems.org"
gem 'puma'
gem 'puma-heroku'
```

Looks like it works, this is the output;
```
prune % bundle exec puma -C settings.rb hello.ru
[21874] * Pruning Bundler environment
[21874] Puma starting in cluster mode...
[21874] * Version 4.2.1 (ruby 2.5.0-p0), codename: Distant Airhorns
[21874] * Min threads: 5, max threads: 5
[21874] * Environment: development
[21874] * Process workers: 2
[21874] * Phased restart available
[21874] * Listening on tcp://0.0.0.0:3000
[21874] Use Ctrl-C to stop
[22030] + Gemfile in context: /Users/haroon/projects/pumatests/prune/Gemfile
[22031] + Gemfile in context: /Users/haroon/projects/pumatests/prune/Gemfile
WARN: Unresolved specs during Gem::Specification.reset:
      puma (< 5.0, >= 3.0)
WARN: Clearing out unresolved specs.
Please report a bug if this causes problems.
WARN: Unresolved specs during Gem::Specification.reset:
      puma (< 5.0, >= 3.0)
WARN: Clearing out unresolved specs.
Please report a bug if this causes problems.
[21874] - Worker 0 (pid: 22030) booted, phase: 0
[21874] - Worker 1 (pid: 22031) booted, phase: 0
^C[21874] - Gracefully shutting down workers...
[21874] === puma shutdown: 2019-10-31 21:38:15 +0000 ===
[21874] - Goodbye!
```

The min, max threads have been updated as per heroku plugin.

If I update the code in launcher to;
```
plugin_paths = [] 

      @config.all_plugins.compact.each do |plugin_name|
        plugin = Bundler.rubygems.loaded_specs("puma-" + plugin_name.to_s)
        plugin_paths << plugin.require_paths.map { |x| File.join(plugin.full_gem_path, x) }
      end
```

(I'm having to add the prefix i.e. `"puma-" + plugin_name.to_s`) which seems a bit weird to me, but here is the output from that;

```
prune % bundle exec puma -C settings.rb hello.ru
[20934] * Pruning Bundler environment
[20934] Puma starting in cluster mode...
[20934] * Version 4.2.1 (ruby 2.5.0-p0), codename: Distant Airhorns
[20934] * Min threads: 5, max threads: 5
[20934] * Environment: development
[20934] * Process workers: 2
[20934] * Phased restart available
[20934] * Listening on tcp://0.0.0.0:3000
[20934] Use Ctrl-C to stop
[21090] + Gemfile in context: /Users/haroon/projects/pumatests/prune/Gemfile
[21091] + Gemfile in context: /Users/haroon/projects/pumatests/prune/Gemfile
[20934] - Worker 0 (pid: 21090) booted, phase: 0
[20934] - Worker 1 (pid: 21091) booted, phase: 0
^C[20934] - Gracefully shutting down workers...
[20934] === puma shutdown: 2019-10-31 21:34:30 +0000 ===
[20934] - Goodbye!
```

When I do an inspect on the plugin and the array this is what I get;
```
#<Bundler::StubSpecification:0x00007ff4e4398dd0 @name="puma-heroku", @version=#<Gem::Version "1.1.0">, @platform="ruby", @spec_fetcher=nil, @dependencies=[<Gem::Dependency type=:runtime name="puma" requirements="< 5.0, >= 3.0">], @stub=#<Gem::StubSpecification:0x00007ff4e423fec0 @extension_dir=nil, @full_gem_path="/Users/haroon/.asdf/installs/ruby/2.5.0/lib/ruby/gems/2.5.0/gems/puma-heroku-1.1.0", @gem_dir=nil, @ignored=nil, @loaded_from="/Users/haroon/.asdf/installs/ruby/2.5.0/lib/ruby/gems/2.5.0/specifications/puma-heroku-1.1.0.gemspec", @data=#<Gem::StubSpecification::StubLine:0x00007ff4e4134af8 @name="puma-heroku", @version=#<Gem::Version "1.1.0">, @platform="ruby", @extensions=[], @full_name="puma-heroku-1.1.0", @require_paths=["lib"]>, @name=nil, @spec=nil, @base_dir="/Users/haroon/.asdf/installs/ruby/2.5.0/lib/ruby/gems/2.5.0", @gems_dir="/Users/haroon/.asdf/installs/ruby/2.5.0/lib/ruby/gems/2.5.0/gems", @default_gem=false, @activated=true, @full_require_paths=["/Users/haroon/.asdf/installs/ruby/2.5.0/lib/ruby/gems/2.5.0/gems/puma-heroku-1.1.0/lib"]>, @source=#<Bundler::Source::Rubygems:0x70344888516040 rubygems repository https://rubygems.org/ or installed locally>>
[["/Users/haroon/.asdf/installs/ruby/2.5.0/lib/ruby/gems/2.5.0/gems/puma-heroku-1.1.0/lib"]]
```

Hope I understood your question?

hahmed:
Actually, now that I have looked at this again with master this looks like it works to me, here is a config (settings.rb);

``` 
threads 3, 5
plugin :heroku
workers 2
prune_bundler
```
When I run `bundle exec puma -C settings.rb hello.ru`

```
[1743] * Pruning Bundler environment
[1743] Puma starting in cluster mode...
[1743] * Version 4.2.1 (ruby 2.5.0-p0), codename: Distant Airhorns
[1743] * Min threads: 5, max threads: 5
[1743] * Environment: development
[1743] * Process workers: 2
[1743] * Phased restart available
[1743] * Listening on tcp://0.0.0.0:3000
[1743] Use Ctrl-C to stop
[1898] + Gemfile in context: /Users/haroon/projects/pumatests/prune/Gemfile
[1899] + Gemfile in context: /Users/haroon/projects/pumatests/prune/Gemfile
WARN: Unresolved specs during Gem::Specification.reset:
WARN: Unresolved specs during Gem::Specification.reset:
      puma (< 5.0, >= 3.0)
WARN: Clearing out unresolved specs.
      puma (< 5.0, >= 3.0)
Please report a bug if this causes problems.
WARN: Clearing out unresolved specs.
Please report a bug if this causes problems.
[1743] - Worker 1 (pid: 1899) booted, phase: 0
[1743] - Worker 0 (pid: 1898) booted, phase: 0
^C[1743] - Gracefully shutting down workers...
[1743] === puma shutdown: 2019-11-02 10:37:39 +0000 ===
[1743] - Goodbye!
```
We can see the min, max threads equal to the amount in the Heroku plugin (5,5)
`[1743] * Min threads: 5, max threads: 5`

This tells me that master already works, and that issue is resolved? Not too sure about the spec warnings that are appearing though. If I made a few changes to my PR in the launcher;

```
      @config.all_plugins.compact.each do |plugin_name|
        plugin = Bundler.rubygems.loaded_specs("puma-" + plugin_name.to_s)
        plugin_paths << plugin.require_paths.map { |x| File.join(plugin.full_gem_path, x) }
      end
```

Those warnings also go away, 

```
prune % bundle exec puma -C settings.rb hello.ru
[2093] * Pruning Bundler environment
[2093] Puma starting in cluster mode...
[2093] * Version 4.2.1 (ruby 2.5.0-p0), codename: Distant Airhorns
[2093] * Min threads: 5, max threads: 5
[2093] * Environment: development
[2093] * Process workers: 2
[2093] * Phased restart available
[2093] * Listening on tcp://0.0.0.0:3000
[2093] Use Ctrl-C to stop
[2249] + Gemfile in context: /Users/haroon/projects/pumatests/prune/Gemfile
[2250] + Gemfile in context: /Users/haroon/projects/pumatests/prune/Gemfile
[2093] - Worker 0 (pid: 2249) booted, phase: 0
[2093] - Worker 1 (pid: 2250) booted, phase: 0
```

but I really don't like appending `puma` onto the plugin name, as users will likely name their plugin anything they like;

`plugin = Bundler.rubygems.loaded_specs("puma-" + plugin_name.to_s)`

hahmed:
On master could not replicate this issue, so closing.

PR #2035, opened 2019-10-15
Update pumactl to remove development as default environment

### LINKED ISSUE #2023: Pumactl broken environment option
**Describe the bug**

Our production is in `development` mode now. `environment` method in `config/puma.rb` is broken (after #1885, I guess).

**Puma config:**

```ruby
# frozen_string_literal: true

environment 'production'

app do |env|
  [200, {}, ["embedded app"]]
end
```

**To Reproduce**

```ruby
# Gemfile

# frozen_string_literal: true

source 'https://rubygems.org'

gem 'puma', '< 4.2'
```

```
> bundle exec pumactl start -F puma_config.rb
* Version 4.1.1 (ruby 2.6.5-p114), codename: Fourth and One
* Environment: production
* Listening on tcp://0.0.0.0:9292
Use Ctrl-C to stop
```

```ruby
# Gemfile

# frozen_string_literal: true

source 'https://rubygems.org'

gem 'puma', '>= 4.2'
```

```
> bundle exec pumactl start -F puma_config.rb
* Version 4.2.0 (ruby 2.6.5-p114), codename: Distant Airhorns
* Environment: development
* Listening on tcp://0.0.0.0:9292
Use Ctrl-C to stop
```

**Expected behavior**

`production` in production, not `development` after any value for `environment` method.

 - OS: Linux
 - Puma Version ≥ 4.2

AndrewSpeed:
# What changed?

This PR:
 - removes the default environment of `development` from `pumactl`
 - defaults to development when searching for config files to load, if one is not provided to `pumactl`
 - passes the value passed to the `environment` option over `RACK_ENV` to `pumactl`, if provided

fixes #2023

### DIFF
--- a/test/helpers/config_file.rb
+++ b/test/helpers/config_file.rb
@@ -1,0 +1,16 @@
+class TestConfigFileBase < Minitest::Test
+  private
+
+  def with_env(env = {})
+    original_env = {}
+    env.each do |k, v|
+      original_env[k] = ENV[k]
+      ENV[k] = v
+    end
+    yield
+  ensure
+    original_env.each do |k, v|
+      ENV[k] = v
+    end
+  end
+end
--- a/test/test_config.rb
+++ b/test/test_config.rb
@@ -1,25 +1,9 @@
 # frozen_string_literal: true
 
 require_relative "helper"
+require_relative "helpers/config_file"
 
 require "puma/configuration"
-
-class TestConfigFileBase < Minitest::Test
-  private
-
-  def with_env(env = {})
-    original_env = {}
-    env.each do |k, v|
-      original_env[k] = ENV[k]
-      ENV[k] = v
-    end
-    yield
-  ensure
-    original_env.each do |k, v|
-      ENV[k] = v
-    end
-  end
-end
 
 class TestConfigFile < TestConfigFileBase
   parallelize_me!
--- a/test/test_pumactl.rb
+++ b/test/test_pumactl.rb
@@ -24,6 +24,17 @@ class TestPumaControlCli < Minitest::Test
     server.close
   end
 
+  def with_config_file(path_to_config, port)
+    path = Pathname.new(path_to_config)
+    Dir.mktmpdir do |tmp_dir|
+      Dir.chdir(tmp_dir) do
+        FileUtils.mkdir_p(path.dirname)
+        File.open(path, "w") { |f| f << "port #{port}" }
+        yield
+      end
+    end
+  end
+
   def test_config_file
     control_cli = Puma::ControlCLI.new ["--config-file", "test/config/state_file_testing_config.rb", "halt"]
     assert_equal "t3-pid", control_cli.instance_variable_get("@pidfile")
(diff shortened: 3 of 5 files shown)


test/test_pumactl.rb
```diff
@@ -29,16 +29,54 @@ def test_config_file
...
-    ENV.delete 'RACK_ENV' # remove from travis
+  def test_rack_env_without_environment
+    ENV.update("RACK_ENV" => "test")
     control_cli = Puma::ControlCLI.new ["halt"]
-    assert_equal "development", control_cli.instance_variable_get("@environment")
+    assert_equal "test", control_cli.instance_variable_get("@environment")
+  end
+
+  def test_environment_without_rack_env
+    ENV.delete "RACK_ENV" # remove from travis
```
composerinteralia: Instead of mutating the environment, which could lead to order-dependent test failures if there is another test that unexpectedly relies on a specific environment variable being set, it might be nice to reuse the [`with_env` helper](https://github.com/puma/puma/blob/860c17557c5a9d56e4d04189d637ec414ecd72f8/test/test_config.rb#L10-L22).

Or, if we are OK with adding an additional dependency, at thoughtbot we often use [`climate_control`](https://github.com/thoughtbot/climate_control) for this sort of thing. It has the advantage of being threadsafe, which might help if we plan to parallelize these tests.

test/test_pumactl.rb
```diff
@@ -29,16 +29,54 @@ def test_config_file
...
     control_cli = Puma::ControlCLI.new ["-e", "test", "halt"]
     assert_equal "test", control_cli.instance_variable_get("@environment")
   end
 
-  def test_config_file_exist
-    ENV.delete 'RACK_ENV' # remove from travis
+  def test_environment_specific_config_file_exist
+    ENV.delete "RACK_ENV"
+    port = 6002
+    Dir.mktmpdir do |tmp_dir|
```
composerinteralia: I think some of this file manipulation makes the test a bit more difficult to read. What do you think about extracting a helper like:

```rb
def with_config_file(path_to_config, port)
  path = Pathname.new(path_to_config)
  Dir.mktmpdir do |tmp_dir|
    Dir.chdir(tmp_dir) do
      FileUtils.mkdir(path.dir)
      File.open(path, "w") { |f| f << "port #{port}" }
      yield
    end
  end
end

def test_environment_specific_config_file_exist
  port = 6002
  path_to_config = "config/puma.rb"

  with_config_file(path_to_config, port) do
    control_cli = Puma::ControlCLI.new ["-e", "production", "halt"]
    assert_equal path_to_config,
      control_cli.instance_variable_get("@config_file")
  end
end
```

test/test_pumactl.rb
nateberkopec: Let's just use with_env for now, but keep our eyes on climate_control for the future.

test/helpers/config_file.rb
```diff
@@ -0,0 +1,16 @@
+class TestConfigFileBase < Minitest::Test
```
AndrewSpeed: This seemed like the most appropriate place for this file, but let me know if there's another convention I should be following

test/test_config.rb
```diff
@@ -1,26 +1,10 @@
 # frozen_string_literal: true
 
 require_relative "helper"
+require_relative "helpers/config_file"
```
nateberkopec: Works for me!

nateberkopec:
Solid PR, thanks for the contribution 👍

PR #2036, opened 2019-10-17
Address thread pool test TODOs [changelog skip]

composerinteralia:
These TODOs were added in 9f4edf4c while cleaning up the tests a bit.

For the first TODO, it looks like there was a point to those two lines: this and the previous test call `Thread.current.kill` in the thread pool block so the threads will die once some work is added. I tried to make this more obvious by starting the `auto_reap!` earlier and checking that it hasn't reaped any threads that are still alive.

For the other TODO I replaced `sleep 10` with the `Thread.pass until
finish` pattern we use in some other tests in this file. Now if `pool.shutdown(0)` fails to raise an error we finish right away rather than waiting for the 5 second ConditionVariable timeout. I also added an assertion to ensure we rescue the `ForceShutdown` error.

### DIFF
--- a/test/test_thread_pool.rb
+++ b/test/test_thread_pool.rb
@@ -223,13 +223,14 @@ class TestThreadPool < Minitest::Test
   def test_auto_reap_dead_threads
     pool = new_pool(2,2) { Thread.current.kill }
 
+    pool.auto_reap! 0.1
+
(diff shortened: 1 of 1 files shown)


nateberkopec:
Gorgeous. Thanks!

PR #2037, opened 2019-10-17
Refactor tests, fix errors shown [changelog skip]

MSP-Greg:
1. Clean up leaked fd's both in test & lib code

2. Delete control io's and unix paths

3. Fix code differences between signal and control_cli behavior

4. control_cli.rb - refactor, add thread-status using sock control

5. launcher.rb - add thread-status via sock control

6. detect.rb - Add Puma::HAS_FORK and Puma::HAS_UNIX, adjust tests to match

7. Fixed connection handling during stop (both Single & Clustered)

8. split some test files into two classes, one running parallel, one 'serial'

### DIFF
--- a/lib/puma/app/status.rb
+++ b/lib/puma/app/status.rb
@@ -16,47 +16,51 @@ module Puma
 
       def call(env)
         unless authenticate(env)
-          return rack_response(403, 'Invalid auth token', 'text/plain')
+          return rack_response 403, 'Invalid auth token', 'text/plain'
         end
 
         case env['PATH_INFO']
         when /\/stop$/
           @cli.stop
-          rack_response(200, OK_STATUS)
+          rack_response 200, OK_STATUS
 
         when /\/halt$/
           @cli.halt
-          rack_response(200, OK_STATUS)
+          rack_response 200, OK_STATUS
 
         when /\/restart$/
           @cli.restart
-          rack_response(200, OK_STATUS)
+          rack_response 200, OK_STATUS
 
         when /\/phased-restart$/
           if !@cli.phased_restart
-            rack_response(404, '{ "error": "phased restart not available" }')
+            rack_response 404, '{ "error": "phased restart not available" }'
           else
-            rack_response(200, OK_STATUS)
+            rack_response 200, OK_STATUS
           end
 
         when /\/reload-worker-directory$/
           if !@cli.send(:reload_worker_directory)
-            rack_response(404, '{ "error": "reload_worker_directory not available" }')
+            rack_response 404, '{ "error": "reload_worker_directory not available" }'
           else
-            rack_response(200, OK_STATUS)
+            rack_response 200, OK_STATUS
           end
 
         when /\/gc$/
           GC.start
-          rack_response(200, OK_STATUS)
+          rack_response 200, OK_STATUS
 
         when /\/gc-stats$/
-          rack_response(200, GC.stat.to_json)
+          rack_response 200, GC.stat.to_json
 
         when /\/stats$/
-          rack_response(200, @cli.stats)
+          rack_response 200, @cli.stats
+
+        when /\/thread-status$/
+          rack_response 200, @cli.thread_status, 'text/plain'
+
         else
-          rack_response 404, "Unsupported action", 'text/plain'
+          rack_response 404, 'Unsupported action', 'text/plain'
         end
       end
 
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -28,10 +28,12 @@ module Puma
 
       @phased_state = :idle
       @phased_restart = false
+      @restart = false
     end
 
     def stop_workers
       log "- Gracefully shutting down workers..."
+      @launcher.close_binder_listeners unless @restart
       @workers.each { |x| x.term }
 
       begin
@@ -333,7 +335,10 @@ module Puma
     def stop_blocked
       @status = :stop if @status == :run
       wakeup!
-      @control.stop(true) if @control
+      if @control
+        @control.stop(true)
+        close_control_io
+      end
       Process.waitall
     end
 
@@ -364,23 +369,23 @@ module Puma
     # We do this in a separate method to keep the lambda scope
     # of the signals handlers as small as possible.
     def setup_signals
-      Signal.trap "SIGCHLD" do
+      Signal.trap :CHLD do
         wakeup!
       end
 
-      Signal.trap "TTIN" do
+      Signal.trap :TTIN do
         @options[:workers] += 1
         wakeup!
       end
 
-      Signal.trap "TTOU" do
+      Signal.trap :TTOU do
         @options[:workers] -= 1 if @options[:workers] >= 2
         wakeup!
       end
 
       master_pid = Process.pid
 
-      Signal.trap "SIGTERM" do
+      Signal.trap :TERM do
         # The worker installs their own SIGTERM when booted.
         # Until then, this is run by the worker and the worker
         # should just exit if they get it.
@@ -388,8 +393,6 @@ module Puma
           log "Early termination of worker"
           exit! 0
         else
-          @launcher.close_binder_listeners
-
           stop_workers
           stop
 
@@ -522,7 +525,6 @@ module Puma
             @status = :stop
           end
         end
-
         stop_workers unless @status == :halt
       ensure
         @check_pipe.close
--- a/lib/puma/control_cli.rb
+++ b/lib/puma/control_cli.rb
@@ -11,7 +11,20 @@ require 'socket'
 module Puma
   class ControlCLI
 
-    COMMANDS = %w{halt restart phased-restart start stats status stop reload-worker-directory gc gc-stats}
+    CMDS_SGNLS = {
+      'halt'           => :QUIT,
+      'phased-restart' => :USR1,
+      'restart'        => :USR2,
+      'stop'           => :TERM,
+      'thread-status'  => :INFO,
+      'gc'       => nil,
+      'gc-stats' => nil,
+      'start'    => nil,
+      'stats'    => nil,
+      'status'   => nil
+    }.freeze
+
+    COMMANDS = CMDS_SGNLS.keys.sort.freeze
 
     def initialize(argv, stdout=STDOUT, stderr=STDERR)
       @state = nil
@@ -181,7 +194,7 @@ module Puma
         end
 
         message "Command #{@command} sent success"
-        message response.last if @command == "stats" || @command == "gc-stats"
+        message response.last if %w!stats gc-stats thread-status!.include? @command
       end
     ensure
       server.close if server && !server.closed?
@@ -193,42 +206,22 @@ module Puma
       end
 
       begin
-
-        case @command
-        when "restart"
-          Process.kill "SIGUSR2", @pid
-
-        when "halt"
-          Process.kill "QUIT", @pid
-
-        when "stop"
-          Process.kill "SIGTERM", @pid
-
-        when "stats"
-          puts "Stats not available via pid only"
-          return
-
-        when "reload-worker-directory"
-          puts "reload-worker-directory not available via pid only"
-          return
-
-        when "phased-restart"
-          Process.kill "SIGUSR1", @pid
-
-        when "status"
-          begin
-            Process.kill 0, @pid
-            puts "Puma is started"
-          rescue Errno::ESRCH
-            raise "Puma is not running"
-          end
-
-          return
-
+        if sgnl = CMDS_SGNLS[@command]
+          Process.kill sgnl, @pid
         else
-          return
+          case @command
+          when "status"
+            begin
+              Process.kill 0, @pid
+              puts "Puma is started"
+            rescue Errno::ESRCH
+              raise "Puma is not running"
+            end
+            return
+          else
+            return
+          end
         end
-
       rescue SystemCallError
         if @command == "restart"
           start
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -2,7 +2,7 @@
 
 require 'puma/events'
 require 'puma/detect'
-require 'puma/cluster'
+require 'puma/cluster' if ::Puma::HAS_FORK
 require 'puma/single'
 require 'puma/const'
 require 'puma/binder'
@@ -65,7 +65,7 @@ module Puma
 
       generate_restart_data
 
-      if clustered? && !Process.respond_to?(:fork)
+      if clustered? && !HAS_FORK
         unsupported "worker mode not supported on #{RUBY_ENGINE} on this platform"
       end
 
@@ -169,6 +169,7 @@ module Puma
 
       setup_signals
       set_process_title
+
       @runner.run
 
       case @status
@@ -184,7 +185,9 @@ module Puma
       when :exit
         # nothing
       end
-      @binder.close_unix_paths
+        @binder.close
+        @binder.close_unix_paths
+        @runner.end_control @status
     end
 
     # Return which tcp port the launcher is using, if it's using TCP
@@ -195,7 +198,7 @@ module Puma
     def restart_args
       cmd = @options[:restart_cmd]
       if cmd
-        cmd.split(' ') + @original_argv
+        cmd.split + @original_argv
       else
         @restart_argv
       end
@@ -203,6 +206,24 @@ module Puma
 
     def close_binder_listeners
       @binder.close_listeners
+    end
+
+    def thread_status
+      out = ''.dup
+      Thread.list.each do |thread|
+        oid = thread.object_id.to_s 36
+        out <<   "Thread  TID-#{oid} #{thread['label']}\n"
+        logstr = "Thread: TID-#{oid}"
+        logstr += " #{thread.name}" if thread.respond_to?(:name)
+        out << "#{logstr}\n"
+
+        if thread.backtrace
+          out << thread.backtrace.join("\n") + "\n"
+        else
+          out << "<no backtrace available>\n"
+        end
+      end
+      out
     end
 
     private
@@ -324,18 +345,7 @@ module Puma
     end
 
     def log_thread_status
-      Thread.list.each do |thread|
-        log "Thread TID-#{thread.object_id.to_s(36)} #{thread['label']}"
-        logstr = "Thread: TID-#{thread.object_id.to_s(36)}"
-        logstr += " #{thread.name}" if thread.respond_to?(:name)
-        log logstr
-
-        if thread.backtrace
-          log thread.backtrace.join("\n")
-        else
-          log "<no backtrace available>"
-        end
-      end
+      log thread_status
     end
 
     def set_process_title
@@ -408,7 +418,7 @@ module Puma
 
     def setup_signals
       begin
-        Signal.trap "SIGUSR2" do
+        Signal.trap :USR2 do
           restart
         end
       rescue Exception
@@ -417,7 +427,7 @@ module Puma
 
       unless Puma.jruby?
         begin
-          Signal.trap "SIGUSR1" do
+          Signal.trap :USR1 do
             phased_restart
           end
         rescue Exception
@@ -426,7 +436,7 @@ module Puma
       end
 
       begin
-        Signal.trap "SIGTERM" do
+        Signal.trap :TERM do
           graceful_stop
 
           raise(SignalException, "SIGTERM") if @options[:raise_exception_on_sigterm]
@@ -436,7 +446,7 @@ module Puma
       end
 
       begin
-        Signal.trap "SIGINT" do
+        Signal.trap :INT do
           stop
         end
       rescue Exception
@@ -444,7 +454,7 @@ module Puma
       end
 
       begin
-        Signal.trap "SIGHUP" do
+        Signal.trap :HUP do
           if @runner.redirected_io?
             @runner.redirect_io
           else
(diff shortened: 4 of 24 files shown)


MSP-Greg:
This is a lot of code.  It started when I was working on tests, noticed odd things happening, and decided to write some code to check leaked fd's.  There were quite a few.  Some were from the tests, some were from the code.  The test code will dump fd's on test failures and at the end of testing if there are more than three open fd's (STDOUT, STDERR, & STDIN).

Also, tried to minimize the 'test timed out' issues by cleaning up some of the code that was testing io read values, etc.

While working on test coverage, I was too lazy to shut off Actions, as the coverage was running in Travis.  This has passed a gazillion builds without failure, and a minimum number re-tried tests.

Some of the code updates the 'integration' tests, which are typically 'front to back' tests, but many utilize signals, which means the Puma instance must be in a separate process.  Another reason is 'restart', which has `Kernel.exec` calls, which don't work to well in threads...

Several of the 'helper methods' help with things like closing sockets, etc, which are easy to overlook when writing tests.  I started documenting the 'helper' methods and classes, but haven't finished.  These could be included in normal Puma docs, kind of like RubyGems TestCase.

One thing that should be added is binding to an ssl socket, but I thought this was far enough along for a draft PR.

Similar code could/should be added to tests that are running a Puma instance in a thread, as several tests do that.  That can wait until another PR.

Lastly, with large problems, I often do not work in a linear fashion, but I seem to converge on a solution.  Happy to divide this up, but need suggestions as to how...

nateberkopec:
Will look more later but I don’t like running tests on all branches for pushes. You end up with two builds when making PRs from this remote.

MSP-Greg:
I removed the change re branches and the additional rake file.

lib/puma/app/status.rb
```diff
@@ -16,47 +16,51 @@ def initialize(cli, token = nil)
...
+            rack_response 200, OK_STATUS
           end
 
         when /\/reload-worker-directory$/
           if !@cli.send(:reload_worker_directory)
-            rack_response(404, '{ "error": "reload_worker_directory not available" }')
+            rack_response 404, '{ "error": "reload_worker_directory not available" }'
           else
-            rack_response(200, OK_STATUS)
+            rack_response 200, OK_STATUS
```
nateberkopec: If we're gonna do this (also changing to single-quotes), we should also be added Rubocop rules.

lib/puma/control_cli.rb
```diff
@@ -191,42 +204,22 @@ def send_signal
...
-          begin
-            Process.kill 0, @pid
-            puts "Puma is started"
-          rescue Errno::ESRCH
-            raise "Puma is not running"
-          end
-
-          return
-
+        if sgnl = CMDS_SGNLS[@command]
```
nateberkopec: Mistake? `==`.

nateberkopec:
I'm conceptually OK with all of this, but I think we should be adding Rubocop rules any time we're making style changes.

MSP-Greg:
I'm sorry for the style changes.  I can remove them tomorrow/today.  Been looking at monitors since before Apple II days, and I'm 60+.  After a long day, this:
```ruby
rack_response 200, OK_STATUS
```
is easier to read than:
```ruby
rack_response(200, OK_STATUS)
```
Re RuboCop, I've seen some of the discussions in ruby/ruby about it, and it's a pretty solid NO there...

JFYI, this makes JRuby more stable, but stable in the sense that it consistently fails, but seems to more often get to the test summary.  I tried, I give up.  Someone familiar with JRuby will have to look at it.  It's rock solid on Ubuntu & macOS.

Also, I've got inherited tests (tcp, unix & ssl) working in TestBinder.  Much better coverage.  Setup to go on top of this.

lib/puma/app/status.rb
MSP-Greg: BTW, I didn't change all of them (currently in master), note both single & double quotes.  So I just made them consistent...

https://github.com/puma/puma/blob/1d97e42dde2bcc737f8d01f6c2292f6845c9c504/lib/puma/app/status.rb#L59

MSP-Greg:
Just rebased on master, should be passing on Windows...

test/test_pumactl.rb
```diff
@@ -106,50 +97,49 @@ def test_default_config_file_exist
...
-    shutdown_cmd = Puma::ControlCLI.new(opts + ["halt"])
-    out, _ = capture_subprocess_io do
-      shutdown_cmd.run
-    end
-    assert_match "Command halt sent success\n", out
+    Puma::ControlCLI.new(opts + ['stop'], @ready, @ready).run
+    assert_io 'Goodbye', io: @wait
+
+    #Puma::ControlCLI.new(opts + ['halt'], @ready, @ready).run
+    #assert_io 'Stopping immediately!', io: @wait, debug: true
```
nateberkopec: Should this still be commented?

MSP-Greg: IDK.  You may have seen errors where Events is trying to write but the pipe/IO is closed?  Those were part of what I was trying to clean up, so I may have switched to stop early in the work on it.

Also, IDK if we've got halt covered in all of the four classes that include it.  I'll look at tomorrow.

This PR was meant as a start to improving the test framework, both in terms of stability and ease of use.  I wanted it to be trivial to do tests for the same functionality, with one using a control socket and one using a signal.  I didn't want to mess with UNIX paths, remembering to close all the IO's, have leaky fd's, etc.  Also, no more naked read statements leading to lengthy 'test took too long' timeouts when a good method should limit them to 5 or 10 seconds, etc, etc..

I did have to use it in enough test files to see if it was functional, and it changed/expanded as I was developing it.  So, I'm sure this will change in the future, but I think it's a good first step.  

And, given how far I got into things with the Binder work and the systemd stuff, I feel pretty confident about the changes in the lib files.

Sorry for the extended post, I've wrestled with some messy things today.  There are a few small changes I'll push over the weekend.

MSP-Greg: @nateberkopec

Hate to ask, can you try the above commented 'halt' lines with the two 'stop' lines commented?  Locally it freezes, probably why I commented it (a reminder for later).  I can push on my fork to Actions also...

MSP-Greg: I rebased and deleted the comments.  We should see halt no being called in test coverage.

MSP-Greg:
One thing not explicitly mentioned in the above comments is that current behavior shows reset connections during `term_closes_listeners` tcp tests.  The tests check connection handling after a stop command is issued.  I think it was discussed when the debug output was added to test logging.  This PR adds more tests for this behavior.

This PR also fixes the 'reset' issue.  Connections are either 'successes' or 'refused', we no longer have 'resets'...

MSP-Greg:
Due to intermittent (but okay) results in the `stop_closes_listeners` tests, one more change.

The tests throw 40 connections at a server being stopped, 10 before, 30 after.  On macOS, occasionally 11 connections are correctly responded to.  Other platforms it seems to always be 10.  

Adjusted asserts so the test is not run again.

see https://github.com/puma/puma/runs/268563842#step:9:288

Besides, showing that the PR is certainly stable.  Sorry, the one time I didn't run rubocop first...

MSP-Greg:
Thanks for reminding me about that comment.  Fixed `halt`, added a test for anytime the integration tests use a UNIX socket for control that Puma deletes the file, and added several `halt` tests.

With the changes, this is done from my perspective.

lib/puma/launcher.rb
```diff
@@ -184,7 +185,9 @@ def run
       when :exit
         # nothing
       end
-      @binder.close_unix_paths
+        @binder.close
+        @binder.close_unix_paths
```
nateberkopec: I just realized - do we really want to close all sockets when restarting? Isn't that exactly the opposite of what we're supposed to do?

lib/puma/launcher.rb
```diff
@@ -195,7 +198,7 @@ def connected_port
     def restart_args
       cmd = @options[:restart_cmd]
       if cmd
-        cmd.split(' ') + @original_argv
+        cmd.split + @original_argv
```
nateberkopec: I don't think we should remove this, as it depends on `$;` being nil.

lib/puma/reactor.rb
```diff
@@ -244,7 +244,8 @@ def run_internal
 
                 cert = ssl_socket.peercert
 
-                c.close
+                # may generate another SSL error, continue to remove the monitor
+                c.close rescue nil
```
nateberkopec: Was this in response to a specific issue?

lib/puma/reactor.rb
```diff
@@ -302,6 +303,7 @@ def clear_monitor(mon)
     def run
       run_internal
     ensure
+      @selector.close
```
nateberkopec: Same here, was this in response to a specific issue?

lib/puma/server.rb
```diff
@@ -258,6 +258,7 @@ def handle_servers_lopez_mode
         STDERR.puts e.backtrace
       ensure
         begin
+          @notify.close rescue nil
```
nateberkopec: Makes sense, seems like an oversight. Is this tested?

lib/puma/server.rb
```diff
@@ -428,8 +429,8 @@ def handle_servers
         STDERR.puts "Exception handling servers: #{e.message} (#{e.class})"
         STDERR.puts e.backtrace
       ensure
-        @check.close
-        @notify.close
+        @check.close  rescue nil
```
nateberkopec: What was the motivation for adding rescue nil?

lib/puma/single.rb
```diff
@@ -30,13 +30,14 @@ def stop
     end
 
     def halt
-      @server.halt
+      @server.halt if @server
```
nateberkopec: How does `@server` become nil?

lib/puma/runner.rb
```diff
@@ -88,6 +88,20 @@ def start_control
...
 
+    def end_control(type)
+      if @control && [:halt, :stop].include?(type)
+        @control.send type, true
+        @control.binder.close
+        @control.binder.close_unix_paths
+      end
+    end
+
+    def close_control_io
```
nateberkopec: The control server itself should just handle this. I can't think of a circumstance where we want to stop the control server but want to keep its IOs open, right?

nateberkopec: i.e. no one should call close_control_io, everywhere should call end_control(type).

test/helper.rb
```diff
@@ -45,6 +49,35 @@ def hit(uris)
...
+  def assert_io(re, io: @server, debug: false, timeout: 10)
+    regex = Regexp === re ? re : Regexp.new(Regexp.escape re)
+
+    STDOUT.puts('',
+      "#{' ' * 20}*** #{full_name} assert_io #{re} ***") if debug
+    l = ''
+
+    ttf = Time.now.to_f + 10
+
+    while IO.select([io], nil, nil, timeout) do
```
nateberkopec: I feel like we could use more descriptive variable names in this method, I think you slipped into writing C here 😆

test/helper.rb
```diff
@@ -79,12 +112,11 @@ module TestSkips
...
   # usage: skip NO_FORK_MSG unless HAS_FORK
   # windows >= 2.6 fork is not defined, < 2.6 fork raises NotImplementedError
-  HAS_FORK = ::Process.respond_to? :fork
+  HAS_FORK = ::Puma::HAS_FORK
   NO_FORK_MSG = "Kernel.fork isn't available on the #{RUBY_PLATFORM} platform"
 
-  # socket is required by puma
-  # usage: skip UNIX_SKT_MSG unless UNIX_SKT_EXIST
-  UNIX_SKT_EXIST = Object.const_defined? :UNIXSocket
+  # usage: skip UNIX_SKT_MSG unless HAS_UNIXHAS_UNIX
```
nateberkopec: typo at EOL

test/helper.rb
```diff
@@ -125,6 +157,31 @@ def skip_unless(eng, bt: caller)
     end
     skip skip_msg, bt if skip_msg
   end
+
+  def out_io
```
nateberkopec: This is gonna need a comment. It looks like you're trying to make sure all ios are closed after the suite ends, right?

In that case, I think this needs assertions (not just debug output) and instead of returning against length, it would be much more descriptive if `ary` contained STDOUT/STDERR/whatever the third thing is.

test/helper.rb
```diff
@@ -135,18 +192,35 @@ def self.run(reporter, options = {}) # :nodoc:
     super
   end
 
+  def after_teardown
+    return if passed? || skipped?
+    out_io unless Puma.jruby? or RUBY_VERSION < '2.3'
```
nateberkopec: What's different about jruby and Ruby 2.2?

test/helper.rb
```diff
@@ -135,18 +192,35 @@ def self.run(reporter, options = {}) # :nodoc:
...
   unless $debugging_hold
-    out = $debugging_info.strip
-    unless out.empty?
+    out = "#{$debugging_info.rstrip}\n".dup
+    files = Dir.glob('tmp/*').map { |s| s.sub 'tmp/', '' }
+      .join("\n").sub(RUBY_PLATFORM, '')
+
+    out << files
+
+    unless out.strip.empty?
```
nateberkopec: I think this just be an assertion, not debug output.

test/helper.rb
```diff
@@ -135,18 +192,35 @@ def self.run(reporter, options = {}) # :nodoc:
...
+    "#{cls_name}__#{name}"
+  end
 end
 
 Minitest.after_run do
   # needed for TestCLI#test_control_clustered
   unless $debugging_hold
-    out = $debugging_info.strip
-    unless out.empty?
+    out = "#{$debugging_info.rstrip}\n".dup
```
nateberkopec: Should be extracted into its own method, like `assert_no_open_tmp_files`

test/test_cli.rb
```diff
@@ -9,34 +9,31 @@ class TestCLI < Minitest::Test
...
+    @tmp_path2 = "#{@tmp_path}_2"
 
     File.unlink @tmp_path  if File.exist? @tmp_path
     File.unlink @tmp_path2 if File.exist? @tmp_path2
 
     @wait, @ready = IO.pipe
 
     @events = Puma::Events.strings
-    @events.on_booted { @ready << "!" }
+    @events.on_booted { @ready.syswrite '!' }
```
nateberkopec: What was wrong with the shovel?

test/test_cli.rb
```diff
@@ -9,34 +9,31 @@ class TestCLI < Minitest::Test
...
   end
 
   def teardown
-    File.unlink @tmp_path if File.exist? @tmp_path
+    File.unlink @tmp_path  if File.exist? @tmp_path
     File.unlink @tmp_path2 if File.exist? @tmp_path2
 
-    @wait.close
-    @ready.close
+    @ready.close rescue nil
```
nateberkopec: I think you can probably understand why I don't think just adding `rescue nil` everywhere is fixing anything, right? 😆

test/test_integration_cluster.rb
```diff
@@ -2,102 +2,122 @@
...
 class TestIntegrationCluster < TestIntegration
-  parallelize_me!
-
-  DARWIN = !!RUBY_PLATFORM[/darwin/]
-
-  def setup
-    skip NO_FORK_MSG unless HAS_FORK
-    super
-  end
+  parallelize_me! unless Puma.jruby?
```
nateberkopec: This line isn't ever run on JRuby anyway, is it?

test/test_integration_cluster.rb
```diff
@@ -2,102 +2,122 @@
...
 
   def teardown
     return if skipped?
     super
   end
 
   def test_pre_existing_unix
-    skip UNIX_SKT_MSG unless UNIX_SKT_EXIST
+    skip UNIX_SKT_MSG unless HAS_UNIX
+    setup_puma bind: :unix, ctrl: :unix
```
nateberkopec: This reads like you're setting up a control server and then not using it. Is that correct?

lib/puma/launcher.rb
MSP-Greg: I don't think the code ever reaches here in restarts.  It hits the Kernel.exec statements.

lib/puma/launcher.rb
MSP-Greg: I've seen that a lot of places, including Ruby core, but I'll change it.

lib/puma/cluster.rb
```diff
@@ -28,10 +28,12 @@ def initialize(cli, events)
 
       @phased_state = :idle
       @phased_restart = false
+      @restart = false
     end
 
     def stop_workers
       log "- Gracefully shutting down workers..."
+      @launcher.close_binder_listeners unless @restart
```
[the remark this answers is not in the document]
MSP-Greg: Some of these changes were related to leaked fd's, which is why I added the log output for it.  This may have been part of the fix of the macOS issue with ':reset' connections, which are now zero.  Previously they hovered around +/- 7.

lib/puma/reactor.rb
MSP-Greg: Yes.  Where, don't recall.  I can comment it out and run the tests.

test/test_cli.rb
MSP-Greg: Sometime ago I came across info implying that mixing 'sys' & 'normal' on the same IO may cause odd issues.  Ruby version or platform specific, I don't recall, nor how long ago...

test/test_cli.rb
MSP-Greg: It's in a test.  If something else goes wrong before it...

test/test_integration_cluster.rb
MSP-Greg: Ok, so maybe I forgot that JRuby doesn't support fork.  I did that before I decided I was tired of seeing all the skip messages and added the class end statement conditional.  I'll remove.

test/test_integration_cluster.rb
MSP-Greg: stop_server_wait calls `run_pumactl 'stop'`

lib/puma/reactor.rb
MSP-Greg: nio4r doc for the method: 'Close the selector and free system resources'

I don't recall whether it affected leaky fd's or not.  It should be done somewhere...

lib/puma/server.rb
MSP-Greg: The test suite did not previously test for leaky fd's.  Right or wrong, I decided that doing a `GC.start` & `ObjectSpace.each_object(IO)` for most tests might set the CI time out too far.  So I just did it when the tests finished, and made it easy to check individual tests if needed.

When I first added it, the fd's were well over 100.

Summing up, there are no asserts for leaky fd's at present, just the check at the end.

lib/puma/server.rb
MSP-Greg: Exactly, I don't recall.  We're in an ensure clause at shutdown ?

lib/puma/single.rb
MSP-Greg: Good question.  I saw it used elsewhere, and again, this is 'exit' code.

lib/puma/runner.rb
MSP-Greg: > The control server itself should just handle this

The problem is that there isn't a control server class, and servers can also inherit binders.  So closing anything in the `Server` class is dicey.  Won't make that mistake again...

Re `close_control_io`, I think I can get rid of it, I'll need to add a `sync` arg to the `end_control` method.

test/helper.rb
MSP-Greg: Yeah, I started writing that while frustrated.

Question: I've got STDERR going to the IO.pipe output for integration tests, and many of the other tests map Events IO to one IO.

Should the method dump all the read data to the console if the assert fails?

test/helper.rb
MSP-Greg: I knew that...  Thanks,

test/helper.rb
MSP-Greg: First: Correct.

The third thing (STDIN)?  Sorry.  I've never checked into whether one can fail in `Minitest.after_run`....

test/helper.rb
MSP-Greg: I'll double check.  Something must have happened, don't recall what.

test/helper.rb
MSP-Greg: Again, `Minitest.after_run` isn't in a test.  I'll check

test/helper.rb
MSP-Greg: See above

nateberkopec:
I'm going to close this spike because it's too big to merge as-is. Please break this up into ~100-line chunks. Suggestions for chunks:

* Style changes
* 1 PR for each file changed in lib/, with corresponding tests
* 1 PR that adds the new helpers in tests
* 1 PR for each changed test file

lib/puma/cluster.rb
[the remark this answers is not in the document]
MSP-Greg: > I'm wondering what we were doing before that was wrong.

Remember your question about how some of the connections were being 'dumped' during a stop?  They were already accepted but then they were reset?  This fixes that, but we can't close them if a restart is being done, otherwise Binder#redirects_for_restart won't work because the sockets need to stay open.

MSP-Greg:
@nateberkopec

Most of the changes requested in your review can be found here:

https://github.com/MSP-Greg/puma/commit/a352600c0f253b5629153e4ea3aae5f6bc36baaa

Items that were left as-is I can explain if you'd like.

PR #2039, opened 2019-10-17
Remove broken puma/convenient [changelog skip]

composerinteralia:
This file was added in 2016 in 663666c5, then broken in 2017 in
cda9317e. `c` was renamed to `user_config`, but one reference to `c` was
missed.

To confirm the method is indeed broken I ran:

```rb
require 'puma/convenient'
Puma.run
# => NameError (undefined local variable or method `c' for Puma:Module)
```

Since this has been broken for more than 2 years, I think it is pretty safe
to say nobody is using it. That is why I propose removing instead of
fixing it.


composerinteralia:
Windows builds are broken: https://github.com/puma/puma/issues/2040

Issue #1937, opened 2019-08-28
Add documentation for TCP mode (Lopez Express)


Jesus:
Puma has the `--tcp-mode` option which isn't known by many people perhaps because the only public documentation about it is this help from the CLI:

```
Run the app in raw TCP mode instead of HTTP mode
```

It might be good to have a short paragraph in the Readme (or somewhere else) describing how to use this cool feature.

Feel free to close this ticket if you don't want to clutter the Readme file with this.

I can't do this right now but I might be able in a few weeks. Also this maybe a good first contribution.

PR #2045, opened 2019-10-19
[README] Add section about TCP mode [changelog skip]

andrykonchin:
Document in short launching Puma in the TCP mode

Description may be loo long so please let me know if it should be moved to some other place, e.g. `docs` directory.

Related issue https://github.com/puma/puma/issues/1937

### DIFF
--- a/docs/tcp_mode.md
+++ b/docs/tcp_mode.md
@@ -1,0 +1,96 @@
+# TCP mode
+
+Puma also could be used as a TCP server to process incoming TCP
+connections.
+
+
+## Configuration
+
+TCP mode can be enabled with CLI option `--tcp-mode`:
+
+```
+$ puma --tcp-mode
+```
+
+Default ip and port to listen to are `0.0.0.0` and `9292`. You can configure
+them with `--port` and `--bind` options:
+
+```
+$ puma --tcp-mode --bind tcp://127.0.0.1:9293
+$ puma --tcp-mode --port 9293
+```
+
+TCP mode could be set with a configuration file as well with `tcp_mode`
+and `tcp_mode!` methods:
+
+```
+# config/puma.rb
+tcp_mode
+```
+
+When Puma starts in the TCP mode it prints the corresponding message:
+
+```
+puma --tcp-mode
+Puma starting in single mode...
+...
+* Mode: Lopez Express (tcp)
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -230,6 +230,91 @@ $ puma -C "-"
...
+
+TCP mode could be set with a configuration file as well with `tcp_mode`
+and `tcp_mode!` methods:
+
+```
+# config/puma.rb
+tcp_mode
+```
+
+When Puma starts in the TCP mode it prints corresponded message:
```
nateberkopec: "it prints the corresponding message"

README.md
```diff
@@ -230,6 +230,91 @@ $ puma -C "-"
...
+
+```
+puma --tcp-mode
+Puma starting in single mode...
+...
+* Mode: Lopez Express (tcp)
+```
+
+Application to process TCP connections should be declared in the same
```
nateberkopec: "An application"

README.md
```diff
@@ -230,6 +230,91 @@ $ puma -C "-"
...
+way as an ordinary HTTP application. It's a callable object which accepts
```
nateberkopec: I don't think this is really accurate (that it's created in the same way), because for HTTP Puma serves Rack apps, but for TCP, it's a different structure.

README.md
```diff
@@ -230,6 +230,91 @@ $ puma -C "-"
...
+telnet 0.0.0.0 9293
+Trying 0.0.0.0...
+Connected to 0.0.0.0.
+Escape character is '^]'.
+sssss
+Echo sssss
+^CConnection closed by foreign host.
+```
+
+After application finishes processing Puma closes a socket. In order to
```
nateberkopec: After the application finishes, Puma closes the socket.

README.md
```diff
@@ -230,6 +230,91 @@ $ puma -C "-"
...
+prevent this application should set `env['detach'] = true`.
```
nateberkopec: this, the application should set

nateberkopec:
Thank you! A few small changes.

I think we should move this to docs/tcp_mode.md.

andrykonchin:
@nateberkopec Done

nateberkopec:
Woo hoo! Thank you!

Issue #2015, opened 2019-10-07
SSL support for activate_control_app


Nowaker:
Valid:
```
activate_control_app 'tcp://127.0.0.1:9000', auth_token: token
```
Also valid:
```
ssl_bind '127.0.0.1', 8443, cert: 'config/cert-dev.pem', key: 'config/cert-dev.key'
```

Invalid:
```
activate_control_app 'ssl://127.0.0.1:9000', auth_token: token

[44959] ERROR: Invalid control URI: ssl://127.0.0.1:9000
```

It would be great for Puma to allow control app to be run behind a TLS certificate to ensure confidentiality of the token.

I know I can TLS-ify it using nginx but so we could use the same argument against `ssl_bind` :-)

nateberkopec:
Closed by #2046

PR #2046, opened 2019-10-20
Add SSL support for the control app

composerinteralia:
This PR includes two commits, one for the refactor and one for the change. I can open two separate PRs if it is easier to review that way.

---

### Extract class for building SSL context
 
This commit extracts the `MiniSSL::Context` creation into its own
`MiniSSL::ContextBuilder` class along the same lines as in #1989.

This will allow us to reuse this code for adding SSL support to the
control app (issue #2015). Since we will need the `MiniSSL` require
and check in both places, I moved that into the `ContextBuilder` class
as well.

----

### Add SSL support for the control app

This starts to address #2015. I think we will need to add SSL support
to the control cli as well.

### DIFF
--- a/lib/puma/runner.rb
+++ b/lib/puma/runner.rb
@@ -2,6 +2,7 @@
 
 require 'puma/server'
 require 'puma/const'
+require 'puma/minissl/context_builder'
 
 module Puma
   # Generic class that is used by `Puma::Cluster` and `Puma::Single` to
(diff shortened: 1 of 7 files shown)


composerinteralia:
Oops, I forgot to add a changelog entry. I will be sure to add that if this change looks OK.

lib/puma/runner.rb
```diff
@@ -2,6 +2,7 @@
 
 require 'puma/server'
 require 'puma/const'
+require 'puma/minissl/context_builder'
```
nateberkopec: Nice call. I'm not gonna review that commit because it looks like a straight copy-paste into the new class.

nateberkopec:
Nice! LGTM.

composerinteralia:
Thanks! Changelog entry added.

PR #2047, opened 2019-10-21
4.3.0

nateberkopec:
Placeholder. This will be out by end of week.

Name by @MSP-Greg

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,6 +1,14 @@
 ## Master
 
 * Features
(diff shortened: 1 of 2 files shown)


jdoss:
Any ETA on this? I would love to not have to work around this bug https://github.com/puma/puma/pull/2035 in 4.2.1 anymore on my deployments. :smile:

PR #2051, opened 2019-10-23
PR template: link to contributing, changelog

olleolleolle:
Point to the "blob" on master, so that the link is clickable from the PR

### Description

I tried to use the links, and the links were ineffective.

This PR updates the PR template to make the links point directly to the documents.


nateberkopec:
Oops, thx!

PR #2052, opened 2019-10-24
Add ssl support to the control_cli

composerinteralia:
If I have a server running:

```sh
bundle exec bin/puma test/rackup/hello.ru \
  --control-url="ssl://127.0.0.1:9000?key=examples/puma/puma_keypair.pem&cert=examples/puma/cert_puma.pem" \
  --control-token="token"
```

This commit will allow me to restart that server (or run any other
control cli command) with:

```sh
bundle exec bin/pumactl restart \
  --control-url="ssl://127.0.0.1:9000" \
  --control-token="token"
```

Before this commit the pumactl restart command would have raised an
error: `"Invalid scheme: ssl"`

- [X] I have reviewed the [guidelines for contributing](../CONTRIBUTING.md) to this repository.
- [X] I have added an entry to [HISTORY.md](../HISTORY.md) if this PR fixes a bug or adds a feature. If it doesn't need an entry to HISTORY.md, I have added `[changelog skip]` to all commit messages.
- [X] If this PR doesn't need tests (docs change), I added `[ci skip]` to the commit messages.

### DIFF
--- a/lib/puma/control_cli.rb
+++ b/lib/puma/control_cli.rb
@@ -141,6 +141,12 @@ module Puma
 
       # create server object by scheme
       server = case uri.scheme
+                when "ssl"
+                  require 'openssl'
+                  OpenSSL::SSL::SSLSocket.new(
+                    TCPSocket.new(uri.host, uri.port),
+                    OpenSSL::SSL::SSLContext.new
+                  ).tap(&:connect)
                 when "tcp"
                   TCPSocket.new uri.host, uri.port
                 when "unix"
--- a/test/test_pumactl.rb
+++ b/test/test_pumactl.rb
@@ -1,16 +1,19 @@
 require_relative "helper"
 require_relative "helpers/config_file"
+require_relative "helpers/ssl"
 
 require 'puma/control_cli'
 
 class TestPumaControlCli < TestConfigFileBase
+  include SSLHelper
+
   def setup
     # use a pipe to get info across thread boundary
     @wait, @ready = IO.pipe
   end
 
   def wait_booted
-    line = @wait.gets until line =~ /Listening on/
+    line = @wait.gets until line =~ /Use Ctrl-C to stop/
   end
 
   def teardown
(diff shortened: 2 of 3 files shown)


lib/puma/control_cli.rb
```diff
@@ -141,6 +141,12 @@ def send_request
 
       # create server object by scheme
       server = case uri.scheme
+                when "ssl"
+                  require 'openssl'
```
composerinteralia: I'm not totally sure if this was the right thing to have done here. I started out with:

```sh
require 'puma/puma_http11'
require 'puma/minissl'
MiniSSL.check
MiniSSL::Socket.new(
  TCPSocket.new(uri.host, uri.port),
  MiniSSL::Engine.client
)
```

but that was raising an "Unknown OpenSSL error: 2" at https://github.com/puma/puma/blob/3294c82037f60da8c5dc810a7ff3a61ed11ee707/ext/puma_http11/mini_ssl.c#L324 and both my C and SSL knowledge are probably a bit too limited to know what to do with that without some help.

test/test_pumactl.rb
```diff
@@ -132,24 +135,53 @@ def test_control_url_and_status
...
     out, _ = capture_subprocess_io do
-      shutdown_cmd.run
+      cmd.run
     end
-    assert_match "Command halt sent success\n", out
+    assert_match expected_out, out
+  end
 
-    assert_kind_of Thread, t.join, "server didn't stop"
+  def assert_app_running(host)
```
nateberkopec: I'm not sure this is a useful abstraction, because you're not actually asserting that "an app" is "running,", you're asserting that a puma server running the "embedded app" config file returns a 200 when connected to.

nateberkopec: Doesn't wait_booted perform a check something like this, as well?

composerinteralia: Good point. This code didn't seem important to what either test was actually testing, so I tried to hide it away in a helper method. Digging further, I see now that `wait_booted` on was waiting for the embedded app to boot, but not for the control app. Connecting to the "embedded app" took long enough that the control app was booted by the time we needed it. `sleep 0.1` would also have worked.

I am going to update `wait_booted` (used only in these two tests) to wait for the control app instead, and then get rid of the "embedded app" assertion for the new spec. I will the extra assertions it in the existing spec in case they are testing something that isn't covered elsewhere.

Issue #1964, opened 2019-09-12
Thread backtrace printing - additional features


nateberkopec:
#1320 adds a SIGINFO handler that prints all current thread backtraces.

* [ ] If the master of a cluster receives SIGINFO, it should fan out that signal to all of its children.
* [x] Add a pumactl command that does this, so platforms without SIGINFO can use this feature.

Fudoshiki:
puma master

```zsh
=> Booting Puma
=> Rails 6.1.0.alpha application starting in development 
=> Run `rails server --help` for more startup options
Puma starting in single mode...
* Version 4.1.1 (ruby 2.6.4-p104), codename: Fourth and One
* Min threads: 5, max threads: 5
* Environment: development
* Listening on tcp://0.0.0.0:3000
Use Ctrl-C to stop
#<Thread:0x00000001180f7990@/Users/sharevari/.rvm/gems/ruby-2.6.4@hospital40/bundler/gems/puma-7e704b553c73/lib/puma/plugin.rb:66 run> terminated with exception (report_on_exception is true):
Traceback (most recent call last):
/Users/sharevari/.rvm/gems/ruby-2.6.4@hospital40/bundler/gems/puma-7e704b553c73/lib/puma/plugin.rb:67:in `block (2 levels) in fire_background': undefined method `set_thread_name' for #<Puma::PluginRegistry:0x0000000117b8c668> (NoMethodError)
Did you mean?  set_trace_func
```

dentarg:
@Fudoshiki I think you should open up a new issue (or PR), I believe it comes from this change: https://github.com/puma/puma/pull/1968/files#diff-e897820765f1098c2b56bc7417b49cdd

dentarg:
@nateberkopec this looks done?

> [ ] If the master of a cluster receives SIGINFO, it should fan out that signal to all of its children.


```shell
$ puma -C test/config/settings.rb test/rackup/hello.ru
Puma starting in single mode...
* Puma version: 5.2.1 (ruby 2.7.1-p83) ("Fettisdagsbulle")
*  Min threads: 3
*  Max threads: 5
*  Environment: development
*          PID: 31798
* Listening on http://0.0.0.0:3000
Use Ctrl-C to stop
<kill -SIGINFO 31798 in another terminal window>
Thread: TID-pk
  /Users/dentarg/.gem/ruby/2.7.1/gems/puma-5.2.1/lib/puma/launcher.rb:233:in `backtrace'
[...]

PR #2054, opened 2019-10-25
Add pumactl command to print thread backtraces

composerinteralia:
Completes 1 of 2 items from #1964

This commit adds an endpoint to the status app to print thread
backtraces, and control cli command to call that endpoint.

I tried this locally by starting a server with:

```sh
bundle exec bin/puma test/rackup/hello.ru \
  --control-url="unix://test.sock" \
  --control-token="token"
```

and then printing the backtraces with:

```sh
bundle exec bin/pumactl thread-backtraces \
  --control-url="unix://test.sock" \
  --control-token="token"
```

- [X] I have added an entry to [History.md](../blob/master/History.md) if this PR fixes a bug or adds a feature. If it doesn't need an entry to HISTORY.md, I have added `[changelog skip]` to all commit messages.
- [X] If this PR doesn't need tests (docs change), I added `[ci skip]` to the commit messages.

### DIFF
--- a/lib/puma/app/status.rb
+++ b/lib/puma/app/status.rb
@@ -57,9 +57,12 @@ module Puma
           rack_response(200, @cli.stats)
 
         when /\/thread-backtraces$/
-          strings = Puma::Events.strings
-          @cli.log_thread_status(strings)
-          rack_response(200, strings.stdout.string)
+          backtraces = []
+          @cli.thread_status do |name, backtrace|
+            backtraces << { name: name, backtrace: backtrace }
+          end
+
+          rack_response(200, backtraces.to_json)
         else
           rack_response 404, "Unsupported action", 'text/plain'
         end
--- a/lib/puma/app/status.rb
+++ b/lib/puma/app/status.rb
@@ -55,6 +55,11 @@ module Puma
 
         when /\/stats$/
           rack_response(200, @cli.stats)
+
+        when /\/thread-backtraces$/
+          strings = Puma::Events.strings
+          @cli.log_thread_status(strings)
+          rack_response(200, strings.stdout.string)
         else
           rack_response 404, "Unsupported action", 'text/plain'
         end
--- a/lib/puma/control_cli.rb
+++ b/lib/puma/control_cli.rb
@@ -11,7 +11,8 @@ require 'socket'
 module Puma
   class ControlCLI
 
-    COMMANDS = %w{halt restart phased-restart start stats status stop reload-worker-directory gc gc-stats}
+    COMMANDS = %w{halt restart phased-restart start stats status stop reload-worker-directory gc gc-stats thread-backtraces}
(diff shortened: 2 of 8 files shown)


lib/puma/app/status.rb
```diff
@@ -55,6 +55,11 @@ def call(env)
 
         when /\/stats$/
           rack_response(200, @cli.stats)
+
+        when /\/thread-backtraces$/
+          strings = Puma::Events.strings
+          @cli.log_thread_status(strings)
+          rack_response(200, strings.stdout.string)
```
composerinteralia: Should this be JSON instead of the raw string?

nateberkopec: Yes. All current responses are json, so we're going to need to turn this into json object.

Maybe something like:

```
threads: [
  { name: "", backtrace: "" },
  { name: "", backtrace: "" }
]
```

lib/puma/control_cli.rb
```diff
@@ -11,7 +11,8 @@
 module Puma
   class ControlCLI
 
-    COMMANDS = %w{halt restart phased-restart start stats status stop reload-worker-directory gc gc-stats}
+    COMMANDS = %w{halt restart phased-restart start stats status stop reload-worker-directory gc gc-stats thread-backtraces}
+    PRINTABLE_COMMANDS = %w{gc-stats stats thread-backtraces}
```
nateberkopec: Nice.

lib/puma/launcher.rb
```diff
@@ -205,6 +205,21 @@ def close_binder_listeners
       @binder.close_listeners
     end
 
+    def log_thread_status(events)
```
nateberkopec: Looks like you'll need to refactor this out a bit so that you have two methods - one that prepares a Ruby object representation of threads and backtraces, then one method that logs that info to events.

composerinteralia: @nateberkopec Should we always use the object representation, or do you want to preserve the existing format for the SIGINFO and use the object representation only for the status app?

nateberkopec: I think what matters is that the current logged output remains the same, the status app outputs JSON, and that they both use the same method underneath for obtaining that info.

composerinteralia:
Ready for another review

nateberkopec:
Nice work!

composerinteralia:
Woah, that was fast! Thanks!

olleolleolle:
From the sidelines: this is super helpful! The tab that was added to each backtrace line makes for an ergonomic developer experience.

Littlejd1997:
Is there a timeline for releasing this?

nateberkopec:
This was merged 3 days after 4.3 was released, so I realize it's been in the backlog for a while. You can see progress on 5.0 here: https://github.com/puma/puma/milestone/7

PR #2055, opened 2019-10-27
Update README With More Explanation of Setting RACK_ENV [ci skip]

### LINKED ISSUE #1795: Documentation Around Setting Puma Environment
### Steps to reproduce

1) Puma docs don't really specify why we need to set the Environment in the puma config.

### Expected behavior
In the environment section of the documentation, it basically says "set environment here."

### Actual behavior
 I would expect the documentation to explain a little more the impact that setting environment has.

### System configuration
N/A

### Context
We recently ran into an issue where we discovered that we weren't properly setting RACK_ENV in our production environment, so our Puma servers were being booted with "development" environment. I went to investigate what impact this might be having, and all I could find in reference to the environment was that it uses that environment configuration to look for where to find its configuration file. This is also a little confusing, as you can set the environment inside the config file, but if it's only using the environment to reconcile the location of the config file, which it must have found already to be reading from, I'm not sure it's actually even doing anything (unless there are undocumented speed side effects, hence this issue). *EDIT* I realized that it might be using that environment to find a config to override defaults.

We're loading puma with the -C option, so that it wasn't really affecting us, but I couldn't find any other documentation explaining if there were any performance enhancements or side effects by running Puma in "production" versus "development".

I would open an PR to add to the ReadMe that the environment is only used to help Puma locate the appropriate config files, but I don't actually know that for sure.

dlinch:
### Description
Documentation update addressing my issue (Closes #1795).

Specifically, around the side effects of setting the rack environment on the library.

Adding how it could affect middleware, and how it will show stack traces in dev & test. Also specifying the default value with a link to show all config default values.

- [x] If this PR doesn't need tests (docs change), I added `[ci skip]` to the commit messages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,6 +228,8 @@ If you want to prevent Puma from looking for a configuration file in those locat
 $ puma -C "-"
 ```
 
+The other side-effects of setting the environment are whether to show stack traces (in `development` or `test`), and setting RACK_ENV may potentially affect middleware looking for this value to change their behavior. The default puma RACK_ENV value is `development`. You can see all config default values [here](https://github.com/puma/puma/blob/12d1706ddc71b89ed2ee26275e31c788e94ff541/lib/puma/configuration.rb#L170).
+
 Check out [dsl.rb](https://github.com/puma/puma/blob/master/lib/puma/dsl.rb) to see all available options.
 
 ## Restart


nateberkopec:
Looks like you didn’t complete your sentence: “The default puma RACK_ENV value is”

dlinch:
@nateberkopec My bad, lost the word `development` when I reworded it to include all default config values. It should be fixed now. I also rearranged it slightly to keep the flow better.

nateberkopec:
Thanks!

PR #2058, opened 2019-10-29
[ci skip] fix readme single file test example

spuyet:
Not a big deal, but the provided test file does not exist anymore :wink:


nateberkopec:
Heh, thanks

PR #2063, opened 2019-10-30
[wip] Fail build if compiling c extension fails

### LINKED ISSUE #1953: Fail build if compiling extensions raises warnings

hahmed:
### Description
Fixes #1953

This is how the build looks within the current workflow, I'm not sure if you wanna compile the c code for every ruby/windows version in the matrix 🤷‍♂️

Screenshot 2019-10-30 at 22 59 14

I had a spike on a repo I did and managed to get this to work by throwing some random errors in the code which caused an error, here is my config and the screenshot of the build;

```
    steps:
    - uses: actions/checkout@v1
    - name: load ruby
      uses: actions/setup-ruby@v1
    - name: make make
      run: ruby ext/puma/http_parser11/extconf.rb
    - name: check those errors
      run: make
```

Screenshot 2019-10-30 at 23 04 35

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -40,7 +40,10 @@ jobs:
       uses: actions/setup-ruby@v1
       with:
         ruby-version: ${{ matrix.ruby }}
-
+    - name: create makefile for c extension
+      run: ruby ext/puma_http11/extconf.rb
+    - name: compile and check warnings in c extension
+      run: make
     - name: ubuntu & macos - install ragel
       if:   startsWith(matrix.os, 'ubuntu') || startsWith(matrix.os, 'macos')
       run:  |
--- a/test/test_app_status.rb
+++ b/test/test_app_status.rb
@@ -43,7 +43,6 @@ class TestAppStatus < Minitest::Test
     @app.instance_variable_set(:@auth_token, "abcdef")
 
     status, _, _ = lint('/whatever')
-    assert false
 
     assert_equal 403, status
   end
--- a/test/test_app_status.rb
+++ b/test/test_app_status.rb
@@ -43,6 +43,7 @@ class TestAppStatus < Minitest::Test
     @app.instance_variable_set(:@auth_token, "abcdef")
 
     status, _, _ = lint('/whatever')
+    assert false
 
     assert_equal 403, status
   end


MSP-Greg:
Why compile twice?  You're also compiling before build tools are updated?

hahmed:
> Why compile twice? You're also compiling before build tools are updated?

Should we run after this step: 
```
- name: compile
      run:  bundle exec rake compile
```

How can we catch ext errors without creating the makefile and running `make`?

MSP-Greg:
> How can we catch ext errors

These are not errors.  They're warnings.  Big difference.  Two versions of Ubuntu, macOS, & Windows.  

Ubuntu 16.04's Rubies were compiled with gcc 5.4.0, Ubuntu 18.04's Rubies were compiled with gcc 7.4.0.  Not sure what the OS has installed.  Window is using gcc 9.2.0.  And also whatever macOS is using.

Hence, getting all four to not throw any warnings might be tricky.

It might be helpful to throw the flags at the console though...

JFYI, with MinGW gcc 9.2.0, no warnings.  See:  https://github.com/puma/puma/runs/280509985#step:8:0

> without creating the makefile and running make?

That's what `bundle exec rake compile` does.

hahmed:
Thanks @MSP-Greg I'm going to go ahead and close this as I feel #1953 is fixed by `bundle exec rake compile`

hahmed:
> These are not errors. They're warnings. Big difference. 

Sorry my bad, I meant to try and fail the build when warnings were being shown...

Will look at the rake task compile, totally forgot that - will see if I can pass in some flags to fail the build on warning.

MSP-Greg:
@hahmed 

No problem.

CI is kind of it's own dsl.  In the past (with Travis & AppVeyor), there were two yml files.  Actions can be used with one workflow (yml file).  Some repos are doing that, some use separate workflows.

When using one workflow, there are some operations that need to be done for windows before compiling.  That's what the `MSP-Greg/msys2-action@master` action does.

Actions Windows images have multiple compilers, multiple OpenSSL locations, etc, and the action cleans that up, otherwise really odd things can happen.

> fail the build when warnings were being shown..

As mentioned above, I'm not sure if that's possible when using all the different compilers.  I do know that the job will fail if compiling errors or doesn't complete...

PR #2065, opened 2019-11-02
Add TruffleRuby to (Travis) CI [changelog skip]

nateberkopec:


nateberkopec:
Something weird is happening with the jaro_winkler install, which doesn't repro for me locally.

nateberkopec:
Just gonna put this in allow_failures anyway.

Issue #852, opened 2015-12-19
Sinatra / Puma duplicate logs


Edwardzyc:
A default sinatra/puma setup results in duplicate logs

http://stackoverflow.com/questions/31206060/sinatra-with-puma-gives-twice-the-output-in-the-terminal

There is a solution given - is this the generally accepted workaround?

Issue #2074, opened 2019-11-13
Setting `log_requests` in puma config file


jchristie55332:
**Describe the bug**

When `log_requests` is set in a puma configuration file it is not respected and as such the requests are not logged to standard out.

**Puma config:**

```
DATETIME_FORMAT = '%Y-%m-%d %H:%M:%S %z'.freeze

environment ENV.fetch('RAILS_ENV')

log_requests

log_formatter do |str|
  "#{Time.now.strftime DATETIME_FORMAT} : |Puma| #{str}"
end
```

**To Reproduce**

Install Puma into rails project and add puma config file
Boot application using `rails s`

**Expected behavior**

Application to log all requests to STDOUT

**Desktop (please complete the following information):**
 - OS: Mac OS 10.14.06
 - Puma Version: 4.3.0

composerinteralia:
I tried this out in a non-rails context on puma `master` by running `RAILS_ENV=development bundle exec bin/puma test/rackup/hello.ru -C config.rb` where `config.rb` contained the config from this comment. I did see logs for requests when I ran this, so there may be something going on here outside puma. Are you able to provide a sample Rails application that reproduces this problem?

---

Curiously, while the standard puma logs did use the custom `log_formatter`, the request logs did not:

```rb
2019-11-13 20:00:08 -0500 : |Puma| Puma starting in single mode...
2019-11-13 20:00:08 -0500 : |Puma| * Version 4.3.0 (ruby 2.6.3-p62), codename: Mysterious Traveller
2019-11-13 20:00:08 -0500 : |Puma| * Min threads: 0, max threads: 16
2019-11-13 20:00:08 -0500 : |Puma| * Environment: development
2019-11-13 20:00:08 -0500 : |Puma| * Listening on tcp://0.0.0.0:9292
2019-11-13 20:00:08 -0500 : |Puma| Use Ctrl-C to stop
127.0.0.1 - - [13/Nov/2019:20:00:12 -0500] "GET / HTTP/1.1" 200 - 0.0001
127.0.0.1 - - [13/Nov/2019:20:03:15 -0500] "GET / HTTP/1.1" 200 - 0.0002
```
I wonder if that is expected, or if there is a separate bug here with the request logs not using the `log_formatter`

jchristie55332:
[...]

PR #2075, opened 2019-11-13
remove setting quiet requests inside puma rack handler

jchristie55332:
### Description
Remove setting quiet in user settings by default

Closes `#2074`

### DIFF
--- a/test/test_rack_handler.rb
+++ b/test/test_rack_handler.rb
@@ -227,20 +227,17 @@ class TestUserSuppliedOptionsIsNotPresent < Minitest::Test
     end
   end
 
-  def test_file_log_requests_wins_over_deafult_config
+  def test_file_log_requests_wins_over_default_config
     file_log_requests_config = true
 
-    Dir.mktmpdir do |d|
-      Dir.chdir(d) do
-        FileUtils.mkdir("config")
-        File.open("config/puma.rb", "w") { |f| f << "log_requests #{file_log_requests_config}" }
+    @options[:config_files] = [
+      'test/shell/t1_conf.rb'
+    ]
 
-        conf = Rack::Handler::Puma.config(->{}, @options)
-        conf.load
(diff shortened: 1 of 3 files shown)


jchristie55332:
> It looks like this was added in [03ed48c](https://github.com/puma/puma/commit/03ed48ca903879f3b9a649f5df1cf9d6577ef3d2) to fix #852. Will we reintroduce that issue if we change the default here?

In [03ed48c](https://github.com/puma/puma/commit/03ed48ca903879f3b9a649f5df1cf9d6577ef3d2)] the default configuration was updated, which I agree with, however forcing to quiet requests in the rack handler seems unnecessary to me (as this has been defaulted already)

I have added a couple of test cases to ensure that both the user and file defaults are working as expected.  Note that the file configuration does not currently work in master branch

test/test_rack_handler.rb
```diff
@@ -226,4 +226,39 @@ def test_user_port_wins_over_config
       end
     end
   end
+
+  def test_file_log_requests_wins_over_deafult_config
```
nateberkopec: typo default

test/test_rack_handler.rb
```diff
@@ -226,4 +226,39 @@ def test_user_port_wins_over_config
...
+    file_log_requests_config = true
+
+    Dir.mktmpdir do |d|
+      Dir.chdir(d) do
```
nateberkopec: I'd rather just have a new, pre-baked config file on the filesystem, rather than create it dynamically. That way we can run the test in parallel.

nateberkopec:
So just to confirm, this PR does _not_ cause double logging with Sinatra?

nateberkopec:
I can confirm that this does not reintroduce #852. I still don't really understand what it fixes though because I haven't reproduced #2074  locally, still working on that.

nateberkopec:
Tested in Rails using provided repro and it LGTM

Issue #1996, opened 2019-09-26
Remove Puma::Server#connected_port


nateberkopec:
If it's only used in tests, it shouldn't be a public method.

* [ ] Make connected_port not public
* [ ] Provide a different way of allowing tests to see what ports we're connected to. It should support ALL ports we are currently bound to w/tcp AND ssl, not just the last one we bound to w/TCP.

nateberkopec:
A big upside of this will be that we can then rm UniquePort and use port 0 everywhere in the tests, using connected_ports (? suggested name) to figure out what we got bound to.

MSP-Greg:
> A big upside of this will be that we can then rm UniquePort

I'm not thinking about this right now (systemd).  But, what may work with 'normal Puma use'  may not work with parallel testing.

Personally, I always considered sockets to be two objects communicating on an agreed upon channel/port.  So 'port 0' always seemed odd.  Puma is only creating one of the objects.  If it was creating both, that would be different...

PR #2076, opened 2019-11-19
Removes connected_port, implements connected_ports

drews256:
### Description
This work starts the effort of removing connected_port. 

Referencing this issue:
https://github.com/puma/puma/issues/1996

I have moved the code around a bit to get all connected_ports, the tests seem
to pass in the appropriate areas.

I want to verify this path before attempting to remove all UniquePort calls in
the tests.

@nateberkopec Also, I wanna talk about removing UniquePort, as it seems like it is being used for some CLI specs and the `control-url`. Is that approriate? Do we expect to be able to pass `0` in as the port for the control and be able to get it's connected port? 

### DIFF
--- a/lib/puma/binder.rb
+++ b/lib/puma/binder.rb
@@ -226,19 +226,22 @@ module Puma
       end
 
       host = host[1..-2] if host and host[0..0] == '['
-      s = TCPServer.new(host, port)
+      tcp_server = TCPServer.new(host, port)
       if optimize_for_latency
-        s.setsockopt(Socket::IPPROTO_TCP, Socket::TCP_NODELAY, 1)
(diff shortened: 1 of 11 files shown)


nateberkopec:
> Do we expect to be able to pass 0 in as the port for the control and be able to get it's connected port?

Control servers should be able to "bind" to port 0, just like app servers, yes.

lib/puma/binder.rb
```diff
@@ -226,19 +226,22 @@ def add_tcp_listener(host, port, optimize_for_latency=true, backlog=1024)
...
+      tcp_server.listen backlog
 
-      @ios << s
-      s
+      @ios << tcp_server
+      tcp_server
     end
 
-    attr_reader :connected_port
+    attr_reader :connected_ports
```
nateberkopec: Not necessary as you define the method on the next line

Issue #2078, opened 2019-11-21
Puma does not prefer idle-workers in cluster mode


ayufan:
**Describe the bug**

We at GitLab.com are production testing Puma with prepartions for switching to it.
However, we noticed a few oddness when running Puma vs Unicorn in mostly similar configurations.

An worker of Puma when running in Cluster mode has no knowledge about siblings and their utilisation. This results in some of the sockets to be accepted by sub-optimal workers, that are already processing requests. This does have a statistical significance as in my local testing, and GitLab.com testing in the above workload we see an increase of around 20-30% of durations on P60%. Which is a significant increase.

**Capacity of Puma**

Capacity of Puma web-server is defined by `workers * threads`. Each of these
form a slot that can accept a new request. It means that each slot can accept
a new request at random (effectively round-robin).

Now, what happens if we have 2 requests waiting to be processed and two workers,
and two threads:

1. Our total capacity is 4 (W2 * T2),
2. Each request can be assigned at random to any worker, even the worker that is
   processing currently request, as we do not control that,
3. It means that in ideal scenario if we have 2 requests, for the optimal performance
   they should be assigned to two separate workers,
4. Puma does not implement any mechanism for 3., rather it is first accepting, first wins.
   It does mean that it is plausible that the two requests will be assigned in sub-optimal
   way: to the single worker, but multiple threads.
5. If the two requests are being processed by the same worker, they do share CPU-time,
   it means that they processing time is increased by the noisy neighbor due to Ruby MRI GVL.

Interestingly, the same latency impact is present on `Sidekiq`, we just don't see it,
as we do not care about real-time aspect of background processing that much.

However, the scheduling changes can improve `Sidekiq` performance as well if we would
target `sidekiq` as well.

This is more described in detail here: https://gitlab.com/gitlab-com/gl-infra/infrastructure/issues/8334#note_247859173

**Our Configuration**

We do run Puma and Unicorn in the following configurations:

1. Puma runs in `W16/T2`, the `16` corresponds to 16 CPUs available, nodes do not process anything else,
2. Unicorn runs in `W30`, the `30` seems like some artificial factor taken to have some sort of node saturation,
3. The nodes are lightly utilised, peeking at around 45% of CPU usage,
4. We seem to process around 70-80rq/sec.

We did notice oddness in Puma scheduling, as it was sometimes hitting the workers that are already processing requests. This resulted in increase of latency and duration of requests processing, even though we had a ton of spare capacity on other workers that were simply idle.

[...]

PR #2079, opened 2019-11-21
Inject small delay for busy workers to improve requests distribution

ayufan:
**Describe the bug**

We at GitLab.com are production testing Puma with prepartions for switching to it.
However, we noticed a few oddness when running Puma vs Unicorn in mostly similar configurations.

An worker of Puma when running in Cluster mode has no knowledge about siblings and their utilisation. This results in some of the sockets to be accepted by sub-optimal workers, that are already processing requests. This does have a statistical significance as in my local testing, and GitLab.com testing in the above workload we see an increase of around 20-30% of durations on P60%. Which is a significant increase.

**Capacity of Puma**

Capacity of Puma web-server is defined by `workers * threads`. Each of these
form a slot that can accept a new request. It means that each slot can accept
a new request at random (effectively round-robin).

Now, what happens if we have 2 requests waiting to be processed and two workers,
and two threads:

1. Our total capacity is 4 (W2 * T2),
2. Each request can be assigned at random to any worker, even the worker that is
   processing currently request, as we do not control that,
3. It means that in ideal scenario if we have 2 requests, for the optimal performance
   they should be assigned to two separate workers,
4. Puma does not implement any mechanism for 3., rather it is first accepting, first wins.
   It does mean that it is plausible that the two requests will be assigned in sub-optimal
   way: to the single worker, but multiple threads.
5. If the two requests are being processed by the same worker, they do share CPU-time,
   it means that they processing time is increased by the noisy neighbor due to Ruby MRI GVL.

Interestingly, the same latency impact is present on `Sidekiq`, we just don't see it,
as we do not care about real-time aspect of background processing that much.

However, the scheduling changes can improve `Sidekiq` performance as well if we would
target `sidekiq` as well.

This is more described in detail here: https://gitlab.com/gitlab-com/gl-infra/infrastructure/issues/8334#note_247859173

**Our Configuration**

We do run Puma and Unicorn in the following configurations:

1. Puma runs in `W16/T2`, the `16` corresponds to 16 CPUs available, nodes do not process anything else,
2. Unicorn runs in `W30`, the `30` seems like some artificial factor taken to have some sort of node saturation,
3. The nodes are lightly utilised, peeking at around 45% of CPU usage,
4. We seem to process around 70-80rq/sec.

We did notice oddness in Puma scheduling, as it was sometimes hitting the workers that are already processing requests. This resulted in increase of latency and duration of requests processing, even though we had a ton of spare capacity on other workers that were simply idle.

The nodes are configured today like that, due to graceful restart/overwrite/shutdown of Unicorn, to have a spare capacity to handle as many as twice amount of regular `W30`. They are highly underutilized on average due to that, but this is something to change. 

**Data**

The exact data are presented here: https://gitlab.com/gitlab-com/gl-infra/infrastructure/issues/8334#note_247824814 with the description of what we are seeing.

**Expected behavior: Ideal-scheduling algorithm in multi-threaded scenario**

In ideal scenario we would always want to use our capacity efficiently:

1. Assign to free workers first, as they do offer the best latency,
1. Assign to workers with the least amount of other requests being processed, as they are the least busy,
1. Assign to workers that are close to finish the requests being processed, as they will have a free capacity in the future (this is hard to implement, as this becomes a quite complex heuristic to understand the request completion time).

Currently, Puma does nothing from that. For round-robin we pay around 20% of ~performance penalty due to sub-optimal request processing assignment.

It is expected that the closer we get to 100% CPU-usage the more expected is that the Puma-non-scheduling algorithm will not be a problem, due to node being saturated, and threads by balanced by the kernel.

### Workaround

We can insert very small delay for busy workers, by assuming that they are already busy doing other work, and even if we schedule requests on them it will not really make them process requests faster.

The value of `5ms` is tested to provide a compromise between throughput and a delay, as we don't really want to delay forever. There's a test that tries to test that value and provide a benchmark which shows when this delay makes a difference.

### DIFF
--- a/History.md
+++ b/History.md
@@ -10,6 +10,7 @@
   * Increases maximum URI path length from 2048 to 8196 bytes (#2167)
   * Force shutdown responses can be overridden by using the `lowlevel_error_handler` config (#2203)
   * Faster phased restart and worker timeout (#2121)
+  * Inject small delay for busy workers to improve requests distribution (#2079)
 
 * Deprecations, Removals and Breaking API Changes
   * `Puma.stats` now returns a Hash instead of a JSON string (#2086)
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -664,6 +664,18 @@ module Puma
       @options[:shutdown_debug] = val
     end
 
+    # Controls an injected delay (in seconds) before
+    # accepting new socket
+    # if we are already processing requests,
+    # it gives a time for less busy workers
+    # to pick workbefore us
+    #
+    # This only affects Ruby MRI implementation
+    # The best value is between 0.001 (1ms) to 0.010 (10ms)
+    def wait_for_less_busy_worker(val)
+      @options[:wait_for_less_busy_worker] = val.to_f
+    end
+
     # Control how the remote address of the connection is set. This
     # is configurable because to calculate the true socket peer address
     # a kernel syscall is required which for very fast rack handlers
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -283,6 +283,9 @@ module Puma
               else
                 begin
                   pool.wait_until_not_full
+                  pool.wait_for_less_busy_worker(
+                    @options[:wait_for_less_busy_worker].to_f)
+
                   if io = sock.accept_nonblock
                     client = Client.new io, @binder.env(sock)
                     if remote_addr_value
--- a/lib/puma/thread_pool.rb
+++ b/lib/puma/thread_pool.rb
@@ -228,6 +228,25 @@ module Puma
       end
     end
 
+    def wait_for_less_busy_worker(delay_s)
+      # Ruby MRI does GVL, this can result
+      # in processing contention when multiple threads
+      # (requests) are running concurrently
+      return unless Puma.mri?
+      return unless delay_s > 0
+
+      with_mutex do
+        return if @shutdown
+
+        # do not delay, if we are not busy
+        return unless busy_threads > 0
+
+        # this will be signaled once a request finishes,
+        # which can happen earlier than delay
+        @not_full.wait @mutex, delay_s
+      end
+    end
+
     # If there are any free threads in the pool, tell one to go ahead
     # and exit. If +force+ is true, then a trim request is requested
     # even if all threads are being utilized.
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -116,6 +116,7 @@ module TestSkips
       when :darwin  then "Skip unless darwin"  unless RUBY_PLATFORM[/darwin/]
       when :jruby   then "Skip unless JRuby"   unless Puma.jruby?
       when :windows then "Skip unless Windows" unless Puma.windows?
+      when :mri     then "Skip unless MRI"     unless Puma.mri?
       else false
     end
     skip skip_msg, bt if skip_msg
--- a/benchmarks/wrk/cpu_spin.sh
+++ b/benchmarks/wrk/cpu_spin.sh
@@ -1,0 +1,102 @@
+#!/bin/bash
+
+set -eo pipefail
+
+ITERATIONS=400000
+HOST=127.0.0.1:9292
+URL="http://$HOST/cpu/$ITERATIONS"
+
+MIN_WORKERS=1
+MAX_WORKERS=4
+
+MIN_THREADS=4
+MAX_THREADS=4
+
+DURATION=2
+MIN_CONCURRENT=1
+MAX_CONCURRENT=8
+
+retry() {
+  local tries="$1"
+  local sleep="$2"
+  shift 2
+
+  for i in $(seq 1 $tries); do
+    if eval "$@"; then
+      return 0
+    fi
+
+    sleep "$sleep"
+  done
+
+  return 1
+}
+
+ms() {
+  VALUE=$(cat)
+  FRAC=${VALUE%%[ums]*}
+  case "$VALUE" in
+    *us)
+      echo "scale=1; ${FRAC}/1000" | bc
+      ;;
+
+    *ms)
+      echo "scale=1; ${FRAC}/1" | bc
+      ;;
+
+    *s)
+      echo "scale=1; ${FRAC}*1000/1" | bc
+      ;;
+  esac
+}
+
+run_wrk() {
+  result=$(wrk -H "Connection: Close" -c "$wrk_c" -t "$wrk_t" -d "$DURATION" --latency "$@" | tee -a wrk.txt)
+  req_sec=$(echo "$result" | grep "^Requests/sec:" | awk '{print $2}')
+  latency_avg=$(echo "$result" | grep "^\s*Latency.*%" | awk '{print $2}' | ms)
+  latency_stddev=$(echo "$result" | grep "^\s*Latency.*%" | awk '{print $3}' | ms)
+  latency_50=$(echo "$result" | grep "^\s*50%" | awk '{print $2}' | ms)
+  latency_75=$(echo "$result" | grep "^\s*75%" | awk '{print $2}' | ms)
+  latency_90=$(echo "$result" | grep "^\s*90%" | awk '{print $2}' | ms)
+  latency_99=$(echo "$result" | grep "^\s*99%" | awk '{print $2}' | ms)
+
+  echo -e "$workers\t$threads\t$wrk_c\t$wrk_t\t$req_sec\t$latency_avg\t$latency_stddev\t$latency_50\t$latency_75\t$latency_90\t$latency_99"
+}
+
+run_concurrency_tests() {
+  echo
+  echo -e "PUMA_W\tPUMA_T\tWRK_C\tWRK_T\tREQ_SEC\tL_AVG\tL_DEV\tL_50%\tL_75%\tL_90%\tL_99%"
+  for wrk_c in $(seq $MIN_CONCURRENT $MAX_CONCURRENT); do
+    wrk_t="$wrk_c"
+    eval "$@"
+    sleep 1
+  done
+  echo
+}
+
+with_puma() {
+  # start puma and wait for 10s for it to start
+  bundle exec bin/puma -w "$workers" -t "$threads" -b "tcp://$HOST" -C test/config/cpu_spin.rb &
+  local puma_pid=$!
+  trap "kill $puma_pid" EXIT
+
+  # wait for Puma to be up
+  if ! retry 10 1s curl --fail "$URL" &>/dev/null; then
+    echo "Failed to connect to $URL."
+    return 1
+  fi
+
+  # execute testing command
+  eval "$@"
+  kill "$puma_pid" || true
+  trap - EXIT
+  wait
+}
+
+for workers in $(seq $MIN_WORKERS $MAX_WORKERS); do
+  for threads in $(seq $MIN_THREADS $MAX_THREADS); do
+    with_puma \
+      run_concurrency_tests \
+      run_wrk "$URL"
+  done
+done
--- a/test/config/cpu_spin.rb
+++ b/test/config/cpu_spin.rb
@@ -1,0 +1,17 @@
+# call with "GET /cpu/<d> HTTP/1.1\r\n\r\n",
+# where <d> is the number of iterations
+
+require 'benchmark'
+
+# configure `wait_for_less_busy_workers` based on ENV, default `true`
+wait_for_less_busy_worker ENV.fetch('WAIT_FOR_LESS_BUSY_WORKERS', '0.005').to_f
+
+app do |env|
+  iterations = (env['REQUEST_PATH'][/\/cpu\/(\d.*)/,1] || '1000').to_i
+
+  duration = Benchmark.measure do
+    iterations.times { rand }
+  end
+
+  [200, {"Content-Type" => "text/plain"}, ["Run for #{duration.total} #{Process.pid}"]]
+end
--- a/test/test_busy_worker.rb
+++ b/test/test_busy_worker.rb
@@ -1,0 +1,106 @@
+require_relative "helper"
+require "puma/events"
+
+class TestBusyWorker < Minitest::Test
+  parallelize_me!
+
+  def setup
+    @ios = []
+    @server = nil
+  end
+
+  def teardown
+    @server.stop(true) if @server
+    @ios.each {|i| i.close unless i.closed?}
+  end
+
+  def new_connection
+    TCPSocket.new('127.0.0.1', @server.connected_ports[0]).tap {|s| @ios << s}
+  rescue IOError
+    Thread.current.purge_interrupt_queue if Thread.current.respond_to? :purge_interrupt_queue
+    retry
+  end
+
+  def send_http(req)
+    new_connection << req
+  end
+
+  def send_http_and_read(req)
+    send_http(req).read
+  end
+
+  def with_server(**options, &app)
+    @requests_count = 0 # number of requests processed
+    @requests_running = 0 # current number of requests running
+    @requests_max_running = 0 # max number of requests running in parallel
+    @mutex = Mutex.new
+
+    request_handler = ->(env) do
+      @mutex.synchronize do
+        @requests_count += 1
+        @requests_running += 1
+        if @requests_running > @requests_max_running
+          @requests_max_running = @requests_running
+        end
+      end
+
+      begin
+        yield(env)
+      ensure
+        @mutex.synchronize do
+          @requests_running -= 1
+        end
+      end
+    end
+
+    @server = Puma::Server.new request_handler, Puma::Events.strings, **options
+    @server.min_threads = options[:min_threads] || 0
+    @server.max_threads = options[:max_threads] || 10
+    @server.add_tcp_listener '127.0.0.1', 0
+    @server.run
+  end
+
+  # Multiple concurrent requests are not processed
+  # sequentially as a small delay is introduced
+  def test_multiple_requests_waiting_on_less_busy_worker
+    skip_unless :mri
+
+    with_server(wait_for_less_busy_worker: 1.0) do |_|
+      sleep(0.1)
+
+      [200, {}, [""]]
+    end
+
+    n = 2
+
+    Array.new(n) do
+      Thread.new { send_http_and_read "GET / HTTP/1.0\r\n\r\n" }
+    end.each(&:join)
+
+    assert_equal n, @requests_count, "number of requests needs to match"
+    assert_equal 0, @requests_running, "none of requests needs to be running"
+    assert_equal 1, @requests_max_running, "maximum number of concurrent requests needs to be 1"
+  end
+
+  # Multiple concurrent requests are processed
+  # in parallel as a delay is disabled
+  def test_multiple_requests_processing_in_parallel
+    skip_unless :mri
+
+    with_server(wait_for_less_busy_worker: 0.0) do |_|
+      sleep(0.1)
+
+      [200, {}, [""]]
+    end
+
+    n = 4
+
+    Array.new(n) do
+      Thread.new { send_http_and_read "GET / HTTP/1.0\r\n\r\n" }
+    end.each(&:join)
+
+    assert_equal n, @requests_count, "number of requests needs to match"
+    assert_equal 0, @requests_running, "none of requests needs to be running"
+    assert_equal n, @requests_max_running, "maximum number of concurrent requests needs to match"
+  end
+end


lib/puma/thread_pool.rb
```diff
@@ -213,6 +217,25 @@ def wait_until_not_full
...
+
+          # if we do wait for more than busy threads
+          # and what is in backlog, stop
+          break if tick >= busy + @todo.size
+
+          # DEBUG info
+          puts "Injecting latency >>> tick:#{tick} busy:#{busy} backlog:#{@todo.size}."
+
+          # this will be signaled once a request finishes,
+          # which can happen ealier than tick time
```
olleolleolle: typo: ealier -> earlier

evanphx:
This along with the original Issue are well researched and explained, thanks so much for that!

Would you be wiling to test a suite of values for latency and tick so we can get a feel for how they influence the average response time?

Could you run your test against a request that only lasts 150ms so we can get a feel for if your patch causes a meaningful slowdown in it? If it does, one option would be to sample average time to complete a job and feed it back into the wait algorithm so that it only ever attempts to sleep for, say, 1% or less of the average request time.

ayufan:
I work on doing some tests on GitLab.com to validate it :)

lib/puma/server.rb
```diff
@@ -214,6 +214,13 @@ def run_lopez_mode(background=true)
       end
     end
 
+    def wait_for_threads_to_finish
+      @thread_pool.wait_for_threads_to_finish(
+        tick_time: ENV.fetch('PUMA_INJECT_LATENCY', '0.001').to_f,
+        max_wait_ticks: ENV.fetch('PUMA_INJECT_WAIT_TICKS', '3').to_i
```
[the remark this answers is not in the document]
ayufan: Not really. I was yet testing at all behaviour of max ticks or even if it is needed or if it helps with anything. I was assuming that after some time it simply does not matter. Worker is already so busy, that we are likely hitting saturation at this point.

The best would be to have some heuristic based on the current workload to know estimate of worker being freed, but this becomes too complex to handle. So, being dumb, likely doing 1 or 2 ticks should suffice in such cases.

ayufan:
Yes, this approach particularly idle workers, or semi-idle. I did not see the other one to have big benefits for workloads and our configuration. I think it is too late to make that decision. The other one just prevents workers to pick new connections if they are already fully utilised (it seems), so it might seem that the other one is beneficial as well.

wjordan:
> I did not see the other one to have big benefits for workloads and our configuration.

By 'the other one' did you mean specifically the code in #1646 or the different code in #1920? Note that the code in #1920 is incorrectly implemented as I mentioned, and would have no effect.

ayufan:
Let me check my notes @wjordan, but also I was testing a very specific condition :) 

Yes. I was testing the #1920: https://docs.google.com/spreadsheets/d/1y2YqrPPgZ-RtjKiCGplJ7YkwjMXkZx6vEPq7prFbUn4/edit#gid=0. It shows no effect on my tests.

wjordan:
I tested out this PR on the Sinatra-MySQL TechEmpower benchmark mentioned in https://github.com/puma/puma/pull/1646#issuecomment-420845786:
- `m4.10xlarge` EC2 instance with 40 vCPUs
- Ruby 2.5.7-p206
- Puma configured with 40 workers, 5 threads (1 thread for comparison)
- Test configured for 40 concurrent requests, 20 queries per request (`./tfb --test sinatra --concurrency-levels 40 --type query --query-levels 20`)

Results (request/sec):

1 thread: **3917.86**

5 threads:
- `master`: **3119.52**
- `ayufan-research:gitlab-latency` (#2079, this PR): **3402.19**
- `wjordan:balance_requests` (#1646): **3932.26**

ayufan:
Thanks @wjordan. Let me check what happens with your changes when I run against my testing suite :)

ayufan:
## Puma BASELINE without patches

```
➜  gitlab-rails ./run-ab-projects
100	1	8.757	87.572
200	2	9.589	95.886
300	3	10.418	104.176
400	4	11.556	115.563
500	5	13.994	139.937
600	6	15.488	154.883
700	7	18.259	182.589
800	8	20.339	203.389
➜  gitlab-rails ./run-ab-projects
100	1	8.891	88.907
200	2	9.325	93.254
300	3	10.471	104.706
400	4	12.004	120.038
500	5	13.212	132.125
600	6	15.558	155.583
700	7	18.037	180.374
800	8	20.100	201.004
➜  gitlab-rails ./run-ab-projects
100	1	8.967	89.672
200	2	9.489	94.889
300	3	10.759	107.591
400	4	11.720	117.197
500	5	13.355	133.551
600	6	15.439	154.391
700	7	18.122	181.223
800	8	20.236	202.364
```

## PUMA with PR 2079

> https://github.com/puma/puma/commit/30f2206bb9d31bc9527b697f6948f7963315df88.patch

```
curl https://github.com/puma/puma/commit/30f2206bb9d31bc9527b697f6948f7963315df88.patch | patch -p1 -d /opt/gitlab/embedded/lib/ruby/gems/2.6.0/gems/puma-4.3.1/
```

```
➜  gitlab-rails ./run-ab-projects
100	1	9.506	95.056
200	2	9.586	95.860
300	3	10.281	102.806
400	4	10.811	108.114
500	5	13.427	134.265
600	6	16.204	162.037
700	7	18.505	185.049
800	8	21.025	210.249
➜  gitlab-rails ./run-ab-projects
100	1	8.932	89.320
200	2	9.015	90.146
300	3	9.749	97.495
400	4	10.317	103.170
500	5	12.891	128.913
600	6	15.084	150.837
700	7	17.667	176.668
800	8	20.239	202.388
➜  gitlab-rails ./run-ab-projects
100	1	9.185	91.852
200	2	9.243	92.427
300	3	9.557	95.567
400	4	10.269	102.689
500	5	12.625	126.253
600	6	15.184	151.837
700	7	17.795	177.945
800	8	19.925	199.245
➜  gitlab-rails ./run-ab-projects
100	1	8.879	88.789
200	2	9.056	90.557
300	3	9.596	95.961
400	4	10.370	103.698
500	5	13.024	130.237
600	6	15.276	152.763
700	7	17.490	174.896
800	8	20.303	203.026
```

### Puma with PR 1646

> https://patch-diff.githubusercontent.com/raw/puma/puma/pull/1646.patch

```
# I needed to fix manually, as it is not compatible with Puma 4.3.1
curl https://patch-diff.githubusercontent.com/raw/puma/puma/pull/1646.patch | patch -p1 -d /opt/gitlab/embedded/lib/ruby/gems/2.6.0/gems/puma-4.3.1/
```

```
➜  gitlab-rails ./run-ab-projects
100	1	9.353	93.534
200	2	10.097	100.967
300	3	11.138	111.383
400	4	12.623	126.234
500	5	14.353	143.528
600	6	16.087	160.866
700	7	18.354	183.539
800	8	21.022	210.224
➜  gitlab-rails ./run-ab-projects
100	1	8.861	88.609
200	2	9.608	96.078
300	3	11.007	110.065
400	4	11.523	115.233
500	5	13.487	134.872
600	6	15.591	155.913
700	7	18.424	184.245
800	8	20.288	202.876
➜  gitlab-rails ./run-ab-projects
100	1	8.918	89.179
200	2	9.255	92.552
300	3	10.456	104.563
400	4	11.964	119.635
500	5	13.714	137.145
600	6	16.010	160.098
700	7	17.923	179.234
800	8	20.582	205.822
➜  gitlab-rails ./run-ab-projects
100	1	8.928	89.279
200	2	9.604	96.035
300	3	10.933	109.333
400	4	11.741	117.414
500	5	13.128	131.279
600	6	15.790	157.905
700	7	18.070	180.696
800	8	20.592	205.915
```

## `run-ab-projects`

```bash
#!/bin/bash

run_ab() {
    result=$(ab -q -n "$1" -c "$2" http://localhost:4080/api/v4/projects)
    time_taken=$(echo "$result" | grep "Time taken for tests:" | cut -d" " -f7)
    time_per_req=$(echo "$result" | grep "Time per request:" | grep "(mean)" | cut -d" " -f10)

    echo -e "$1\t$2\t$time_taken\t$time_per_req"
}

for i in 1 2 3 4 5 6 7 8; do
    run_ab $((i*100)) $i
    sleep 1
done
```

ayufan:
I also put the results in: https://docs.google.com/spreadsheets/d/1y2YqrPPgZ-RtjKiCGplJ7YkwjMXkZx6vEPq7prFbUn4/edit#gid=1614859337.

ayufan:
We are running the `cond.wait` patch (this PR) on `GitLab.com` and we see a performance penalty in our testing. This is a change to before running patch.

ayufan:
We are running Puma on `gitlab.com` on 25% of fleet. So, far it really seems that patch does make Puma to be equal to Unicorn in our workload.

We are discussing that in: https://gitlab.com/gitlab-com/gl-infra/infrastructure/issues/8334#note_277452635.

The results from running today will be posted tomorrow, but so far distributions look pretty much equal between Puma/Unicorn. It means that threading does not affect for us on the request processing.

ayufan:
@nateberkopec @olleolleolle @wjordan 

It seems that simple change has a big impact for us: https://gitlab.com/gitlab-com/gl-infra/infrastructure/issues/8334#note_279516488. Doing extensive tests on `GitLab.com` shows no noticeable difference between Unicorn (W30/W40 (depends on service)) and Puma (W16/T4).

All our servers have 16 CPUs assigned.

I will look more into understanding @wjordan performance measurements.

nateberkopec:
I'm caught up on your thread on Gitlab now, but it looks like you still have some measurement issues re: the patch's effectiveness, correct?

ayufan:
@nateberkopec 

I would say not really, just comparison between one node vs another gives wrong results for various reasons: like traffic distribution, or variability in performance between nodes. This was also the reason why we started running at 23% of fleet to reduce these variability. 

Anyway, we will do another set of disables to have bigger sample sets. Now, it seems that the impact is substantial, and makes all execution duration to be pretty consistent with Unicorn (at least for us).

ayufan:
As of today GitLab.com runs fully on Puma with this patch: https://gitlab.com/gitlab-com/gl-infra/production/issues/1684.

I will now spend some time on comparing and figuring out which solution to ask for upstreaming :)

Thanks for your help!

lib/puma/const.rb
```diff
@@ -100,7 +100,7 @@ class UnsupportedOption < RuntimeError
   # too taxing on performance.
   module Const
 
-    PUMA_VERSION = VERSION = "4.3.0".freeze
+    PUMA_VERSION = VERSION = "4.3.1".freeze
```
schneems: I think you may need to rebase with master to get rid of this change. In general we'll rev the version.

History.md
```diff
@@ -1,11 +1,16 @@
...
+  * Inject small latency to make Puma to prefer workers that are more idle (#2079)
 
 * Bugfixes
   * Your bugfix goes here (#Github Number)
 
+## 4.3.1 and 3.12.2 / 2019-12-05
+
+* Security
+  * Fix: a poorly-behaved client could use keepalive requests to monopolize Puma's reactor and create a denial of service attack. CVE-2019-16770.
+
```
schneems: I think you may need to rebase with master to get rid of this change.

lib/puma/server.rb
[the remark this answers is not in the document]
schneems: Env.fetch has some perf penalties. Can we read it from the environment once and store somwhere in a variable?

```ruby
require 'benchmark/ips'


variable = 0.001

Benchmark.ips do |x|
  x.report("env fetch") { ENV.fetch("FOO", 0.001) }
  x.report("variable") { variable }
  x.compare!
end

# Warming up --------------------------------------
#            env fetch   247.018k i/100ms
#             variable   382.650k i/100ms
# Calculating -------------------------------------
#            env fetch      5.572M (± 3.3%) i/s -     27.913M in   5.015719s
#             variable     21.348M (± 3.7%) i/s -    106.759M in   5.008674s

# Comparison:
#             variable: 21348220.1 i/s
#            env fetch:  5571895.6 i/s - 3.83x  slower
```

lib/puma/const.rb
```diff
@@ -118,6 +118,13 @@ module Const
     # sending data back
     WRITE_TIMEOUT = 10
 
+    # How many requests to attempt inline before sending a client back to
+    # the reactor to be subject to normal ordering. The idea here is that
+    # we amortize the cost of going back to the reactor for a well behaved
+    # but very "greedy" client across 10 requests. This prevents a not
+    # well behaved client from monopolizing the thread forever.
+    MAX_FAST_INLINE = 10
+
```
schneems: Need rebase

lib/puma/server.rb
```diff
@@ -479,7 +490,19 @@ def process_client(client, buffer)
...
+
+            if requests >= MAX_FAST_INLINE
+              # This will mean that reset will only try to use the data it already
+              # has buffered and won't try to read more data. What this means is that
+              # every client, independent of their request speed, gets treated like a slow
+              # one once every MAX_FAST_INLINE requests.
+              check_for_more_data = false
+            end
+
+            unless client.reset(check_for_more_data)
```
schneems: Need rebase.

lib/puma/thread_pool.rb
```diff
@@ -213,6 +217,22 @@ def wait_until_not_full
...
+          # and what is in backlog, stop
+          break if tick >= busy + @todo.size
+
+          # this will be signaled once a request finishes,
+          # which can happen earlier than tick time
+          @not_full.wait @mutex, tick_time
+        end
+      end
+    end
+
```
schneems: This is the main meat-and-potatoes. The wait with timeout makes sense to me. Why do we need the "tick" here? What's the logic/reason behind doing this multiple times?

ayufan:
@schneems 

Thanks for review, this is to showcase approach, not something yet to be merged in that form (WIP), even though we already use it :) It was done like that to give us an ability to easily tinker via ENVs without changing any code.

I will remove the aspects that are not relevant: `env.fetch`, `ticks`, etc. to keep it truly minimal change.

lib/puma/thread_pool.rb
nateberkopec: @schneems https://gitlab.com/gitlab-com/gl-infra/infrastructure/issues/8334#note_247859173

ayufan:
The failure of `out_of_bound_stream` is due to: https://github.com/puma/puma/pull/2218#discussion_r413982179.

The test assumes that all requests are processed in parallel, ie. multiple threads is processing data in parallel, which might not always be true:
- scheduler might delay acceptance of the socket, till it is free even if it is has more threads
- test does fail if run `max_threads=1` which creates the sequence of execution, as test works on a failing edge
- the `out_of_bound` seems to hold a mutex which might result in the same contention problem that likely was tried to be solved before

lib/puma/thread_pool.rb
```diff
@@ -208,6 +213,24 @@ def wait_until_not_full
...
+      with_mutex do
+        return if @shutdown
+
+        # do not delay, if we are not busy
+        return unless busy_threads > 0
+
+        # this will be signaled once a request finishes,
+        # which can happen earlier than delay
+        @not_full.wait @mutex, BUSY_WORKER_DELAY
```
nateberkopec, line 230: Are you running this particular variation in prod? It looks like you decided the ticks concept/more-loaded-workers-wait-longer idea was not necessary.

nateberkopec: It would be cool if we could set/get watcher priority in libev through nio4r, then we could do this w/o any latency at all. But that's `TODO`.

nateberkopec:
I need to actually run the bm myself and then check out the `oob` interaction, but overall the change seems very simple and makes sense to me.

lib/puma/thread_pool.rb
ayufan: Not this particular one, once we agree that this is OK. I will push this change to `gitlab.com` and measure it there.

ayufan:
I updated specs for OOB to have less likely of hitting race condition in tests. Adding small delay seems to do the trick and allows us to properly test all scenarios, by ensuring that requests processing is not "so immediate", as currently `/` for OOB was immediate.

@wjordan @nateberkopec WDYT?

lib/puma/dsl.rb
```diff
@@ -664,6 +664,14 @@ def shutdown_debug(val=true)
       @options[:shutdown_debug] = val
     end
 
+    # Controls whether to inject small delay before
+    # accepting new socket
+    # if we are already processing requests
+    # this allows less busy workers to pick before us
+    def wait_for_less_busy_worker(val=true)
```
dentarg, line 671: Should we mention here this only does something on MRI?

test/helper.rb
```diff
@@ -116,6 +116,7 @@ def skip_unless(eng, bt: caller)
       when :darwin  then "Skip unless darwin"  unless RUBY_PLATFORM[/darwin/]
       when :jruby   then "Skip unless JRuby"   unless Puma.jruby?
       when :windows then "Skip unless Windows" unless Puma.windows?
+      when :mri     then "Skip unless MRI" unless Puma.mri?
```
dentarg, line 119:
```suggestion
      when :mri     then "Skip unless MRI"     unless Puma.mri?
```

lib/puma/dsl.rb
ayufan: Good idea.

test/helper.rb
ayufan: Sorry, yes, my omission. Fixing it :)

ayufan:
@nateberkopec I asked our production team to change params of old patch: https://gitlab.com/gitlab-org/gitlab/-/issues/196002#note_330784539. The old patch is still equally the same, as we can control all aspects of it, with configuring params. This PR is simplified version with these values hardcoded.

ayufan:
Some results from running the benchmark (included): https://docs.google.com/spreadsheets/d/1y2YqrPPgZ-RtjKiCGplJ7YkwjMXkZx6vEPq7prFbUn4/edit#gid=1510215191. The exact numbers are fully dependent on a workload, this hopefully somehow shows that interleaved workflow.

ayufan:
@nateberkopec WDYT? What would be the next steps? :)

nateberkopec:
I'm gonna take some time this weekend to review 👍 The code change is simple, so the main thing is figuring out the impact and benchmarking.

schneems:
> figuring out the impact and benchmarking

Agreed, that's a tall order. We could publish this in a pre-release and then solicit people to try it out, though finding the people and getting meaningful feedback there has always been an issue.  We can promote via twitter and slack etc. it's better than nothing.

nateberkopec:
Most def. Once the 5.0 milestone is complete I'll release 5.0.0.beta1

ayufan:
@nateberkopec

I rebased on `master` as there were some changes merge related to OOB:

- I added a config setting where you specify delay, which is `off` by default
- This makes it not yet default, but allows us to merge/test without the impact on existing users without them doing a very simple addition
- I reworked a `cpu_spin.rb` to make it easier
- I added additional `test.rb` to very simply test that, since this is `sleep-based` I had to use small `sleeps`, this test is pretty fast anyway: `~0.238589s`

Do you think this to be acceptable in that form to merge and continue testing this non-blocking/non-breaking change?

lib/puma/thread_pool.rb
```diff
@@ -228,6 +228,25 @@ def wait_until_not_full
       end
     end
 
+    def wait_for_less_busy_worker(delay_s)
+      # Ruby MRI does GVL, this can result
+      # in processing contention when multiple threads
+      # (requests) are running concurrently
+      return unless Puma.mri?
```
nateberkopec, line 235: This is adding two method calls to the hot path, but for now I'll leave them in. Just pointing out for future optimiation that we could metaprogram these out based on config.

nateberkopec:
Excellent work @ayufan. I'm going to write up a guide to 5.0 and the beta which will provide more guidance for using this feature, but excited to see reports/data back when this goes live.

ayufan:
Thank you @nateberkopec !

mwpastore:
Amazing! Thank you all!

lib/puma/thread_pool.rb
ghiculescu: It might be worth swapping the order of calls.

```
require "benchmark/ips"

def mri?
  RUBY_ENGINE == 'ruby' || RUBY_ENGINE.nil?
end

num = rand(-10..10)

Benchmark.ips do |b|
  b.report("mri?") { mri? }

  b.report("delay_s > 0") { num > 0 }

  b.compare!
end
```

```
                mri?   714.254k i/100ms
         delay_s > 0     1.675M i/100ms
                mri?      7.118M (± 3.3%) i/s -     35.713M in   5.023186s
         delay_s > 0     15.494M (±10.6%) i/s -     77.054M in   5.055471s

Comparison:
         delay_s > 0: 15493847.0 i/s
                mri?:  7118414.2 i/s - 2.18x  (± 0.00) slower
```

(I'm not sure how hot this path runs / how much it matters.)

lib/puma/server.rb
```diff
@@ -283,6 +283,9 @@ def handle_servers
               else
                 begin
                   pool.wait_until_not_full
+                  pool.wait_for_less_busy_worker(
+                    @options[:wait_for_less_busy_worker].to_f)
```
ghiculescu, line 287: There's a micro-optimisation here, which I think is worth flagging only because this happens in a while loop. `to_f` is already called when the option's value is set, [here](https://github.com/puma/puma/pull/2079/files#diff-122c03b9c96f060a9cc29b5ed3446467e1909b0054b4f5cfd95ee41cfff832e7R676).

```ruby
require "benchmark/ips"

options = {foo: 0.5}

Benchmark.ips do |b|
  b.report("to_f on option") { options[:foo].to_f }

  b.report("option directly") { options[:foo] }

  b.compare!
end
```

```
      to_f on option     1.593M i/100ms
     option directly     2.169M i/100ms
      to_f on option     16.280M (± 1.2%) i/s -     82.842M in   5.089545s
     option directly     20.139M (± 8.5%) i/s -    101.943M in   5.103727s

Comparison:
     option directly: 20138636.0 i/s
      to_f on option: 16279512.8 i/s - 1.24x  (± 0.00) slower
```

lib/puma/thread_pool.rb
ghiculescu: https://github.com/puma/puma/pull/2579

matthiassb: Would ruby 2.2.1 be impacted by MRI GVL as well?

schneems:
I wrote some more details about this behavior in https://github.com/puma/puma/issues/2078. I've also been writing about this optimization and a lot of people are confused about how exactly it works. I want to give a brief recap to attempt to explain how I see it now:

## Puma accept loop explained

Puma has an "accept loop" (in default boot configuration) where it will continuously call "accept" on the socket independently of the rest of the server. I originally thought it would call once, and block, yielding like IO until the OS got a request and woke it up. It does not do that. Instead, it calls this logic with a timeout of 0, which returns immediately if there are no queued requests. Since it's in a loop, it calls some other housekeeping logic, and then immediately goes back to calling "accept" with a zero timeout.

That means that by default when there are no requests, all workers are racing to be the first one to call "accept" on the socket after the request has been received.

## How the "sleep sort" of this PR works

**Before this change:** All workers called "accept" just as fast whether they were fully overloaded (all threads busy) or if they were sitting idle with no requests.

**After this change:** Workers that are 100% idle now loop as fast as possible to race for the incoming connection. Workers who have even 1 busy thread now have a sleep in their loop. This encourages (but does not guarantee) that a completely idle worker will get any new connections first. This is what is meant by "sleep sort."

## Puma 7 sleep sort changes

With Puma 7, we changed the logic such that, instead of sleeping for a fixed amount, the accept loop sleeps proportionally to the load it's under. I.e. a worker with 4 requests being processed will sleep twice as long as a worker with 2. You can see the most recent logic in my PR https://github.com/puma/puma/pull/3746. 

Fundamentally, this series of "sleep sort" optimizations is about prioritizing workers who are less busy to call "accept" more. There might be second and third order interactions such as between the GIL/GVL and OS scheduler, and things like CPU affinity and CPU cache, etc. However, the main logic is centered on slowing down work ingestion for resources that are busier.

PR #2086, opened 2019-12-09
Puma.stats is now a hash, not JSON.

bdewater:
## Description
Having access to the hash allows to produce stats in other ways (such as StatsD) without having to parse JSON of data that is available in memory. Examples of this workaround are:
- https://github.com/yob/puma-plugin-statsd/blob/fa6ba1f5074473618643ef7cf99747801a001dec/lib/puma/plugin/statsd.rb#L112-L114
- https://github.com/appsignal/appsignal-ruby/pull/488/files#diff-72a8958f52e3339e140e6434a391eba0R49
- https://gitlab.com/gitlab-org/gitlab-foss/blob/2c72668c3bd2838ff101815259d0bb704e845493/lib%2Fgitlab%2Fmetrics%2Fsamplers%2Fpuma_sampler.rb#L29

Changed `Puma.stats` to `Puma.stats_json` to help disambiguate. An alias is provided for backwards compatibility. In a new major release `Puma.stats` could point to the hash version or be removed altogether.

### DIFF
--- a/History.md
+++ b/History.md
@@ -3,6 +3,7 @@
 * Features
   * Add pumactl `thread-backtraces` command to print thread backtraces (#2053)
   * Configuration: `environment` is read from `RAILS_ENV`, if `RACK_ENV` can't be found (#2022)
+  * `Puma.stats` now returns a Hash instead of a JSON string (#2086)
 
 * Bugfixes
   * Your bugfix goes here (#Github Number)
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -73,7 +73,7 @@ module Puma
         @first_term_sent = nil
         @started_at = Time.now
         @last_checkin = Time.now
-        @last_status = '{}'
+        @last_status = {}
         @term = false
       end
 
@@ -94,7 +94,11 @@ module Puma
 
       def ping!(status)
         @last_checkin = Time.now
-        @last_status = status
+        captures = status.match(/{ "backlog":(?<backlog>\d*), "running":(?<running>\d*), "pool_capacity":(?<pool_capacity>\d*), "max_threads": (?<max_threads>\d*) }/)
+        @last_status = captures.names.inject({}) do |hash, key|
+          hash[key.to_sym] = captures[key].to_i
+          hash
+        end
       end
 
       def ping_timeout?(which)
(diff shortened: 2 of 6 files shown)


nateberkopec:
>  In a new major release Puma.stats could point to the hash version or be removed altogether.

Luckily for us the next release will probably be 5.0.

Anyone else have thoughts on this? Consider the proposal basically changing `Puma.stats` to a hash, so the upgrade path will be "everyone using Puma.stats should change to `Puma.stats.to_json`".

lib/puma.rb
```diff
@@ -19,10 +19,16 @@ def self.stats_object=(val)
     @get_stats = val
   end
 
-  def self.stats
-    @get_stats.stats
+  def self.stats_hash
```
nateberkopec: FWIW I don't like the rename. We should either add stats_hash or just change stats to a hash and bump major.

bdewater: I was guessing this would land on a 4.x release, but happy to change :)

bdewater:
> Consider the proposal basically changing `Puma.stats` to a hash, so the upgrade path will be "everyone using Puma.stats should change to `Puma.stats.to_json`".

Done :)

lib/puma.rb
nateberkopec: Yeah no way for you to know really but I'm targeting the next release to be a major bump 👍

History.md
```diff
@@ -3,6 +3,7 @@
 * Features
   * Add pumactl `thread-backtraces` command to print thread backtraces (#2053)
   * Configuration: `environment` is read from `RAILS_ENV`, if `RACK_ENV` can't be found (#2022)
+  * `Puma.stats` now returns a Hash instead of a JSON string
```
nateberkopec: Needs a link to your PR number (#2086).

lib/puma/cluster.rb
```diff
@@ -94,7 +94,11 @@ def term?
 
       def ping!(status)
         @last_checkin = Time.now
-        @last_status = status
+        captures = status.match(/{ "backlog":(?<backlog>\d*), "running":(?<running>\d*), "pool_capacity":(?<pool_capacity>\d*), "max_threads": (?<max_threads>\d*) }/)
```
nateberkopec: Shouldn't we just deserialize the JSON response, really?

lib/puma/cluster.rb
```diff
@@ -352,9 +356,26 @@ def reload_worker_directory
     # the master process.
     def stats
```
nateberkopec: this is so much more readable now ❤️

lib/puma/cluster.rb
bdewater: There seems to have been an intentional decision made in the past to not let Puma rely on the JSON gem. I don't have the context as to why but since this is an internal message being passed I figured it wasn't worth adding the dependency as we don't need more robust parsing. I can change it if you still think it is better to do so.

History.md
nateberkopec: ^^^^ Still needs this.

nateberkopec:
Needs the changelog mod, then it's g2g.

bdewater:
Changelog added & squashed to one commit :)

nateberkopec:
Thanks! I'll be writing an upgrade guide later, will be sure to mention this.

schneems:
I'm going to move to remove this change. It will break the `barnes` gem or anyone and only mysteriously in production. I'm proposing we introduce a new method so we can preserve backward compatibility I'm thinking `Puma.stats_hash` and preserving the current interface. If we still want to remove `Puma.stats` and it's current behavior, we should go through a deprecation cycle i.e. cut a release to start emitting warnings right now before releasing a major breaking change.

bdewater:
That gem has the exact behaviour why I opened this PR, why parse JSON of what you have in memory? It's trivial to work around.

schneems:
> That gem has the exact behaviour why I opened this PR, why parse JSON of what you have in memory? It's trivial to work around.

It's already existing behavior and would break countless apps if it changed. The proposed behavior is good, but lets make a non breaking change to get it

schneems:
I've got a spiked PR of this in https://github.com/puma/puma/pull/2253. Tests are passing, can y'all take a look?

PR #2087, opened 2019-12-15
GitHub Actions no longer have Ruby 2.3.x, checkout w/RVM [changelog skip]

dentarg:
Using RVM to install 2.3.8 like RubyGems does: https://github.com/rubygems/rubygems/blob/v3.1.2/.github/workflows/ubuntu-rvm.yml

...with OpenSSL 1.0.2 like @MSP-Greg says: #2087 (comment)

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -16,21 +16,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ 'ubuntu-16.04', 'ubuntu-18.04', 'macos', 'windows-latest' ]
-        ruby: [ '2.3.x', '2.4.x', '2.5.x', '2.6.x' ]
-        exclude:
-        - os: ubuntu-16.04
-          ruby: 2.4.x
-        - os: ubuntu-16.04
-          ruby: 2.5.x
-        - os: ubuntu-16.04
-          ruby: 2.6.x
-        - os: ubuntu-18.04
-          ruby: 2.3.x
-        - os: macos
-          ruby: 2.3.x
-        - os: windows-latest
-          ruby: 2.3.x
+        os: [ 'ubuntu-18.04', 'macos', 'windows-latest' ]
+        ruby: [ '2.4.x', '2.5.x', '2.6.x' ]
     steps:
     - name: repo checkout
       uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


nateberkopec:
Lame. We'll need to add it back to the Travis matrix or use Docker on Github Actions to do this build then. I don't want to merge until we have a replacement.

dentarg:
Sure...

MSP-Greg:
Should pass without it, but it can't find OpenSSL...

@nateberkopec  I can add Ruby 2.3 & master to Windows, but best done with a separate workflow...

dentarg:
@MSP-Greg I tried running the tests locally with Docker (`docker run -it --rm -v $(pwd):/app ruby:2.3 bash`) but they just hang... 

Hmm... it looks like the image do have OpenSSL

```
root@1633c1df42ae:/app# dpkg -l | grep openssl
ii  libcurl4-openssl-dev:amd64         7.52.1-5+deb9u9                amd64        development files and documentation for libcurl (OpenSSL flavour)
ii  libevent-openssl-2.0-5:amd64       2.0.21-stable-3                amd64        Asynchronous event notification library (openssl)
ii  openssl                            1.1.0j-1~deb9u1                amd64        Secure Sockets Layer toolkit - cryptographic utility
```

MSP-Greg:
@dentarg

Pushing my memory a little, but Ruby 2.3 might not compile with OpenSSL 1.1.0, as that's when they changed the file names.

But, the compile step is showing:
```
checking for BIO_read() in -lcrypto... no
checking for BIO_read() in -llibeay32... no
```
and the top of the test section isn't showing any OpenSSL version info?

Ruby 2.3 might need the OpenSSL gem installed also...

dentarg:
Thanks @MSP-Greg 

> and the top of the test section isn't showing any OpenSSL version info?

Yeah, I did wonder why this didn't show... guess I know now

`gem install openssl` fails in the Docker image (also tried Alpine (after `apk add git make gcc libc-dev`)), I can look into it later...

MSP-Greg:
Thanks for investigating.

>Yeah, I did wonder why this didn't show... guess I know now

I think it was my code, probably should add some kind of 'no OpenSSL' message...

MSP-Greg:
@dentarg 

Sorry, just did something similar this weekend:
```
    - name: install ragel and OpenSSL
      run:  |
        apt-get -qy update
        apt-get -qy install ragel libssl1.0-dev
```

dentarg:
I went with RVM [like RubyGems do](https://github.com/rubygems/rubygems/blob/v3.1.2/.github/workflows/ubuntu-rvm.yml) for 2.3.8. WDYT @nateberkopec? 

I see that the Travis setup is using RVM a lot, maybe we can try to move all or some to this setup going forward (but in separate pull requests).

The Travis failure is fixed by https://github.com/puma/puma/pull/2097

dentarg:
> I see that the Travis setup is using RVM a lot, maybe we can try to move all or some to this setup going forward

However, as @MSP-Greg [have already realized](https://github.community/t5/GitHub-Actions/continue-on-error-allow-failure-UI-indication/td-p/37033), GitHub Actions doesn't seem to have any equivalent for the `allow_failures` matrix that Travis has, so we might have to keep Travis around for a while.

MSP-Greg:
@dentarg

Yes.  Travis & AppVeyor most would consider to be mature platforms, Actions still has some things that are really quirky.  The real issue with `allow_failures` in Actions is that it takes a few mouse clicks to determine what happened...

Give me a little bit and I'll have a suggestion or two.  We shouldn't be building with OpenSSL 1.1.1 on Ruby 2.3 (as it's using 1.0.2), and I think we can combine the 2.7/head rvm with the 2.3 rvm.

MSP-Greg:
@dentarg

See:
https://github.com/MSP-Greg/puma/blob/rvm-2-3/.github/workflows/ruby.yml

Tests at:
https://github.com/MSP-Greg/puma/runs/366225386

Note that 2.3.8 uses OpenSSL 1.0.2
https://github.com/MSP-Greg/puma/runs/366225386#step:10:17

Ruby 2.7.0 uses OpenSSL 1.1.1
https://github.com/MSP-Greg/puma/runs/366225393#step:10:17

I noticed you have two PR's, one for 2.3 and one for 2.7.  Feel to grab the changes and place them wherever...

MSP-Greg:
@dentarg

A while back the CI was set up to test on OpenSSL 1.0.1, 1.0.2, & 1.1.1.  I'm not sure if we should bother with testing on 1.0.1, so I removed Ubuntu-16.04, as it uses the same version of OpenSSL as Ubuntu-latest.  It was just there because of Ruby 2.3, since removed...

dentarg:
Now using OpenSSL 1.0.2 for Ruby 2.3.8, thanks @MSP-Greg

MSP-Greg:
@dentarg

You're welcome.  For those who have never dug into OpenSSL, depending on the OS and the Ruby version, Ruby's OpenSSL stdlib/gem may be built with OpenSSL version 1.0.1, 1.0.2, 1.1.0, or 1.1.1.

Generally, Puma (and any extension gems using OpenSSL) should be built with the same version as the Ruby OpenSSL stdlib/gem was built with.

Also, between OpenSSL versions 1.0.2 and 1.1.0, the names of the libraries changed.

PR #2092, opened 2019-12-17
Least-busy load balancing in cluster mode

wjordan:
### Description

Proof-of-concept HTTP request load-balancing using a 'least-busy' algorithm for cluster mode. Instead of having each worker-process run a select-loop concurrently on the listener and race to accept incoming connections, this PR runs a select-loop on the listener in the parent process and forwards individual client TCP connections to worker-processes by sending the file descriptor through a unix domain socket-pair unique to each worker. With this approach, the parent gets to pick which worker gets the connection using a specific routing algorithm.

In order to implement 'least busy' routing for the parent process to select the worker to send each client connection, each worker updates the parent with its current capacity whenever its threads start or finish processing, and the parent selects the worker with the highest capacity. (If two workers report the same capacity, it chooses the one with the oldest last-request.)

This PR was designed to solve #1254 / #2078 as an alternative approach to #1646, #1920 and #2079. The configuration of custom load-balancing algorithms via DSL could support other use-cases as well.

### Performance

Runs the benchmark noted in https://github.com/puma/puma/pull/2079#issuecomment-564809228 with **4013** requests/sec.

[todo - will put the results of some other micro-benchmarks/comparisons here]

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -28,6 +28,8 @@ module Puma
 
       @phased_state = :idle
       @phased_restart = false
+      @not_full = ConditionVariable.new
+      @mutex = Mutex.new
     end
 
     def stop_workers
@@ -63,7 +65,7 @@ module Puma
     end
 
     class Worker
-      def initialize(idx, pid, phase, options)
+      def initialize(idx, pid, phase, socket, options)
         @index = idx
         @pid = pid
         @phase = phase
@@ -75,9 +77,13 @@ module Puma
         @last_checkin = Time.now
         @last_status = '{}'
         @term = false
+        @socket = socket
+        @capacity = 0
+        @last_request = Time.now
       end
 
-      attr_reader :index, :pid, :phase, :signal, :last_checkin, :last_status, :started_at
+      attr_reader :index, :pid, :phase, :signal, :last_checkin, :last_status, :started_at, :last_request
+      attr_accessor :capacity
 
       def booted?
         @stage == :booted
@@ -97,6 +103,13 @@ module Puma
         @last_status = status
       end
 
+      def accept(connection)
+        @capacity -= 1
+        @last_request = Time.now
+        @socket.send_io(connection)
+        connection.close
+      end
+
       def ping_timeout?(which)
         Time.now - @last_checkin > which
       end
@@ -108,6 +121,7 @@ module Puma
           else
             @term ||= true
             @first_term_sent ||= Time.now
+            @socket.close
           end
           Process.kill @signal, @pid
         rescue Errno::ESRCH
@@ -133,6 +147,7 @@ module Puma
 
       diff.times do
         idx = next_worker_index
+        parent_socket, @worker_socket = UNIXSocket.pair
         @launcher.config.run_hooks :before_worker_fork, idx
 
         pid = fork { worker(idx, master) }
@@ -143,7 +158,7 @@ module Puma
         end
 
         debug "Spawned worker: #{pid}"
-        @workers << Worker.new(idx, pid, @phase, @options)
+        @workers << Worker.new(idx, pid, @phase, parent_socket, @options)
 
         @launcher.config.run_hooks :after_worker_fork, idx
       end
@@ -267,6 +282,8 @@ module Puma
       @launcher.config.run_hooks :before_worker_boot, index
 
       server = start_server
+      server.update_capacity = Proc.new {|c| @worker_write << "!c#{Process.pid},#{c}\n" }
+      server.inherit_binder Binder.new(@launcher.events).tap {|b| b.ios << @worker_socket}
 
       Signal.trap "SIGTERM" do
         @worker_write << "e#{Process.pid}\n" rescue nil
@@ -474,6 +491,29 @@ module Puma
 
       @launcher.events.fire_on_booted!
 
+      Thread.new do
+        Puma.set_thread_name "acceptor"
+
+        # Chooses a load-balancing algorithm from a set of predefined symbol constants or a custom Proc.
+        algorithms = {
+          least_busy: lambda {|workers| workers.min_by {|w| [-w.capacity, w.last_request]}},
+          round_robin: lambda {|workers| workers.min_by(&:last_request)},
+          pile_up: lambda {|workers| workers.select {|w| w.capacity > 0}.min_by {|w| w.index} || workers.min_by(&:last_request)},
+          random: lambda {|workers| workers.sample}
+        }.freeze
+        routing = @options[:load_balancing]
+        debug "Using load balancing algorithm: #{routing}"
+        routing = algorithms[routing] unless routing.is_a?(Proc)
+        raise "Invalid value for load_balancing - #{@options[:load_balancing]}" unless routing.is_a?(Proc)
+
+        Socket.accept_loop(*@launcher.binder.ios) do |connection, _|
+          routing.call(@workers.reject(&:term?)).accept(connection)
+          until @workers.any? {|w| w.capacity > 0 }
+            @mutex.synchronize {@not_full.wait(@mutex)}
+          end
+        end
+      end
+
       begin
         force_check = false
 
@@ -512,6 +552,9 @@ module Puma
                   force_check = true
                 when "p"
                   w.ping!(result.sub(/^\d+/,'').chomp)
+                when "c"
+                  w.capacity = result.sub(/^\d+,/,'').chomp.to_i
+                  @not_full.signal
                 end
               else
                 log "! Out-of-sync worker list, no #{pid} worker"
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -736,5 +736,26 @@ module Puma
       end
     end
 
+    # Control the load-balancing algorithm used to route incoming connections
+    # to workers.
+
+    # Built-in symbol values:
+    #
+    # * :least_busy  - [default] Select worker with least number of busy threads,
+    #                  using the oldest last request as a tie-breaker.
+    # * :round_robin - Select worker with the oldest last request, which will
+    #                  distribute requests evenly among all workers.
+    # * :pile_up     - Select a worker with non-zero capacity that has the lowest index.
+    #                  This will cause requests to 'pile up' on workers until
+    #                  they reach capacity.
+    # * :random      - Select a random worker.
+    #
+    # You can also pass a Proc to specify your own custom routing algorithm. The proc is
+    # passed an Array[Worker] argument and expects a Worker return value.
+    #
+    # @note Cluster mode only.
+    def load_balancing(algorithm=:least_busy)
+      @options[:load_balancing] = algorithm
+    end
   end
 end
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -44,6 +44,7 @@ module Puma
     attr_accessor :auto_trim_time
     attr_accessor :reaping_time
     attr_accessor :first_data_timeout
+    attr_writer :update_capacity
 
     # Create a server for the rack app +app+.
     #
@@ -85,6 +86,7 @@ module Puma
       @mode = :http
 
       @precheck_closing = true
+      @update_capacity = nil
     end
 
     attr_accessor :binder, :leak_stack_on_error, :early_hints
@@ -334,6 +336,7 @@ module Puma
       end
 
       @thread_pool.clean_thread_locals = @options[:clean_thread_locals]
+      @thread_pool.update_capacity = @update_capacity
 
       if queue_requests
         @reactor = Reactor.new self, @thread_pool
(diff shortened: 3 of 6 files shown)


lib/puma/cluster.rb
```diff
@@ -267,6 +282,8 @@ def worker(index, master)
       @launcher.config.run_hooks :before_worker_boot, index
 
       server = start_server
+      server.update_capacity = Proc.new {|c| @worker_write << "!c#{Process.pid},#{c}\n" }
```
wjordan: Note there is a bit of synchronization overhead in this operation, which starts to become visible at extremely low-latency response times / high request rates (e.g., ~1ms/req, ~10k req/sec). An alternative is to use shared memory (e.g. the [raindrops](https://bogomips.org/raindrops/) gem), see 486b4f9206767c1e6f08b6697b2d8695246c6629 for a commit implementing this alternative approach. The tradeoff is pulling in an extra dependency (or writing extra C-extension code) for that extra bit of performance.

lib/puma/cluster.rb
```diff
@@ -75,9 +77,13 @@ def initialize(idx, pid, phase, options)
         @last_checkin = Time.now
         @last_status = '{}'
         @term = false
+        @socket = socket
+        @capacity = 0
```
nateberkopec: Is there a reason we can't re-use information from the threadpool, I wonder, rather than setting a separate counter?

test/test_integration_cluster.rb
```diff
@@ -140,7 +140,7 @@ def test_stuck_phased_restart
   def term_closes_listeners(unix: false)
     skip_unless_signal_exist? :TERM
 
-    cli_server "-w #{WORKERS} -t 0:6 -q test/rackup/sleep_step.ru", unix: unix
+    cli_server "-w #{WORKERS} -t 1:6 -q test/rackup/sleep_step.ru", unix: unix
```
nateberkopec: Why was this req'd?

lib/puma/thread_pool.rb
```diff
@@ -76,7 +77,13 @@ def backlog
     end
 
     def pool_capacity
-      waiting + (@max - spawned)
+      waiting + (@max - spawned) - @todo.size
+    end
+
+    attr_writer :update_capacity
+
+    def update_capacity
```
nateberkopec: Not a huge fan of this. Would be better if we could leave the threadpool alone.

lib/puma/thread_pool.rb
```diff
@@ -115,13 +122,15 @@ def spawn_thread
                 break
               end
 
-              @waiting += 1
+              @waiting += 1; update_capacity
```
evanphx: Please move these calls to their own line

lib/puma/thread_pool.rb
```diff
@@ -115,13 +122,15 @@ def spawn_thread
...
               not_full.signal
               not_empty.wait mutex
               @waiting -= 1
             end
 
-            work = todo.shift if continue
+            if continue
+              work = todo.shift; update_capacity
```
evanphx: DItto, own line.

lib/puma/thread_pool.rb
```diff
@@ -157,7 +166,7 @@ def <<(work)
           raise "Unable to add work while shutting down"
         end
 
-        @todo << work
+        @todo << work; update_capacity
```
evanphx: Ditto, own line.

lib/puma/cluster.rb
```diff
@@ -474,6 +491,29 @@ def run
...
+      Thread.new do
+        Puma.set_thread_name "acceptor"
+
+        # Chooses a load-balancing algorithm from a set of predefined symbol constants or a custom Proc.
+        algorithms = {
+          least_busy: lambda {|workers| workers.min_by {|w| [-w.capacity, w.last_request]}},
+          round_robin: lambda {|workers| workers.min_by(&:last_request)},
+          pile_up: lambda {|workers| workers.select {|w| w.capacity > 0}.min_by {|w| w.index} || workers.min_by(&:last_request)},
+          random: lambda {|workers| workers.sample}
+        }.freeze
```
evanphx: This should be moved out of here into a constant rather than hidden here, probably even in it's own file.

lib/puma/cluster.rb
```diff
@@ -474,6 +491,29 @@ def run
...
+        routing = @options[:load_balancing]
+        debug "Using load balancing algorithm: #{routing}"
+        routing = algorithms[routing] unless routing.is_a?(Proc)
+        raise "Invalid value for load_balancing - #{@options[:load_balancing]}" unless routing.is_a?(Proc)
+
+        Socket.accept_loop(*@launcher.binder.ios) do |connection, _|
+          routing.call(@workers.reject(&:term?)).accept(connection)
+          until @workers.any? {|w| w.capacity > 0 }
+            @mutex.synchronize {@not_full.wait(@mutex)}
```
evanphx: Please break the block out to a separate line with `do`

lib/puma/cluster.rb
```diff
@@ -474,6 +491,29 @@ def run
...
+        algorithms = {
+          least_busy: lambda {|workers| workers.min_by {|w| [-w.capacity, w.last_request]}},
+          round_robin: lambda {|workers| workers.min_by(&:last_request)},
+          pile_up: lambda {|workers| workers.select {|w| w.capacity > 0}.min_by {|w| w.index} || workers.min_by(&:last_request)},
+          random: lambda {|workers| workers.sample}
+        }.freeze
+        routing = @options[:load_balancing]
+        debug "Using load balancing algorithm: #{routing}"
+        routing = algorithms[routing] unless routing.is_a?(Proc)
+        raise "Invalid value for load_balancing - #{@options[:load_balancing]}" unless routing.is_a?(Proc)
```
evanphx: We should probably be checking this in the config DSL rather than at this level.

lib/puma/dsl.rb
```diff
@@ -736,5 +736,26 @@ def set_remote_address(val=:socket)
...
+    #                  This will cause requests to 'pile up' on workers until
+    #                  they reach capacity.
+    # * :random      - Select a random worker.
+    #
+    # You can also pass a Proc to specify your own custom routing algorithm. The proc is
+    # passed an Array[Worker] argument and expects a Worker return value.
+    #
+    # @note Cluster mode only.
+    def load_balancing(algorithm=:least_busy)
+      @options[:load_balancing] = algorithm
```
evanphx: Here is where we should validate the argument and throw an error.

lib/puma/server.rb
```diff
@@ -386,7 +389,8 @@ def handle_servers
                 break if handle_check
               else
                 begin
-                  if io = sock.accept_nonblock
+                  receive_socket = sock.is_a?(UNIXSocket) && @update_capacity
+                  if (io = receive_socket ? sock.recv_io(TCPSocket) : sock.accept_nonblock)
```
evanphx: While this will work because sock will be a `UNIXServer` in the case of binding on a unix domain socket as the user requested, it's a bit clunky. I'd rather we plumb through a flag to indicated that `recv_io` should be used rather than relying on it being a UNIXSocket.

evanphx:
I fully realize this is a POC btw and that a lot of the comments I left are more about a delivered, prod ready PR. I think the approach is totally valid and you should go ahead and turn this into a prod ready PR.

ayufan:
I'm not sure if this is OK.

This makes the `master` to be responsible for handing off the connections to workers, and makes the `master` to be a single point of contention in such case. The currently model is equal in behaviour to `Unicorn`, and makes this effectively a distributed solution, where we have multiple processes waiting for connection, instead of single one.

This does add complexity, and might have unintended side-effects, if for any reason anything else works on `master`.

nateberkopec:
Closing because I think we have a prod-ready approach in @ayufan's PR and there are two big drawbacks to this approach:

* Master process becomes an acceptor process
* Probably no JRuby or Windows support.

sayap:
Just want to chime in that this latency issue is preventing us from migrating from passenger (OSS, single-threaded) to puma. As noted by @wjordan in https://github.com/puma/puma/issues/1254#issuecomment-420405204, passenger uses the exact same least-busy load balancing approach as this PR, and that has been working exceptionally well us.

Regarding the drawback about making the master process as a single point of failure, I think the fear is rather unfounded, as the master-workers approach works just fine for passenger.

As for the second drawback with cross-platform support, it looks like passenger does work fine on JRuby, but doesn't support Windows. So the concern is indeed valid.

Anyway, the condvar approach also has its own drawback, e.g. there could be significant performance difference across different ruby versions: https://bugs.ruby-lang.org/issues/16255

Furthermore, the latency issue would be more apparent at P99 or P999. If we look at the graphs at https://gitlab.com/gitlab-com/gl-infra/infrastructure/issues/8334, the only one that covers the full range is "Zooming into just /jwt/auth calls", and it does look like the long tail latency is worse with puma + #2079 in that graph?

ayufan:
> Furthermore, the latency issue would be more apparent at P99 or P999. If we look at the graphs at https://gitlab.com/gitlab-com/gl-infra/infrastructure/issues/8334, the only one that covers the full range is "Zooming into just /jwt/auth calls", and it does look like the long tail latency is worse with puma

From my perspective moving all logic to master changes significantly behaviour of Puma, and makes us to re-learn how to run Puma effectively on a scale. The biggest issue is that something can run on `master` process that would slowdown request acceptance, and this could cause performance regression for some.

After doing extensive tests across board we see significantly better latency, and consistent with Unicorn. We were running a big scale tests with and without patch disabled, and the conclusion was that we did not see any difference between Puma and Unicorn after applying patch, on all P-xs that we looked (50, 95, 99, etc.). 

Anyway, it is still on my agenda for this cycle to re-try both of them. I'm quite certain that I will not be able to run this PR on a scale, as we are this user "that we have something running on master" :) I want to see if there's something that we can improve with `cond.wait` approach, because to be fair I did not focus on optimising that approach yet.

@sayap Did you maybe try my patch how it works for you?

wjordan:
I agree this PR isn't production-ready yet and should be shelved for now given the limitations mentioned. I thought I'd add some detailed notes in case I or anyone else wanted to revisit this approach in the future:

#### 1. Master process becomes an acceptor process
While I agree with @sayap's [comment](https://github.com/puma/puma/pull/2092#issuecomment-591286853) that the concern is mostly theoretical at this point and that other web-servers use a similar request-balancing approach of using a dedicated process to pass accepted sockets to workers (see e.g., [`kfcgi`](https://kristaps.bsd.lv/kcgi/extending01.html#fig4) and [`treasure-data/serverengine`](https://github.com/treasure-data/serverengine) in addition to passenger), what's unique here is that Puma runs other code in the same master process (including user-defined hooks), and the Ruby's global VM lock could cause one thread in the same process to block another. Forking a dedicated acceptor _process_ (as opposed to an acceptor thread) could address this concern, and probably wouldn't be too difficult.

#### 2. Probably no JRuby or Windows support
- JRuby shouldn't have any issues, according to its [test coverage](https://github.com/jruby/jruby/blob/54b5c3b626b85941cd5d66fb59e1c6091d0d8b5e/spec/ruby/library/socket/unixsocket/send_io_spec.rb#L26) it supports passing file descriptors via `send_io`.
- Windows support is possible but would take extra work. Although Windows doesn't support passing descriptors over a socket directly, it provides a [`WSADuplicateSocket` API](https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-wsaduplicatesocketa) which creates a new descriptor for a shared socket that can then be passed through a pipe to the child process. Python wraps this up nicely via [`socket.share()`](https://docs.python.org/3/library/socket.html#socket.socket.share) / [`socket.fromshare()`](https://docs.python.org/3/library/socket.html#socket.fromshare), but Ruby doesn't have any equivalent, so either a C extension or ffi/fiddle magic is required. [`treasure-data/serverengine`](https://github.com/treasure-data/serverengine) includes a [Windows-socket wrapper using fiddle](https://github.com/treasure-data/serverengine/blob/master/lib/serverengine/winsock.rb) to implement this approach which could probably be adapted.

I would add a third limitation, the synchronization overhead of passing busy counters through the existing pipe channel after each request, which I mentioned above in https://github.com/puma/puma/pull/2092#discussion_r358989306. This should be replaced with shared-memory counters for best performance but that also adds further complexity to the implementation.

For high-performance use-cases, another possibly simpler approach would be to have each worker listen to a separate socket/port, and use a separate load-balancing proxy such as nginx or haproxy to balance requests across them. This offers best performance, and doesn't add any complexity for setups which already have nginx in their application stack anyway. I'll open a PR demonstrating this approach.

ayufan:
> For high-performance use-cases, another possibly simpler approach would be to have each worker listen to a separate socket/port, and use a separate load-balancing proxy such as nginx or haproxy to balance requests across them. This offers best performance, and doesn't add any complexity for setups which already have nginx in their application stack anyway. I'll open a PR demonstrating this approach.

This seems like an interesting idea @wjordan. If we could allow to have each worker to have their own socket, that would be LB externally.

Issue #2081, opened 2019-12-03
separate tokens for control and status actions


tkishel:
**Is your feature request related to a problem? Please describe.**

Authentication for control and status actions in `Puma::App::Status` is implemented via the same `control-token` option, while the security requirements related to control actions like `halt` and  status actions`stats` differ significantly.

**Describe the solution you'd like**

Separate tokens for control and status actions.

**Describe alternatives you've considered**

An option to set the authorization level `(control | status)` for the `control-token` option.

nateberkopec:
What if, instead, we allowed you to configure a callback lambda/proc?

tkishel:
I'm open to any implementation that allows users to configure querying status independently from controlling the server :) Might you have an example that illustrates what you are thinking about?

tkishel:
How about this ...

config

```
control-url "ssl://0.0.0.0:12345? ..."
control-actions ['stats']
```

status.rb

```
      def initialize(cli, token = nil, actions = [])
        @cli = cli
        @auth_token = token
        @auth_actions = actions
      end

      def authorized(action)
        return true if @auth_actions.empty?
        @auth_actions.include? action
      end

      def not_authorized(action)
        rack_response(401, "Action '#{action}' not authorized", 'text/plain')      
      end
```

```
      def call(env)
        unless authenticate(env)
          return rack_response(403, 'Invalid auth token', 'text/plain')
        end

        case env['PATH_INFO']
        when /\/(stop)$/
          action = $1         
          return not_authorized(action) unless authorized(action)
          @cli.stop
          rack_response(200, OK_STATUS)

        when /\/(stats)$/
          action = $1         
          return not_authorized(action) unless authorized(action)
          rack_response(200, @cli.stats)
```

resulting in

```
# curl https://puma.example.com:12345/stop?token=hello
Action 'stop' not authorized

# curl https://puma.example.com:12345/stats?token=hello
{ "backlog": 0, "running": 0, "pool_capacity": 10, "max_threads": 10 }
```

tkishel:
See https://github.com/puma/puma/pull/2083

bdewater:
> I'm open to any implementation that allows users to configure querying status independently from controlling the server :)

Isn't this possible to implement in your application? 

You might need to `JSON.parse(Puma.stats)` in your own application code to get a hash to for data manipulation/extraction, but that requirement will go away with https://github.com/puma/puma/pull/2086

tkishel:
It is, but ...

[...]

PR #2098, opened 2019-12-31
Exclude status from control-token check (#2081)

tkishel:
Prior to this commit ...

Puma allows access to all (control and status) actions via the same token.

With this commit ...

Puma does not require a token to access status actions.

### Description

### DIFF
--- a/History.md
+++ b/History.md
@@ -5,6 +5,7 @@
   * Configuration: `environment` is read from `RAILS_ENV`, if `RACK_ENV` can't be found (#2022)
   * `Puma.stats` now returns a Hash instead of a JSON string (#2086)
   * `GC.compact` is called before fork if available (#2093)
+  * Exclude status from control-token check (#2081)
 
 * Bugfixes
   * Your bugfix goes here (#Github Number)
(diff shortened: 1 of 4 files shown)


tkishel:
As per this comment in the Issue:

> No, it just means we're over-securing status.

nateberkopec:
Sorry, I should have been more clear. Over-securing status != status should be open to the world. I was more trying to agree with your assertion that there are two separate authorization levels here that we need to distinguish.

I am still in favor of my original proposal of keeping the original behavior and then allowing a configurable proc/lambda. That lets people implement whatever scheme they want.

TBH when it comes to security behavior, everyone has weird opinions and you have to let people do what they want. One-size-fits-all authorization schemes are guaranteed to piss someone off.

tkishel:
I'm happy to contribute towards this, but reluctant to second-guess an implementation.

Could you provide some psuedo-code, or more specifics?

nateberkopec:
In your config, you should be able to provide a lambda. The lambda receives some arguments (I guess: the action name, the provided token, the configured 'correct' token. Maybe we can provide more too?). Returning false means that you are not authorized, and we return a 403. Returning true means you are authorized and we continue.

nateberkopec:
BTW thanks for all of your contributions on this. You've done a lot of work here!

tkishel:
I appreciate you saying that. I'll see what I can do ...

PR #2103, opened 2020-01-09
Update Actions ruby.yml [changelog skip]

MSP-Greg:
Current Rubies, 2.4 thru 2.7 on all platforms, plus 2.3 on Ubuntu & macOS

Ubuntu 2.3, macOS 2.3, & Windows 2.4 use OpenSSL 1.0.2, all others are 1.1.1

JRuby fails tests

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -1,30 +1,31 @@
 name: Puma
 
-on:
-  push:
-    branches:
-      - '*'
-  pull_request:
-    branches:
-      - '*'
+on: [push, pull_request]
 
 jobs:
   build:
     name: >-
-      OS: ${{ matrix.os }}  Ruby: ${{ matrix.ruby }}
+      ${{ matrix.os }}  Ruby: ${{ matrix.ruby }}
+    env:
+      CI: true
+      TESTOPTS: -v
+
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
-        os: [ 'ubuntu-18.04', 'macos', 'windows-latest' ]
-        ruby: [ '2.4.x', '2.5.x', '2.6.x' ]
+        os: [ ubuntu-18.04, macos, windows-latest ]
+        ruby: [ 2.3, 2.4, 2.5, 2.6, 2.7 ]
+        exclude:
+          - os: windows-latest
+            ruby: 2.3
+
     steps:
     - name: repo checkout
-      uses: actions/checkout@v1
-      with:
-        fetch-depth: 10
+      uses: actions/checkout@v2
+
     - name: load ruby
-      uses: actions/setup-ruby@v1
+      uses: eregon/use-ruby-action@v1
       with:
         ruby-version: ${{ matrix.ruby }}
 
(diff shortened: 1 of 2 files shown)


MSP-Greg:
Interesting.  Passed in my fork a few times.  The PR CI passed all except ubuntu-18.04 Ruby: 2.4.

.github/workflows/ruby.yml
```diff
@@ -1,30 +1,31 @@
...
     steps:
     - name: repo checkout
-      uses: actions/checkout@v1
-      with:
-        fetch-depth: 10
+      uses: actions/checkout@v2
+
     - name: load ruby
-      uses: actions/setup-ruby@v1
+      uses: eregon/use-ruby-action@master
```
dentarg: I'll contribute some background links on this change for anyone curious to know more

* The repo for this is action is https://github.com/eregon/use-ruby-action
* It is using https://github.com/eregon/ruby-install-builder
* And it might become next version of the official action, see more in https://github.com/actions/setup-ruby/issues/44

nateberkopec:
Nice. Let's remove the few overlapping builds from Travis as well.

eregon:
@MSP-Greg Thanks for this PR, I was also thinking it would be a nice simplification for repos currently using setup-ruby + RVM + windows separately.

eregon:
The test failure looks like a transient to me (i.e., I would think unrelated to the action).

olleolleolle:
> Nice. Let's remove the few overlapping builds from Travis as well.

From a cursory look, that seems to be only:

- `rvm: 2.7.0`

MSP-Greg:
Thanks for all the feedback.

@eregon

> The test failure looks like a transient to me (i.e., I would think unrelated to the action).

Yes, unrelated to the action.  Over the weekend I'd like to check it using the seed value.  If it is repeatable, is it Puma, or the test suite?  OS or Ruby version specific?  Then, how hard to track down & fix?

@nateberkopec

> Nice.

Thanks to eregon for the action.

> Let's remove the few overlapping builds from Travis as well.

OpenSSL 1.0.2 is EOL.  Personally, and given that this happened recently, I don't have issues with using it in CI.  But, 1.0.1 is just too old.  So, should the Travis test using 1.0.1 be removed?  Or, since it is passing, leave it?

Re OpenSSL 1.0.2, it is tested here on all three platforms. Hence, they can be removed also?

The workflow in this PR can have 'release' JRuby & TruffleRuby added, but Actions currently is not 'allow failure' friendly.  One needs to open the job then open its steps to see failure.

So, for now, I'd leave 'allow failure' jobs in Travis.

.github/workflows/ruby.yml
eregon: @MSP-Greg Could you update to `uses: eregon/use-ruby-action@v1` here?

eregon:
```suggestion
      uses: eregon/use-ruby-action@v1
```

MSP-Greg: Done

eregon:
@MSP-Greg Could you remove the redundant `rvm: 2.7.0` job from TravisCI?
Others jobs don't seem affected by this PR, so I'd leave them as a separate cleanup.

@nateberkopec Could you merge this?

MSP-Greg:
@eregon

Thanks for the suggestions.

Removed three jobs from Travis, two are not **exact** matches with an Action job, but some were added to test various OpenSSL/OS combinations, and they're being tested in Actions.

I have now had enough morning coffee...

eregon:
Thanks for merging!
I made a follow-up PR to follow the transfer of that action to the ruby organization: https://github.com/puma/puma/pull/2110

PR #2110, opened 2020-02-07
eregon/use-ruby-action moved to ruby/setup-ruby

eregon:
* See https://github.com/ruby/setup-ruby/releases/tag/v1.13.0

Follow-up of https://github.com/puma/puma/pull/2103


nateberkopec:
Thanks, @dentarg got there first!

PR #2111, opened 2020-02-07
Simplify Runner#start_control URL parsing

wjordan:
### Description
Simplifies some code in `Runner#start_control` by reusing similar logic from `Binder#parse`.

~Besides simplifying code, this fixes a bug/issue where a Unix socket created for a control server (e.g., `--control-url unix://control.sock`) was not being properly removed on server shutdown.~
~This is because `Binder#parse` additionally adds its urls to `@listeners`, which allows them to be automatically cleaned up by `Binder#close_listeners` when the server shuts down.~

(Properly fixing the control-socket cleanup issue requires additional code changes not in this PR, but this simplification makes the fix easier to apply.)

### Your checklist for this pull request
- [x] I have reviewed the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.
- [x] I have added an entry to [History.md](../blob/master/History.md) if this PR fixes a bug or adds a feature. If it doesn't need an entry to HISTORY.md, I have added `[changelog skip]` the pull request title.
- [x] I have added appropriate tests if this PR fixes a bug or adds a feature.
- [x] My pull request is 100 lines added/removed or less so that it can be easily reviewed.
- [x] If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.
- [x] If this closes any issues, I have added "Closes `#issue`" to the PR description or my commit messages.
- [ ] I have updated the documentation accordingly.
- [x] All new and existing tests passed, including Rubocop.

PR #2112, opened 2020-02-08
Refactor listener closing code

wjordan:
### Description
This PR fixes a bug/issue where a Unix socket created for a control server (e.g., `--control-url unix://control.sock`) was not being properly removed on server shutdown.

This fix has two parts:
1. `Binder#parse` adds its urls to `@listeners` allowing them to be automatically cleaned up by `Binder#close_listeners` when the server shuts down, and the code path setting up the control-server listeners was not adding its urls to this array. #2111 (which this PR extends) fixes this part by simplifying the control-server code to call `Binder#parse` instead.
2. The control server was only being stopped on shutdown, but the associated listeners were never closed. This PR adds a call to `close_binder_listeners` to always close the control listeners as well.

I've modified the pumactl integration tests to ensure the unix paths are cleaned up after getting shutdown by `halt` and `stop` commands. (Both tests fail without the fix in this PR.)

Finally, for related cleanup the call to `@binder.close_unix_paths` on launcher shutdown was no longer needed (because closing listeners is handled more consistently by calls to `close_binder_listeners`). I've removed this method to simplify the interface.


lib/puma/server.rb
```diff
@@ -217,6 +217,7 @@ def run_lopez_mode(background=true)
     def handle_servers_lopez_mode
       begin
         check = @check
+        handle_check
```
wjordan: These calls to `handle_check` (and the change to make `handle_check` non-blocking) works around a bug triggered by the `TestIntegrationSingle#test_term_suppress` test, where the TERM signal is sent to the process before the server is fully booted up, causing it to enter an endless loop because the calls to `IO.select` raise an error (because the IO is already closed), so the 'stop server' command is never received.

nateberkopec:
Confirmed test fails on master 👍

MSP-Greg:
> Confirmed test fails on master

Actually, Actions can't download the Ruby binary, although all of the Ruby binaries are downloaded on install..

Both the Ubuntu & macOS files come from the same place, https://github.com/ruby/ruby-dev-builder/releases.  Why it can retrieve the macOS fine, but not Ubuntu is odd.  The code used is standard GitHub code, it may not have any retry code, I'm not sure...

nateberkopec:
@MSP-Greg sorry, I meant I checkout that test locally against the master branch, and it fails on master branch

nateberkopec:
This may have broken the 2.2 build though, ugh.

nateberkopec:
2.2 on Travis will be the death of me, I'm sure.

MSP-Greg:
> 2.2 on Travis will be the death of me, I'm sure.

Don't know if I mentioned it, but 2.2 was left out of Actions CI because...

See https://bugs.ruby-lang.org/issues/13239

It was not backported to 2.2.  Maybe it's time?

MSP-Greg:
JFYI, I added two 'if <io>.closed?' statements, and Actions Ruby 2.2 was failing on:
```
TestLauncher#test_dependencies_and_files_to_require_after_prune_is_correctly_built_with_extra_deps [/home/runner/work/puma/puma/test/test_launcher.rb:30]:
Expected /puma\/lib$/ to match # encoding: US-ASCII
#    valid: true
"/home/runner/work/puma/extensions/x86_64-linux/2.2.0-static/puma-4.3.3".

  2) Failure:
TestLauncher#test_dependencies_and_files_to_require_after_prune_is_correctly_built_for_no_extra_deps [/home/runner/work/puma/puma/test/test_launcher.rb:14]:
Expected /puma\/lib$/ to match # encoding: US-ASCII
#    valid: true
"/home/runner/work/puma/extensions/x86_64-linux/2.2.0-static/puma-4.3.3".
```

PR #2114, opened 2020-02-12
Split ruby.yml, revise Actions, all OS's test from 2.3 thru master/head [changelog skip]

MSP-Greg:
### Description

Revisions to Actions workflows.

1. All OS's test on Ruby 2.3 thru master/head.  Previously, many master/head builds were not fully tested.  As currently set up, all are fully tested.

2. Bundler is pre-installed on all builds, so it is no longer installed/updated.  Because of that, 'frozen string' does not work in older Rubies.  Added conditional steps to only add 'frozen string' where it will pass.

3. There are several places in the GitHub UI where lists of jobs appear.  I reformatted the string shown to hopefully make it not truncated in any of those places...

4. All OS's (including Windows) can have Ruby 2.2 added, but it is currently failing.  Not sure if it's code or tests.  Another PR can fix that.

5. Puma does compile and test successfully with a Windows mswin build, but changes are needed for it, as jaro_winkler won't compile on it, and it's a dependency for RuboCop.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -5,7 +5,7 @@ on: [push, pull_request]
 jobs:
   build:
     name: >-
-      ${{ matrix.os }}  Ruby: ${{ matrix.ruby }}
+      ${{ matrix.os }}, ${{ matrix.ruby }}
     env:
       CI: true
       TESTOPTS: -v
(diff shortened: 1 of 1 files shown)


dentarg:
> 3\. There are several places in the GitHub UI where lists of jobs appear.  I reformatted the string shown to hopefully make it not truncated in any of those places...

👍 ⭐️

nateberkopec:
giphy

.github/workflows/ruby.yml
```diff
@@ -14,45 +14,93 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-18.04, macos, windows-latest ]
-        ruby: [ 2.3, 2.4, 2.5, 2.6, 2.7 ]
-        exclude:
-          - os: windows-latest
-            ruby: 2.3
+        os: [ ubuntu-18.04, macos ]
+        ruby: [ 2.3, 2.4, 2.5, 2.6, 2.7, ruby-head ]
```
nateberkopec: So, does this add ruby-head as a must-pass? Is that a good idea?

MSP-Greg: > Is that a good idea?

What's Don think?  You don't really want to get me going on Ruby master builds, but...

Without pointing fingers, some of the previous builds used in CI were not fully tested, especially with the main MRI test suite 'test-all'.  So, a daily cron job would run and create a build based on a commit that didn't pass ruby/ruby CI.

All the builds used with current actions are fully tested.  If a build doesn't pass CI, the last passing build will be kept in place for CI use.

Hence, I think it's a risk worth taking.  If it breaks, ping me.  JFYI, that's what I've been working on for while, Windows builds (converting to Actions, adding mswin, etc)  and custom actions scripts.



> does this add ruby-head as a must-pass?

Due to the Actions UI, etc, there isn't a worthwhile way to do 'allow failure'.

nateberkopec: Oh, interesting. So it's not _really_ ruby-head, it's actually "the last ruby trunk build that passed it's own CI?"

MSP-Greg: Correct.  If you can make up a good abbreviation for that, be happy to use it.

Since ruby/ruby switched to git, trunk is no more, although it might be an alias just because...

nateberkopec: Cool. I’m still not sure I want to make uncaught bugs on Ruby head crash our build and make it fail until they’re fixed...

MSP-Greg:
JFYI, the following are the Ruby 2.2 failures on macOS & Ubuntu.  On Windows, there is code in the extconf.rb file 9`append_cflags`) that won't work with mkmf.rb from 2.2.

```
  2) Failure:
TestLauncher#test_dependencies_and_files_to_require_after_prune_is_correctly_built_with_extra_deps [/home/runner/work/puma/puma/test/test_launcher.rb:30]:
Expected /puma\/lib$/ to match # encoding: US-ASCII
#    valid: true
"/home/runner/work/puma/extensions/x86_64-linux/2.2.0-static/puma-4.3.1".

  3) Failure:
TestLauncher#test_dependencies_and_files_to_require_after_prune_is_correctly_built_for_no_extra_deps [/home/runner/work/puma/puma/test/test_launcher.rb:14]:
Expected /puma\/lib$/ to match # encoding: US-ASCII
#    valid: true
"/home/runner/work/puma/extensions/x86_64-linux/2.2.0-static/puma-4.3.1".
```

.github/workflows/ruby.yml
MSP-Greg: How about we call it 'master'?

> I want to make uncaught bugs on Ruby head crash

I think that's unlikely.  It would certainly be a good idea to test master sometime in Q4.  But if you're going to do that, might as well go year round.

I've been building/using master for 3+ years, and 'uncaught bugs' are very rare.  Puma is an important gem, and if it exposes a bug in master, isn't it better that they be identified before a Ruby release in December?

In the unlikely event that it does cause CI failures here, it's only two line of code to change...

nateberkopec: We'll try it, but I reserve the right to rollback 😆

MSP-Greg: Please ping me first on any failures.  Thanks.

Issue #2115, opened 2020-02-13
Puma errors and performs badly when receiving a non-HTTPS request while TLSv1.3-configured 


nakwa:
**The bug: Timeout / Puma throwing an error when querying Rails through a reverse proxy.**

This is similar to https://github.com/puma/puma/issues/1708 but I thought it was worth summarizing/detailing again, in a separate issue (also because the context/usage is different).

``` 
~$ bundle exec rails s -b 'ssl://localhost:3000?key=$SSL_KEY&cert=$SSL_CERT'
=> Booting Puma
=> Rails 6.0.2.1 application starting in development
=> Run `rails server --help` for more startup options
W, [2020-02-13T21:27:50.354171 #6604]  WARN -- : Meta::Organization: overriding method 'invalid?'!
Puma starting in single mode...
* Version 4.3.1 (ruby 2.6.5-p114), codename: Mysterious Traveller
* Min threads: 0, max threads: 16
* Environment: development
* Listening on ssl://localhost:3000?key=/Users/[REDACTED]/app/private/ssl-dev/dev-key.key&cert=/Users/[REDACTED]/app/private/ssl-dev/dev-certificate.crt
Use Ctrl-C to stop
Error in reactor loop escaped: System error: Undefined error: 0 - 0 (Puma::MiniSSL::SSLError)
/Users/[REDACTED]/.rvm/gems/ruby-2.6.5/gems/puma-4.3.1/lib/puma/minissl.rb:43:in `read'
/Users/[REDACTED]/.rvm/gems/ruby-2.6.5/gems/puma-4.3.1/lib/puma/minissl.rb:43:in `engine_read_all'
/Users/[REDACTED]/.rvm/gems/ruby-2.6.5/gems/puma-4.3.1/lib/puma/minissl.rb:54:in `read_nonblock'
/Users/[REDACTED]/.rvm/gems/ruby-2.6.5/gems/puma-4.3.1/lib/puma/minissl.rb:128:in `read_and_drop'
/Users/[REDACTED]/.rvm/gems/ruby-2.6.5/gems/puma-4.3.1/lib/puma/minissl.rb:145:in `close'
/Users/[REDACTED]/.rvm/gems/ruby-2.6.5/gems/puma-4.3.1/lib/puma/client.rb:145:in `close'
/Users/[REDACTED]/.rvm/gems/ruby-2.6.5/gems/puma-4.3.1/lib/puma/reactor.rb:266:in `rescue in block in run_internal'
[...]

Issue #1708, opened 2019-01-15
Error in reactor loop escaped: System error: Undefined error: 0 - 0 (Puma::MiniSSL::SSLError)


coneybeare:
I am setting up capybara to use SSL through puma. It all works in terms of connectivity, but this gets `puts`ed out to `STDOUT` and I can find no way to fix it or silence it. I'm ok with either as a solution.

### Steps to reproduce

1) Setup SSL as such:

````ruby
# https://gist.github.com/tadast/9932075
if Rails.env.test?
  unless File.exist?(Rails.root.join('config', 'ssl', 'localhost.test.key'))
    def generate_root_cert(root_key)
      root_ca = OpenSSL::X509::Certificate.new
      root_ca.version = 2 # cf. RFC 5280 - to make it a "v3" certificate
      root_ca.serial = 0x0
      root_ca.subject = OpenSSL::X509::Name.parse "/C=BE/O=A1/OU=A/CN=localhost.test"
      root_ca.issuer = root_ca.subject # root CA's are "self-signed"
      root_ca.public_key = root_key.public_key
      root_ca.not_before = Time.now
      root_ca.not_after = root_ca.not_before + 2 * 365 * 24 * 60 * 60 # 2 years validity
      root_ca.sign(root_key, OpenSSL::Digest::SHA256.new)
      root_ca
    end

    root_key = OpenSSL::PKey::RSA.new(2048)
    file = File.new( Rails.root.join('config', 'ssl', 'localhost.test.key'), "wb")
    file.write(root_key)
    file.close

    root_cert = generate_root_cert(root_key)

    file = File.new( Rails.root.join('config','ssl', 'localhost.test.crt'), "wb")
    file.write(root_cert)
    file.close
  end
    
  ssl_bind 'localhost.test', 3000, {
    key: Rails.root.join('config','ssl', 'localhost.test.key'),
    cert: Rails.root.join('config','ssl', 'localhost.test.crt'),
    verify_mode: 'none'
  }
end
````

2) Run Rspec feature specs using capybara, headless chrome, and puma

````ruby
# https://robots.thoughtbot.com/headless-feature-specs-with-chrome
Capybara.register_driver :chrome do |app|
  Capybara::Selenium::Driver.new(app, browser: :chrome)
end

Capybara.register_driver :headless_chrome do |app|
  capabilities = Selenium::WebDriver::Remote::Capabilities.chrome(
    chromeOptions: {
      args: %w(headless disable-gpu),
[...]

PR #2116, opened 2020-02-18
Fix rejecting HTTP requests when TLS1.3 is used by server

MSP-Greg:
### Description

Closes https://github.com/puma/puma/issues/2115

### DIFF
--- a/lib/puma/minissl.rb
+++ b/lib/puma/minissl.rb
@@ -5,8 +5,18 @@ begin
 rescue LoadError
 end
 
+# need for Puma::MiniSSL::OPENSSL constants used in `HAS_TLS1_3`
+require 'puma/puma_http11'
+
 module Puma
   module MiniSSL
+
+    # define constant at runtime, as it's easy to determine at built time,
+    # but Puma could (it shouldn't) be loaded with an older OpenSSL version
+    HAS_TLS1_3 = !IS_JRUBY &&
+      (OPENSSL_VERSION[/ \d+\.\d+\.\d+/].split('.').map(&:to_i) <=> [1,1,1]) != -1 &&
+      (OPENSSL_LIBRARY_VERSION[/ \d+\.\d+\.\d+/].split('.').map(&:to_i) <=> [1,1,1]) !=-1
+
     class Socket
       def initialize(socket, engine)
         @socket = socket
@@ -21,6 +31,24 @@ module Puma
       def closed?
         @socket.closed?
       end
+
+      # returns a two element array
+      # first is protocol version (SSL_get_version)
+      # second is 'handshake' state (SSL_state_string)
+      #
+      # used for dropping tcp connections to ssl
+      # see OpenSSL ssl/ssl_stat.c SSL_state_string for info
+      #
+      def ssl_version_state
+        IS_JRUBY ? [nil, nil] : @engine.ssl_vers_st
+      end
+
+      # used to check the handshake status, in particular when a TCP connection
+      # is made with TLSv1.3 as an available protocol
+      def bad_tlsv1_3?
+        HAS_TLS1_3 && @engine.ssl_vers_st == ['TLSv1.3', 'SSLERR']
+      end
+      private :bad_tlsv1_3?
 
       def readpartial(size)
         while true
@@ -41,6 +69,7 @@ module Puma
 
       def engine_read_all
         output = @engine.read
+        raise SSLError.exception "HTTP connection?" if bad_tlsv1_3?
         while output and additional_output = @engine.read
           output << additional_output
         end
@@ -167,7 +196,7 @@ module Puma
       end
     end
 
-    if defined?(JRUBY_VERSION)
+    if IS_JRUBY
       class SSLError < StandardError
         # Define this for jruby even though it isn't used.
       end
@@ -184,7 +213,7 @@ module Puma
         @no_tlsv1_1 = false
       end
 
-      if defined?(JRUBY_VERSION)
+      if IS_JRUBY
         # jruby-specific Context properties: java uses a keystore and password pair rather than a cert/key pair
         attr_reader :keystore
         attr_accessor :keystore_pass
--- a/History.md
+++ b/History.md
@@ -26,6 +26,7 @@
   * Configuration: `environment` is read from `RAILS_ENV`, if `RACK_ENV` can't be found (#2022)
 
 * Bugfixes
+  * Close client http connections made to an ssl server with TLSv1.3 (#2116)
   * Do not set user_config to quiet by default to allow for file config (#2074)
   * Always close SSL connection in Puma::ControlCLI (#2211)
   * Windows update extconf.rb for use with ssp and varied Ruby/MSYS2 combinations (#2069)
--- a/ext/puma_http11/mini_ssl.c
+++ b/ext/puma_http11/mini_ssl.c
@@ -301,6 +301,7 @@ void raise_error(SSL* ssl, int result) {
   char msg[512];
   const char* err_str;
   int err = errno;
+  int mask = 4095;
   int ssl_err = SSL_get_error(ssl, result);
   int verify_err = (int) SSL_get_verify_result(ssl);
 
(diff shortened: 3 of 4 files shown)


lib/puma/minissl.rb
```diff
@@ -22,6 +33,46 @@ def closed?
...
 
+      # returns a two element array
+      # first is protocol version (SSL_get_version)
+      # second is 'handshake' state (SSL_state_string)
+      #
+      # used for dropping tcp connections to ssl
+      # see OpenSSL ssl/ssl_stat.c SSL_state_string for info
+      #
+      def ssl_version_state
+        ::Puma::IS_JRUBY ? ['', ''] : @engine.ssl_vers_st.map(&:rstrip)
```
nateberkopec: This is because you didn't implement a JRuby version, right? Not because the JRuby SSL version should actually return different values?

nateberkopec: Or our JRuby extension doesn't support TLS 1.3?

nateberkopec:
Wondering what your response is to https://github.com/puma/puma/issues/2115#issuecomment-591266614****

lib/puma/minissl.rb
MSP-Greg: I wrote the code quite a while ago.  At the time, JRuby didn't support TLSv1.3 yet, and and I opened an issue somewhere asking about it, didn't really get a response.

nateberkopec: Ah, got it.

MSP-Greg:
> Wondering what your response is to

Part of the issue is slow connections, so the code has to wait for some interval to determine if a handshake has completed.  The test seems to show that it's rejected, whereas currently it is not.  Also, I don't recall what 'state' is as a protocol is negotiated down from TLSv1.3 to TLSv1.2.

What should the code do with the connection as far a response or logging?  It's kind like dictionary attacks on mail servers.  What's the best way to respond?   Let it disappear, respond with unknown whatever, etc...

lib/puma/minissl.rb
dentarg: @MSP-Greg is this the issue? https://github.com/jruby/jruby-openssl/issues/184

MSP-Greg: @dentarg Yes

MSP-Greg:
As to why I made this a draft PR...

This is one potential fix for what could be an attack vector, in that http requests to a Puma https server do need to be 'cleaned out' in a timely fashion.  But:

1. Does it affect normal throughput?

2. How should these https requests be handled?  Should the handling change depending on whether Puma is the 'front end' server or behind something like NGINX?  If so, how to allow configuration of the handling?

lib/puma/minissl.rb
```diff
@@ -40,7 +91,18 @@ def readpartial(size)
       end
 
       def engine_read_all
-        output = @engine.read
+        begin
+          output = @engine.read
+          return output if bad_tlsv1_3?
+        rescue Puma::MiniSSL::SSLError => e
+          if e.message.end_with? 'System error: Undefined error: 0 - 0'
```
dentarg, line 98: This doesn't look robust to me, matching against the message... do we know where does this error come from? What does `Undefined error: 0 - 0` mean?

At https://github.com/puma/puma/issues/1708#issuecomment-552079685 I saw the similar but different error

    Error in reactor loop escaped: System error: Success - 0 (Puma::MiniSSL::SSLError)

MSP-Greg: @dentarg

> matching against the message

I wasn't happy with using a string match either.  Let me get back into this and see what can be done.

> What does Undefined error: 0 - 0 mean?

Similar to what happened with 1.1.1e.  The error previously had a numeric code of 0, then with 1.1.1e they changed the code, which was a breaking change.  I'll look at the code for `Puma::MiniSSL::SSLError` and see if better info can be passed from the c code.

This issue is messy, as flooding an https server with http connections could be used as an DOS attack.

We want Puma to reject http connections, and we want that done as quickly as possible so a minimal amount of resources are used by them.  We also don't want to slow down real https connections.

The issue is made worse by the fact that TLSv1.3 and TLSv1.2 handle http connections differently, and right now, I can't recall that exact difference.  As I recall, TLSv1.2 handled the problem itself, TLSv1.3 forces Puma to deal with it...

nateberkopec:
> How should these https requests be handled? Should the handling change depending on whether Puma is the 'front end' server or behind something like NGINX? If so, how to allow configuration of the handling?

@MSP-Greg Any further thoughts on this front?

MSP-Greg:
@nateberkopec

> Any further thoughts on this front?

Maybe.  Question - OpenSSL 1.0.2 is EOL.  Are you ok with adding methods that will only be available with OpenSSL 1.1.0 or later?  I don't think many people are using 1.1.0, so it's really about 1.1.1.  For instance, the context min/max protocol methods in newer Ruby versions maps directly to new OpenSSL methods, and translating them back into 1.0.2 is a PITA.

I think I mentioned it somewhere else, let me look at the SSL error issue and see what can be done.  I've got time this weekend...

nateberkopec:
This is related to the Ruby version issue, no? What's the compat between Ruby versions and OpenSSL versions look like?

MSP-Greg:
> What's the compat between Ruby versions and OpenSSL versions look like?

It's a little grey.

Generally, Ruby 2.2 and 2.3 are considered incompatible with OpenSSL 1.1.0.  The 2.2 and 2.3 builds available on Actions for **all OS's** use 1.0.2.

Re Ruby 2.4, the Actions Ubuntu & macOS builds are done using 1.1.1, but interestingly, you cannot force a TLSv1.3 connection without an error.  Whether they'll connect when left alone, I haven't determined.

But, the current Ruby OpenSSL gem (2.1.2) is compatible with Ruby 2.3.  So, a crazy person could probably install it using OpenSSL 1.1.1 on Ruby 2.3.  I haven't tried that.

Since I wanted to double check some things, I finally created some notes on release dates, etc.
```
       OpenSSL
Ruby   stdlib               OpenSSL
                            1.1.0 pre 2015-Dec-10
2.3     1.1.1  2015-Dec
                            1.1.0     2016-Aug-25
2.4     2.0.9  2016-Dec
2.5     2.1.2  2017-Dec
                            1.1.1 pre 2018-Feb-13  * includes TLSv1.3
                            1.1.1     2018-Sep-11
2.6     2.1.2  2018-Dec
2.7     2.1.2  2019-Dec
2.8     2.2.0
```

nateberkopec:
> Are you ok with adding methods that will only be available with OpenSSL 1.1.0 or later?

So then IMO the answer is "only if you can still run Puma with SSL on on ruby 2.3+". Adding things in a backwards-compatible way or gating new features behind OpenSSL 1.1.0 is IMO fine.

MSP-Greg:
> Adding things in a backwards-compatible way or gating new features behind OpenSSL 1.1.0 is IMO fine.

Just to be clear, if one writes code making use of the new features (using a Ruby compiled with OpenSSL 1.1.1), then one tries to run the same code with an older version of OpenSSL, it's ok that it won't run, since the methods won't be defined?

BTW, I don't recall how much this affects the changes I'd like to make to the SSL error handling.

nateberkopec:
> it's ok that it won't run

Maybe even stronger, perhaps, we should blow up and exception in that situation. If you say something in your config explicitly that uses Open SSL 1.1+, and then in production you only have 1.0 available at runtime, we should probably kick and scream about that.

MSP-Greg:
I got a little side tracked with MSYS2 jumping from gcc 9.2.0 to 10.1.0 this weekend...

On the one hand, I agree.  On the other hand, if someone is touching a production server and they don't understand the importance of having similar OpenSSL versions for development and production, maybe they need to learn a lesson?

Anyway, we could set it up so the methods that require 1.1.1 are undefined if compiled with 1.0?

nateberkopec:
> Anyway, we could set it up so the methods that require 1.1.1 are undefined if compiled with 1.0?

Sure, let's give it a shot.

MSP-Greg:
1. Yanked out some code that wasn't needed.
2. macOS JRuby failure is a stable test issue?
3. Haven't rerun PR #2117, but it failed locally without the lib/ext code changes.

nateberkopec:
JRuby definitely a stability issue, made a new issue about it

PR #2117, opened 2020-02-18
Run test_http_rejection without code updates [changelog skip]

MSP-Greg:
This PR is only the test component of PR https://github.com/puma/puma/pull/2116

It should fail on any CI jobs using OpenSSL 1.1.1, which is Ruby 2.4 and later on all platforms, with one exception of Ruby 2.4 on Windows.

Sorry, meant to make this a draft.  **It should not be merged.**

### DIFF
--- a/test/test_puma_server_ssl.rb
+++ b/test/test_puma_server_ssl.rb
@@ -199,6 +199,44 @@ class TestPumaServerSSL < Minitest::Test
       assert_match(msg, @events.error.message) if @events.error
     end
   end
+
+  def test_http_rejection
+    body_http  = nil
+    body_https = nil
+
+    start_server
+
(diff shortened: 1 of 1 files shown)


nateberkopec:
It passed on 2.2 on Travis.

MSP-Greg:
> It passed on 2.2 on Travis.

Sorry, wasn't clear.

It should pass on any Ruby build using OpenSSL 1.0.2, and Ruby 2.2 uses 1.0.2.  It will not pass on any build using 1.1.1, which is the first OpenSSL version with TLSv1.3.

MSP-Greg:
@nateberkopec

I'm ok to close this, I thought showing that the test failed on current master was worth having record of...

Issue #2018, opened 2019-10-08
phased-restart: Could not find nio4r-2.5.1 in any of the sources (Bundler::GemNotFound)


gingerlime:
**Describe the bug**

This seems to happen when doing phased-restarts using versioned deploy folders with symlinks.

We use a folder structure similar to this:

```
/app/app.d/{version (int)}
/app/app -> /app/app.d/{version} # symlink
/app/prev -> /app/app.d/{version-1} # symlink
```

When we deploy, we increment the version, run bundle etc, then switch the symlink and do a phased-restart of puma.

All seems to work fine, but after a few deploys, we also start pruning older versions, and then we see this error in the puma.log

```
[12357] - Starting phased worker restart, phase: 4
[12357] - Starting phased worker restart, phase: 4
[12357] + Changing to /var/local/app/app
[12357] + Changing to /var/local/app/app
[12357] - Stopping 28107 for phased upgrade...
[12357] - Stopping 28107 for phased upgrade...
[12357] - TERM sent to 28107...
[12357] - TERM sent to 28107...
[7886] + Gemfile in context: /var/local/app/app.d/909/Gemfile
[7886] + Gemfile in context: /var/local/app/app.d/909/Gemfile
[7886] Early termination of worker
[7886] Early termination of worker
[7944] + Gemfile in context: /var/local/app/app.d/909/Gemfile
[7944] + Gemfile in context: /var/local/app/app.d/909/Gemfile
/usr/lib/ruby/site_ruby/2.5.0/bundler/spec_set.rb:91:in `block in materialize': Could not find nio4r-2.5.1 in any of the sources (Bundler::GemNotFound/usr/lib/ruby/site_ruby/2.5.0
/bundler/spec_set.rb:91:in `block in materialize': Could not find nio4r-2.5.1 in any of the sources (Bundler::GemNotFound)
)
        from /usr/lib/ruby/site_ruby/2.5.0/bundler/spec_set.rb:85:in `map!'
        from /usr/lib/ruby/site_ruby/2.5.0/bundler/spec_set.rb:85:in `materialize'
        from /usr/lib/ruby/site_ruby/2.5.0/bundler/definition.rb:170:in `specs'
        from /usr/lib/ruby/site_ruby/2.5.0/bundler/definition.rb:237:in `specs_for'
        from /usr/lib/ruby/site_ruby/2.5.0/bundler/definition.rb:226:in `requested_specs'
[...]

Issue #653, opened 2015-02-10
Puma 2.11 does not start with bundler 1.8 & prune_bundler


dvrensk:
I created a new server today, and RVM installed bundler 1.8 for me.  This is what happens when I try to start puma:

```
~/app/current$ bundle exec puma 
* Pruning Bundler environment
/usr/local/rvm/rubies/ruby-2.1.1/lib/ruby/2.1.0/rubygems/dependency.rb:298:in `to_specs': Could not find 'rack' (= 1.5.2) among 15 total gem(s) (Gem::LoadError)
	from /usr/local/rvm/rubies/ruby-2.1.1/lib/ruby/2.1.0/rubygems/dependency.rb:309:in `to_spec'
	from /usr/local/rvm/rubies/ruby-2.1.1/lib/ruby/2.1.0/rubygems/core_ext/kernel_gem.rb:53:in `gem'
	from /home/simba/app/shared/bundle/ruby/2.1.0/gems/puma-2.11.0/bin/puma-wild:20:in `block in <main>'
	from /home/simba/app/shared/bundle/ruby/2.1.0/gems/puma-2.11.0/bin/puma-wild:18:in `each'
	from /home/simba/app/shared/bundle/ruby/2.1.0/gems/puma-2.11.0/bin/puma-wild:18:in `<main>'
```

This is my `config/puma.rb`:

```ruby
workers 3
threads 2, 6
directory "/home/simba/app/current"
pidfile "/home/simba/app/shared/pids/puma.pid"

rackup      DefaultRackup
bind 'unix:///tmp/.rails3app'
environment 'production'
prune_bundler

# Use this with phased restarts (not preload_app)
on_worker_boot do
  $start_order_queue_processor = true
  $start_queue_processor = true
end
```

I've manually downgraded bundler to 1.7.12 and everything works.

NickLaMuro:
We ran into the same issue on our end.  For what it is worth, removing the `prune_bundler` option seemed to allow us to deploy again.

evanphx:
Sounds like perhaps prune_bundler broke. I'll take a look.

SavithaK:
FWIW, we are seeing the same issue as well.

evanphx:
@indirect would you be willing to chime in here? Did bundler change anything that might break this code: https://github.com/puma/puma/blob/master/lib/puma/cli.rb#L462-L489 ?

evanphx:
[...]

PR #2120, opened 2020-02-20
Replace deprecated `Bundler.with_clean_env` with `with_unbundled_env`

### LINKED ISSUE #2133: BUNDLE_GEMFILE is incorrectly set to parent Gemfile in worker processes (when not setting BUNDLE_GEMFILE explicitly)
**Describe the bug**

Before #1893, if you used `prune_bundler` and started `bundle exec puma` without explicitly setting `BUNDLE_GEMFILE`, workers would not have a value set for the environment variable `BUNDLE_GEMFILE`. This is desirable when you want workers to use the Gemfile located in the application directory, especially if that directory is different from the directory that you started the puma master process in (such as with a phased restart).

After #1893, workers have `BUNDLE_GEMFILE` unexpectedly set to the path of the Gemfile used by the master process. I think this forces `bundler` in the workers to use that Gemfile instead of the one in the worker's working directory as it would if the environment variable was unset.

I think the intent of that PR was that if you start your server with `BUNDLE_GEMFILE=Gemfile.rails6 bundle exec puma`, that workers would have `BUNDLE_GEMFILE` set to the same value. But it ignores the fact that `Bundler` itself sets a value for `ENV['BUNDLE_GEMFILE']` even if it wasn't specified at startup time. Workers now end up inheriting this value, when they didn't before.

One consequence of this bug is that it is no longer possible to delete from your server the Gemfile that was used when you booted the puma master process. If you do, workers fail with the next phased restart. I imagine there are other bugs related to being able to change gems in the Gemfile for new versions of your application, but I haven't been able to reproduce those.

**Puma config:**

```
port 3000
workers 1
prune_bundler
directory '/usr/src/releases/current'

on_worker_boot do
  pp "BUNDLE_GEMFILE: #{ENV['BUNDLE_GEMFILE']}"
end
```

**To Reproduce**

I wrote a script that reproduces the error here: https://github.com/cjlarose/puma-phased-restart-could-not-find-gem-errors/tree/regression-1893 (I reused a repo from an unrelated issue)

Before #1893 , workers boot and print `BUNDLE_GEMFILE: `. After that PR, workers boot and print `BUNDLE_GEMFILE: /usr/src/releases/1582850222334837300/Gemfile`, referring to the path of the Gemfile that was used by the puma master process.

**Expected behavior**

If `BUNDLE_GEMFILE` was unspecified when you started the master process, it should remain unspecified in the worker processes.

 - OS: Linux
 - Puma Version 8c9b3ebc7a86bdc5488704a0225b76409b9f555f

AlexWayfer:
Also add tests for `GEM_HOME` environment variable preserving.

Rework of #1794

Also drop `BUNDLE_GEMFILE` preserving, resolve #2133

Also disable local `path` for `bundle install` in CI, it makes `nio4r` unavailable.

Also install (update) `bundler` in CI, old versions are incompatible.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -40,8 +40,11 @@ jobs:
         run:  |
           gem update --system 2.7.10 --no-document
 
+      - name: Update Bundler
+        run: gem install bundler
+
       - name: bundle install
-        run:  bundle install --jobs 4 --retry 3 --path=.bundle/puma
+        run:  bundle install --jobs 4 --retry 3
       - name: compile
         run:  bundle exec rake compile
 
@@ -87,7 +90,7 @@ jobs:
           ruby-version: ${{ matrix.ruby }}
 
       - name: bundle install
-        run:  bundle install --jobs 4 --retry 3 --path=.bundle/puma
+        run:  bundle install --jobs 4 --retry 3
 
       - name: compile
         if:   matrix.ruby >= '2.4'
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -297,11 +297,9 @@ module Puma
       deps, dirs = dependencies_and_files_to_require_after_prune
 
       log '* Pruning Bundler environment'
-      home = ENV['GEM_HOME']
-      bundle_gemfile = ENV['BUNDLE_GEMFILE']
-      Bundler.with_clean_env do
-        ENV['GEM_HOME'] = home
-        ENV['BUNDLE_GEMFILE'] = bundle_gemfile
+      gem_home = ENV['GEM_HOME']
+      Bundler.with_unbundled_env do
+        ENV['GEM_HOME'] = gem_home
         ENV['PUMA_BUNDLER_PRUNED'] = '1'
         args = [Gem.ruby, puma_wild_location, '-I', dirs.join(':'), deps.join(',')] + @original_argv
         # Ruby 2.0+ defaults to true which breaks socket activation
--- a/test/test_preserve_bundler_env.rb
+++ b/test/test_preserve_bundler_env.rb
@@ -12,9 +12,12 @@ class TestPreserveBundlerEnv < TestIntegration
     skip_unless_signal_exist? :USR2
 
     @tcp_port = UniquePort.call
+    gem_home = "/home/bundle_env_preservation_test"
+    bundle_gemfile = "Gemfile.bundle_env_preservation_test"
     env = {
+      "GEM_HOME" => gem_home,
       # Intentionally set this to something we wish to keep intact on restarts
-      "BUNDLE_GEMFILE" => "Gemfile.bundle_env_preservation_test",
+      "BUNDLE_GEMFILE" => bundle_gemfile,
       # Don't allow our (rake test's) original env to interfere with the child process
       "BUNDLER_ORIG_BUNDLE_GEMFILE" => nil
     }
@@ -27,9 +30,11 @@ class TestPreserveBundlerEnv < TestIntegration
     @pid = @server.pid
     connection = connect
     initial_reply = read_body(connection)
-    assert_match("Gemfile.bundle_env_preservation_test", initial_reply)
+    refute_match(bundle_gemfile, initial_reply)
+    assert_match(gem_home, initial_reply)
     restart_server connection
     new_reply = read_body(connection)
-    assert_match("Gemfile.bundle_env_preservation_test", new_reply)
+    refute_match(bundle_gemfile, new_reply)
+    assert_match(gem_home, initial_reply)
   end
 end
--- a/History.md
+++ b/History.md
@@ -17,6 +17,7 @@
   * Preserve `BUNDLE_GEMFILE` env var when using `prune_bundler` (#1893)
   * Send 408 request timeout even when queue requests is disabled (#2119)
   * Rescue IO::WaitReadable instead of EAGAIN for blocking read (#2121)
+  * Replace deprecated `Bundler.with_clean_env` with `with_unbundled_env` (#2120)
 
 * Refactor
   * Remove unused loader argument from Plugin initializer (#2095)
--- a/test/bundle_preservation_test/config.ru
+++ b/test/bundle_preservation_test/config.ru
@@ -1,1 +1,11 @@
-run lambda { |env| [200, {'Content-Type'=>'text/plain'}, [ENV['BUNDLE_GEMFILE'].inspect]] }
+run(
+  lambda do |env|
+    [
+      200,
+      {'Content-Type'=>'text/plain'},
+      [
+        [ENV['BUNDLE_GEMFILE'], ENV['GEM_HOME']].inspect
+      ]
+    ]
+  end
+)


AlexWayfer:
Strange fail: https://travis-ci.org/puma/puma/jobs/652963711?utm_medium=notification&utm_source=github_status

AlexWayfer:
I don't see new checks, and I don't understand `maintenance` label (it has no description), but I'll mark this PR as "Ready for review".

AlexWayfer:
I don't understand why Ruby 2.2 fails, [the job](https://travis-ci.org/puma/puma/jobs/652970484?utm_medium=notification&utm_source=github_status) doesn't look like [mentioned problem](https://github.com/puma/puma/pull/1893/files#r379535248).

lib/puma/launcher.rb
```diff
@@ -296,11 +296,7 @@ def prune_bundler
       deps, dirs = dependencies_and_files_to_require_after_prune
 
       log '* Pruning Bundler environment'
-      home = ENV['GEM_HOME']
```
nateberkopec: I thought original_env only looked at `BUNDLE_` environment variables.

AlexWayfer: Do we have any test for this? Or can we write it?

nateberkopec: https://github.com/puma/puma/issues/653 looks like no test exists

AlexWayfer: I'll try to implement tests for this, but if somebody can help me — I appreciate it, I'm not too familiar with Puma sources and environments.

nateberkopec:
Questions/concerns:

1) Does the new Bundler method mess with GEM_HOME? May be able to read Bundler source to determine.
2) Can we write a test to catch our own behavior re: GEM_HOME?
3) Can we fix the failure on 2.2 (it is related IMO)?

lib/puma/launcher.rb
AlexWayfer: Also there are no tests for #1893. And I've found `prune_bundler` in `test/` directory only in `test/config/prune_bundler_with_deps.rb`, and I see no usage of this file in `test/`.

I've added `p ENV['GEM_HOME']` inside `Launcher#prune_bundler` and there is no output from tests.

AlexWayfer: Oh, I've found `test/test_preserve_bundler_env.rb`, OK.

nateberkopec: You won't see that output because the Puma tests create new processes whose output is not piped to STDOUT. Take a look at the helpers/integration test code for more info.

The test for #1893 is here: https://github.com/puma/puma/pull/1893/files#diff-1a1a7e313ee72a0ca0e8eb22a2524879

AlexWayfer: > Take a look at the helpers/integration test code for more info.

I've found `:log` option for `wait_for_server_to_boot`, yes, thank you.

Current results:

* `GEM_HOME` was `/home/alex/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0` and becomes `nil`
* `BUNDLE_GEMFILE` was `/home/alex/Projects/ruby/puma/test/bundle_preservation_test/Gemfile.bundle_env_preservation_test` and becomes `Gemfile.bundle_env_preservation_test`.

I'll try to add correct test for `GEM_HOME`. I think the change of `BUNDLE_GEMFILE` is not critical (tests passing).

AlexWayfer:
> Does the new Bundler method mess with GEM_HOME? May be able to read Bundler source to determine.

[`Bundler.with_original_env`](https://github.com/rubygems/bundler/blob/13eefdc8fd166862ded7425c45270dab810003b2/lib/bundler.rb#L369-L371) uses [`Bundler.original_env`](https://github.com/rubygems/bundler/blob/13eefdc8fd166862ded7425c45270dab810003b2/lib/bundler.rb#L329-L331), which uses [`EnvironmentPreserver` with `BUNDLER_KEYS`](https://github.com/rubygems/bundler/blob/13eefdc8fd166862ded7425c45270dab810003b2/lib/bundler.rb#L37-L38), which [includes `BUNDLE_GEMFILE` and `GEM_HOME`](https://github.com/rubygems/bundler/blob/708bb7b6088cff9cb5cfd82f2424b50f8b1d49bd/lib/bundler/environment_preserver.rb#L6-L17).

But tests are somewhy failing without manual reassign at least `GEM_HOME`.

Oh… it's passing without `bundle exec` for tests running. So, the original env is when you calling `bundle exec rake`. What we should to do?

I think, we can re-do something similar to `with_original_env`, but with env before the call, not the first one. And it's possible, but `with_env(our_collected_env)` is private method. I'll try to contribute to Bundler itself to add something like `with_previous_env`. Sorry, but I can't mark this PR as Draft, as I know.

> Can we write a test to catch our own behavior re: GEM_HOME?

Yes, done.

> Can we fix the failure on 2.2 (it is related IMO)?

Let's look. Maybe it's not related, due to #2018

AlexWayfer:
I've removed "WIP" status because I don't think that we'll get `Bundler.with_previous_env` shortly, and now code is correct as I understand.

We can patch or work-around private `Bundler.with_env` method, at least with `send :with_env, env`. As you wish.

AlexWayfer:
OK, I've understood that I'm not completely understand `prune-bundler` option. What kind of `ENV` is expected inside.

AlexWayfer:
`master` process of `puma` should be… pristine by which bundler env? Should it be original or clean at all? With `unbundled` tests are failing ([example of job](https://github.com/puma/puma/pull/2120/checks?check_run_id=467461504)).

Like… when we use `bundle exec puma --prune-bundler` — the master process is always in current (local) bundle context and it's incorrect, conflicting. And `bundle exec rake` (for tests) has similar behavior of this conflict.

AlexWayfer:
Right now we have 100% reproducable case of #2018. :sweat_smile:

nateberkopec:
Is it? I think this is a different exception class: https://github.com/puma/puma/pull/2120/checks?check_run_id=467555536#step:8:78

AlexWayfer:
> Adding a "review" to clear my request.

Sorry, I just don't know what should I do now.

I see problems, but I don't know in which direction I should move. My questions from February 25th are still actual.

nateberkopec:
Yikes, I just started running tests with a newer bundler version, and the output from all deprecation notices is so ugly. This just got bumped up a notch on the priority list!

nateberkopec:
I'm honestly not sure what to do here, but I think we should be crosstalking with the people in #2133.

AlexWayfer:
I don't know what to do (I hate old versions of software): with (unsupported) Ruby 2.2 this PR is excess:

```
ERROR:  Error installing bundler:
	The last version of bundler (>= 0) to support your Ruby & RubyGems was 1.17.3. Try installing it with `gem install bundler -v 1.17.3`
	bundler requires Ruby version >= 2.3.0. The current ruby version is 2.2.0.
```

So… we should use an old version of Bundler for Ruby 2.2, but we should change a code for a new version of Bundler for Ruby 2.7. It seems like two chairs, I suggest to drop Ruby 2.2 support, or at least left this PR open until it'll be happen.

test/test_preserve_bundler_env.rb
```diff
@@ -27,9 +30,11 @@ def test_usr2_restart_preserves_bundler_environment
     @pid = @server.pid
     connection = connect
     initial_reply = read_body(connection)
-    assert_match("Gemfile.bundle_env_preservation_test", initial_reply)
+    refute_match(bundle_gemfile, initial_reply)
```
seven1m: Seems this is a full revert of the work done in #1893, yes? The goal of #1893 was to allow a custom Gemfile name for apps that don't use the default.

AlexWayfer: Seems #2133 is conflicting with #1893, yes? I'm not sure what to do.

seven1m: Here is the situation as I understand it:

1. Prior to #1893, it was not possible to run your app using a non-default Gemfile. Setting `BUNDLE_GEMFILE=Gemfile.rails6` (for example, as a way to test an app using a different version of Rails), would work at first, but as soon as the app was reloaded/restarted, then the environment variable would be reset to default. #1893 fixed that, but also introduced a bug: #2133.
2. #2133 details the bug where child processes inherit the **full path** of the Gemfile of the master process. Instead of the child process seeing the value of `BUNDLE_GEMFILE` as simply "Gemfile" or "Gemfile.rails6", the child process gets the full path to the gemfile, after its path has been expanded. That's bad.

If there is a way to have both the value preservation of #1893, but without the full path sharing detailed in #2133, then we'd like to have both.

Whew! I hope that makes sense.

.github/workflows/ruby.yml
```diff
@@ -40,8 +40,11 @@ jobs:
         run:  |
           gem update --system 2.7.10 --no-document
 
+      - name: Update Bundler
+        run: gem install bundler
```
nateberkopec: Your PR fails on Ruby 2.2 because you're install Bundler 2 on Ruby 2.2. Instead, you need to change the Github action so that you're running `gem install bundler -v 1.17.3` when running Ruby 2.2.

MSP-Greg: The 2.2 Rubies have Bundler 1.7.3 installed already.  See the 2.2 jobs at:
https://github.com/ruby/setup-ruby/runs/491284436

nateberkopec:
Oh I see.

1. Ruby 2.2 does not support Bundler 2. It supports 1.17.3.
2. In Bundler 1.17.3, [`clean_env` is basically the original environment variables from process start MINUS any env variables starting with `BUNDLE_`](https://github.com/rubygems/bundler/blob/d7089abb6a3c922bfe74199c103e745bb11bd46f/lib/bundler.rb#L285)
3. So, Ruby 2.2 won't work with unbundled_env, because it won't exist. That's fine, we'll just need a conditional on the Bundler 2.1+. Also, lots of people still use Bundler 1 and Bundler 2.0.
4. For Bundler 2.1+ users, we can use with_unbundled_env, and I believe the implementation is exactly the same [as far as I can tell](https://github.com/rubygems/bundler/blob/v2.1.0/lib/bundler.rb#L340)

IMO we should scale back this PR to the above and leave out any other changes.

nateberkopec:
Closing in favor of #2151, lets keep discussion about BUNDLE_GEMFILE on #2133

nateberkopec:
ktvfudccklfjttvnilehbecjbcgjcfbfnjjged

AlexWayfer:
There are tests… but, as you wish. Thank you for clarification and suggestions.

PR #2121, opened 2020-02-20
Rescue IO::WaitReadable instead of EAGAIN for blocking read

wjordan:
### Description

Rescue `IO::WaitReadable` instead of `EAGAIN` on calls to `read_nonblock` when a blocking read would occur.

On Windows, `read_nonblock` raises `Errno::EWOULDBLOCK`, which is a different value on this platform from `Errno::EAGAIN`. Both of these errors are extended by `IO::WaitReadable`, which is Ruby's recommended exception class for retrying `read_nonblock`.

I've included a test `test_open_connection_wait_no_queue` which [fails on windows](https://github.com/wjordan/puma/runs/458869490) without this PR applied.

### DIFF
--- a/History.md
+++ b/History.md
@@ -13,6 +13,7 @@
   * Windows update extconf.rb for use with ssp and varied Ruby/MSYS2 combinations (#2069)
   * Preserve `BUNDLE_GEMFILE` env var when using `prune_bundler` (#1893)
   * Send 408 request timeout even when queue requests is disabled (#2119)
+  * Rescue IO::WaitReadable instead of EAGAIN for blocking read (#2121)
 
 * Refactor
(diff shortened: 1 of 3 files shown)


nateberkopec:
@MSP-Greg since windows is mentioned, thoughts?

MSP-Greg:
LGTM. The set of `Errno` errors varies by OS, so I'm good.  Ruby has several 'adjustments' for things like this in their socket related tests, which run parallel.

Thanks to @wjordan for this.  Test suites tend to be written with 'good input' tests.  Almost more important are 'bad input' tests, which are needed to make sure the app recovers correctly.  Note that 'bad input' can be accidental or deliberate...

PR #2122, opened 2020-02-20
Pass queued requests to thread pool on server shutdown

wjordan:
### Description

Fixes a concurrency bug during server shutdown when `queue_requests` is enabled (the default), which can can be minimally reproduced by the following sequence:

- s1 and s2 both open client connections
- s1 sends a request
- server shutdown is triggered (but does not complete while s1's request is still in process)
- s2 sends a request

Expected: s2 receives a successful response
Actual: s2 receives a 500 response

Here is the test implementation, currently failing on `master`: https://github.com/puma/puma/blob/17af6cfa8a2fa245b2153479448f54505e7b5de4/test/test_puma_server.rb#L771-L788

The underlying issue is that in a 'graceful' shutdown, the server stops accepting connections and the thread pool stops accepting new work while it waits up to `force_shutdown_after` seconds to finish processing currently active requests. However, a client request queued in the reactor will be rejected if it gets received after the thread pool stops accepting new work.

To fix, this PR makes a small change to shut down the reactor _before_ the thread pool, and to disable `queue_requests` and pass along any remaining non-'idle' client connections to the thread pool when the reactor shuts down. This allows graceful shutdown to behave properly without prematurely aborting any requests it shouldn't.

The logic for determining 'idle' clients that can be closed immediately follows guidance from [RFC 2616 section 8.1.4](https://tools.ietf.org/html/rfc2616#section-8.1.4):

> Servers SHOULD always respond to at least one request per connection,
> if at all possible.

The added test depends on the small fix in #2121 to pass properly on Windows, so this branch currently includes that commit as well.

### DIFF
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -278,15 +278,16 @@ module Puma
       @peerip ||= @io.peeraddr.last
     end
 
-    # Returns true if the connection is idle and may be closed.
-    #
-    # From RFC 2616 section 8.1.4:
-    # Servers SHOULD always respond to at least one request per connection,
-    # if at all possible. Servers SHOULD NOT close a connection in the
-    # middle of transmitting a response, unless a network or client failure
-    # is suspected.
-    def idle?
-      @requests_served > 0 && !in_data_phase
+    # Returns true if the persistent connection can be closed immediately
+    # without waiting for the configured idle/shutdown timeout.
+    def can_close?
+      # Allow connection to close if it's received at least one full request
+      # and hasn't received any data for a future request.
+      #
+      # From RFC 2616 section 8.1.4:
+      # Servers SHOULD always respond to at least one request per connection,
+      # if at all possible.
+      @requests_served > 0 && @parsed_bytes == 0
     end
 
     private
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -273,6 +273,17 @@ module Puma
       @peerip ||= @io.peeraddr.last
     end
 
+    # Returns true if the connection is idle and may be closed.
+    #
+    # From RFC 2616 section 8.1.4:
+    # Servers SHOULD always respond to at least one request per connection,
+    # if at all possible. Servers SHOULD NOT close a connection in the
+    # middle of transmitting a response, unless a network or client failure
+    # is suspected.
+    def idle?
+      @requests_served > 0 && !in_data_phase
+    end
+
     private
 
     def setup_body
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -246,7 +246,12 @@ module Puma
     def finish(timeout)
       return true if @ready
       until try_to_finish
-        unless IO.select([@to_io], nil, nil, timeout)
+        can_read = begin
+          IO.select([@to_io], nil, nil, timeout)
+        rescue ThreadPool::ForceShutdown
+          nil
+        end
+        unless can_read
           write_error(408) if in_data_phase
           raise ConnectionError
         end
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -185,8 +185,6 @@ module Puma
 
       @status = :run
 
-      queue_requests = @queue_requests
-
       @thread_pool = ThreadPool.new(@min_threads,
                                     @max_threads,
                                     ::Puma::IOBuffer) do |client, buffer|
@@ -197,7 +195,7 @@ module Puma
         process_now = false
 
         begin
-          if queue_requests
+          if @queue_requests
             process_now = client.eagerly_finish
           else
             client.finish(@first_data_timeout)
@@ -230,7 +228,7 @@ module Puma
 
       @thread_pool.clean_thread_locals = @options[:clean_thread_locals]
 
-      if queue_requests
+      if @queue_requests
         @reactor = Reactor.new self, @thread_pool
         @reactor.run_in_thread
       end
@@ -314,11 +312,12 @@ module Puma
 
         @events.fire :state, @status
 
-        graceful_shutdown if @status == :stop || @status == :restart
         if queue_requests
+          @queue_requests = false
           @reactor.clear!
           @reactor.shutdown
         end
+        graceful_shutdown if @status == :stop || @status == :restart
       rescue Exception => e
         STDERR.puts "Exception handling servers: #{e.message} (#{e.class})"
         STDERR.puts e.backtrace
@@ -388,6 +387,7 @@ module Puma
             end
 
             unless client.reset(check_for_more_data)
+              return unless @queue_requests
               close_socket = false
               client.set_timeout @persistent_timeout
               @reactor.add client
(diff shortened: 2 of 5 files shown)


lib/puma/server.rb
```diff
@@ -293,8 +293,6 @@ def run(background=true)
         return run_lopez_mode(background)
       end
 
-      queue_requests = @queue_requests
```
wjordan: This removes a change from cc4ad1751977e55163ef583f05152288b4d23186, since we need to change behavior based on the updated value of the `@queue_requests` variable after the reactor is shutdown. I assume this was a performance optimization, but I don't know if it actually makes any measurable difference.

lib/puma/client.rb
```diff
@@ -273,6 +273,17 @@ def peerip
...
 
+    # Returns true if the connection is idle and may be closed.
+    #
+    # From RFC 2616 section 8.1.4:
+    # Servers SHOULD always respond to at least one request per connection,
+    # if at all possible. Servers SHOULD NOT close a connection in the
+    # middle of transmitting a response, unless a network or client failure
+    # is suspected.
+    def idle?
+      @requests_served > 0 && !in_data_phase
```
nateberkopec: Why would we not be idle if `@requests_served` is 0?

wjordan: I interpreted the RFC statement "Servers SHOULD always respond to at least one request per connection, if at all possible" to mean that the server shouldn't consider unused (not served at least one request) connections to be 'idle', so shouldn't force-close them during a graceful shutdown. This makes sense to me- if a client opens a connection it usually makes at least one request pretty soon after.

Maybe `#idle?` isn't the clearest term for this method? I used it here because it the same [RFC section](https://tools.ietf.org/html/rfc2616#section-8.1.4) uses it to refer to a connection 'the server has decided to close':

>    A client, server, or proxy MAY close the transport connection at any
   time. For example, a client might have started to send a new request
   at the same time that the server has decided to close the "idle"
   connection. From the server's point of view, the connection is being
   closed while it was idle, but from the client's point of view, a
   request is in progress.

I also considered `#can_close?`, `#should_close?` or `#closeable?` if any of those seem more fitting.

lib/puma/server.rb
```diff
@@ -422,11 +420,12 @@ def handle_servers
 
         @events.fire :state, @status
 
-        graceful_shutdown if @status == :stop || @status == :restart
         if queue_requests
```
evanphx: I presume this needs to be @queue_requests given your change. But I'd also rather you not change this to an ivar read and leave it as a local var.

lib/puma/client.rb
nateberkopec: Riiight, so if we haven't served any requests, we can just shut down. Got it.

Any of the `close` methods you proposed I think would be more fitting.

nateberkopec:
* [ ] Needs a change on the idle method name 
* [ ] Leave queue requests as local var

lib/puma/client.rb
schneems: I don't think `@requests_served` is a good variable name here. It is (from what I understand) a count of the number of requests that have been received from the client, and not the number of requests served from the server. I think a better name might be `@client_request_count`.

Given that tidbit, I don't think the logic here is entirely correct. I do want to take a step back from the specific implementation though and make sure we're on the same page at a high level.

wjordan:
> My high level question is: Did this used to work? Is this a regression or is this something new we're trying to find/fix?

As far as I can tell, this never used to work since the reactor implementation has existed. For example, on bb2aeb8df7c5282ca4aa3a431d4d427de5548b1b (first commit where the `queue_requests` configuration option is available), the test in this PR fails with `queue_requests = true` and passes with `queue_requests = false`.

I only encountered this issue as an intermittent unit-test failure while working on #2123, and worked backwards from there to create a minimal reproducible failing test and fix. In any case it seems possible that #2200 is caused by this underlying issue, though I can't be sure.

> Regardless of if this was broken or not, another question is: What do we want to happen?
> What do we want to do when the client can be fully buffered? (scenario B)
> What do we want to do when the client cannot be fully buffered? (scenario C)

I wasn't sure how Scenario 'A' differed from 'B' (identical except for the 'desired' part?), so I'll just address B and C.

- For **B**, I agree on _behavior_ - for a graceful shutdown, the server should wait to finish receiving the request, and send a response.

  However, the _implementation_ is a bit different. When the server gets a shutdown request, it _immediately_ shuts the reactor down (passing pending requests to the thread pool) and sets `queue_requests` to `false`. The thread pool then continues reading the partially-buffered client requests and processing responses in its active threads, using the `queue_requests = false` code path as though the reactor never existed.

- For **C**, the difference from B is that the server doesn't wait forever for the request to finish, it closes the connection after a configurable timeout. (HTTP allows that ['A client, server, or proxy MAY close the transport connection at any time.'](https://tools.ietf.org/html/rfc2616#section-8.1.4))

  In Puma, this timeout is configurable by the lesser of `first_data_timeout` (default `30` seconds) or `force_shutdown_after` (default `forever`)

  Also, historically (since #451), if a request is in the 'data phase' (request headers have been sent and a request body is in-flight) when it times out, Puma returns a 408 response before it drops the connection instead of just dropping without any response. (Call this Scenario **C2**) I think this is a minor detail, but ideally behavior during shutdown should be consistent with this existing behavior.

- Finally, worth noting that if the server is busy with a _response_ when the force-shutdown timeout completes, the server responds with a 503 error (uncaught `Puma::ThreadPool::ForceShutdown exception`). Worth pointing this out as a separate related scenario (call it Scenario **D**).

To clarify the behavior across all combinations of scenarios, I added a few extra test cases to this PR. Let me know if these are clear enough and if the expected behavior is reasonable, or if there are any additional scenarios you think should be covered.

> If you pass a non-fully buffered client request to the thread pool what happens? Does it try to buffer it? cc/ @nateberkopec @evanphx

Yes, this is what happens when `queue_requests` is set to `false`, and what this PR does when the server gets a shutdown request.

lib/puma/server.rb
wjordan: There were actually two separate `queue_requests` local vars- one within `Server#run`, and one within `Server#handle_servers`.

The local var in `#run` had to be removed in order for this PR to work at all- see my comment at https://github.com/puma/puma/pull/2122#pullrequestreview-362317379. However, this one in `#handle_servers` actually doesn't need to be changed as well, so I left it as-is to minimize the diff.

That said, since this local var in `#handle_servers` is only ever read on server shutdown and probably has no performance impact, I would lean towards removing it for consistency/simplicity if that's your preference.

lib/puma/client.rb
nateberkopec: Given that this on client.rb, I'm OK with the variable name.

lib/puma/server.rb
nateberkopec: This explanation works for me.

nateberkopec:
👏

aaronjensen:
🎊 Is there a timeframe for the next release of puma?

nateberkopec:
Basically the two PRs listed in the milestone have to get merged: https://github.com/puma/puma/milestone/7

aaronjensen:
Ah, major release is next. Okay, thanks.

PR #2124, opened 2020-02-21
JSON parse cluster worker stats instead of regex

wjordan:
### Description

Refactoring PR to simplify the cluster worker stats code using `JSON.parse` and `.to_json` in place of regex-parsing and string interpolation.

Also includes refactoring of duplicated server-stats logic into a `Server#stats` method.

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -5,6 +5,7 @@ require 'puma/util'
 require 'puma/plugin'
 
 require 'time'
+require 'json'
 
 module Puma
   # This class is instantiated by the `Puma::Launcher` and used
@@ -94,11 +95,7 @@ module Puma
 
       def ping!(status)
         @last_checkin = Time.now
-        captures = status.match(/{ "backlog":(?<backlog>\d*), "running":(?<running>\d*), "pool_capacity":(?<pool_capacity>\d*), "max_threads": (?<max_threads>\d*), "requests_count": (?<requests_count>\d*) }/)
-        @last_status = captures.names.inject({}) do |hash, key|
-          hash[key.to_sym] = captures[key].to_i
-          hash
-        end
+        @last_status = JSON.parse(status, symbolize_names: true)
       end
 
       def ping_timeout?(which)
@@ -287,18 +284,11 @@ module Puma
 
       Thread.new(@worker_write) do |io|
         Puma.set_thread_name "stat payload"
-        base_payload = "p#{Process.pid}"
 
         while true
           sleep Const::WORKER_CHECK_INTERVAL
           begin
-            b = server.backlog || 0
-            r = server.running || 0
-            t = server.pool_capacity || 0
-            m = server.max_threads || 0
-            rc = server.requests_count || 0
-            payload = %Q!#{base_payload}{ "backlog":#{b}, "running":#{r}, "pool_capacity":#{t}, "max_threads": #{m}, "requests_count": #{rc} }\n!
-            io << payload
+            io << "p#{Process.pid}#{server.stats.to_json}\n"
           rescue IOError
             Thread.current.purge_interrupt_queue if Thread.current.respond_to? :purge_interrupt_queue
             break
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -949,5 +949,10 @@ module Puma
       HTTP_INJECTION_REGEX =~ header_value.to_s
     end
     private :possible_header_injection?
+
+    def stats
+      stat_names = %i(backlog running pool_capacity max_threads requests_count)
+      stat_names.map {|name| [name, send(name) || 0]}.to_h
+    end
   end
 end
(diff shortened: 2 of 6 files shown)


nateberkopec:
Hmmm... I wonder if those windows failures are related or not. I feel like I've seen stalls on that test before, though not specifically on Windows.

nateberkopec:
Pinging @MSP-Greg for the possible test stability issue here.

lib/puma/cluster.rb
```diff
@@ -5,6 +5,7 @@
 require 'puma/plugin'
 
 require 'time'
+require 'json'
```
nateberkopec: Currently we only require `json` for the status app, which isn't that big of a deal. This would definitely require `json` for most people running Puma.

nateberkopec: For example, this would, I believe, introduce a new native extension during startup, which I believe affects other open PRs.

lib/puma/server.rb
```diff
@@ -1045,5 +1045,10 @@ def self.current
     def shutting_down?
       @status == :stop || @status == :restart
     end
+
+    def stats
+      stat_names = %i(backlog running pool_capacity max_threads requests_count)
```
nateberkopec: Would prefer these stats in a constant somewhere.

test/test_cli.rb
```diff
@@ -93,7 +93,7 @@ def test_control_for_ssl
 
     expected_stats = /{"started_at":"\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z","backlog":0,"running":0,"pool_capacity":16,"max_threads":16,"requests_count":0}/
     assert_match(expected_stats, body.split(/\r?\n/).last)
-    assert_equal([:started_at, :backlog, :running, :pool_capacity, :max_threads, :requests_count], Puma.stats.keys)
+    assert_equal(JSON.parse(body.split(/\r?\n/).last, symbolize_names: true), Puma.stats)
```
nateberkopec: We already checked the entire response on the previous line. IMO, this test is redundant.

test/test_cli.rb
```diff
@@ -137,6 +137,7 @@ def test_control_clustered
     s.close
 
     assert_match(/\{"started_at":"\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z","workers":2,"phase":0,"booted_workers":2,"old_workers":0,"worker_status":\[\{"started_at":"\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z","pid":\d+,"index":0,"phase":0,"booted":true,"last_checkin":"[^"]+","last_status":\{"backlog":0,"running":2,"pool_capacity":2,"max_threads":2,"requests_count":0\}\},\{"started_at":"\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z","pid":\d+,"index":1,"phase":0,"booted":true,"last_checkin":"[^"]+","last_status":\{"backlog":0,"running":2,"pool_capacity":2,"max_threads":2,"requests_count":0\}\}\]\}/, body.split("\r\n").last)
+    assert_equal(Puma.stats, JSON.parse(body.split("\r\n").last, symbolize_names: true))
```
nateberkopec: Same here, redundant test.

nateberkopec: Maybe we can test Puma.stats output separately in another test if we don't do that explicitly anywhere else?

MSP-Greg:
@nateberkopec

I did notice those tests failures.  I'll have a look.

lib/puma/cluster.rb
wjordan: I thought requiring `json` would be straightforward as part of the [standard library](https://ruby-doc.org/stdlib/libdoc/json/rdoc/JSON.html), but I did find it unusual that it was being intentionally avoided. If loading `json` does introduce some known stability/performance issue on any of the supported platforms (I'm less familiar with non-linux/MRI platforms) it makes sense to not accept this change.

MSP-Greg:
@nateberkopec

I grabbed the PR, and ran tests in my fork, both standard and using the `--seed` value from the failed CI here.  Both passed.

I then quickly tried using 2.6.5 locally, and things got weird.  It had Bundler 1.7.2 installed, when I upgraded both RubyGems & Bundler with `gem update --system`, everything was fine.  Hence, I can't determine what happened here. Maybe some moon phase thing.  Sorry.

But, all the master/head are passing.  Just thought I'd point that out.  Also, maybe PR https://github.com/puma/puma/pull/2126

nateberkopec:
Starting a test re-run. Still needs my code concerns addressed. I'm discussing the json thing with Evan.

test/test_cli.rb
wjordan: Yeah, I added these lines because my first iteration of this PR unintentionally changed keys in `Puma.stats` from symbols to strings (`JSON.parse` default) without failing any tests, so I thought that behavior should be covered by an extra assertion somewhere. I can move it to a new test to make it more clear.

lib/puma/cluster.rb
nateberkopec: I checked with Evan and there was no special reason. This is good to 🚢

PR #2130, opened 2020-02-27
Use faster `Regexp#match?` to reject possible injected headers

carlosantoniodasilva:
### Description

When looking for possible injected headers with CRLF, use faster `Regexp#match?` which doesn't update Ruby $~ and other related variables.

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -687,7 +687,7 @@ module Puma
 
           return :async if req.hijacked
           # Checking to see if an attacker is trying to inject headers into the response
-          headers.reject! { |_k, v| CRLF_REGEX =~ v.to_s }
+          headers.reject! { |_k, v| CRLF_REGEX.match? v.to_s }
 
           status = status.to_i
 


MSP-Greg:
I don't think it's available with Ruby 2.3, only `#match`...

nateberkopec:
Thanks, but we support Ruby 2.2+, so this won't work for us. 

I tried a version with match for 2.2-2.3, and match? for 2.4+, but it doesn't make a speed difference. The reason is that `=~` only sets globals on a _successful_ match, which 99.999999% of the time won't happen on this code path. 

So, since they're the same perf 99.999% of the time, a branching definition isn't worth the maintenance and readability cost.

To test for yourself, see the benchmarks in the benchmarks folder 👍

nateberkopec:
Actually, slightly incorrect: [match/=~ sets the globals to `nil`](https://github.com/ruby/ruby/blob/master/re.c#L3122), which, as my benchmarks showed, is basically free because it doesn't allocate anything.

carlosantoniodasilva:
Oh I missed the support on Ruby 2.2/3 was still on Puma, thanks for pointing it out. :+1:

bdewater:
It may not matter here performance-wise, but in case it would have: refinements are an excellent way of polyfilling this for older Rubies as shown in https://github.com/ruby/csv/pull/30

PR #2143, opened 2020-03-02
add RAILS_MIN_THREADS, RAILS_MAX_THREADS, set default worker, preload…

### LINKED ISSUE #1949: Make "puma-heroku" the default config

jalevin:
… if using workers

### Description
closes #1949 

- [ x] I have reviewed the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.
- [ x] I have added appropriate tests if this PR fixes a bug or adds a feature.
- [ x] My pull request is 100 lines added/removed or less so that it can be easily reviewed.


History.md
```diff
@@ -10,6 +10,12 @@
   * Changed #connected_port to #connected_ports (#2076)
   * `--control` has been removed. Use `--control-url` (#1487)
   * `worker_directory` has been removed. Use `directory`
+  * min_threads now set by environment variables RAILS_MIN_THREADS and MIN_THREADS
```
nateberkopec: Let's make the other env variable PUMA_MIN_THREADS/PUMA_MAX_THREADS

History.md
```diff
@@ -10,6 +10,12 @@
...
+  * max_threads now set by environment variables RAILS_MAX_THREADS and MAX_THREADS
+  * max_threads default to 5 in MRI or 16 for all other interpretters
+  * default to 1 worker if running on MRI or 0 if jruby/windows
```
nateberkopec: Oh, this is in puma-heroku? hm...

lib/puma/configuration.rb
```diff
@@ -106,7 +106,7 @@ def finalize_values
   #
   # It also handles loading plugins.
   #
-  # > Note: `:port` and `:host` are not valid keys. By the time they make it to the
+  # > Note: `:port` and `:host` are not valid keys. By they time they make it to the
```
nateberkopec: Rebase?

lib/puma/configuration.rb
```diff
@@ -167,14 +169,24 @@ def flatten!
       self
     end
 
+    def default_workers
+      return 0 if Puma.jruby? || Puma.windows?
+      1
+    end
+
+    def default_max_threads
+      return 5 if RUBY_ENGINE.nil? || RUBY_ENGINE == 'ruby'
```
nateberkopec: Hmmm...is this is best way to detect MRI?

Also, we should abstract this behind `Puma.mri?`

nateberkopec:
My guess on test failures would be that they're related to the change of changing the default from single mode to cluster mode, which I'm not even sure is a good idea.

olleolleolle:
I liked the PUMA_MIN_THREADS naming.

Would it be easier to create this PR if it was only about the ENV vars, and left out the other stuff? Could make this change smaller and easier to review.

PS: this is awesome, thanks for taking the time to work on it!

nateberkopec:
I'm fine with the size-factor of this PR. Need to think about the Single vs Cluster for MRI thing.

dentarg:
Regarding "changing the default from single mode to cluster mode", I assume it would be a change for Puma 5, but nonetheless I guess the [`before_fork`](https://github.com/puma/puma/tree/8f282206dcdd07d765553179fe31d61acfd08dda#clustered-mode) stuff (e.g. you should do `DB.disconnect`) could be a surprise for people running the default and not knowing about cluster mode.

nateberkopec:
OK. Let's keep Puma in singlemode by default. No need to force cluster by changing workers to 1. With that plus the other changes outlined above, LGTM.

jalevin:
fyi: for some reason test/test_config.rg:43 is a flaky test. I need to research this more, but occasional this returns 2 instead of 0. I'm not sure if that is coming from a test helper or something

jalevin:
Builds failing with `TimeoutEveryTestCase::TestTookTooLong: execution expired... `

History.md
```diff
@@ -10,6 +10,11 @@
   * Changed #connected_port to #connected_ports (#2076)
   * `--control` has been removed. Use `--control-url` (#1487)
   * `worker_directory` has been removed. Use `directory`
+  * min_threads now set by environment variables RAILS_MIN_THREADS and MIN_THREADS
```
nateberkopec: Sorry if I wasn't clear - `/min_threads/puma_min_threads/` and `/max_threads/puma_max_threads` please.

lib/puma/configuration.rb
```diff
@@ -137,6 +137,8 @@ def initialize(user_options={}, default_options = {}, &block)
       @file_dsl    = DSL.new(@options.file_options, self)
       @default_dsl = DSL.new(@options.default_options, self)
 
+      default_options[:preload_app] = (default_options[:workers] > 1 && Puma::Plugin.new.workers_supported?)
```
nateberkopec: Come to think of it, this could break things for people doing the "don't preload but do USR2 restarts instead" path. 

I think we need to check to see if other settings indicating people are doing that are set, like `prune_bundler`. So if `prune_bundler` is true, don't preload_app by _default_. But I'm not sure if we can detect it otherwise? 

+ Also let's find a way to check if we can do workers without calling into `Puma::Plugin`, that looks ugly.

nateberkopec: This breakage may be what the tests are showing you.

ukolovda:
Will using `PUMA_MIN_THREADS` and `PUMA_MAX_THREADS` instead of `RAILS_MIN_THREADS` (and MAX) be better?
I use Puma for `Sinatra` application, and `RAILS_*` variables will be confuse.

lib/puma/configuration.rb
jalevin: How do we feel about this?

```			
# 1. workers_supported - #Puma::Plugin.new.workers_supported?)
workers_supported = !(Puma.jruby? || Puma.windows?)

# 2. preload app
if !@options[:prune_bundler]
  default_options[:preload_app] =(@options[:workers] > 1) && workers_supported 
end
```

nateberkopec: LGTM!

lib/puma/configuration.rb
```diff
@@ -137,6 +137,12 @@ def initialize(user_options={}, default_options = {}, &block)
       @file_dsl    = DSL.new(@options.file_options, self)
       @default_dsl = DSL.new(@options.default_options, self)
 
+      workers_supported = !(Puma.jruby? || Puma.windows?)
```
nateberkopec, line 140: Check ```::Process.respond_to? :fork``` instead.

lib/puma/configuration.rb
```diff
@@ -167,14 +173,20 @@ def flatten!
       self
     end
 
+    def default_max_threads
+      return 5 if Puma.mri?
```
nateberkopec, line 177: I would prefer a ternary here instead of return.

lib/puma/configuration.rb
nateberkopec: Also no need to assign to a variable, just put it on line 143

lib/puma/configuration.rb
```diff
@@ -167,14 +173,20 @@ def flatten!
...
-        :min_threads => 0,
-        :max_threads => 16,
+        :min_threads => Integer(ENV['PUMA_MIN_THREADS'] || ENV['MIN_THREADS'] || 0),
+        :max_threads => Integer(ENV['PUMA_MAX_THREADS'] || ENV['MAX_THREADS'] || default_max_threads),
         :log_requests => false,
         :debug => false,
         :binds => ["tcp://#{DefaultTCPHost}:#{DefaultTCPPort}"],
-        :workers => 0,
+        :workers => Integer(ENV['WEB_CONCURRENCY'] || 0),
+        :daemon => false,
```
nateberkopec, line 189: Looks like this was accidentally added back in.

test/test_config.rb
```diff
@@ -9,6 +9,58 @@
...
+    max_threads = 16
+    max_threads = 5 if RUBY_ENGINE.nil? || RUBY_ENGINE == 'ruby'
+    assert_equal max_threads, Puma::Configuration.new.default_max_threads
+  end
+
+  def test_config_loads_correct_min_threads
+    conf = Puma::Configuration.new
+    assert_equal 0, conf.options.default_options[:min_threads]
+
+    ENV['MIN_THREADS'] = '7'
```
nateberkopec, line 22: Setting a global and running tests in parallel threads, you're gonna have a Bad Time (as you've seen on your tests).

Maybe just removing parallelize_me will fix everything here. Put these tests into a separate test class which is not paralellized.

jalevin: Doesn't look like removing `parallelize_me!` works here. Any other ideas?

jalevin: Scratch that- extrapolated into it's own class and used the with_env helper. All seems fine.

jalevin:
Does truffleruby support workers?

lib/puma/plugin.rb
```diff
@@ -109,7 +109,7 @@ def in_background(&blk)
     end
 
     def workers_supported?
```
jalevin, line 111: This method feels like it should be in the lib/puma/detect.rb

lib/puma/plugin.rb
```diff
@@ -109,7 +109,7 @@ def in_background(&blk)
...
-      return false if Puma.jruby? || Puma.windows?
+      return false if Puma.jruby? || Puma.windows? || Puma.truffle?
```
jalevin, line 112: should we check ::Process.respond_to?(:fork) as well?
Does that single check eliminate the need to check for jruby/windows/truffle?

jalevin: @nateberkopec

lib/puma/plugin.rb
jalevin: @nateberkopec

lib/puma/plugin.rb
nateberkopec: Fixed in my nitpick commit. Thanks for the work on this PR!

jalevin: Sweet!

fabioperrella:
Hi folks! Is there a reason to set `max_threads` default to 5 in MRI?

I couldn't find the why in the discussions or related issues.

nateberkopec:
@fabioperrella long version: https://www.speedshop.co/2020/05/11/the-ruby-gvl-and-scaling.html

short version: high threadcounts increase contention for the GVL. too much GVL contention == increased latency.

jsmpereira:
@nateberkopec Thanks for that resource :)

@fabioperrella I was going to mention, from https://github.com/puma/puma/blob/master/5.0-Upgrade.md#upgrade:

> If you are running MRI, default thread count on Puma is now 5, not 16. This may change the amount of threads running in your threadpool. We believe 5 is a better default for most Ruby web applications on MRI. Higher settings increase latency by causing GVL contention.

fabioperrella:
great, tks!!

PR #2154, opened 2020-03-07
Leave BUNDLE_GEMFILE unset in workers if it was unset in the master

### LINKED ISSUE #2133: BUNDLE_GEMFILE is incorrectly set to parent Gemfile in worker processes (when not setting BUNDLE_GEMFILE explicitly)

cjlarose:
### Description

Closes #2133


lib/puma/launcher.rb
```diff
@@ -298,7 +298,8 @@ def prune_bundler
 
       log '* Pruning Bundler environment'
       home = ENV['GEM_HOME']
-      bundle_gemfile = ENV['BUNDLE_GEMFILE']
+      bundle_gemfile = ENV['BUNDLER_ORIG_BUNDLE_GEMFILE']
+      bundle_gemfile = nil if bundle_gemfile == 'BUNDLER_ENVIRONMENT_PRESERVER_INTENTIONALLY_NIL'
```
nateberkopec: What's this about?

cjlarose: Removed in 0b47735f78c370d103af5c1b8b581948f56cd688

seven1m:
This is solid work. Love the test!

nateberkopec:
I'm gonna play around with the test locally but the code itself LGTM

PR #2156, opened 2020-03-07
Binder testcov

nateberkopec:


lib/puma/binder.rb
```diff
@@ -43,7 +43,8 @@ def initialize(events)
       @ios = []
     end
 
-    attr_reader :ios, :listeners, :unix_paths, :proto_env, :envs
+    attr_reader :ios, :listeners, :unix_paths, :proto_env, :envs, :activated_sockets, :inherited_fds
```
nateberkopec: Making all of this stuff public for a future refactor

lib/puma/binder.rb
```diff
@@ -238,7 +233,7 @@ def add_tcp_listener(host, port, optimize_for_latency=true, backlog=1024)
     end
 
     def connected_ports
-      ios.map { |io| io.addr[1] }
+      ios.map { |io| io.addr[1] }.uniq
```
nateberkopec: Bugfix - ipv6 binds would make ports show up twice

PR #2165, opened 2020-03-09
[changelog skip] Backfill tests for stdout/stderr redirection on SIGHUP

cjlarose:
### Description

I noticed that there weren't any tests for what is supposed to happen when `puma` receives the `SIGHUP` signal. #561 was the PR that introduced the behavior. I didn't really understand what the use case was at first, but [an issue on a unrelated project](https://github.com/influxdata/telegraf/issues/5128) gave a good explanation for why this functionality is desirable.

### DIFF
--- a/test/test_redirect_io.rb
+++ b/test/test_redirect_io.rb
@@ -7,19 +7,15 @@ class TestRedirectIO < TestIntegration
   def setup
     super
 
-    @out_file = Tempfile.new 'puma-out'
-    @out_file_path = @out_file.path
-    @err_file = Tempfile.new 'puma-err'
-    @err_file_path = @err_file.path
+    @out_file_path = Tempfile.new('puma-out').path
+    @err_file_path = Tempfile.new('puma-err').path
   end
 
   def teardown
     super
 
-    @out_file.close
-    @err_file.close
-    File.unlink @out_file_path
-    File.unlink @err_file_path
+    paths = [@out_file_path, @err_file_path, @old_out_file_path, @old_err_file_path]
+    File.unlink(*paths)
   end
 
   def test_sighup_redirects_io_single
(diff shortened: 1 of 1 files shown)


cjlarose:
Tests for macOS on Ruby < 2.4 are failing with `bundle install` errors. Seems unrelated. FWIW tests pass on the same commit on my fork https://github.com/cjlarose/puma/actions/runs/52368829

test/test_redirect_io.rb
```diff
@@ -0,0 +1,103 @@
+require_relative "helper"
+require_relative "helpers/integration"
+
+class TestRedirectIO < TestIntegration
+  parallelize_me!
+
+  def setup
+    super
+
+    @out_file = Tempfile.new 'puma-out'
```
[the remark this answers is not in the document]
nateberkopec: Maybe use the current threads name or object_id or something

test/test_redirect_io.rb
```diff
@@ -0,0 +1,103 @@
...
+    @out_file_path = @out_file.path
+    @err_file = Tempfile.new 'puma-err'
+    @err_file_path = @err_file.path
+  end
+
+  def teardown
+    super
+
+    @out_file.close
```
nateberkopec: You can use `close!` and then you don't have to unlink manually.

nateberkopec: or `close(true)`

test/test_redirect_io.rb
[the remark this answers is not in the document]
cjlarose: Under what circumstances would you expect a collision? `Tempfile` generates a unique filename.

> Tempfile's filename picking method is both thread-safe and inter-process-safe: it guarantees that no other threads or processes will pick the same filename.

https://ruby-doc.org/stdlib-2.4.0/libdoc/tempfile/rdoc/Tempfile.html#class-Tempfile-label-Minor+notes

test/test_redirect_io.rb
cjlarose: `@out_file` is an instance of `File < IO`. There is no `close!` method and `IO#close` takes no arguments

https://ruby-doc.org/core-2.7.0/IO.html#method-i-close

test/test_redirect_io.rb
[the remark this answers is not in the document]
nateberkopec: TIL, nvm

test/test_redirect_io.rb
nateberkopec: Oh tricky tricky, I didn't follow that through the test. Why replace a `Tempfile` with a `File`?

cjlarose: After `log_rotate_output_files`, we reopen the file so we get a new handle to same path (otherwise the handle would still be pointing to the old file).

This was all to avoid having to create file handles repeatedly, but it was not really necessary and pretty confusing. I pushed a change to make it more clear.

nateberkopec:
Brilliant! Thanks mate. Your PRs are really top-quality 👍

PR #2168, opened 2020-03-09
Increase max URI path length to 8196, closes #2134

### LINKED ISSUE #2134: Add configurable max URI length to meet HTTP/1.1 spec
## Problem

I am developing an endpoint to `GET` information on a large number of values which are supplied as array parameters `/get_thing?ids[]=1&ids[]=2...` but the number of values I can retrieve is limited by Puma's [max URI limit](https://github.com/puma/puma/pull/408/files#diff-a59f94a30a1e4fd79523ef77ba8fa711) which was set based on what Internet Explorer would accept in 2013. This is no longer a reasonable metric against which to set the URI length. The [IETF updated HTTP/1.1 spec](https://tools.ietf.org/html/rfc7230) from 2014 says in [section 3.1.1](https://tools.ietf.org/html/rfc7230#section-3.1.1):

> HTTP does not place a predefined limit on the length of a
> request-line, as described in Section 2.5. [...]
> Various ad hoc limitations on request-line length are found in
> practice.  It is RECOMMENDED that all HTTP senders and recipients
> support, at a minimum, request-line lengths of 8000 octets.

In [section 2.5](https://tools.ietf.org/html/rfc7230#section-2.5):

> HTTP does not have specific length limitations for many of its
> protocol elements because the lengths that might be appropriate will
> vary widely, depending on the deployment context and purpose of the
> implementation.  Hence, interoperability between senders and
> recipients depends on shared expectations regarding what is a
> reasonable length for each protocol element.  Furthermore, what is
> commonly understood to be a reasonable length for some protocol
> elements has changed over the course of the past two decades of HTTP
> use and is expected to continue changing in the future.
> 
> At a minimum, a recipient MUST be able to parse and process protocol
> element lengths that are at least as long as the values that it
> generates for those same protocol elements in other messages.  For
> example, an origin server that publishes very long URI references to
> its own resources needs to be able to parse and process those same
> references when received as a request target.

## The solution I'd like

To comply with RFC7230 and the HTTP/1.1 spec the server _should_ have a configurable `MAX_REQUEST_PATH_LENGTH`. This could be set in the `config\puma.rb` file if required.

## Alternatives considered

At a minimum the server should implement the recommended 8000 octets (which for ASCII URLs I _think_ would be 8000 characters), which would reduce the number of calls my application would have to make. This is imperfect however as it doesn't meet the requirement that the server should have no hard upper limit.

Laukei:
### Description
Increases the maximum path length for a resource from 2048 to 8196 to fulfill the resolution in #2134 to bring Puma in line with the minimum recommended path length limit from [the HTTP/1.1 spec](https://tools.ietf.org/html/rfc7230#section-3.1.1):

> It is RECOMMENDED that all HTTP senders and recipients
> support, at a minimum, request-line lengths of 8000 octets.

**Design choices**

The HTTP/1.1 spec says a request-line length of 8000 octets should be the minimum. An octet is 8 bits, which we would typically call a byte. In the worst-case scenario, if the path length is in ASCII-encoded text or the UTF-8 code pages that align with it, one byte represents one character, so 8000 octets represents 8000 characters. To round this number to the nearest integer multiple of 1024 I chose 8196, purely as multiples of 1024 defined the previous values, and the recommendation is for _a minimum_ limit.


Laukei:
I haven't updated any documentation with this change as I couldn't find anywhere the `MAX_REQUEST_PATH_LENGTH` was mentioned in docs!

nateberkopec:
Thanks!

michaelize:
1024 * 8 = 8192

I see there are two values used in this PR, 8192 and 8196. Should they both be 8192?

michaelize:
One more question, it seems MAX_REQUEST_URI_LENGTH = MAX_REQUEST_PATH_LENGTH + MAX_QUERY_STRING_LENGTH, should we also increase the value for MAX_REQUEST_URI_LENGTH?

Appreciate any feedback on this

dentarg:
Yeah, I don't get it either... is it a typo? (8192 vs 8196) @nateberkopec @Laukei

Laukei:
Yep! They should be the same.

nateberkopec:
Anyone wanna PR that fix?

Issue #2177, opened 2020-03-12
out_of_band hook not working when using multiple threads


GuiTeK:
**Describe the bug**
The `out_of_band` hook (C.F. https://github.com/puma/puma/pull/1648) seems not to work when the number of threads specified with `threads N, N` is `> 1`.

**Puma config:**

```
threads 2, 2
port        ENV.fetch("PUMA_PORT") { 3000 }
environment ENV.fetch("RAILS_ENV") { "development" }
workers ENV.fetch("PUMA_WORKERS") { 1 }

out_of_band do
  puts 'OOB Hook Working'
end

preload_app!

plugin :tmp_restart
```

Command: `bundle exec puma -C config/puma.rb oob.ru`

Contents of `oob.ru`:
```
run lambda { |env| [200, {"Content-Type" => "text/plain"}, ["Hello World"]] }
```

**To Reproduce**
1. Make sure `config/puma.rb` and `oob.ru` have the contents showed above
2. Run `bundle exec puma -C config/puma.rb oob.ru`
3. Make some requests: `curl http://127.0.0.1:3000/`
4. Note that `OOB Hook Working` does NOT get printed in Puma output

Now, to show it works with `threads 1, 1`:
1. Stop the Puma server if already running
2. Replace `threads 2, 2` in the Puma config with `threads 1, 1`
3. Run `bundle exec puma -C config/puma.rb oob.ru`
4. Make some requests: `curl http://127.0.0.1:3000/`
5. Note that `OOB Hook Working` IS present in Puma output

**Expected behavior**
I expect `out_of_band` hook to be called even when Puma is configured to run with multiple threads.

**Desktop:**
 - OS: Mac OS X 10.14
 - Puma Version: 4.3.3

nateberkopec:
@GuiTeK Just curious, what do you use `out_of_band` for?

GuiTeK:
@nateberkopec I'd like to use it for Garbage Collection (to run `GC.start`).

nateberkopec:
I see. So, if we provide a feature in Puma, it should obviously work, and this definitely doesn't, so I want to fix that.

_However_, OOBGC is an outdated technique since the changes introduced in the Ruby 2.3 GC. I can't recommend it. GitHub disabled it after they realized it was causing an additional 10% CPU usage.

wjordan:
[...]

PR #2178, opened 2020-03-12
Fix out_of_band hook never executed when multiple worker threads (clo…

GuiTeK:
### Description

**Alternative solution** to this PR: https://github.com/puma/puma/pull/2180
I believe this one should be merged and https://github.com/puma/puma/pull/2180 should be discarded but the maintainers are probably wiser than me on Puma!

#### Issue description
C.F. https://github.com/puma/puma/issues/2177

#### Fix description
Before this patch, the `out_of_band` hook was executed only when *all* worker threads were free:
```ruby
pool << client
busy_threads = pool.wait_until_not_full
if busy_threads == 0
  @options[:out_of_band].each(&:call) if @options[:out_of_band]
end
```

In addition, due to how the function `wait_until_not_full` (in lib/puma/thread_pool.rb) works, namely:
* If not all threads are busy, return the number of busy threads
* Otherwise, wait (loop) for the condition above

The condition `if busy_threads == 0` could never be satisfied if the number of worker threads was > 1.

EDIT: it could be satisfied if the worker finished its job before the call to `wait_until_not_full`, which is *very unlikely* because this function is called right after adding the client to the pool. 

I believe that the `out_of_band` hook should be allowed to run if there is *at least 1* available worker thread. It seems reasonable to me and provides an easy fix for this bug.

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -291,7 +291,7 @@ module Puma
 
                     pool << client
                     busy_threads = pool.wait_until_not_full
-                    if busy_threads == 0
+                    if busy_threads < @max_threads
                       @options[:out_of_band].each(&:call) if @options[:out_of_band]
                     end
                   end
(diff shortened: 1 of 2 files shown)


lib/puma/server.rb
```diff
@@ -291,7 +291,7 @@ def handle_servers
 
                     pool << client
                     busy_threads = pool.wait_until_not_full
-                    if busy_threads == 0
```
nateberkopec: Well, the problem here is that if busy_threads != 0, it's not really out of band anymore, is it?

nateberkopec:
I think this is a non-starter because it makes the OOB hook no longer OOB and just a background thread.

PR #2180, opened 2020-03-12
Fix out_of_band hook never executed when multiple worker threads (clo…

GuiTeK:
### Description

This is an **alternative solution** for https://github.com/puma/puma/issues/2177.
The other PR is https://github.com/puma/puma/pull/2178. The difference is that **this PR keeps strictly the same behaviour as before** (i.e. runs OOB hook only if *all* threads are free).

#### Issue description
C.F. https://github.com/puma/puma/issues/2177

#### Fix description
Here is the code responsible for executing the `out_of_band` hook:
```ruby
pool << client
busy_threads = pool.wait_until_not_full
if busy_threads == 0
  @options[:out_of_band].each(&:call) if @options[:out_of_band]
end
```

Due to how the function `wait_until_not_full` (in lib/puma/thread_pool.rb) works, namely:
* If not all threads are busy, return the number of busy threads
* Otherwise, wait (loop) for the condition above

The condition `if busy_threads == 0` could never be satisfied if the number of worker threads was > 1. It could be satisfied if the worker finished its job before the call to `wait_until_not_full`, which is *very unlikely* because this function is called right after adding the client to the pool. 

This PR fixes this problem by giving a callback function to `ThreadPool.initialize` to signal the `Server` when all worker threads are free (using the boolean variable `@all_worker_threads_free`).

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -63,6 +63,9 @@ module Puma
 
       @status = :stop
 
+      @all_worker_threads_free = false
+      @all_worker_threads_free_mutex = Mutex.new
+
       @min_threads = 0
       @max_threads = 16
       @auto_trim_time = 30
@@ -189,6 +192,7 @@ module Puma
 
       @thread_pool = ThreadPool.new(@min_threads,
                                     @max_threads,
+                                    Proc.new { signal_all_worker_threads_free },
                                     ::Puma::IOBuffer) do |client, buffer|
 
         # Advertise this server into the thread
@@ -290,10 +294,6 @@ module Puma
                     end
 
                     pool << client
-                    busy_threads = pool.wait_until_not_full
-                    if busy_threads == 0
-                      @options[:out_of_band].each(&:call) if @options[:out_of_band]
-                    end
                   end
                 rescue SystemCallError
                   # nothing
@@ -309,6 +309,19 @@ module Puma
             end
           rescue Object => e
             @events.unknown_error self, e, "Listen loop"
+          end
+
+          if @all_worker_threads_free_mutex.try_lock # Do not block the critical path if mutex isn't available
+            if @all_worker_threads_free
+              begin
+                @options[:out_of_band].each(&:call) if @options[:out_of_band]
+              rescue Exception => e
+                STDERR.puts "Exception handling OOB callbacks: #{e.message} (#{e.class})"
+                STDERR.puts e.backtrace
+              end
+              @all_worker_threads_free = false
+            end
+            @all_worker_threads_free_mutex.unlock
           end
         end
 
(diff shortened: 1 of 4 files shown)


nateberkopec:
This approach makes a bit more sense as it's still truly OOB. Need to think about implementation...

nateberkopec:
Rather than use the pipe to communicate intra-process, couldn't we use a condition variable to signal to a thread in the server to run the oob calls?

GuiTeK:
Hey! Thanks for the feedback.

I'm not sure about the `ConditionVariable`, especially because the "main" thread (by that, I mean the thread in which `handle_servers` is running) will not always be `wait`ing for it for the `signal` emitted from the worker threads (and in that case, it will be lost).

If you really want to implement it that way, can you please explain me how you would do it?

For now, I implemented it using a boolean & a mutex.

GuiTeK:
I removed a now useless `include Puma::Const` I had added in a previous commit.

nateberkopec:
Oh awesome! Thanks for correcting me - as weird as this may sound coming from a current maintainer of Puma, I'm not an expert on the Mutex/CV classes (that code is almost entirely still Evan's). But it seemed to me like the checkpipe was unnecessary when the threadpool and the runner were always going to be in the same process - you figured it out.

lib/puma/server.rb
```diff
@@ -189,6 +192,7 @@ def run(background=true)
 
       @thread_pool = ThreadPool.new(@min_threads,
                                     @max_threads,
+                                    Proc.new { signal_all_worker_threads_free },
```
nateberkopec: Why did you choose to pass a Proc here instead of just passing the mutex?

lib/puma/thread_pool.rb
```diff
@@ -116,6 +117,10 @@ def spawn_thread
               end
 
               @waiting += 1
+              busy_threads = @spawned - @waiting
+              if busy_threads == 0
```
nateberkopec: Let's extract these 4 lines into a private method, called `signal_threadpool_empty`.

nateberkopec:
I'm wondering about how we could test this.

lib/puma/server.rb
GuiTeK: Because I needed to pass at least 2 args: the mutex + the boolean itself. And this way the `ThreadPool` doesn't need to know what the `Server` is doing, it just tells "_hey, all my workers are free, do whatever you want_".

lib/puma/thread_pool.rb
GuiTeK: Done in https://github.com/puma/puma/commit/64c98c33796253a52bb5a600bf2e4ac15fcd587d

GuiTeK:
> I'm wondering about how we could test this.

I don't know either, it seems there were no tests at all for the OOB hook.

nateberkopec:
Thanks for your contribution. #2218 has tests and is a little cleaner IMO, so I'm going to close this PR.

PR #2181, opened 2020-03-13
Introduce Netty OpenSSL engine

fzakaria:
### Description

Add the ability for MiniSSL to try and create Netty's SSLEngine
using OpenSSL bindings through https://netty.io/wiki/forked-tomcat-native.html

It provides much better performance than Java's default SSL engine
as outlined in https://netty.io/wiki/requirements-for-4.x.html#benefits-of-using-openssl

> Speed: In local testing, we've seen performance improvements of 3x over the JDK. GCM, which is used by the only cipher suite required by the HTTP/2 RFC, is 10-500x faster.

The fallback is kept if the proper Netty libraries are not found on
the classpath.

This pull-request also adds some missing _constants_ & _methods_ that were needed on MIniSSL in order for the tests to pass successfully.

#### JDK Engine

```
  2 threads and 4 connections
    Latency     3.58ms    2.47ms  87.61ms   98.72%
    Req/Sec   580.07     62.48   666.00     90.30%
  Latency Distribution
     50%    3.31ms
     75%    3.49ms
     90%    3.85ms
     99%    6.44ms
  34553 requests in 30.10s, 6.49GB read
Requests/sec:   1147.91
Transfer/sec:    220.86MB
```

#### OpenSSL engine

```
    Latency    14.71ms   27.64ms 149.69ms   83.57%
    Req/Sec     2.54k     1.01k    4.97k    69.00%
  Latency Distribution
     50%  477.00us
     75%   11.29ms
     90%   65.76ms
     99%   98.47ms
  151625 requests in 30.07s, 28.49GB read
Requests/sec:   5042.61
Transfer/sec:      0.95GB
```

### Detailed Information

This is a little information for Ruby-ists that might not be familiar with Java/JRuby.
If you are, feel free to skip ahead to the pull request!

Java is a compiled language; meaning that any _Classes_ used must resolve at the moment of compilation -- the end result is a ".class" file. The classes which were present at compilation must also be present at Runtime if you'd like to resolve them through discovery via the Classpath (LOAD_PATH for Java). If they are not found, you are presented with a "ClassNotFoundException".

This runtime mechanism allows code to be compiled against an API but at runtime swap out the implementation or perhaps take a different course of action if it is not included.

This pull-request takes this approach to try and _upgrade_ to using Netty's OpenSSL if it's discovered (i.e. resolves) on the Classpath; if not it degrades back to using Java's Native SSL engine.

As long as the API is stable, other versions at runtime for Netty could be used aside from the version that was used to compile the JAR.

### DIFF
--- a/JRUBY.md
+++ b/JRUBY.md
@@ -32,3 +32,10 @@ however they boil down to:
 
 Reading the [documentation](https://netty.io/wiki/forked-tomcat-native.html) is a great way to make sure the requirements
 are set.
+
+## Development
+
+### Logging
+
+If you'd like to see finer log details, consider changing _logging.properties_
+to change the default log level to **FINE**
--- a/JRUBY.md
+++ b/JRUBY.md
@@ -1,0 +1,34 @@
+# JRuby
+
+Puma supports JRuby and has a few additional _goodies_ depending on what is available on the classpath.
+
+## Netty OpenSSL
+
+For those interested in improving the performance of the SSLEngine; Puma has support to leverage Netty's
+OpenSSL native bindings through [netty-tcnative](https://netty.io/wiki/forked-tomcat-native.html).
+
+> Speed: In local testing, we've seen performance improvements of 3x over the JDK.
+> GCM, which is used by the only cipher suite required by the HTTP/2 RFC, is 10-500x faster.
+
+If all the requirments have been met (see below); you can enable Netty OpenSSL via the system property **puma.ssl.use-netty**
+
+```
+ENV_JAVA["puma.ssl.use-netty"] = true
+or
+-J-Dpuma.ssl.use-netty=true
+```
+
+Currently, Puma has been compiled against version **4.1.47.Final** of Netty; however it is likely that other
+versions of Netty 4.X will function.
+
+### Requirements
+
+Requirements for the use of OpenSSL can be found [here](https://netty.io/wiki/requirements-for-4.x.html#tls-with-openssl)
+however they boil down to:
+
+1. OpenSSL version >= 1.0.2 for ALPN support, or version >= 1.0.1 for NPN.
+2. netty-tcnative version >= 1.1.33.Fork7 must be on classpath.
+3. Supported platforms: linux-x86_64, mac-x86_64, windows-x86_64. Supporting other platforms will require manually building netty-tcnative.
+
+Reading the [documentation](https://netty.io/wiki/forked-tomcat-native.html) is a great way to make sure the requirements
+are set.
--- a/Rakefile
+++ b/Rakefile
@@ -51,8 +51,11 @@ else
   task :java => :maven_dependencies
   # create a task which will download the dependencies if they do not exist already in the working directory
   task :maven_dependencies do
-    %w(https://repo1.maven.org/maven2/io/netty/netty-buffer/4.1.47.Final/netty-buffer-4.1.47.Final.jar
-       https://repo1.maven.org/maven2/io/netty/netty-handler/4.1.47.Final/netty-handler-4.1.47.Final.jar).each do |dependency|
+    %w(https://repo1.maven.org/maven2/io/netty/netty-common/4.1.47.Final/netty-common-4.1.47.Final.jar
+       https://repo1.maven.org/maven2/io/netty/netty-codec/4.1.47.Final/netty-codec-4.1.47.Final.jar
+       https://repo1.maven.org/maven2/io/netty/netty-buffer/4.1.47.Final/netty-buffer-4.1.47.Final.jar
+       https://repo1.maven.org/maven2/io/netty/netty-handler/4.1.47.Final/netty-handler-4.1.47.Final.jar
+       https://repo1.maven.org/maven2/io/netty/netty-tcnative-boringssl-static/2.0.29.Final/netty-tcnative-boringssl-static-2.0.29.Final.jar).each do |dependency|
       filename = URI(dependency).path.split('/').last
       next if File.exists?(filename)
       puts URI(dependency)
@@ -81,7 +84,9 @@ file "lib/puma/puma_http11.rb" do |t|
   end
 end
 
-Rake::TestTask.new(:test)
+Rake::TestTask.new(:test) do |t|
+  t.ruby_opts = ["-J-Djava.util.logging.config.file=logging.properties -J-Dpuma.ssl.use-netty=true"]
+end
 
 # tests require extension be compiled, but depend on the platform
 if Puma.jruby?
--- a/Rakefile
+++ b/Rakefile
@@ -48,7 +48,7 @@ if !Puma.jruby?
     CLEAN.include "lib/puma/puma_http11.rb"
   end
 else
-  task :java => :maven_dependencies
+  task :compile => :maven_dependencies
   # create a task which will download the dependencies if they do not exist already in the working directory
   task :maven_dependencies do
     %w(https://repo1.maven.org/maven2/io/netty/netty-common/4.1.47.Final/netty-common-4.1.47.Final.jar
--- a/Rakefile
+++ b/Rakefile
@@ -84,7 +84,10 @@ file "lib/puma/puma_http11.rb" do |t|
 end
 
 Rake::TestTask.new(:test) do |t|
-  t.ruby_opts = ["-J-Djava.util.logging.config.file=logging.properties -J-Dpuma.ssl.use-netty=true"]
+  # if we are testing on JRuby; let's enable Netty OpenSSL for MiniSSL
+  # this is controlled via the system property below.
+  # We also set the logging.properties file; so that we can tune Java Logging
+  t.ruby_opts = ["-J-Djava.util.logging.config.file=logging.properties -J-Dpuma.ssl.use-netty=true"] if Puma.jruby?
 end
 
 # tests require extension be compiled, but depend on the platform
--- a/Rakefile
+++ b/Rakefile
@@ -49,6 +49,7 @@ else
   # Java (JRuby)
   Rake::JavaExtensionTask.new("puma_http11", gemspec) do |ext|
     ext.lib_dir = "lib/puma"
+    ext.classpath = "netty-handler-4.1.47.Final.jar:netty-buffer-4.1.47.Final.jar"
   end
 end
 
--- a/Rakefile
+++ b/Rakefile
@@ -7,7 +7,7 @@ require_relative 'lib/puma/detect'
 require 'rubygems/package_task'
 require 'bundler/gem_tasks'
 require "uri"
-require 'net/http'
+require "net/http"
 
 gemspec = Gem::Specification.load("puma.gemspec")
 Gem::PackageTask.new(gemspec).define
--- a/Rakefile
+++ b/Rakefile
@@ -8,6 +8,7 @@ require 'rubygems/package_task'
 require 'bundler/gem_tasks'
 require "uri"
 require "net/http"
+require "fileutils"
 
 gemspec = Gem::Specification.load("puma.gemspec")
 Gem::PackageTask.new(gemspec).define
@@ -48,7 +49,9 @@ if !Puma.jruby?
     CLEAN.include "lib/puma/puma_http11.rb"
   end
 else
+  task :java => :maven_dependencies
   task :compile => :maven_dependencies
+
   # create a task which will download the dependencies if they do not exist already in the working directory
   task :maven_dependencies do
     %w(https://repo1.maven.org/maven2/io/netty/netty-common/4.1.47.Final/netty-common-4.1.47.Final.jar
@@ -56,8 +59,12 @@ else
        https://repo1.maven.org/maven2/io/netty/netty-buffer/4.1.47.Final/netty-buffer-4.1.47.Final.jar
        https://repo1.maven.org/maven2/io/netty/netty-handler/4.1.47.Final/netty-handler-4.1.47.Final.jar
        https://repo1.maven.org/maven2/io/netty/netty-tcnative-boringssl-static/2.0.29.Final/netty-tcnative-boringssl-static-2.0.29.Final.jar).each do |dependency|
-      filename = URI(dependency).path.split('/').last
+      jar_filename = URI(dependency).path.split('/').last
+      filename = "tmp/java/puma_http11/#{jar_filename}"
       next if File.exists?(filename)
+
+      # make sure the directory exists
+      FileUtils.mkdir_p "tmp/java/puma_http11"
       response = Net::HTTP.get_response(URI(dependency))
       open(filename, "wb") { |file|
         file.write(response.body)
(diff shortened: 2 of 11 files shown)


Rakefile
```diff
@@ -49,6 +49,7 @@ else
   # Java (JRuby)
   Rake::JavaExtensionTask.new("puma_http11", gemspec) do |ext|
     ext.lib_dir = "lib/puma"
+    ext.classpath = "netty-handler-4.1.47.Final.jar:netty-buffer-4.1.47.Final.jar"
```
fzakaria: This was just added so that compilation through the _JavaExtensionTask_ succeeds.
The documentation should outline which version of Netty it is compiled against.
Luckily, Netty is very aggressive about keeping API compatibility -- so I suspect all 4.X to work.

ext/puma_http11/org/jruby/puma/MiniSSL.java
```diff
@@ -196,6 +192,61 @@ public IRubyObject initialize(ThreadContext threadContext, IRubyObject miniSSLCo
...
+   * than the JDK native one.
+   * Fall back to the default engine otherwise.
+   */
+  private static SSLEngine tryCreateEngine(KeyManagerFactory keyManagerFactory,
+                                               TrustManagerFactory trustManagerFactory,
+                                               String[] protocols,
+                                               long verify_mode,
+                                               String[] ciphers) throws NoSuchAlgorithmException, KeyManagementException {
+    try {
+      io.netty.handler.ssl.ClientAuth clientAuth = io.netty.handler.ssl.ClientAuth.NONE;
```
fzakaria: If Netty libraries are not available this will throw a _ClassNotFoundException_ and default back to the original implementation.

ext/puma_http11/org/jruby/puma/MiniSSL.java
```diff
@@ -196,6 +192,61 @@ public IRubyObject initialize(ThreadContext threadContext, IRubyObject miniSSLCo
...
+      if ((verify_mode & 0x1) != 0) {
+        clientAuth = io.netty.handler.ssl.ClientAuth.OPTIONAL;
+      }
+      if ((verify_mode & 0x2) != 0) {
```
fzakaria: https://github.com/netty/netty/blob/2f32e0b8adb63decd9031e26fa5dd4154d93ce97/handler/src/main/java/io/netty/handler/ssl/JdkSslContext.java#L346
Are the associated mappings to verify.

nateberkopec:
Thanks for your contribution 🙇 

If any other JRuby users are watching, would love your thoughts as well.

fzakaria:
Tagging @headius but not sure if he has the bandwidth;

Rakefile
nateberkopec: Hmmm, any way we can not point directly to an exact version here? Would be better if could just specify netty 4, as you suggested.

ext/puma_http11/org/jruby/puma/MiniSSL.java
nateberkopec: Does this work? In CI, the build simply fails https://travis-ci.org/github/puma/puma/jobs/661773151#L326

nateberkopec:
Also, do we need to make this configurable? I don't really have an opinion, but just noticing here that we're using Netty _as long as it exists_, which may not be appropriate for a security-related feature.

Rakefile
fzakaria: Hey @nateberkopec ; not sure how familiar you are with Java. When compiling the JAR for distribution you have to compile the Java files into Class files.
In order to achieve this, the Netty library usage within MiniSSL.java needs to be resolvable at compile time.
At **runtime** however, the dependency can be omitted at which point the JVM throws an exception (ClassNotFoundException).

This classpath library here is not analogous to a Gemfile listing a version range; so there is no equivalent to "~> 4.0"; it has to be an actual JAR to build against.

ext/puma_http11/org/jruby/puma/MiniSSL.java
fzakaria: Looks like the JARs were not included in the PR; I will add them.

Rakefile
nateberkopec: Very unfamiliar with Java. I'm also going to have to figure out how this works with our rake-compile and release flow.

JRUBY.md
```diff
@@ -0,0 +1,41 @@
...
+
+If all the requirments have been met (see below); you can enable Netty OpenSSL via the system property **puma.ssl.use-netty**
+
+```
+ENV_JAVA["puma.ssl.use-netty"] = true
+or
+-J-Dpuma.ssl.use-netty=true
+```
+
+Currently, Puma has been compiled against version **4.1.47.Final** of Netty; however it is likely that other
```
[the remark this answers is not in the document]
fzakaria: Again this is what it is "compiled against"; but Java resolves dependencies at runtime via the classpath.
Customers are free to change the version they run against.

Rakefile
```diff
@@ -46,10 +48,29 @@ if !Puma.jruby?
...
 else
-  # Java (JRuby)
+  task :java => :maven_dependencies
+  # create a task which will download the dependencies if they do not exist already in the working directory
+  task :maven_dependencies do
+    %w(https://repo1.maven.org/maven2/io/netty/netty-common/4.1.47.Final/netty-common-4.1.47.Final.jar
+       https://repo1.maven.org/maven2/io/netty/netty-codec/4.1.47.Final/netty-codec-4.1.47.Final.jar
+       https://repo1.maven.org/maven2/io/netty/netty-buffer/4.1.47.Final/netty-buffer-4.1.47.Final.jar
+       https://repo1.maven.org/maven2/io/netty/netty-handler/4.1.47.Final/netty-handler-4.1.47.Final.jar
+       https://repo1.maven.org/maven2/io/netty/netty-tcnative-boringssl-static/2.0.29.Final/netty-tcnative-boringssl-static-2.0.29.Final.jar).each do |dependency|
```
fzakaria: netty-tcnative-boringssl-static is a really useful dependency. It makes pulling in boringssl (a fork of OpenSSL) extremely easy.

> This is an uber jar that contains the statically linked libraries for all supported netty-tcnative-boringssl-static-{os_arch}. While this tends to be a fairly large jar, it greatly simplifies the getting started experience since an application need not be concerned with obtaining the correct jar for the platform.

ext/puma_http11/org/jruby/puma/MiniSSL.java
```diff
@@ -143,6 +161,7 @@ public IRubyObject initialize(ThreadContext threadContext, IRubyObject miniSSLCo
     KeyStore ks = KeyStore.getInstance(KeyStore.getDefaultType());
     KeyStore ts = KeyStore.getInstance(KeyStore.getDefaultType());
 
+    String alias = miniSSLContext.callMethod(threadContext, "alias").convertToString().asJavaString();
```
fzakaria: Consumers must now provide the _alias_ of the certificate name within the KeyStore.
So that we know which PrivateKey to pull out to setup as the TLS for the server.

```
keytool -list -v -keystore examples/puma/keystore.jks -storepass blahblah                                                    
Keystore type: jks
Keystore provider: SUN

Your keystore contains 1 entry

Alias name: mydomain
Creation date: Aug 23, 2012
Entry type: PrivateKeyEntry
```

NB* I would have preferred to avoid KeyStore entirely and work directly with PEM files; but I find this to be less disruptive.

ext/puma_http11/org/jruby/puma/MiniSSL.java
```diff
@@ -47,6 +65,36 @@ public static void createMiniSSL(Ruby runtime) {
...
+        ssl.defineConstant(
+                "OPENSSL_VERSION",
+                runtime.newString(io.netty.handler.ssl.OpenSsl.versionString())
+        );
+      } catch (Throwable t) {
+        LOGGER.log(Level.INFO, "Failed to use Netty OpenSSL " + t.getMessage(), t);
+        ssl.defineConstant("OPENSSL_LIBRARY_VERSION", runtime.newString("Unknown"));
+        ssl.defineConstant("OPENSSL_VERSION", runtime.newString("Unknown"));
+      }
+    }
```
fzakaria: These constants were missing which would actually cause the test file **test_puma_server_ssl.rb** to actually skip itself for JRuby because _DISABLE_SSL_ was true due to the missing constants.

We now have test coverage again for JRuby + SSL !
🍾 🎆 

> rake test TEST=test/test_puma_server_ssl.rb
> Fabulous run in 0.895517s, 12.2834 runs/s, 10.0501 assertions/s.
> 11 runs, 9 assertions, 0 failures, 0 errors, 2 skips

nateberkopec:
@fzakaria Can you rebase? I moved the JRuby build to Github Actions.

JRUBY.md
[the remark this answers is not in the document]
nateberkopec: Is there really no better way in Java to deal with this? "We compiled against X.Y.Z but other versions _may_ work" is not great.

Rakefile
```diff
@@ -46,10 +48,29 @@ if !Puma.jruby?
...
+      filename = URI(dependency).path.split('/').last
+      next if File.exists?(filename)
+      response = Net::HTTP.get_response(URI(dependency))
+      open(filename, "wb") { |file|
```
nateberkopec: It would be better if we wrote this to the `pkg` directory (which you'll have to make if it doesnt exist ofc).

Rakefile
nateberkopec: I'm confused on why we have to add it?

ext/puma_http11/org/jruby/puma/MiniSSL.java
nateberkopec: Thanks for adding this!

logging.properties
```diff
@@ -0,0 +1,4 @@
+handlers= java.util.logging.ConsoleHandler
```
nateberkopec: Why does this file have to be added?

test/test_puma_server_ssl.rb
```diff
@@ -55,6 +57,7 @@ def start_server
     if Puma.jruby?
       ctx.keystore =  File.expand_path "../../examples/puma/keystore.jks", __FILE__
       ctx.keystore_pass = 'blahblah'
+      ctx.alias = 'mydomain'
```
nateberkopec: I think we should add something about this in the HISTORY ("`alias must be provided`")

nateberkopec:
Thanks for your patience - your explanations are really helping me out @fzakaria!

JRUBY.md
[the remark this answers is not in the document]
fzakaria: It's actually very similar to C in that regard.

For C, the macro _OPENSSL_VERSION_NUMBER_ defines the library Puma was compiled against and _OpenSSL_version(OPENSSL_VERSION)_ is the runtime call to check the version it was loaded with at runtime.

it's a side effect of dynamic libraries.

Rakefile
fzakaria: I supply this to _RakeTestTask_ so that the Puma SSL tests on JRuby can easily use OpenSSL without any setup needed for the environment.

The alternative is to dynamically link against OpenSSL on the system but that's more fragile than statically compiling in BoringSSL directly.

NB: BoringSSL is just a slimmed down fork of OpenSSL by Google.

logging.properties
fzakaria: MiniSSL.java utilizes some logging methods via JUL (Java Utility Logging).
I found it useful for testing to include this properties file so I can turn on more finer log statements as I developed this change.

I describe it in the _JRUBY.md_. It would not be distributed with Puma and is merely a development productivity file.

fzakaria:
@nateberkopec thanks for all the great feedback.

1. I've rebased with the latest changes
2. I've reworked the code so there's no need for _alias_ any longer.
3. I've included a benchmark! 🔥  (please see update description)

But the numbers are impressive at least for these 100kB documents.

**OpenSSL** vs **JDK Engine**
477.00us **vs** 3.31ms
151625 requests in 30.07s, 28.49GB read **vs** 34553 requests in 30.10s, 6.49GB read
5042.61 requests/sec **vs** 1147.91 requests/sec
0.95GB Transfer/sec **vs** 220.86MB transfer/sec

Rakefile
fzakaria: wrote it to the _tmp/java/puma_http11_ directory

olleolleolle:
Noted that the GitHub Actions builds for JRuby have the failure described in https://github.com/jruby/jruby-openssl/issues/172 in them:

> Calling SSLContext#min_version= or max_version= results in call to undefined method set_minmax_proto_version

MSP-Greg:
Re the `SSLContext#min_version=` issues, see PR https://github.com/puma/puma/pull/2186. Came across this issue in another repo.

JRuby is using much of the MRI ruby code for SSLContext, and the ruby code calls a  private method ([`SSLContext#set_minmax_proto_version`](https://msp-greg.github.io/ruby_trunk/openssl/OpenSSL/SSL/SSLContext.html#set_minmax_proto_version-instance_method)) that is a c method/function, which doesn't exist in JRuby.  That method calls an OpenSSL method that was added in 1.1.0.

Not sure if the fix will work with Netty...

fzakaria:
I will kick off the test again with the rebase.

MSP-Greg:
@fzakaria

I just looked at BoringSSL's [ssl.h](https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html), and there are differences between it and OpenSSL as to the min/max protocol selection API.  Not sure how that affects JRuby's min/max methods, or the PR I did...

fzakaria:
Looking at the remaining failure; I don't think it's related to the change but the test itself is flaky

The test does:
```ruby
system "ruby -rrubygems -Ilib bin/pumactl -F test/shell/t2_conf.rb start &"
sleep 5
system "curl http://localhost:10103/"
```
so it just waits 5 seconds before trying to curl.

However we can see in the output of the GitHub action that it took longer than 5 seconds.
You can see the startup message occurs after the cURL
```
curl: (7) Failed to connect to localhost port 10103: Connection refused
* Version 4.3.3 (jruby 9.2.11.0 - ruby 2.5.7), codename: Mysterious Traveller
* Environment: development
* Listening on tcp://[::]:10103
``

logging.properties
kares: are you sure `LOGGER.info("Using JDK SSL engine");` won't get distributed?
instead of these you might consider not using JUL and simply using (J)Ruby's `debug/verbose` flag.
or a system property `-Dpuma.ssl.debug=true` to turn on simple (std-out) logging.

nateberkopec:
The integration tests that run with the script file are super flaky. There's an open issue to rewrite those into proper tests. They're relics from before we had a good integration test suite.

nateberkopec:
I think for now, we can't merge. Given the discussion on #2188, we should first attempt a stdlib implementation, which then may lead to changes in JRuby itself on SSL optimization. 

We may come back to this after that.

Thanks for all of your hard work and fast responses on this PR 🙏 

I've brought in a similar benchmark to what you wrote to master. Still working on it.

PR #2182, opened 2020-03-13
Add customizable error messages on thread shutdown

zanker:
### Description
When threads are forced to shutdown, the response is static and cannot be configured. For our use case, we want to always guarantee a JSON response is returned on the rare case this comes up.

I didn't make this a proc like `lowlevel_error_handler`, since the intent is to be shutting down quickly, a static response felt more appropriate. While I was here, I also added a test for the base case (default message) as I couldn't find any tests which exercised that.

### DIFF
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -539,6 +539,22 @@ module Puma
       @options[:lowlevel_error_handler] = obj
     end
 
+    # When force_shutdown_after is set, this allows you to control the message returned
+    # on a force shutdown to the client.
+    #
+    # @example
+    #   force_shutdown_error_response(500, {"Content-Type" => "application/json"}, [JSON.generate({message: "Server shutdown."})])
+    def force_shutdown_error_response(status, headers, response)
+      raise "Headers must be a hash" unless headers.is_a?(Hash)
+      raise "Response must be enumerable" unless response.is_a?(Enumerable)
+
+      @options[:force_shutdown_error_response] = [
+        Integer(status),
+        headers,
+        response
+      ]
+    end
+
     # This option is used to allow your app and its gems to be
     # properly reloaded when not using preload.
     #
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -545,8 +545,8 @@ module Puma
     # @example
     #   force_shutdown_error_response(500, {"Content-Type" => "application/json"}, [JSON.generate({message: "Server shutdown."})])
     def force_shutdown_error_response(status, headers, response)
-      raise "Headers must be a hash" unless headers.is_a?(Hash)
-      raise "Response must be enumerable" unless response.is_a?(Enumerable)
+      raise "Not a Rack-compatible header (must respond to each)" unless headers.respond_to?(:each)
+      raise "Not a Rack-compatible response (must respond to each)" unless response.respond_to?(:each)
 
       @options[:force_shutdown_error_response] = [
         Integer(status),
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -593,9 +593,13 @@ module Puma
           @events.log "Detected force shutdown of a thread, returning 503"
           @events.unknown_error self, e, "Rack app"
 
-          status = 503
-          headers = {}
-          res_body = ["Request was internally terminated early\n"]
+          if @options[:force_shutdown_error_response]
+            status, headers, res_body = @options[:force_shutdown_error_response]
+          else
+            status = 503
+            headers = {}
+            res_body = ["Request was internally terminated early\n"]
+          end
 
         rescue Exception => e
           @events.unknown_error self, e, "Rack app", env
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -593,14 +593,7 @@ module Puma
           @events.log "Detected force shutdown of a thread, returning 503"
           @events.unknown_error self, e, "Rack app"
 
-          if @options[:force_shutdown_error_response]
-            status, headers, res_body = @options[:force_shutdown_error_response]
-          else
-            status = 503
-            headers = {}
-            res_body = ["Request was internally terminated early\n"]
-          end
-
+          status, headers, res_body = @options[:force_shutdown_error_response]
         rescue Exception => e
           @events.unknown_error self, e, "Rack app", env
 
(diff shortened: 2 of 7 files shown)


lib/puma/dsl.rb
```diff
@@ -539,6 +539,22 @@ def lowlevel_error_handler(obj=nil, &block)
...
     end
 
+    # When force_shutdown_after is set, this allows you to control the message returned
+    # on a force shutdown to the client.
+    #
+    # @example
+    #   force_shutdown_error_response(500, {"Content-Type" => "application/json"}, [JSON.generate({message: "Server shutdown."})])
+    def force_shutdown_error_response(status, headers, response)
+      raise "Headers must be a hash" unless headers.is_a?(Hash)
+      raise "Response must be enumerable" unless response.is_a?(Enumerable)
```
nateberkopec: Actually, it doesn't have to be an Enumerable. It just has to respond to `each`.

lib/puma/dsl.rb
```diff
@@ -539,6 +539,22 @@ def lowlevel_error_handler(obj=nil, &block)
       @options[:lowlevel_error_handler] = obj
     end
 
+    # When force_shutdown_after is set, this allows you to control the message returned
+    # on a force shutdown to the client.
+    #
+    # @example
+    #   force_shutdown_error_response(500, {"Content-Type" => "application/json"}, [JSON.generate({message: "Server shutdown."})])
+    def force_shutdown_error_response(status, headers, response)
+      raise "Headers must be a hash" unless headers.is_a?(Hash)
```
nateberkopec: Same here. It just has to respond to each, it doesn't have to be a hash.

nateberkopec: Maybe `raise "Not a Rack-compatible response" unless [headers, response].all? {|x| x.respond_to?(:each)}"`

lib/puma/server.rb
```diff
@@ -593,9 +593,13 @@ def handle_request(req, lines)
           @events.log "Detected force shutdown of a thread, returning 503"
           @events.unknown_error self, e, "Rack app"
 
-          status = 503
-          headers = {}
-          res_body = ["Request was internally terminated early\n"]
+          if @options[:force_shutdown_error_response]
+            status, headers, res_body = @options[:force_shutdown_error_response]
+          else
+            status = 503
```
nateberkopec: Let's set this as the default value of the `@options`, not hard-coded here.

nateberkopec:
Provided some code feedback but I'm not 100% sure on the usecase.

"we want to always guarantee a JSON response" -> Why? Isn't the 503 status code enough here?

nateberkopec:
Then, if we decide that yes, it's important that Puma returns a particular content-type on low-level failures, I think instead we should change all low-level errors. There's a lot of them and this only configures one, which isn't that helpful.

Something like:

```ruby
lowlevel_error_content_type = :json
```

Maybe we can support :json or :none for now.

Existing sites where we render lowlevel errors would be changed to look something like:

```ruby
if @leak_stack_on_error
    generate_error_response(status: 500, body: "Puma caught this error: #{e.message} (#{e.class})\n#{e.backtrace.join("\n")}")
else
  generate_error_reponse(status: 500, body: "An unhandled lowlevel error occurred. The application logs may have details").
end
```

dentarg:
> Provided some code feedback but I'm not 100% sure on the usecase.
> 
> "we want to always guarantee a JSON response" -> Why? Isn't the 503 status code enough here?

I can understand the use-case, if you are building an API, and you document that your API return responses in JSON, it is unexpected when the response isn't in JSON. Maybe you want to include more information than the HTTP status code, e.g. your application can return HTTP status 503 for number of different errors, and you want to provide a certain error code in the error response. Then it makes sense to return the body in JSON, including that information.

lib/puma/dsl.rb
zanker: Ah interesting, I didn't realize Rack solely cared about `each` for the headers too. I went with:

```ruby
      raise "Not a Rack-compatible header (must respond to each)" unless headers.respond_to?(:each)
      raise "Not a Rack-compatible response (must respond to each)" unless response.respond_to?(:each)
```

To err on the side of being a bit more specific/helpful on errors.

zanker:
@dentarg has it right. Returning static JSON of Puma's choosing would be an improvement over static plaintext, but it's not compatible with a JSON API if the API expects `{error: true, msg: "foo bar"}` and Puma returns `{message: "foo bar"}`.

I think it's ok if `lowlevel_error` has a static plaintext response, since you can always override it to be JSON, but I don't have strong feelings.

I made the changes requested, let me know if you have anymore feedback!

nateberkopec:
Cool, ppl on Twitter agreed this is useful so we'll put it in. I need to pull it down and play w/implementation to see if there's a better way to integrate it w/the rest of our error handling.

zanker:
Sounds good! Thanks for the fast response

zanker-stripe:
Just to double check, I assume you don't need anything from me at this point since it's approved. You're just waiting on having time to play around with alternatives?

nateberkopec:
@zanker-stripe yup, thanks!

nateberkopec:
OK, so I took another look and here's what I think needs to happen:

1. Thread shutdown is getting a special-case path and avoiding lowlevel_error_handler. It should not do that and use lowlevel_error_handler instead.
2. To do that, we need lowlevel_error_handler to get a new argument, status, which should default to 500.
3. To deal with the new argument, lowlevel_error_handler needs a case to deal with an arity == 3 handler.

```ruby
def lowlevel_error(e, env, status = 500)
      if handler = @options[:lowlevel_error_handler]
        if handler.arity == 1
          return handler.call(e)
        elsif handler.arity == 2
          return handler.call(e, env)
        elsif handler.arity == 3
          return handler.call(e, env, status)
        end
      end
    end
```

and the default config should have the following:

```ruby
  DefaultLowlevelErrorHandler = lambda do |_e, _env, status|
      [status, {}, ["An unhandled lowlevel error occurred. The application logs may have details.\n"]].freeze
    end
```

nateberkopec:
This way, if you want JSON responses, you can do that consistently for all error types.

zanker:
@nateberkopec Do you have any concerns with the fact that this means people can accidentally run slower code during exit and essentially delay the shutdown timer?

I'm fine with it if you are, and I'll do a PR for that.

nateberkopec:
I'm OK with that. We can document that this should return ASAP or the shutdown timer may be affected.

nateberkopec:
People do talk to external APIs in their lowlevel_handlers often (Sentry, Rollbar, etc) but I don't think that possibly delaying the shutdown timer by up to a second is all that critical of an issue 🤷‍♂

zanker-stripe:
Fair enough! #2203 will handle it that way.

PR #2185, opened 2020-03-15
Actions - Add nonMRI workflow - allow failure - JRuby, TruffleRuby [changelog skip]

MSP-Greg:
### Description

Add JRuby & TruffleRuby to Actions via another 'job', all are 'allow-failure'


MSP-Greg:
At present, Actions does not have an 'allow-failures' similar to Travis/AppVeyor.  Last I checked, one must still use the web UI to actually look at the steps to see whether a failure occurred, as the failure 'indication' won't propagate any further.

So, given the CI here and in my fork, one option is to have a non-MRI workflow that allows failures and runs JRuby and TruffleRuby on Unbuntu.  A separate workflow makes it easier to force a seed value for testing, limit tests, etc.  Once the CI stabilizes, then move JRuby & TruffleRuby back into the main CI.

JFYI, setup-ruby-pkgs uses the code from my fork of setup-ruby, and the selection/installation of Ruby builds is identical to setup-ruby.  If you open the step, you'll see the same output during the 'install' phase...

nateberkopec:
@MSP-Greg But a separate workflow would still have the same issue of requiring us to view the output to see if it failed or not, right?

MSP-Greg:
@nateberkopec

Yes.  The main reason for a separate workflow (could be in the same file) is it would not need logic to determine which jobs failed.

nateberkopec:
Makes sense. Let's do it. Also for now, I think we should be on truffleruby-head because they're fixing a few issues re: frozen strings that our build is surfacing.

MSP-Greg:
> truffleruby-head

Will do, almost done.

eregon:
Agreed, for now there is no point to test with `truffleruby`, just test with `truffleruby-head`.

MSP-Greg:
Any thoughts on Jruby, ie `jruby-9.2` or `jruby-head`?

.github/workflows/ruby.yml
```diff
@@ -40,10 +40,10 @@ jobs:
         run:  bundle exec rake compile
 
       - name: test
-        timeout-minutes: 10
+        timeout-minutes: 8
         run:  |
           # RubyGems < 3, no frozen strings
-          if [[ "2.6" < "${{ matrix.ruby }}.0" ]]; then
+          if [[ "2.6" < "${{ matrix.ruby }}.0" ]] && [[ "${{ matrix.ruby }}" != "truffleruby" ]]; then
```
nateberkopec: truffleruby not needed here, right?

.github/workflows/ruby.yml
```diff
@@ -85,10 +85,56 @@ jobs:
...
+    env:
+      CI: true
+      TESTOPTS: -v
+
+    runs-on: ${{ matrix.cfg.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        cfg:
+          - { os: ubuntu-latest, ruby: jruby-9.2        }
```
nateberkopec: JRuby 9.2 failure should not be allowed - that build passes and is one of our supported platforms.

nateberkopec: Personally I'd just put it in the main action and call this one nonMRIHead or something

MSP-Greg: Done

.github/workflows/ruby.yml
MSP-Greg: Removed here and line 146, as truffleruby-head is ok

.github/workflows/ruby.yml
```diff
@@ -85,10 +85,55 @@ jobs:
...
+        continue-on-error: true
+        run:  bundle exec rake compile
+
+      - name: test
+        timeout-minutes: 8
+        continue-on-error: true
+        if: success()
+        run:  |
+          # RubyGems < 3, no frozen strings
+          if [[ "2.6" < "${{ matrix.cfg.ruby }}.0" ]]; then
```
eregon: This check won't work, since there is no version in the matrix entries above

MSP-Greg: Please check the log of the 'load ruby, ragel' step, and you'll see the correct matrix.cfg.ruby version loaded from setup-ruby's log.  This is another supported way to construct a matrix.

eregon: It results in checks like:
```
# RubyGems < 3, no frozen strings
  if [[ "2.6" < "jruby-head.0" ]]; then
    export RUBYOPT="--enable=frozen-string-literal"
  fi
```
which don't make much sense (https://github.com/puma/puma/pull/2185/checks?check_run_id=509502033).

MSP-Greg: It was my understanding that jruby-head supported `--enable=frozen-string-literal`.  Is that correct?  As I thought the conditional is true...

nateberkopec:
@MSP-Greg In this PR, let's also go ahead and delete Travis.yml 🎉

.github/workflows/ruby.yml
nateberkopec: I don't really care as long it evals correctly FWIW

MSP-Greg: I'm not seeing frozen string errors...

MSP-Greg:
> delete Travis.yml

Done.  Working on getting JRuby to run ssl tests...

nateberkopec:
I'll probably open up a separate PR later to let JRuby skip those integration tests outside the minitest suite that it keeps failing on.

MSP-Greg:
I looked at getting ssl working on JRuby.  There are methods and constants that are defined in MRI Puma that aren't defined in JRuby Puma.  Below is what I found:

```ruby
Puma::MiniSSL::Engine#init?
Puma::MiniSSL::Engine#shutdown

Puma::MiniSSL::OPENSSL_VERSION
Puma::MiniSSL::OPENSSL_LIBRARY_VERSION
Puma::MiniSSL::OPENSSL_NO_SSL3
Puma::MiniSSL::OPENSSL_NO_TLS1
Puma::MiniSSL::OPENSSL_NO_TLS1_1
```

.github/workflows/ruby.yml
eregon: Since this job applies only for jruby and truffleruby, how about simply removing the conditional and always `--enable=frozen-string-literal`?

MSP-Greg:
JRuby - I've got code that allows running all the SSL tests on JRuby.

They pass, but I'd call them false positives.  Hence, maybe JRuby should be moved to the 'allow-failure' until it really passes the SSL tests?

Thoughts?

MSP-Greg:
> as it allows further regressions on JRuby

Good point.

MSP-Greg:
Rebased and made some changes.

1. Added Windows Ruby 2.2, needed minor change in extconf.rb

2. Made adjustments to JRuby so it would pass CI.  It is missing `Puma::MiniSSL::Engine#init?` and `Puma::MiniSSL::Engine#shutdown` methods, which may be affecting it's throughput.

3. I suspect that Actions may be implementing a better 'allow-failure' for jobs, and they're also changing matrix additions.  Hence, the separate job in the workflow may be able to be folded into the normal Ubuntu/macOS job in the future...

I see that the Ubuntu JRuby job failed.  Intermittent, as it passed in my fork.

nateberkopec:
Awesome work @MSP-Greg, thanks,

PR #2187, opened 2020-03-16
Fix warnings for !HAS_FORK [changelog skip]

eregon:
Skip teardown if setup was skipped in test_preserve_bundler_env

* Otherwise we get
puma/test/helpers/integration.rb:27: warning: instance variable @ios_to_close not initialized
puma/test/helpers/integration.rb:34: warning: instance variable @bind_path not initialized
See https://travis-ci.org/github/puma/puma/jobs/662767158#L469
* Same pattern as in test/test_integration_cluster.rb.

### DIFF
--- a/test/test_preserve_bundler_env.rb
+++ b/test/test_preserve_bundler_env.rb
@@ -8,6 +8,7 @@ class TestPreserveBundlerEnv < TestIntegration
   end
 
   def teardown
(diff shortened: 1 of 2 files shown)


eregon:
Thanks for merging!

nateberkopec:
Thanks for the contrib! 👏

PR #2189, opened 2020-03-18
Update Rack::Handler::Puma.run method options param

jkowens:
Rack has opted to use the double splat operator for options.

See: https://github.com/rack/rack/commit/871c82f0ffa269d7e2e237d3ade9a1eac26418fe


nateberkopec:
Thanks!

PR #1320, opened 2017-06-03
SIGINFO prints thread backtraces

nateberkopec:
Closes #1041 

I'm open to suggestions for other signals. SIGINFO is only present on *BSD.

Todo:

* [ ] Figure out how to fan out the SIGINFO from the parent to child processes (currently have to send it to each child
* [ ] Tests

PR #2195, opened 2020-03-22
Use SIGWINCH to print thread backtraces

stanhu:
### Description

https://github.com/puma/puma/pull/1320 added support for using SIGINFO,
but this is only available on BSD-based systems. SIGWINCH is on Linux.

### DIFF
--- a/History.md
+++ b/History.md
@@ -7,6 +7,7 @@
   * `GC.compact` is called before fork if available (#2093)
   * Add `requests_count` to workers stats. (#2106)
   * Increases maximum URI path length from 2048 to 8196 bytes (#2167)
+  * Sending SIGWINCH to any Puma worker now prints currently active threads and their backtraces (#2195)
 
 * Deprecations, Removals and Breaking API Changes
   * `Puma.stats` now returns a Hash instead of a JSON string (#2086)
--- a/docs/signals.md
+++ b/docs/signals.md
@@ -40,7 +40,7 @@ Puma cluster responds to these signals:
 - `USR1` restart workers in phases, a rolling restart. This will not reload configuration file.
 - `HUP`  reopen log files defined in stdout_redirect configuration parameter. If there is no stdout_redirect option provided it will behave like `INT`
 - `INT` equivalent of sending Ctrl-C to cluster. Will attempt to finish then exit.
-- `CHLD`
+- `WINCH` (`INFO` on BSD-based systems) prints a thread backtrace. This is useful for debugging infinite loops or slow performance.
 
 ## Callbacks order in case of different signals
 
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -453,17 +453,29 @@ module Puma
       end
 
       begin
+        Signal.trap "SIGWINCH" do
+          log_backtrace
+        end
+      rescue Exception
+        log "*** SIGWINCH not implemented, signal based thread backtraces unavailable!"
+      end
+
+      begin
         unless Puma.jruby? # INFO in use by JVM already
           Signal.trap "SIGINFO" do
-            thread_status do |name, backtrace|
-              @events.log name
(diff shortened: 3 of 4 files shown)


nateberkopec:
SIGPROF is already used pretty widely in profilers, e.g. [stackprof](https://github.com/tmm1/stackprof/search?q=SIGPROF&type=).

I'm not sure we can use that signal.

nateberkopec:
Just looking at the BSD signal list, maybe SIGVTALRM or SIGWINCH? [I'd like to know what people get for the `Signal.list` on other platforms, btw.](https://twitter.com/nateberkopec/status/1242344897782267904)

nateberkopec:
NVM, Ruby uses SIGVTALRM internally. SIGWINCH however probably won't interfere with anyone's existing handlers in a webapp.

stanhu:
On Linux, I get

```
{"EXIT"=>0, "HUP"=>1, "INT"=>2, "QUIT"=>3, "ILL"=>4, "TRAP"=>5, "ABRT"=>6, "IOT"=>6, "FPE"=>8, "KILL"=>9, "BUS"=>7, "SEGV"=>11, "SYS"=>31, "PIPE"=>13, "ALRM"=>14, "TERM"=>15, "URG"=>23, "STOP"=>19, "TSTP"=>20, "CONT"=>18, "CHLD"=>17, "CLD"=>17, "TTIN"=>21, "TTOU"=>22, "IO"=>29, "XCPU"=>24, "XFSZ"=>25, "VTALRM"=>26, "PROF"=>27, "WINCH"=>28, "USR1"=>10, "USR2"=>12, "PWR"=>30, "POLL"=>29}
```

nateberkopec:
Sounds like SIGWINCH is probably the best option then, yeh?

stanhu:
Yes, thanks.

I've also wondered why attaching `gdb` to Puma and calling `call rb_backtrace()` doesn't work like it does in Unicorn and Sidekiq. Is th

nateberkopec:
> I've also wondered why attaching gdb to Puma and calling call rb_backtrace() doesn't work like it does in Unicorn and Sidekiq.

Can you open a new issue?

nateberkopec:
I think we're going to have to revert. SIGWINCH gets sent all the time when changing window sizes and so this makes terminal output kind of unusable locally.

I'm not sure any other signal would work.

stanhu:
Ok. Is it possible to use `TTIN` as Sidekiq does, and then make it clear that signal does two different things, depending on whether it's sent to the worker process or the cluster worker?

nateberkopec:
Maybe. Do we listen for TTIN in child workers at all right now?

stanhu:
@nateberkopec Not at the moment. I think I got it to work. It would be a breaking change for anyone issuing `TTIN` on the cluster workers, though.

eregon:
FWIW, TruffleRuby prints all thread backtraces on `SIGALRM`.
Maybe we should ask for this feature in MRI instead of adding it here?

dentarg:
@eregon Nate commented "Ruby uses SIGVTALRM internally" above, but I'm not sure for what (from a [really quick look](https://github.com/ruby/ruby/search?q=ALRM&unscoped_q=ALRM))

eregon:
SIGALRM is different than SIGVTALRM (which is reserved in MRI & TruffleRuby).

See https://github.com/ruby/spec/blob/ebdf0d1b06b39374a85baaae53771449ab1fe3d7/core/signal/trap_spec.rb#L118-L133 for the list of reserved signals for each Ruby implementation.

dentarg:
Thanks for clarifying that @eregon 

(I should have search for [`SIGVTALRM`](https://github.com/ruby/ruby/search?q=SIGVTALRM&unscoped_q=SIGVTALRM) not just `ALRM`, the GitHub search doesn't seem to surface partial hits)

PR #2198, opened 2020-03-23
Run tests on TruffleRuby, all tests pass now

eregon:
### Description

This PR improves the reliability of tests and as a result all tests now pass on TruffleRuby.
One extra fix was needed in TruffleRuby (https://github.com/oracle/truffleruby/commit/2e046e4400c10c3ed19d2bd14e75741d21d54931).
A major part of getting TruffleRuby to pass Puma's tests was by done by Shopify contributors to TruffleRuby (especially the signal stuff).

Since tests pass on TruffleRuby, I also moved TruffleRuby along with other tested versions in CI, so failures are no longer ignored.
I ran the CI twice in GitHub Actions and it passed in both cases.

cc @chrisseaton @nateberkopec 

### DIFF
--- a/lib/puma/minissl.rb
+++ b/lib/puma/minissl.rb
@@ -125,11 +125,14 @@ module Puma
 
       def read_and_drop(timeout = 1)
         return :timeout unless IO.select([@socket], nil, nil, timeout)
-        return :eof unless read_nonblock(1024)
-        :drop
-      rescue Errno::EAGAIN
-        # do nothing
-        :eagain
+        case @socket.read_nonblock(1024, exception: false)
+        when nil
+          :eof
+        when :wait_readable
+          :eagain
+        else
+          :drop
+        end
       end
 
       def should_drop_bytes?
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -43,15 +43,10 @@ def hit(uris)
 end
 
 module UniquePort
-  @port  = 3211
-  @mutex = Mutex.new
-
   def self.call
-    @mutex.synchronize {
-      @port += 1
-      @port = 3307 if @port == 3306  # MySQL on Actions
-      @port
-    }
+    TCPServer.open('127.0.0.1', 0) do |server|
+      server.connect_address.ip_port
+    end
   end
 end
 
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -57,7 +57,7 @@ module TimeoutEveryTestCase
   end
 
   def run(*)
-    ::Timeout.timeout(Puma.jruby? ? 120 : 60, TestTookTooLong) { super }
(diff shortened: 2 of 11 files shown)


Rakefile
```diff
@@ -86,4 +86,4 @@ namespace :test do
   end
 end
 
-task :default => [:rubocop, "test:all"]
+task :default => [*(:rubocop if RUBY_ENGINE == "ruby"), "test:all"]
```
nateberkopec: What does RUBY_ENGINE return on TruffleRuby/JRuby?

I guess this is fine, maybe a bit annoying though if you're working on Puma using JRuby and then get surprised by CI failing b/c of whitespace.

test/helper.rb
```diff
@@ -43,15 +43,10 @@ def hit(uris)
...
-  @port  = 3211
-  @mutex = Mutex.new
-
   def self.call
-    @mutex.synchronize {
-      @port += 1
-      @port = 3307 if @port == 3306  # MySQL on Actions
-      @port
-    }
+    TCPServer.open('127.0.0.1', 0) do |server|
```
nateberkopec: 👏

Rakefile
eregon: `RUBY_ENGINE` is simply `"truffleruby"` and `"jruby"` there.
I was tempted to replace `defined?(JRUBY_VERSION)` with `RUBY_ENGINE == "jruby"` in the codebase since that's the portable way to test the Ruby implementation, but avoided so in this PR to keep it smaller.

Given RuboCop runs (from what I can see locally) significantly slower (> 10s) on both of these implementations I think it's likely to be less annoying if it doesn't run on those implementations.

I could do that change only in the CI file, but I thought it's actually helpful also when running `bundle exec rake` locally.

nateberkopec: I think we should probably keep it the same for all implementations. If you personally think RC is too slow, you can still just run `rake test:all`.

eregon: OK, I'll move that check to the CI file, won't be as nice though.

eregon: Done!

nateberkopec:
Thanks so much @eregon and Shopify! 🎉

lib/puma/minissl.rb
```diff
@@ -125,11 +125,14 @@ def flush
...
-        # do nothing
-        :eagain
+        case @socket.read_nonblock(1024, exception: false)
+        when nil
+          :eof
+        when :wait_readable
+          :eagain
+        else
+          :drop
+        end
```
eregon: For the record, this might be related to a bug in OpenSSL, mentioned in https://github.com/ruby/openssl/issues/355#issuecomment-600875524
Anyway, it seems safer to not call to SSL anymore if we are in such a state, as this commit does.

PR #2203, opened 2020-03-26
Handle thread shutdown responses via low level error messages

zanker-stripe:
### Description

This is a follow on from the suggestions in https://github.com/puma/puma/pull/2182#issuecomment-603638985. I did not make the change around `DefaultLowlevelErrorHandler`, because:

https://github.com/puma/puma/blob/0b737cce42746d8226701520f35685e0624e061b/lib/puma/server.rb#L819-L823

The `leak_stack_on_error` option is stored on the server instance and not options, which leaves no way of getting that data into the handler. I wasn't a fan of passing that through, and reorganizing how `leak_stack_on_error` works also felt out of scope.

Functionality wise, there's no changes besides thread shutdown now goes through lowlevel error handler and we pass `status` to the handler based on arity. If `leak_stack_on_error` is set, we include the stack, if not we show the generic error messages. 

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -589,17 +589,15 @@ module Puma
 
             return :async
           end
+        rescue ThreadPool::ForceShutdown => e
+          @events.unknown_error self, e, "Rack app", env
+          @events.log "Detected force shutdown of a thread"
+
+          status, headers, res_body = lowlevel_error(e, env, 503)
         rescue Exception => e
           @events.unknown_error self, e, "Rack app", env
 
-          status = 500
-          if e.is_a?(ThreadPool::ForceShutdown)
-            status = 503
-
-            @events.log "Detected force shutdown of a thread, returning 503"
-          end
-
-          status, headers, res_body = lowlevel_error(e, env, status)
+          status, headers, res_body = lowlevel_error(e, env, 500)
         end
 
         content_length = nil
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -589,18 +589,17 @@ module Puma
 
             return :async
           end
-        rescue ThreadPool::ForceShutdown => e
-          @events.log "Detected force shutdown of a thread, returning 503"
-          @events.unknown_error self, e, "Rack app"
-
-          status = 503
-          headers = {}
-          res_body = ["Request was internally terminated early\n"]
-
         rescue Exception => e
           @events.unknown_error self, e, "Rack app", env
 
-          status, headers, res_body = lowlevel_error(e, env)
+          status = 500
+          if e.is_a?(ThreadPool::ForceShutdown)
+            status = 503
+
+            @events.log "Detected force shutdown of a thread, returning 503"
+          end
+
+          status, headers, res_body = lowlevel_error(e, env, status)
         end
 
         content_length = nil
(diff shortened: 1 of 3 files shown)


lib/puma/server.rb
```diff
@@ -589,18 +589,17 @@ def handle_request(req, lines)
...
-        rescue ThreadPool::ForceShutdown => e
-          @events.log "Detected force shutdown of a thread, returning 503"
-          @events.unknown_error self, e, "Rack app"
-
-          status = 503
-          headers = {}
-          res_body = ["Request was internally terminated early\n"]
-
         rescue Exception => e
           @events.unknown_error self, e, "Rack app", env
```
zanker-stripe: Technically this means we pass `env` on shutdown errors, but that seemed like more of an oversight than an initial design. Happy to unwind if I'm wrong though.

lib/puma/server.rb
```diff
@@ -589,18 +589,17 @@ def handle_request(req, lines)
...
 
-          status, headers, res_body = lowlevel_error(e, env)
+          status = 500
+          if e.is_a?(ThreadPool::ForceShutdown)
```
nateberkopec: I think this is less clear/nice looking than what it was before. Can we just keep the separate handling for ThreadPool::ForceShutdown (it's own rescue block) as it was before?

lib/puma/server.rb
```diff
@@ -807,19 +806,21 @@ def read_body(env, client, body, cl)
...
           return handler.call(e)
-        else
+        elsif handler.arity == 2
           return handler.call(e, env)
+        else
+          return handler.call(e, env, status)
         end
       end
 
       if @leak_stack_on_error
```
nateberkopec: Maybe we should just get rid of this option and tell people to write their own lowlevel_handler that leaks the stack if they prefer. Next release is 5.0 so we can break APIs.

nateberkopec: What do you think?

lib/puma/server.rb
zanker-stripe: Sure, I changed it back.

lib/puma/server.rb
zanker-stripe: In the use cases I've used Puma, we always had `lowlevel_error_handler`s and this default didn't matter. It's not as friendly for out-of-the-box handling though, when somebody is new and trying to do local development. I'd vote to keep good defaults here.

I think the ideal way of doing this would be the default config for `lowlevel_error_handler` is:

```ruby
      if @options[:environment] == "development" || @options[:environment] == "test"
        [status, {}, ["Puma caught this error: #{e.message} (#{e.class})\n#{e.backtrace.join("\n")}"]]
      else
        [status, {}, ["An unhandled lowlevel error occurred. The application logs may have details.\n"]]
      end
```

That way your custom code is in the configuration, where it belongs, and you're not threading this extra stack context for a single bit of functionality.

I'm not sure the best way to do that though, so I'd push back a little and say it's a better change to make outside of this PR.

test/test_puma_server.rb
```diff
@@ -261,6 +261,36 @@ def test_doesnt_print_backtrace_in_production
     assert_match(/HTTP\/1.0 500 Internal Server Error/, data)
   end
 
+  def test_force_shutdown_custom_error_message
+    handler = lambda {|err, env, status| [500, {"Content-Type" => "application/json"}, ["{}\n"]]}
+    @server = Puma::Server.new @app, @events, {:lowlevel_error_handler => handler, :force_shutdown_after => 2}
+
+    server_run app: ->(env) do
+      @server.stop
+      sleep 10
```
nateberkopec: I think we can tighten this up to about `5`, right?

test/test_puma_server.rb
```diff
@@ -261,6 +261,36 @@ def test_doesnt_print_backtrace_in_production
...
+    assert_match(/Content-Type: application\/json/, data)
+    assert_match(/{}\n$/, data)
+  end
+
+  def test_force_shutdown_error_default
+    @server = Puma::Server.new @app, @events, {:force_shutdown_after => 2}
+
+    server_run app: ->(env) do
+      @server.stop
+      sleep 10
```
nateberkopec: Same here

nateberkopec:
1 more thing then this is g2g for me.

zanker-stripe:
Thanks! Changes made.

PR #2214, opened 2020-04-02
Fix the CI [changelog skip]

eregon:
### Description
See https://github.com/puma/puma/runs/553627446 and https://github.com/oracle/truffleruby/issues/1984

The CI fails recently due to `ruby/setup-ruby` defaulting to latest Bundler (= 2) by default.

The failures are rather unclear so I suggest to stick to Bundler 1 to keep CI green, and then investigate what goes wrong with Bundler 2.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -28,6 +28,7 @@ jobs:
         uses: MSP-Greg/setup-ruby-pkgs@v1
         with:
           ruby-version: ${{ matrix.ruby }}
+          bundler: 1
           apt-get: ragel
           brew: ragel
 
--- a/test/test_thread_pool.rb
+++ b/test/test_thread_pool.rb
@@ -87,7 +87,7 @@ class TestThreadPool < Minitest::Test
   end
 
   def test_trim_leaves_min
-    skip_on :truffleruby # Undiagnose thread race. TODO fix
+    skip_on :jruby, :truffleruby # Undiagnose thread race. TODO fix
     pool = new_pool(1, 2) do |work|
       @work_mutex.synchronize do
         @work_done.signal


eregon:
I'm unsure what canceled that job: https://github.com/puma/puma/pull/2214/checks?check_run_id=555689509
I didn't do it AFAIK.

The CI is green on the run in my fork: https://github.com/eregon/puma/actions/runs/69222499

eregon:
FYI the failures on the run on master is due to https://github.com/MSP-Greg/setup-ruby-pkgs/issues/2

MSP-Greg:
Actually, I've seen the error before, often with just a single job in many.

The problem is with http://azure.archive.ubuntu.com/ubuntu bionic/universe
  Could not connect to azure.archive.ubuntu.com:80 (52.177.174.250), connection timed out

I'll update MSP-Greg/setup-ruby-pkgs anyway.

eregon:
@nateberkopec Could you restart the jobs in https://github.com/puma/puma/runs/556273752 to see if it's fixed now?

MSP-Greg:
@eregon @nateberkopec  @dentarg

> FYI the failures on the run on master is due to

If that's the case, how would you explain the failures in:

https://github.com/puma/puma/actions/runs/69538791

`sudo apt-get -qy update` was done in all the Ubuntu tests, and they still failed.

'connection timed out' is not equivalent to a 404 or a similar error message.

Adding ~20 sec to each Ubuntu test is a waste of time.  It's one thing if `apt-get` is being used for a dependency, it's another thing if it's a build utility like `ragel`...

I'll probably revert the change.  If it's still a concern for some, one can always use
```
apt-get: _update_ ragel
```
as the README noted before the change.

I might suggest that people actually read the test logs before jumping to conclusions.

eregon:
@MSP-Greg Sorry, I thought that was the cause.

On the upside, it's clearer in that log what happens. In the previous log it wasn't (to me, at least).

I think not calling `apt-get update` is a bug, unless the azure `apt` servers do something differently than the official servers (I would think not, but would be good to check).
Otherwise, `apt-get install` might fail due to (in my understanding) Ubuntu removing older packages.
You'll notice many Dockerfiles using apt-get also do `apt-get update` before, because it might fail otherwise.
Example: https://github.com/docker-library/ruby/blob/a564feaaee4c8647c299ab11d41498468bb9af7b/2.7/buster/Dockerfile#L20-L21
And in ruby/ruby's actions too: https://github.com/ruby/ruby/blob/810d66f3e7da3a5267974e3a3b952ec3e378049b/.github/workflows/ubuntu.yml#L25-L26

`apt` is slow, I don't like it either, but that's how it is.

Re reverting, I think we should ask recommendations regarding this on the GitHub Actions forum.

dentarg:
I guess something like [`travis_retry`](https://docs.travis-ci.com/user/common-build-problems/#travis_retry) would be good to have for commands involving resources we don't have control over (like the network to Azure apt caches).

Issue #2148, opened 2020-03-05
Tests stall quite often


dleidert:
**Describe the bug**
I'm one of the Debian maintainers of puma. The 4.3 series has been uploaded to experimental. But often the build gets stuck when running the tests and must be terminated after 1,5 hours. `ps` for example shows this:

```
root       28222  0.0  0.0   2388   784 pts/1    S+   02:18   0:00 sh -c /usr/bin/ruby2.5 -w -I"test"  "/usr/lib/ruby/vendor_ruby/rake/rake_test_loader.rb" "test/test_app_status.rb" "test/test_binder.rb" "test/test_cli.rb" "test/test_config.rb" "test/test_events.rb" "test/test_http10.rb" "test/test_http11.rb" "test/test_integration_cluster.rb" "test/test_integration_pumactl.rb" "test/test_integration_single.rb" "test/test_iobuffer.rb" "test/test_launcher.rb" "test/test_minissl.rb" "test/test_null_io.rb" "test/test_persistent.rb" "test/test_plugin.rb" "test/test_puma_server.rb" "test/test_pumactl.rb" "test/test_rack_handler.rb" "test/test_rack_server.rb" "test/test_tcp_logger.rb" "test/test_tcp_rack.rb" "test/test_thread_pool.rb" "test/test_unix_socket.rb" "test/test_web_server.rb" -v
root       28223  9.6  0.2 3615992 47240 pts/1   Sl+  02:18   0:32 puma 4.3.3 (unix:///tmp/puma-test20200305-251-8k6mxn2) [tree]
```

It is not always the same test. Here are a few logs:

<https://salsa.debian.org/ruby-team/puma/-/jobs/594219>
<https://buildd.debian.org/status/package.php?p=puma&suite=experimental>
<https://buildd.debian.org/status/fetch.php?pkg=puma&arch=amd64&ver=4.3.1-1&stamp=1580998562&raw=0>
<https://buildd.debian.org/status/fetch.php?pkg=puma&arch=arm64&ver=4.3.1-1&stamp=1580998668&raw=0>

Any idea how to further track down the cause?

**To Reproduce**
I have no idea.

**Expected behavior**
Tests should run through without getting stuck.

**Desktop (please complete the following information):**
Debian Linux Sid (unstable)
puma 4.3.3

nateberkopec:
Thanks for your work and for all of your reports 👍

nateberkopec:
[...]

PR #2216, opened 2020-04-03
Set `Connection: closed` when running without request queueing

praboud-stripe:
### Description
When running with `queue_requests=false`, the server closes the connection after each request. However, the server would indicate via the `Connection` header that the connection would be persisted. This was happening because the headers got reported in `handle_request`, _before_ the check to `@queue_requests` in `process_client` forces the connection to be closed. This PR moves the check to be in `handle_request` - this ensures that what the server reports in the headers, and what it actually does with the connection can't get out of whack.

Specifically the behaviour that we want here is:
 - for HTTP/1.0, omit the `Connection` header
 - for HTTP/1.1, setting `Connection: closed`. This is required by https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html, specifically:
`HTTP/1.1 applications that do not support persistent connections MUST include the "close" connection option in every message.`

As part of doing this change, I've refactored the logic around setting the connection headers to make it easier to understand why the headers are set the way they are. I've pulled this into a separate commit (345e127, which is intended to purely be a refactor and to have no functional effect). In general, this PR is probably best reviewed commit-by-commit.

The motivation for making this change is that this could previously cause clients which looked at this header to attempt to reuse connections which the server had closed. Clients that do this would get sent a TCP RST packet for their troubles, and would raise an exception.

### DIFF
--- a/test/test_puma_server.rb
+++ b/test/test_puma_server.rb
@@ -882,4 +882,43 @@ EOF
       assert_does_not_allow_http_injection(app)
     end
   end
+
+  def test_http11_connection_header_queue
+    server_run app: ->(_) { [200, {}, [""]] }
+
+    sock = send_http "GET / HTTP/1.1\r\n\r\n"
+    assert_equal ["HTTP/1.1 200 OK", "Content-Length: 0"], header(sock)
+
+    sock << "GET / HTTP/1.1\r\nConnection: close\r\n\r\n"
+    assert_equal ["HTTP/1.1 200 OK", "Connection: close", "Content-Length: 0"], header(sock)
+
+    sock.close
+  end
+
+  def test_http10_connection_header_queue
+    server_run app: ->(_) { [200, {}, [""]] }
+
+    sock = send_http "GET / HTTP/1.0\r\nConnection: keep-alive\r\n\r\n"
+    assert_equal ["HTTP/1.0 200 OK", "Connection: Keep-Alive", "Content-Length: 0"], header(sock)
+
+    sock << "GET / HTTP/1.0\r\n\r\n"
+    assert_equal ["HTTP/1.0 200 OK", "Content-Length: 0"], header(sock)
+    sock.close
+  end
+
+  def test_http11_connection_header_no_queue
+    @server = Puma::Server.new @app, @events, queue_requests: false
+    server_run app: ->(_) { [200, {}, [""]] }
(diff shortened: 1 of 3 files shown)


test/test_puma_server.rb
```diff
@@ -1,4 +1,6 @@
 require_relative "helper"
+require "puma/events"
+require "net/http"
```
praboud-stripe, line 3: This is conceptually unrelated to the rest of the change, but this test file needs these includes to run individually.

test/test_puma_server.rb
```diff
@@ -880,4 +882,43 @@ def assert_does_not_allow_http_injection(app, opts = {})
...
+
+    sock = send_http "GET / HTTP/1.0\r\nConnection: keep-alive\r\n\r\n"
+    assert_equal ["HTTP/1.0 200 OK", "Connection: Keep-Alive", "Content-Length: 0"], header(sock)
+
+    sock << "GET / HTTP/1.0\r\n\r\n"
+    assert_equal ["HTTP/1.0 200 OK", "Content-Length: 0"], header(sock)
+    sock.close
+  end
+
+  def test_http11_connection_header_no_queue
```
praboud-stripe, line 909: This test fails before the bugfix - the other 3 are just to ensure no regressions / generally exercise this code.

praboud-stripe:
Hrm - the builds seem to be flaking. (The builds are failing with timeouts, and don't seem to be failing in a consistent way.) Does anybody have any advice on what to do here?

MSP-Greg:
I've seen this before, but not as bad as it is now.  Before, it was an occasional job in a group of several.  Now, sometimes it's most of the Ubuntu jobs.

I wish I knew where to file an issue for it...

praboud-stripe:
Oof - the latest build failure is an error from inside of Ragel, from a different test: https://github.com/puma/puma/pull/2216/checks?check_run_id=557243347
I don't know if I'm just getting unlucky, or if there's actually a persistent failure on the Ubuntu 2.2 build.

praboud-stripe:
Great! All builds pass - so it seems like I was just getting unlucky. (This issue looks maybe related https://github.com/puma/puma/issues/2148, but honestly I don't really know.) I might try taking a swing at debugging what's going on here - but for now, this PR is ready for review.

nateberkopec:
Thanks! Good work.

PR #2218, opened 2020-04-07
Fix out_of_band hook

### LINKED ISSUE #2177: out_of_band hook not working when using multiple threads

wjordan:
### Description
Fixes #2177. (Thanks @GuiTeK for tracking down this bug!)

Modifies the `out_of_band` hook implementation in the `Server` class to ensure that hooks are fired as expected when the server is configured with >1 max threads.

Compare to #2178, #2180. This implementation is pretty simple. It changes behavior slightly by executing the hooks on the same thread that finished processing the last request (instead of the main acceptor thread), so the server can continue accepting incoming requests if any arrive while the out-of-band hooks are still executing.

I've also added a couple of server tests to ensure out-of-band hooks fire as expected in two scenarios (sequential and parallel requests).

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -224,6 +224,12 @@ module Puma
             @reactor.add client
           end
         end
+
+        if @options[:out_of_band]
+          @thread_pool.with_mutex do
+            notify_safely(OUT_OF_BAND_COMMAND) if @thread_pool.busy_threads == 1
+          end
+        end
       end
 
       @thread_pool.clean_thread_locals = @options[:clean_thread_locals]
@@ -288,10 +294,7 @@ module Puma
                     end
 
                     pool << client
-                    busy_threads = pool.wait_until_not_full
-                    if busy_threads == 0
-                      @options[:out_of_band].each(&:call) if @options[:out_of_band]
-                    end
+                    pool.wait_until_not_full
                   end
                 rescue SystemCallError
                   # nothing
@@ -343,6 +346,9 @@ module Puma
       when RESTART_COMMAND
         @status = :restart
         return true
+      when OUT_OF_BAND_COMMAND
+        @options[:out_of_band].each(&:call) if @options[:out_of_band]
+        return false
       end
 
       return false
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -224,12 +224,6 @@ module Puma
             @reactor.add client
           end
         end
-
-        if @options[:out_of_band]
-          @thread_pool.with_mutex do
-            notify_safely(OUT_OF_BAND_COMMAND) if @thread_pool.busy_threads == 1
-          end
-        end
       end
 
       @thread_pool.clean_thread_locals = @options[:clean_thread_locals]
@@ -346,9 +340,6 @@ module Puma
       when RESTART_COMMAND
         @status = :restart
         return true
-      when OUT_OF_BAND_COMMAND
-        @options[:out_of_band].each(&:call) if @options[:out_of_band]
-        return false
       end
 
       return false
@@ -444,6 +435,12 @@ module Puma
           # Already closed
         rescue StandardError => e
           @events.unknown_error self, e, "Client"
+        end
+
+        if @options[:out_of_band]
+          @thread_pool.with_mutex do
+            @options[:out_of_band].each(&:call) if @thread_pool.busy_threads == 1
+          end
         end
       end
     end
--- a/lib/puma/thread_pool.rb
+++ b/lib/puma/thread_pool.rb
@@ -82,6 +82,10 @@ module Puma
       waiting + (@max - spawned)
     end
 
+    def busy_threads
+      with_mutex { @spawned - @waiting + @todo.size }
+    end
+
     # :nodoc:
     #
     # Must be called with @mutex held!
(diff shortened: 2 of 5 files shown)


MSP-Greg:
I just switched to a different 'apt-get' server, so Ubuntu should pass consistently, I think.

lib/puma/thread_pool.rb
```diff
@@ -79,6 +79,10 @@ def pool_capacity
       waiting + (@max - spawned)
     end
 
+    def busy_threads
+      @spawned - @waiting + @todo.size
```
nateberkopec, line 83: I think we need to wait for the @mutex here if not already held. This will work on MRI but won't work on JRuby/etc.

lib/puma/dsl.rb
```diff
@@ -494,8 +494,9 @@ def after_worker_fork(&block)
 
     # Code to run out-of-band when the worker is idle.
     # These hooks run immediately after a request has finished
-    # processing and there are no busy threads on the worker.
-    # The worker doesn't accept new requests until this code finishes.
```
nateberkopec, line 498: Whoops, that wasn't ever true, was it?

nateberkopec: Also, Q: should this be the actual behavior? Is it really OOB otherwise?

nateberkopec:
Only Q for me is: should the OOB hook work the way it was advertised? "No other requests will be processed until the hook finishes". 

Both approaches have the potential to add latency to requests.

lib/puma/dsl.rb
wjordan: > Whoops, that wasn't ever true, was it?

The oob hook previously ran from the acceptor thread, so I think it _was_ true at least for the case where it was working correctly (max threads 1).

> should this be the actual behavior? Is it really OOB otherwise?

Hmm, I went back and forth on whether it's better to keep accepting new requests while an OOB hook is running. (Might depend on the hook's use case?) For GC optimization at least, I thought continuing to accept requests could potentially be better under high load (though the MRI GVL could make it a wash). But on second thought, _not_ accepting new requests is probably safer for light loads especially while request-balancing is still uneven. Plus it is 'really OOB' for other use-cases. I can change this to run on the acceptor thread as before.

lib/puma/thread_pool.rb
wjordan: Done in ff28636f2ccd274569ed29c7bc5d505d5d96ba87 - not 100% sure if this approach (calling `@mutex.owned?` in a `with_mutex` wrapper method) has any small perf implications we care about. Alternative would be to not use a method at all, or introduce a `in_mutex` argument.

lib/puma/dsl.rb
wjordan: Updated in 7b3639125d8591103db7962f88531208ca716aff

nateberkopec:
Hm. I think that if the method is called `oob`, it needs to actually be `oob` - so:

1. Grab the threadpool mutex. If we are the only busy thread, then:
2. Stop accepting requests.
3. Run callbacks, return.
4. Start accepting requests.

nateberkopec:
@wjordan thoughts on https://github.com/puma/puma/pull/2218#issuecomment-611351606?

nateberkopec:
Oh OK, that's definitely better than interacting with the threadpool 👍

lib/puma/server.rb
```diff
@@ -439,6 +436,12 @@ def process_client(client, buffer)
         rescue StandardError => e
           @events.unknown_error self, e, "Client"
         end
+
+        if @options[:out_of_band]
+          @thread_pool.with_mutex do
```
ayufan, line 441: Does this actually create a mutex now on `thread_pool`? The `OOB` hook is executed within a mutex, rather than dispatched to be executed.

This also, connected with a test `#test_out_of_band_stream` creates a very strong dependency on execution:
the test assumes that all connections are in fact processed concurrently, but this might not always be true, because should we assume that `threads=1`, kind of similar to executing `Puma` with `threads=1`. The test does fail in this condition, as this is being executed on down failing edge of the event.

PR #2221, opened 2020-04-10
Fix ThreadPool#shutdown timeout accuracy

wjordan:
### Description
Fixes up `ThreadPool#shutdown` so that the provided `timeout` value (which corresponds to the `force_shutdown_after` configuration option) is more accurate, and can be set as a float instead of just an integer value.

Previously, the shutdown timeout ran a loop `timeout.times`, which called `t.join 1` on each thread (one after another, not parallel), followed by a `sleep 1` each loop iteration, then `t.join SHUTDOWN_GRACE_TIME` on each thread after raising `ForceShutdown`. Instead of forcing shutdown in `timeout` (+ `SHUTDOWN_GRACE_TIME`) seconds, this forced a shutdown in `timeout * (n + 1)` (+ `n * SHUTDOWN_GRACE_TIME`) seconds (where `n` is the number of running threads). The updated code fixes this by joining a maximum of `timeout` seconds (followed by a maximum of `grace` seconds) across all threads.

I also added an extra step at the end that calls `kill` on any remaining threads still alive after the grace period, plus another short (1-second) join to wait for force-killed threads to (finally) exit, as an extra precaution that threads are fully cleaned up when shutdown returns. I thought this might be helpful, but could be removed as it's not strictly needed for the rest of this PR.

### DIFF
--- a/History.md
+++ b/History.md
@@ -32,6 +32,7 @@
   * Pass queued requests to thread pool on server shutdown (#2122)
   * Fixed a few minor concurrency bugs in ThreadPool that may have affected non-GVL Rubies (#2220)
   * Fix `out_of_band` hook never executed if the number of worker threads is > 1 (#2177)
+  * Fix ThreadPool#shutdown timeout accuracy (#2221)
 
 * Refactor
   * Remove unused loader argument from Plugin initializer (#2095)
--- a/lib/puma/thread_pool.rb
+++ b/lib/puma/thread_pool.rb
@@ -277,6 +277,9 @@ module Puma
     end
 
     # Tell all threads in the pool to exit and wait for them to finish.
+    # Wait +timeout+ seconds then raise +ForceShutdown+ in remaining threads.
+    # Next, wait an extra +grace+ seconds then force-kill remaining threads.
+    # Finally, wait +kill_grace+ seconds for remaining threads to exit.
     #
     def shutdown(timeout=-1)
       threads = with_mutex do
@@ -295,27 +298,26 @@ module Puma
         # Wait for threads to finish without force shutdown.
         threads.each(&:join)
       else
-        # Wait for threads to finish after n attempts (+timeout+).
-        # If threads are still running, it will forcefully kill them.
-        timeout.times do
-          threads.delete_if do |t|
-            t.join 1
-          end
-
-          if threads.empty?
-            break
-          else
-            sleep 1
+        join = ->(timeout) do
+          start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+          threads.reject! do |t|
+            elapsed = Process.clock_gettime(Process::CLOCK_MONOTONIC) - start
+            t.join timeout - elapsed
           end
         end
 
+        # Wait +timeout+ seconds for threads to finish.
+        join.call(timeout)
+
+        # If threads are still running, raise ForceShutdown and wait to finish.
         threads.each do |t|
           t.raise ForceShutdown
         end
+        join.call(SHUTDOWN_GRACE_TIME)
 
-        threads.each do |t|
-          t.join SHUTDOWN_GRACE_TIME
-        end
+        # If threads are _still_ running, forcefully kill them and wait to finish.
+        threads.each(&:kill)
+        join.call(1)
       end
 
       @spawned = 0
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -14,6 +14,7 @@ end
 require "minitest/autorun"
 require "minitest/pride"
 require "minitest/proveit"
+require "minitest/stub_const"
 require_relative "helpers/apps"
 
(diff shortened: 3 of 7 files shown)


History.md
```diff
@@ -29,6 +29,7 @@
   * Read directly from the socket in #read_and_drop to avoid raising further SSL errors (#2198)
   * Set `Connection: closed` header when queue requests is disabled (#2216)
   * Pass queued requests to thread pool on server shutdown (#2122)
+  * Fix ThreadPool#shutdown timeout accuracy
```
nateberkopec, line 32: Needs a PR # link

nateberkopec:
LGTM, anyone else have any thoughts?

lib/puma/thread_pool.rb
```diff
@@ -298,27 +301,25 @@ def shutdown(timeout=-1)
...
-          end
-
-          if threads.empty?
-            break
-          else
-            sleep 1
+        join = ->(timeout) do
+          start = Time.now.to_f
+          threads.reject! do |t|
+            t.join timeout + start - Time.now.to_f
```
dentarg, line 307: I actually had one thought. Should we be using [monotonic time](https://blog.dnsimple.com/2018/03/elapsed-time-with-ruby-the-right-way/)? Or is `#to_f` good enough for us?

lib/puma/thread_pool.rb
```diff
@@ -298,27 +301,26 @@ def shutdown(timeout=-1)
...
-        timeout.times do
-          threads.delete_if do |t|
-            t.join 1
-          end
-
-          if threads.empty?
-            break
-          else
-            sleep 1
+        join = ->(timeout) do
```
evanphx, line 304: I don't understand why this proc is created and then called a few lines later.

evanphx: Is it because you want to call it twice with the same logic?

lib/puma/thread_pool.rb
```diff
@@ -281,8 +281,11 @@ def auto_reap!(timeout=5)
     end
 
     # Tell all threads in the pool to exit and wait for them to finish.
+    # Wait +timeout+ seconds then raise +ForceShutdown+ in remaining threads.
+    # Next, wait an extra +grace+ seconds then force-kill remaining threads.
+    # Finally, wait +kill_grace+ seconds for remaining threads to exit.
     #
-    def shutdown(timeout=-1)
+    def shutdown(timeout=-1, grace: SHUTDOWN_GRACE_TIME, kill_grace: 1)
```
evanphx, line 288: Nothing uses these options it appears except the tests so we should probably not even bother to expose them.

lib/puma/thread_pool.rb
nateberkopec: I thought it was kind of interesting way to avoid creating a method (which would not have been called anywhere else).

lib/puma/thread_pool.rb
nateberkopec: I noticed this but I'm neutral since this is an internal API.

lib/puma/thread_pool.rb
wjordan: Yeah, it's just a lightweight form of local code reuse to avoid repeating the same few lines of code 3 times in a row.

lib/puma/thread_pool.rb
wjordan: This was only to limit the duration of the added test (which would otherwise take 5 seconds), see 09702d19fed95cb4d6b5cc5a478d2d376db8db3e for an alternate approach that doesn't expose any extra arguments and instead uses a simple `stub_const` patch (added to the test helper).

test/helper.rb
```diff
@@ -143,3 +143,16 @@ def full_name
     end
   end
 end
+
+module StubConst
+  # Stub constant for the duration of a block
+  def stub_const(name, val, &block)
```
nateberkopec, line 149: Just use `Minitest::Mock`

nateberkopec: https://github.com/seattlerb/minitest/blob/master/lib/minitest/mock.rb#L212

wjordan: `#stub` doesn't seem to work for constants, only methods.
There is a [`minitest-stub-const`](https://github.com/adammck/minitest-stub-const) extension ([not in core](https://github.com/seattlerb/minitest/pull/230#issuecomment-12869299)) that provides `stub_const`, so I've updated the PR to add that gem-plugin as a dependency if that's preferred to the in-line code.

The alternative would be to replace the `SHUTDOWN_GRACE_TIME` const with a getter-method that can be stubbed with minitest as-is, but I think stubbing the const directly is cleaner. (It  might also be helpful for stubbing other timing-related constants like `WORKER_CHECK_INTERVAL` as well to speed up other tests..)

PR #2224, opened 2020-04-15
The new way of allowing failures on GitHub Actions [changelog skip]

dentarg:
See https://github.blog/changelog/2020-04-15-github-actions-new-workflow-features/

With this, it [looks like](https://github.com/dentarg/puma/actions?query=branch%3Atmp%2Factions-fubar) the workflow is marked as green, even when the build allowed to fail [fails](https://github.com/dentarg/puma/runs/590406607?check_suite_focus=true#step:7:164).

### DIFF
--- a/.github/workflows/puma.yml
+++ b/.github/workflows/puma.yml
@@ -21,10 +21,12 @@ jobs:
         ruby: [ 2.2, 2.3, 2.4, 2.5, 2.6, 2.7, head, jruby, truffleruby-head ]
         include:
           - { os: windows , ruby: mingw }
+          - { os: ubuntu  , ruby: jruby-head, continue-on-error: true }
         exclude:
           - { os: windows , ruby: head }
           - { os: windows , ruby: jruby }
           - { os: windows , ruby: truffleruby-head }
+    continue-on-error: ${{ matrix.continue-on-error || false }}
 
     steps:
       - name: repo checkout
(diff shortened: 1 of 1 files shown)


dentarg:
For anyone interested (@MSP-Greg?) I will note that I got some weird failures when testing this: [ubuntu 2.7](https://github.com/dentarg/puma/runs/590369796?check_suite_focus=true#step:7:171), [macos 2.5](https://github.com/dentarg/puma/runs/590369842?check_suite_focus=true#step:7:275). Some temporary instability in the Actions environment I guess.

dentarg:
> [looks like](https://github.com/dentarg/puma/actions?query=branch%3Atmp%2Factions-fubar) the workflow is marked as green

Ah, that's not enough for PRs it seems, we still need the `if: success()`... too bad

https://github.com/puma/puma/blob/67f9b1f3f2318c14279ba212ebdfaa5bccd9bc08/.github/workflows/puma.yml#L101

I'll leave this open for a while anyway.

MSP-Greg:
@dentarg 

> Some temporary instability in the Actions environment I guess.

Umm, maybe not.  If one runs the CI forcing the seed value for all jobs, and more jobs fail, it's the test suite that is unstable.  I think I've commented on two.

Re the Actions 'allow-failure', the guy who added the new 'matrix' feature is aware of it, but I think it's messy.

JFYI (or others), If one looks at a CI run, and the left hand 'selection' is on the 'root' (which for Puma is 'CI'), the 'annotations' section shows which jobs failed.

nateberkopec:
Doesn't seemed to have worked, as jruby-head failed but I'm still seeing "some checks were not successful".

dentarg:
> Doesn't seemed to have worked, as jruby-head failed but I'm still seeing "some checks were not successful".

Yeah, [the workflow is green](https://github.com/puma/puma/actions?query=workflow%3ACI+event%3Apull_request+branch%3Aactions%2Fcontinue-on-error) but apparently that's not what matters for PRs 😆

I might experiment some more in my fork

PR #2228, opened 2020-04-21
run regular GC if compact not available

jjb:
### Description

While doing some simple experiments with GC.compact in ruby 2.7 I noticed that
GC.start also had some benefits in the pre_fork phase. Since the master process won't be
growing (as much? at all?) as the forked processes, it might not ever get an organic GC, so this
is perhaps a good time to get in a manual one to save a meg or three.

(I actually don't know if the garbage collection done by GC.start reduces forked memory and/or CoW effeciency — if so, an even better idea to do it for all ruby versions).

- [ x] I have added an entry to [History.md](../blob/master/History.md) if this PR fixes a bug or adds a feature. If it doesn't need an entry to HISTORY.md, I have added `[changelog skip]` the pull request title.
- [x ] I have added appropriate tests if this PR fixes a bug or adds a feature.
- [ x] My pull request is 100 lines added/removed or less so that it can be easily reviewed.
- [ x] If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.
- [x ] If this closes any issues, I have added "Closes `#issue`" to the PR description or my commit messages.
- [x ] I have updated the documentation accordingly.
- [x ] All new and existing tests passed, including Rubocop. **i did not run tests because i was unable to install ragel**

### DIFF
--- a/History.md
+++ b/History.md
@@ -5,7 +5,7 @@
   * Add pumactl `thread-backtraces` command to print thread backtraces (#2053)
   * Configuration: `environment` is read from `RAILS_ENV`, if `RACK_ENV` can't be found (#2022)
   * Do not set user_config to quiet by default to allow for file config (#2074)
-  * `GC.compact` is called before fork if available (#2093)
+  * `GC.compact` is called before fork if available (#2093), otherwise GC.start is called
   * Add `requests_count` to workers stats. (#2106)
   * Increases maximum URI path length from 2048 to 8196 bytes (#2167)
   * Force shutdown responses can be overridden by using the `lowlevel_error_handler` config (#2203)
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -459,7 +459,11 @@ module Puma
       @master_read, @worker_write = read, @wakeup
 
       @launcher.config.run_hooks :before_fork, nil, @launcher.events
-      GC.compact if GC.respond_to?(:compact)
+      if GC.respond_to?(:compact)
+        GC.compact
+      else
+        GC.start
+      end
 
       spawn_workers
 


nateberkopec:
I've been thinking about this. See [nakayoshi_fork](https://github.com/ko1/nakayoshi_fork).

I think maybe we should:

```
if configured_to_cow_friendly_fork
3.times { GC.start }
if GC.respond_to?(:compact)
  GC.compact
end
end
```

This will introduce a lag time before forking, maybe up to 1-2 seconds. I'm not sure if that amount of lag would be acceptable for everyone so we may need to put the whole thing behind a config flag.

nateberkopec:
The "3 times" is very important here because you're trying to promote all existing objects to the old generation.

I'm actually not sure if it would be better to compact before or after that, because I don't know how generations interaction w/compaction.

nateberkopec:
If you try combinations of GC.start/GC.compact ordering, and test my thesis that 3 runs is better than 1, and track latency of all these options and report back, that would be 🙇 .

Maybe test against rubygems.org or another decently large open source app.

jjb:
fascinating, I didn't know that multiple runs of GC.start could make a difference. I guess ruby GC makes some sort of time/value compromise and doesn't always do the best it can?

in a test just now with my rails app, I set cache_classes and eager_load to true, and put `binding.irb` inside puma `before_fork`. i observed memory in Activity Monitor:

started at 288.3
after 1 GC.start: 279.1
subsequent GC.start did not change it
GC.compact did not change it! sad trombone
(but maybe it made the internal pages more efficient and better for forking/CoW?)

Then I did this test with 10 workers. I only ran each test once and I didn't put any load on the workers, so didn't really flex the CoW situation:

https://gist.github.com/jjb/d65bc10bb55fed7e4cafb07be035c10a

the original implementation of compact ran a GC before and after https://bugs.ruby-lang.org/issues/15626 I don't know if that made it into the final implementation https://github.com/ruby/ruby/commit/3c55b643aec09bbe779dab25b2397947eded2b9b

nateberkopec:
> guess ruby GC makes some sort of time/value compromise and doesn't always do the best it can?

In a way yes - this is the premise of a [Generational GC.](https://blog.heroku.com/incremental-gc)

In order for the test to be "fair" you'll have to put some load on your server after forking. The premise of nakayoshi_fork and compact-before-fork is that you improve copy-on-write by putting old objects all together on the fewest number of pages/space possible, which means you will do more copying and less writing as the server process takes requests.

jjb:
> In a way yes - this is the premise of a Generational GC.

ah, gotcha

> In order for the test to be "fair" you'll have to put some load on your server after forking

alas I don't have time right now to set something up. would be cool to do a test with https://github.com/noahgibbs/rsb/

nateberkopec:
Adding contrib-wanted if anyone wants to try to run those tests themselves.

nateberkopec:
I'm thinking about putting this + compact behind an option for 5.0 and gathering data that way.

jjb:
sounds good to me!

nateberkopec:
Closing in favor of #2256

PR #2230, opened 2020-04-23
Improve OOB hook

ayufan:
### Description

This moves the hook to be executed outside of busy loop
considering the thread to be truly idle at a time of execution
of the hook, and allowing another thread to accept the connection
while the hook is being called.

### DIFF
--- a/History.md
+++ b/History.md
@@ -38,6 +38,7 @@
   * Fixed a few minor concurrency bugs in ThreadPool that may have affected non-GVL Rubies (#2220)
   * Fix `out_of_band` hook never executed if the number of worker threads is > 1 (#2177)
   * Fix ThreadPool#shutdown timeout accuracy (#2221)
+  * Call `out_of_band` hook outside of busy loop (#2230)
 
 * Refactor
   * Remove unused loader argument from Plugin initializer (#2095)
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -226,6 +226,7 @@ module Puma
         end
       end
 
+      @thread_pool.out_of_band_hook = @options[:out_of_band]
       @thread_pool.clean_thread_locals = @options[:clean_thread_locals]
 
       if @queue_requests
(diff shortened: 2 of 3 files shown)


ayufan:
@nateberkopec @wjordan WDYT about this?

ayufan:
There seems to be also the problem with that test, it seems to be flakky:

```
 17) Failure:
TestPumaServer#test_out_of_band_stream [D:/a/puma/puma/test/test_puma_server.rb:1044]:
Expected: 100
  Actual: 42
```

wjordan:
In #2218 we intentionally kept the behavior that "No other requests will be processed until the hook finishes", in order to keep the hooks truly 'out of band' - see https://github.com/puma/puma/pull/2218#issuecomment-610730736 + https://github.com/puma/puma/pull/2218#discussion_r405249810. This is covered by the added test [`test_out_of_band_overlapping_requests`](https://github.com/puma/puma/pull/2218/files#diff-ad2a310d12c66209944ac6f74d7c3d99R1048) which ensures a new request doesn't begin processing while an OOB hook is still running, and fails in this PR.

If there really are distinct use-cases where a 'concurrent out of band' hook is needed, maybe a second configuration option for the more relaxed mode would be useful?

Other than that, the general idea of moving the out of band hooks from `Server` to `Threadpool` could make sense. It simplifies the implementation a bit since the threadpool-mutex interactions don't exit the class.

There are two additional edge-cases worth considering for improving the OOB implementation:
- Distinguishing between work ending in a partial-request queued in the Reactor and work ending in a finished request-response. Currently, the partial-requests also trigger OOB which isn't ideal since it means it may get triggered multiple times for a single request depending on timing. I think this edge-case is possibly part of the `test_out_of_band_stream` test-flakiness noted above.
- Blocking the server from `accept`ing new connections while OOB is running. Although the threadpool won't start working on new requests until OOB finishes, currently the server will still `accept` incoming connections while OOB is running. Even just accepting a connection mid-OOB could make request-balancing in clustered mode sub-optimal, since we'd rather the request be accepted by a process that can work on it right away versus a process blocked on OOB work.

ayufan:
@wjordan Thanks for awesome explanation. I did not notice that aspects. Thus I will close this, and improve the relevant test.

We still have open question whether we should accept a new `@todos` while executing `OOB`. I would rather say no, so, maybe we should actually move this `OOB` execution in main loop, but under similar conditions?

I will close this for now, so we could discuss this further.

PR #2234, opened 2020-04-24
Improvements to out_of_band hook

wjordan:
### Description
Followup to #2230 (extending the idea to move the OOB hook from `Server` into `ThreadPool`), addressing the two extra edge cases described at https://github.com/puma/puma/pull/2230#issuecomment-618825848.

Also:
- Refactor existing out-of-band tests from the big `TestPumaServer` class into its own `TestOutOfBandServer` class
- Add extra tests covering the new edge-cases handled
- A bit more cleanup/documentation on the test suite (added high-level comments describing the behavior expected by each test)

(Turns out there are a ton of complicated little edge-cases to consider around what seemed like such a simple feature!)


lib/puma/server.rb
```diff
@@ -326,6 +317,20 @@ def handle_servers
...
+    rescue Errno::ECONNABORTED
+      # client closed the socket even before accept
+      begin
+        sock.close
+      rescue
+        Thread.current.purge_interrupt_queue if Thread.current.respond_to? :purge_interrupt_queue
+      end
+    end
+    private :accept_socket
+
```
wjordan, line 333: Extracted this logic into a private method both to simplify the accept-loop, but more importantly to allow the to be stubbed in the `test_blocks_new_connection` unit-test.

lib/puma/server.rb
```diff
@@ -278,27 +281,15 @@ def handle_servers
...
-                  end
-                rescue SystemCallError
-                  # nothing
-                rescue Errno::ECONNABORTED
-                  # client closed the socket even before accept
-                  begin
-                    io.close
-                  rescue
-                    Thread.current.purge_interrupt_queue if Thread.current.respond_to? :purge_interrupt_queue
+                pool.wait_until_not_full
```
wjordan, line 284: Moving the call to `wait_until_not_full` before `accept_socket` prevents connections from being accepted until out_of_band hooks are completed.

(I'm not 100% sure this change won't have any other side-effects, but the test suite seems to think it's okay...)

An alternative would be to just add an extra `pool.with_mutex {}` line above `accept_socket`.

lib/puma/server.rb
```diff
@@ -224,8 +224,11 @@ def run(background=true)
             @reactor.add client
           end
         end
+
+        process_now
```
wjordan, line 228: This return value of the `ThreadPool` block (`process_now` is true when a complete request was processed by the block) is used to determine if an OOB hook should be scheduled to fire once the thread is eventually idle.

lib/puma/server.rb
wjordan: hmm, the original `ECONNABORTED` rescue clause had `io.close` not `sock.close`, I should change this back to the original behavior.

That  said, the line is _ancient_, dating back to 2006 (https://github.com/puma/puma/commit/ab3c8082de82e6fc96838d444be06432620743ab) (!) and never covered by any tests, so it's very likely this is 'superstitious' code that should probably be changed/removed separately anyway. (In my reading of the current code, `ECONNABORTED` would only ever be thrown by `accept_nonblock` if the listener socket is closed, but in that case `io` would be `nil` causing `io.close` to always fail..)

wjordan: I managed to figure out how to stub the `accept` call without needing to extract the method, so I'll move this refactoring/cleanup to a separate PR.

test/test_out_of_band_server.rb
```diff
@@ -0,0 +1,159 @@
...
+        @mutex.synchronize do
+          @oob_count += 1
+          @oob_finished.signal
+          @oob_finished.wait(@mutex, 1) if oob_wait
+        end
+      end
+    end
+    app_wait = options.delete(:app_wait)
+    app = ->(_) do
+      raise 'OOB conflict' if in_oob.locked?
```
nateberkopec, line 53: Nice, I like that approach.

nateberkopec:
Tests are 😍

PR #2250, opened 2020-05-04
Add unified detailed error logging

### LINKED ISSUE #2235: Debug logging for 4xx/5xx
We need opt-in settings for dumping all info we have about bad requests/exceptions. This happens in many places in the code and we should use a unified interface as well.

Related to #2232

alexeevit:
### Description
Hey, it's my humble try to make a unified debug logger for the whole project.
The work isn't finished and I'm not sure if I'm doing what expected so I'm going to just put it here for further discussion.
Closes #2235


acmh:
Please, can you fix the rubocop offenses to the CI run the tests?

alexeevit:
@nateberkopec hey, I've got a couple of questions:
- WDYT about the solution?
- Do I have to put logging to each `rescue` in the project?
- What about the logging format? Do we need to support different formatters?
- Do we have to print headers & body of http requests?

lib/puma/events.rb
```diff
@@ -106,6 +109,7 @@ def parse_error(server, env, error)
     def ssl_error(server, peeraddr, peercert, error)
       subject = peercert ? peercert.subject : nil
       @stderr.puts "#{Time.now}: SSL error, peer: #{peeraddr}, peer cert: #{subject}, #{error.inspect}"
```
nateberkopec, line 111: Basically, I think _all_ error related output should go through this new interface. So no more `stderr.puts`, whatever we're logging today to stderr should be converted to this new format.

lib/puma/events.rb
```diff
@@ -97,6 +99,7 @@ def parse_error(server, env, error)
         "(#{env[HTTP_X_FORWARDED_FOR] || env[REMOTE_ADDR]}#{env[REQUEST_PATH]}): " \
```
nateberkopec, line 99: same here. No more puts. Only calls to the new interface, which should probably just end up calling `stderr.puts` by default.

lib/puma/server.rb
```diff
@@ -919,13 +938,15 @@ def fast_write(io, str)
...
+          @debug_logger.error_dump(e)
           if !IO.select(nil, [io], nil, WRITE_TIMEOUT)
             raise ConnectionError, "Socket timeout writing data"
           end
 
           retry
-        rescue  Errno::EPIPE, SystemCallError, IOError
+        rescue  Errno::EPIPE, SystemCallError, IOError => e
+          @debug_logger.error_dump(e)
           raise ConnectionError, "Socket timeout writing data"
```
nateberkopec, line 950: Do we rescue this connectionerror somewhere higher up in the stack? You should just be able to error_dump from there instead.

lib/puma/server.rb
```diff
@@ -319,8 +329,7 @@ def handle_servers
         end
         graceful_shutdown if @status == :stop || @status == :restart
       rescue Exception => e
-        STDERR.puts "Exception handling servers: #{e.message} (#{e.class})"
```
nateberkopec, line 322: Here you did the right thing - replacing the old interface with the new one 👍

lib/puma/server.rb
```diff
@@ -111,16 +113,18 @@ def inherit_binder(bind)
       def cork_socket(socket)
         begin
           socket.setsockopt(6, 3, 1) if socket.kind_of? TCPSocket
-        rescue IOError, SystemCallError
+        rescue IOError, SystemCallError => e
           Thread.current.purge_interrupt_queue if Thread.current.respond_to? :purge_interrupt_queue
+          @debug_logger.error_dump(e)
```
nateberkopec, line 118: This rescue and the one in `uncork` aren't really worth logging, I think.

lib/puma/server.rb
```diff
@@ -58,6 +59,7 @@ class Server
     def initialize(app, events=Events.stdio, options={})
       @app = app
       @events = events
+      @debug_logger = DebugLogger.stdio
```
nateberkopec, line 62: I think ideally no one would access this interface directly, but instead would access it indirectly through a method on `Events`, like parse_error.

nateberkopec:
Left some comments

> Do I have to put logging to each rescue in the project?

Definitely not, because not all errors are worth logging. Also, I feel like there's a higher-level rescue in Puma somewhere where you can catch most of the exceptions that are re-raised. Instead, concentrate on _where we are logging errors today_, usually to stderr but maybe in other ways, and replacing those with this interface

 > Do we have to print headers & body of http requests?

I think there should be a global option to dump the rack env. So, if this option is on, we dump the rack env to the output every time the error interface is called.

alexeevit:
Well, the CI is weird

lib/puma/server.rb
alexeevit: Yup, removed debug_logger from the server and replaced its invokes with wrappers from the events class

lib/puma/server.rb
alexeevit: Done

lib/puma/server.rb
alexeevit: Yup, added logging higher up in the stack

lib/puma/events.rb
alexeevit: Done

lib/puma/events.rb
alexeevit: Done

alexeevit:
I removed the `server` argument from logging methods in the Events class and unified them with the single order of arguments

lib/puma/events.rb
```diff
@@ -80,51 +79,44 @@ def debug(str)
...
+    # +env+ the request, and +error+ a
     # parsing exception.
     #
-    def parse_error(server, env, error)
-      @stderr.puts "#{Time.now}: HTTP parse error, malformed request " \
-        "(#{env[HTTP_X_FORWARDED_FOR] || env[REMOTE_ADDR]}#{env[REQUEST_PATH]}): " \
-        "#{error.inspect}" \
-        "\n---\n"
+    def parse_error(error, env)
+      @debug_logger.error_dump(error: error, env: env, text: 'HTTP parse error, malformed request', force: true)
```
dentarg, line 102: (This is more a comment on the concept and not this specific line)

"debug logger" and the "force: true" reads a bit strange to me... what about using different [levels](https://ruby-doc.org/stdlib-2.4.0/libdoc/logger/rdoc/Logger.html) instead? 

Just throwing this out as an idea, not demanding anything, this is already great work, thanks for doing it 👍

dentarg:
> Well, the CI is weird

It's because some [whitespace](https://github.com/puma/puma/pull/2253/commits/b39268b05b1117a7a30ce4dbb231dd23c273f6ae) snuck [in the gemspec](https://github.com/puma/puma/runs/655421979) recently (@schneems noticed this and fixed it #2253 but if maintainer can fix it directly in master I think that would be good).

Don't know about the latest failures here though

lib/puma/events.rb
alexeevit: Yeah, you're definitely right. It's strange and needs to be simplified, I'll have a look what I can do.

nateberkopec:
This looks great! I think the approach is now right-on-target. I think I'll wait for you to add the request information, and then I'll add a final round of review.

lib/puma/events.rb
alexeevit: I updated the MR with the new name and the new interface of the logger class.
In my opinion, now it looks better, but there is a couple of things that confuse me:

- `info` and `debug` methods prints headers and body of the request. As for me, this is ok for `debug`, but I'm not sure about `info`
- Since we replaced all `$stderr` invokes with the new interface, `log`, `write`, and `debug` use `$stdout` and have the same format, but `error` uses the new interface and have a different format

@nateberkopec what do you think about it?

lib/puma/error_logger.rb
```diff
@@ -0,0 +1,92 @@
+# frozen_string_literal: true
+
+require 'puma/const'
+
+module Puma
+  # The implementation of a logging in debug mode.
```
nateberkopec, line 6: Comment here is out of date now - I think it is uneccessary anyway and can just be deleted.

nateberkopec:
BTW I think the design now is really top-notch, exactly what it should be.

nateberkopec:
Example output, opening an ssl connection to a non-SSL configured Puma:

```
2020-06-07 09:57:49 +0900 HTTP parse error, malformed request (" " - (-)): #<Puma::HttpParserError: Invalid HTTP format, parsing fails. Are you trying to open an SSL connection to a non-SSL Puma?>
```

nateberkopec:
Not a huge fan of how that "malformed request (" " - (-))" looks.

alexeevit:
> Not a huge fan of how that "malformed request (" " - (-))" looks.

Yup, I totally agree, so I fixed it. Now it looks like this:
2020-06-07 16:21:54 +0300 HTTP parse error, malformed request: #<Puma::HttpParserError: Invalid HTTP format, parsing fails. Are you trying to open an SSL connection to a non-SSL Puma?>

lib/puma/error_logger.rb
alexeevit: > Basically nothing about the request should ever be logged unless the user opted in via PUMA_DEBUG.

Agree, so I made the interface to have 2 methods: `debug` and `info`. Obviously, `info` prints only non-sensitive info and `debug` prints only in debug mode.

nateberkopec:
Looks good. I think we have the infrastructure now to improve specific errors/error messages where needed.

This was really impressive work, doing a refactor like this for your first ever contribution to a project is great! I think you can be very proud of what you've done here.

PR #2256, opened 2020-05-11
Added `nakayoshi_fork` option.

nateberkopec:
Reduce memory usage in preloaded cluster-mode apps by GCing before
fork and compacting, where available.

### Description

### DIFF
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -759,5 +759,18 @@ module Puma
     def fork_worker(after_requests=1000)
       @options[:fork_worker] = Integer(after_requests)
     end
+
+    # When enabled, Puma will GC 4 times before forking workers.
+    # If available (Ruby 2.7+), we will also call GC.compact.
+    # Not recommended for non-MRI Rubies.
+    #
+    # Based on the work of Koichi Sasada and Aaron Patterson, this option may
+    # decrease memory utilization of preload-enabled cluster-mode Pumas. It will
+    # also increase time to boot and fork. See your logs for details on how much
+    # time this adds to your boot process. For most apps, it will be less than one
+    # second.
+    def nakayoshi_fork(enabled=false)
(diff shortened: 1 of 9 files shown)


jjb:
the nakayoshi gem does a max of 4 times, not sure if that's a theoretical ideal to also use here

nateberkopec:
True, changed. I'm going to leave out the full algorithm for now because a) license issues w/incorporating someone else's MIT-licensed algo that I don't feel like dealing with and b) compact implies GC, so we could probably GC fewer times in that case.

I want to punt on all the perf issues and possible optimizations until we figure out if this even decreases memory usage at all, if it does then we can deal with these issues.

lib/puma/dsl.rb
```diff
@@ -759,5 +759,18 @@ def set_remote_address(val=:socket)
...
+    # When enabled, Puma will GC 4 times before forking workers.
+    # If available (Ruby 2.7+), we will also call GC.compact.
+    # Not recommended for non-MRI Rubies.
+    #
+    # Based on the work of Koichi Sasada and Aaron Patterson, this option may
+    # decrease memory utilization of preload-enabled cluster-mode Pumas. It will
+    # also increase time to boot and fork. See your logs for details on how much
+    # time this adds to your boot process. For most apps, it will be less than one
+    # second.
+    def nakayoshi_fork(enabled=false)
```
schneems, line 772: Why are we defaulting this to false? It seems weird that adding this in my config does nothing:

```
nakayoshi_fork
```

I have to 

```
nakayoshi_fork(true)
```

Was this intentional or can we change the default?

nateberkopec: Thaaaaat would be a bug

schneems: Maybe that’s why there was so little change in CodeTriage

PR #2260, opened 2020-05-13
Adds systemd notify and watchdog support

### LINKED ISSUE #2184: Systemd notify and watchdog support
https://github.com/mperham/sidekiq/issues/4488

acmh:
### Description
This PR aims to add systemd notify and watchdog support. This PR was based on [#4488](https://github.com/mperham/sidekiq/issues/4488)(Sidekiq).

It now supports the notification and watchdog functionality that systemd allows.

Add READY=1 notification upon startup
Add STOPPING=1 notification upon terminate
Add WATCHDOG=1 loop

References:
https://github.com/agis/ruby-sdnotify
https://www.freedesktop.org/software/systemd/man/sd_notify.html

This closes #2184 


lib/puma/cluster.rb
```diff
@@ -365,13 +365,15 @@ def phased_restart
     def stop
       @status = :stop
       wakeup!
+      @launcher.events.fire_on_stopped!
```
acmh, line 368: I am not sure if I made the right thing here

lib/puma/sd_notify.rb
```diff
@@ -0,0 +1,124 @@
+# frozen_string_literal: true
+
+require "socket"
+
+# SdNotify is a pure-Ruby implementation of sd_notify(3). It can be used to
```
nateberkopec, line 5: What's the provenance of this code/where'd it come from? Did you write this?

nateberkopec:
This is fantastic. I'm so excited that you implemented this!

lib/puma/sd_notify.rb
dentarg: Looks like it comes from https://github.com/mperham/sidekiq/pull/4489/files#diff-aeecdbde3c294612b90dbc578e369375

nateberkopec: Ah I see from your description now.

I think we will have to include the original LICENSE at the top of the file. https://github.com/agis/ruby-sdnotify/blob/master/LICENSE

nateberkopec: I'm also interested in discussing if we should simply use the gem, and raise/bail out if someone tries to use the feature without the correct constant defined.

acmh: Sorry, I thought that I put the license it was one of my concerns, I will fix that.

acmh: I agree with you to use the gem. Also, the quality of gem's code is very good and the sysmtemd specification is stable so we will not do other changes on the gem's code in the near feature IMHO.

lib/puma/cluster.rb
nateberkopec: Hmmm I'm not sure either, I have to look closer at when we use each method.

lib/puma/launcher.rb
```diff
@@ -309,6 +310,42 @@ def prune_bundler
...
+    end
+
+    #
+    # Puma's systemd integration allows Puma to inform systemd:
+    #  1. when it has successfully started
+    #  2. when it is starting shutdown
+    #  3. periodically for a liveness check with a watchdog thread
+    #
+
+    def start_watchdog
```
nateberkopec, line 332: Let's keep this in the `Puma::SDNotify` class.

test/test_systemd.rb
```diff
@@ -0,0 +1,39 @@
...
+    assert_equal(ENV["NOTIFY_SOCKET"], @sockaddr)
+    assert_equal(count, 7)
+
+    count = Puma::SdNotify.stopping
+    assert_equal(socket_message, "STOPPING=1")
+    assert_equal(ENV["NOTIFY_SOCKET"], @sockaddr)
+    assert_equal(count, 10)
+
+    refute Puma::SdNotify.watchdog?
+  end
```
nateberkopec, line 38: I think we should 1 additional full-up integration test, where you go through the whole process of starting/stopping a Puma server so we can cover the stuff in the Launcher.

nateberkopec:
Great work! Excited to get this done.

* [ ] Integrate with the gem mentioned. If the user has SD_NOTIFY set in ENV, but doesnt have the gem loaded, log error message saying "Systemd integration failed. It looks like you're trying to use systemd notify but don't have X gem installed."
* [ ] Address other comments made on PR.

acmh:
@nateberkopec Just to clarify:

> Integrate with the gem mentioned. If the user has SD_NOTIFY set in ENV, but doesnt have the gem loaded, log error message saying "Systemd integration failed. It looks like you're trying to use systemd notify but don't have X gem installed."

I will add the gem to the Gemfile instead of using the code in the `lib/puma/sd_notify.rb` file, right?

nateberkopec:
@acmh Yes, correct

acmh:
@nateberkopec, I've made the changes requested, but it is important to mention:

1 - I didn't understand the purpose of the SD_NOTIFY env that you mentioned, did you mean NOTIFY_SOCKET instead? Anyway, I've implemented the way I understood.

2 - Do I need to specify the new gem ruby-sdnotify gem as a runtime dependency on `puma.gemspec?` it is not clear for me when adding or not.

Please feel free to ask me something or make some suggestions.

nateberkopec:
> Do I need to specify the new gem ruby-sdnotify gem as a runtime dependency on puma.gemspec? it is not clear for me when adding or not. 

No, not necessary to make it a dependency for all users.

docs/systemd.md
```diff
@@ -24,12 +24,20 @@ After=network.target
 # Requires=puma.socket
 
 [Service]
-# Foreground process (do not use --daemon in ExecStart or config.rb)
-Type=simple
+#Puma automatically supports systemd's `Type=notify` and watchdog service
```
nateberkopec, line 27: Automatically isn't entirely true. You should mention you also need the correct gem installed.

lib/puma/launcher.rb
```diff
@@ -310,6 +312,32 @@ def prune_bundler
...
+    #  2. when it is starting shutdown
+    #  3. periodically for a liveness check with a watchdog thread
+    #
+
+    def integrate_with_systemd
+      return unless ENV["SD_NOTIFY"]
+      return unless ENV["NOTIFY_SOCKET"]
+
+      require 'sd_notify'
+      return log "Systemd integration failed. It looks like you're trying to use systemd notify but don't have ruby-sdnotify gem installed" unless defined?(SdNotify)
```
nateberkopec, line 327: The gem is called `sd_notify`, so this message is a little confusing, right?

lib/puma/systemd.rb
```diff
@@ -0,0 +1,32 @@
...
+
+module Puma
+  class Systemd
+    def initialize(events)
+      @events = events
+    end
+
+    def start_watchdog
+      usec = Integer(ENV["WATCHDOG_USEC"])
+      return log "systemd Watchdog too fast: " + usec if usec < 1_000_000
```
nateberkopec, line 13: I think we should warn rather than just log here, because the watchdog is effectively disabled by this method returning.

test/test_integration_systemd.rb
```diff
@@ -0,0 +1,64 @@
+require_relative "helper"
+require_relative "helpers/integration"
+
+require 'sd_notify'
+
+class TestIntegrationSystemd < TestIntegration
+  def setup
+    skip "Skipped on Windows because it does not support Systemd" if windows?
```
nateberkopec, line 8: Hm, shouldn't this test just be Linux-only, really?

test/test_integration_systemd.rb
```diff
@@ -0,0 +1,64 @@
...
+    @sockaddr = nil
+    ENV["SD_NOTIFY"] = nil
+    ENV["NOTIFY_SOCKET"] = nil
+  end
+
+  def socket_message
+    @socket.recvfrom(10)[0]
+  end
+
+  def test_notify_protocol
```
nateberkopec, line 36: I don't see a purpose to this test, because you're just testing the behavior of the `SdNotify` gem, which, presumably, has it's own tests.

nateberkopec:
I think it's missing a watchdog integration test?

nateberkopec:
Sorry it took so long for me to review, but left some comments 👍

docs/systemd.md
acmh: Ok, but one question, since I've added the gem in Puma's Gemfile, will it not be installed?

dentarg: No, the Gemfile is not the same thing as the gemspec. `sd_notify` will be an optional dependency as it will only be needed in certain environments (e.g. if you run on Heroku, you don't need it and don't want to spend time installing it on each deploy)

acmh: Thanks @dentarg for the clarification!

test/test_integration_systemd.rb
acmh: Can you give me more clarification on that?

dentarg: @acmh systemd only exist for Linux

acmh: @dentarg, that was my intention by adding that line, am I missing something?

dentarg: Puma runs on macOS too: https://github.com/puma/puma/blob/a72c68b350e8c48b6c051ee729dd156c6044ca9d/.github/workflows/puma.yml#L22

acmh:
@nateberkopec Sorry for the delay, a busy week! I've made the requested changes. Hope is clear enough.

test/test_integration_systemd.rb
```diff
@@ -0,0 +1,64 @@
...
+    skip_unless_signal_exist? :TERM
+
+    cli_server "test/rackup/hello.ru"
+    assert_equal(socket_message, "READY=1")
+
+    stop_server
+    assert_match(socket_message, "STOPPING=1")
+  end
+
+  def test_systemd_watchdog
```
acmh, line 49: I want to test the case when the `ENV["WATCHDOG_USEC"]` value is less than the advised value, but I don't know how to achieve that. So I just left this test.

lib/puma/launcher.rb
```diff
@@ -311,6 +313,32 @@ def prune_bundler
...
 
+    #
+    # Puma's systemd integration allows Puma to inform systemd:
+    #  1. when it has successfully started
+    #  2. when it is starting shutdown
+    #  3. periodically for a liveness check with a watchdog thread
+    #
+
+    def integrate_with_systemd
+      return unless ENV["SD_NOTIFY"]
```
ekohl, line 324: Why would you want this non-standard check? Isn't the presence of `NOTIFY_SOCKET` sufficient? If needed, you can also check if the system was booted with systemd. See https://github.com/sj26/puma-plugin-systemd/blob/8490c0e3136ddd201bef93c6747e1057f0d74a03/lib/puma/plugin/systemd.rb#L159-L167

lib/puma/launcher.rb
```diff
@@ -311,6 +313,32 @@ def prune_bundler
...
+
+      require 'sd_notify'
+      return log "Systemd integration failed. It looks like you're trying to use systemd notify but don't have sd_notify gem installed" unless defined?(SdNotify)
+
+      require 'puma/systemd'
+
+      log "* Enabling systemd notification integration"
+
+      @events.on_booted { SdNotify.ready }
+      @events.on_stopped { SdNotify.stopping }
```
ekohl, line 335: Untested, but based on what https://github.com/sj26/puma-plugin-systemd does I think this also works:

```suggestion
      @events.on_booted { SdNotify.ready }
      @events.on_stopped { SdNotify.stopping }
      @events.on_restart { SdNotify.reloading }
```

lib/puma/systemd.rb
```diff
@@ -0,0 +1,32 @@
...
+
+module Puma
+  class Systemd
+    def initialize(events)
+      @events = events
+    end
+
+    def start_watchdog
+      usec = Integer(ENV["WATCHDOG_USEC"])
+      return warn "systemd Watchdog too fast: #{usec}" if usec < 1_000_000
```
ekohl, line 13: I think this should be fatal. If you're not going to respond to the watchdog, systemd will probably enter a restart loop.

lib/puma/systemd.rb
```diff
@@ -0,0 +1,32 @@
...
+
+      sec_f = usec / 1_000_000.0
+      # "It is recommended that a daemon sends a keep-alive notification message
+      # to the service manager every half of the time returned here."
+      ping_f = sec_f / 2
+      log "Pinging systemd watchdog every #{ping_f.round(1)} sec"
+      Thread.new do
```
ekohl, line 20: Does anything ever stop this thread? On a restart, can multiple watchdog threads end up running?

ekohl (commented):
This would make part of https://github.com/sj26/puma-plugin-systemd obsolete. However, that plugin also launches a status thread. I've also left some inline comments about what the plugin does and you could also implement.

nateberkopec:
Thanks a bunch for the review @ekohl!

ekohl:
This PR has stalled, but I'd love to see this functionality in Puma. Anything I can do to help?

ehelms:
@acmh The @theforeman project is keenly interested in seeing this available, today we rely upon https://github.com/sj26/puma-plugin-systemd and first class support through Puma would be a great addition. Is there anything I can do to help you close this out? I'm happy to take over the PR and see it to the finish line for you.

acmh:
@ehelms, of course, you can take this PR and modify what you think necessary. I can't continue working on that. But I am here to provide any context.

PR #2267, opened 2020-05-15
Fix for spawning subprocesses with fork_worker option

wjordan:
### Description
This fixes a bug in `fork_worker` mode (#2099) where an application that spawns subprocesses can sometimes get `nil` instead of an expected `Process::Status` object when waiting on the child (see https://github.com/puma/puma/pull/2099#discussion_r425927136). `test_fork_worker_spawn` demonstrates the regression and currently fails on `master`: https://github.com/puma/puma/blob/2d9b72ecc1e5ae214a964b7d6b692f1740c584a3/test/test_integration_cluster.rb#L171-L183

This PR is one of two approaches for a fix- it stores the list of forked worker pids and the `SIGCHLD` handler calls non-blocking `Process.wait` on each of them, removing the pids that have exited. This is similar to `#wait_workers`.

Worth noting that this approach results in a `wait` syscall for every worker for every subprocess spawned by worker 0's application, which might cause some performance impact for applications which spawn lots of subprocesses and/or have a large number of workers. (I haven't benchmarked in detail so I don't know if it really matters or how big an impact it might be.)

An alternative approach (aacd629575a06eb5c381f57dd39e07b50b337eb2) avoids a `SIGCHLD` handler and instead spawns a separate thread for each forked worker that calls blocking `Process.wait`, to automatically clean up each process as it exits. This is similar to `Process.detach`- it avoids the extra syscall per worker * subprocess, at the cost of an extra thread per worker. That approach was discouraged in https://github.com/puma/puma/pull/2099#discussion_r377972777 but I thought worth mentioning in any case.

### DIFF
--- a/History.md
+++ b/History.md
@@ -45,6 +45,7 @@
   * Fix `UserFileDefaultOptions#fetch` to properly use `default` (#2233)
   * Improvements to `out_of_band` hook (#2234)
   * Prefer the rackup file specified by the CLI (#2225)
+  * Fix for spawning subprocesses with fork_worker option (#2267)
 
 * Refactor
   * Remove unused loader argument from Plugin initializer (#2095)
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -248,6 +248,7 @@ module Puma
       $0 = title
 
       Signal.trap "SIGINT", "IGNORE"
+      Signal.trap "SIGCHLD", "DEFAULT"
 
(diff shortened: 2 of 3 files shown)


nateberkopec:
I'm OK with this approach as fork_worker is experimental. If someone runs into a perf issue w/this fix we can fix it at that point. No need to worry about theoretical bottlenecks on a feature that we're not sure is any good yet.

PR #2269, opened 2020-05-17
Load JSON at runtime

### LINKED ISSUE #2206: Regression in 4.3.3 with require json
**Describe the bug**
Hello, https://github.com/puma/puma/pull/1801 introduced a regression in 4.3.3 when using puma with prune_bundler and control-app if you issue a phased restart with an app with a different json gem version you end up with the following error :

```
/home/ylecuyer/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/bundler-2.1.4/lib/bundler/runtime.rb:312:in `check_for_activated_spec!': You have already activated json 2.3.0, but your Gemfile requires json 2.1.0. Prepending `bundle exec` to your command may solve this. (Gem::LoadError)
```
**Puma config:**

```ruby
#

#
port        ENV.fetch("PORT") { 3000 }

#

# processes).
#

#
# preload_app!

prune_bundler

plugin :tmp_restart
```

This is the default rails new template for config/puma.rb I have only added prune_bundler

**To Reproduce**

https://youtu.be/6VYPTLrFKIg

```
rails new puma-json
cd puma-json
bundle exec puma -p 6000 -w 4 --control-url 'tcp://127.0.0.1:9293' --control-token foo
```

Edit Gemfile and add:

```
gem 'json', '2.1.0'
```

Send phased restart signal to puma

**Expected behavior**

Workers should restart successfully no matter the json version used in the app.

MSP-Greg:
### Description

Fixes loading JSON when Gemfile contains JSON versions that differ from the default gem or previously loaded versions.  Various versions of Bundler have had issues with default gems, as it also loads some for them.

This PR changes JSON loading to runtime.

Closes #2206


ylecuyer:
It is working with my reproduction example ;)

I have mixed feelings about this PR as it fixes the issue but isn't really future proof

MSP-Greg:
> but isn't really future proof

In what sense?  Loading/requiring at runtime is essentially the same as autoload.

ylecuyer:
What I meant was if someone in a future commit adds a `require 'json'` on top of a new file we won't notice it until it breaks again

MSP-Greg:
Given the possible issues with default gems, I think it would be worthwhile to autoload json, psych/yaml, and openssl.  Using autoload removes the need for adding multiple requires when loading at runtime.

I'll have a look at all three.

nateberkopec:
I'm gonna roll with this for now, we can do a longterm fix later.

PR #2241, opened 2020-04-29
[Draft] Various test-suite speed/reliability improvements

wjordan:
### Description

Spike with a bunch of different changes focused on improving the speed and reliability of the test suite. The goal was just to quickly experiment with running tests faster and with fewer flaky failures, and to share some of the high-level ideas for early discussion. Next step would be to break out the most useful fixes/changes into smaller PRs to be merged separately.

#### Highlights

- Higher parallelization (increase `MT_CPU` from default 2 to 10)
- Reduce duration of various sleeps and timeouts in tests (e.g., `sleep 1` to `sleep 0.01`)
- Replace `UniquePort` implementation with port 0 in all tests (`UniquePort` has a race condition that can cause intermittent port conflicts) (#2270)

The following changes have since been broken out into separate smaller PRs:

- ~Fire `on_booted` after server starts (fixes #2212)~ (#2431)
- ~Catch more stream-interruption errors (`EPIPE`, etc) that can happen during shutdown/thread-exit races~ (#2312)
- ~Catch and handle ThreadPool::ForceShutdown errors throughout entire threadpool lifecycle~ (#2271)
- ~Add mutex for adding connections to reactor based on `queue_requests`~ (#2271 / #2377)
- ~Fix some race conditions / edge-cases in Reactor (process multiple triggers in single cycle, re-add timeout when client-request isn't ready to be added back to thread pool [issue #2282])~ (#2279)
- ~Move shell tests into Minitest class so they can be run in parallel~ (#2240, #2346, #2347)

#### Performance

* Ubuntu 2.7
  * Before: [2m 8s](https://github.com/puma/puma/runs/624252018)
  * After: [13s](https://github.com/wjordan/puma/runs/628509391)
* Windows 2.7
  * Before: [53s](https://github.com/puma/puma/runs/624252212)
  * After: [12s](https://github.com/wjordan/puma/runs/628509572)
* MacOS JRuby
  * Before: [3m 8s](https://github.com/puma/puma/runs/624252140)
  * After: [32s](https://github.com/wjordan/puma/runs/628509526)

### Your checklist for this pull request

- [ ] I have reviewed the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.
- [ ] I have added an entry to [History.md](../blob/master/History.md) if this PR fixes a bug or adds a feature. If it doesn't need an entry to HISTORY.md, I have added `[changelog skip]` the pull request title.
- [ ] I have added appropriate tests if this PR fixes a bug or adds a feature.
[...]

PR #2270, opened 2020-05-17
Stabilize CI testing? [changelog skip]

MSP-Greg:
### Description
CI testing is unstable.

The first four commits are each for a single test file and fix small issues.

The fifth commit ('test/helper.rb - revise UniquePort.call, no port 0') always returns a unique port number.  The sixth commit removes the use of 'port 0' from several test files.

### General Info

Currently, 'port 0' is used to determine an open port.  That works fine for normal serial testing.  But, Puma CI is using parallel testing, and ports are opened in both threads and forked or spawned processes.  Combine this with the fact that the job runners are also running code, and the potential for an 'open' port be either opened or repeated exists.  These errors aren't repeatable by using `--seed` values.

### Actions

Yesterday GH Action had some issues.  Today, the CI seems stable, the only thing I've noticed is TruffleRuby failing in the 'test\shell' scripts after the test suite passed.

Let's see if this passes...

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -45,9 +45,25 @@ def hit(uris)
 end
 
 module UniquePort
-  def self.call
-    TCPServer.open('127.0.0.1', 0) do |server|
-      server.connect_address.ip_port
+  @mutex = Mutex.new
+  @port = 9000
+
+  class << self
+    def port_in_use?(port, host = '127.0.0.1')
+      TCPServer.open(host, port).close
+      false
+    rescue Errno::EADDRINUSE
+      true
+    end
+    private :port_in_use?
+
+    def call
+      @mutex.synchronize {
+        begin
+          @port += 1
+        end while port_in_use? @port
+        @port
+      }
     end
   end
 end
--- a/test/test_binder.rb
+++ b/test/test_binder.rb
@@ -15,6 +15,13 @@ class TestBinderBase < Minitest::Test
     @binder = Puma::Binder.new(@events)
   end
 
+  def teardown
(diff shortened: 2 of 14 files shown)


wjordan:
See #2241 for a different pass at fixing the issue by replacing `UniquePort` with using port 0 to allocate a unique available port for each test, then reading the allocated port (either from API or log output) to determine where to send client requests.

I think there is a race condition with `UniquePort` in integration tests where a previously free port passed via configuration to a subprocess may start being used by another parallel test before the subprocess actually binds to it itself. To get around this we either need to use port 0 consistently throughout the test suite (the approach I took in the other PR), or `UniquePort` would need to track available ports it returns to avoid returning the same port twice, in addition to not allowing port 0 to be used in any tests (as done in this PR). This becomes a more prominent issue when increasing parallelization beyond 2 threads (current default).

MSP-Greg:
I'll have a look at #2241.  This PR doesn't change anything in the lib folder, just test files.

I think we're both in agreement that UniquePort either can't exist or it needs to change.

Your PR is probably a better solution, I'll look at it soon.  A lot of code, but with some things that's the only option...

wjordan:
> A lot of code, but with some things that's the only option...

Note that #2241 is still a bit of a mess with various fixes jumbled together- the `UniquePort`-related parts should only affect test files as well, it's just not isolated from all the other work.

nateberkopec:
I liked @wjordan's approach to solving the unique port issue in #2241.

nateberkopec:
Closing in favor of the approach in #2241.

PR #2271, opened 2020-05-17
Better error handling during force shutdown

wjordan:
### Description

Refactors tests covering shutdown behavior to be less flaky (the current tests depend on fragile `sleep` timing that isn't 100% deterministic), and also fixes a couple of subtle timing edge-cases around force-shutdown behavior:

- Only allow `ForceShutdown` to be raised in a thread during specific areas of the connection-processing cycle (marked by `with_force_shutdown` blocks), to ensure that the raised error is always rescued and handled cleanly. (This fixes an issue where the `force_shutdown_after: 0` option throws uncaught exceptions from the threadpool on shutdown)
- ~Add `@queue_mutex` and helper methods `if_queue_requests` and `queue_client` to `Server` to handle an edge case where the reactor shuts down while a thread is concurrently queuing a client. (#2337)~ A fix for this issue was merged in #2377.

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -232,7 +232,9 @@ module Puma
           if @queue_requests
             process_now = client.eagerly_finish
           else
-            client.finish(@first_data_timeout)
+            @thread_pool.with_force_shutdown do
+              client.finish(@first_data_timeout)
+            end
             process_now = true
           end
         rescue MiniSSL::SSLError => e
@@ -244,7 +246,7 @@ module Puma
           client.close
 
           @events.parse_error e, client
-        rescue ConnectionError, EOFError => e
+        rescue ConnectionError, EOFError, ThreadPool::ForceShutdown => e
           client.close
 
           @events.connection_error e, client
@@ -422,7 +424,11 @@ module Puma
               check_for_more_data = false
             end
 
-            unless client.reset(check_for_more_data)
+            next_request_ready = @thread_pool.with_force_shutdown do
+              client.reset(check_for_more_data)
+            end
+
+            unless next_request_ready
               @shutdown_mutex.synchronize do
                 return unless @queue_requests
                 close_socket = false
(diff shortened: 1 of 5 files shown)


lib/puma/client.rb
```diff
@@ -244,19 +244,17 @@ def eagerly_finish
     end # IS_JRUBY
 
     def finish(timeout)
-      return true if @ready
```
nateberkopec, line 247: Hm, what's the impact of changing the return value here? Any?

nateberkopec: Just looking through the rest of the code, seems like we never checked the value.

lib/puma/server.rb
```diff
@@ -344,12 +343,33 @@ def handle_check
       return false
     end
 
+    # Runs the provided block if the Reactor queue is enabled.
+    # Returns block return value if queue is enabled, or nil if queue is not enabled.
+    def if_queue_requests
+      @queue_mutex.synchronize do
+        if @queue_requests
```
nateberkopec, line 350: Since queue_requests is set at configure-time, I think we can move this outside the mutex, right?

lib/puma/server.rb
```diff
@@ -344,12 +343,33 @@ def handle_check
...
+        end
+      end
+    end
+    private :if_queue_requests
+
+    # Queues the provided client
+    def queue_client(client, timeout)
+      if_queue_requests do
+        client.set_timeout timeout
+        @reactor.add client
```
nateberkopec, line 361: Would it be possible to keep concurrency as a concern of the Reactor rather than the Server? It sort of feels like we could refactor this a little further to keep concurrency concerns out of this class.

nateberkopec:
I think my only concerns are the concurrency-awareness being added to Server when I think it hasn't had many concurrency-related reponsibilities yet. I'm wondering if we can shove all of that into Reactor instead so we keep Puma's main "concurrency-aware" classes as Reactor and Threadpool.

nateberkopec:
Sounds good, we can block this PR on that one.

wjordan:
Rebased after #2377 was merged, which fixed the second issue addressed by this PR (I've crossed out that part in the PR description above). This also means that the remaining part of this PR is no longer blocked on #2279 so it's ready for another review.

PR #2279, opened 2020-05-21
Reactor refactor

wjordan:
### Description
The `Reactor` class was getting pretty complicated, hard to reason about and had a tricky bug I was working on fixing (#2282), so this PR is a refactoring pass with a focus on simplicity and more carefully separating concerns between the related classes.

The Reactor has a simple purpose- run a select loop on a collection of IOs, with the added feature of also waking up an IO when a specified timeout has been reached. With the help of ~`SortedSet` and~ `Queue` and using the built-in `Selector#wakeup` feature, the Reactor class can focus on this one task while being a bit easier to understand.


nateberkopec:
I'm gonna wait for Evan on this. One comment from me is that the Reactor class had a lot of docs before and I'd like to keep it a similar level of documentation.

lib/puma/reactor.rb
```diff
@@ -1,405 +1,121 @@
...
-        if diff < 0.0
-          @sleep_for = 0
-        else
-          @sleep_for = diff
-        end
-      end
+    # Wakeup all objects that have timed out.
+    def timeout!
+      # @ios is sorted by timeout so we can efficiently take only timed out objects.
+      @ios.take_while {|timeout_at, _| timeout_in(timeout_at) == 0}.each(&method(:wakeup!))
```
evanphx, line 92: This needs to be broken out into multiple lines and probably expressions. It's quite different to figure out at present.

lib/puma/reactor.rb
```diff
@@ -1,405 +1,121 @@
...
-    # so that the first element to timeout will be at the front of the
-    # array. Then a value to sleep for is derived in the call to `calculate_sleep`
-    def add(c)
-      @mutex.synchronize do
-        @input << c
-        @trigger << "*"
-      end
+    # Register the object with the selector and add to @ios.
+    def register(obj)
+      @ios << @selector.register(obj.last, :r).value = obj
```
evanphx, line 97: The precedence of this line makes my head hurt. I'm unsure what is being passed to <<, please break it up into multiple expressions.

test/test_puma_server.rb
```diff
@@ -374,7 +374,7 @@ def test_status_hook_fires_when_server_changes_states
   end
 
   def test_timeout_in_data_phase
-    @server.first_data_timeout = 2
+    @server.first_data_timeout = 0.01
```
evanphx, line 377: Why was this changed?

wjordan: This was to just help reproduce #2282 more quickly, but doesn't need to be committed so I will remove

lib/puma/client.rb
```diff
@@ -104,6 +104,15 @@ def set_timeout(val)
       @timeout_at = Time.now + val
     end
 
+    # Number of seconds until the timeout elapses.
+    def timeout
+      [@timeout_at - Time.now, 0].max
+    end
+
+    def <=>(other)
```
nateberkopec, line 112: This feels weird to me conceptually. Clients are the same if they time out at the same time?

nateberkopec:
Left a quick comment but I'm gonna need more time to read the changes to reactor + server

lib/puma/client.rb
wjordan: This method was used by `SortedSet` (which calls `#sort!` under the hood to arrange its items), it wasn't meant to convey identity, only ordering. To avoid confusion I changed `@timeouts` back to an Array that calls `#sort_by!` to arrange its items in 51757928a24ae8e3f1980866d1104b79704a758b.

wjordan:
Updated:
- Replaced `SortedSet` with Array based on feedback
- Tweaked the return code path of `Server#process_client` to avoid introducing a `return` from `ensure` - I've learned it is [dangerous](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/EnsureReturn) (it silently eliminates any exceptions passing through including fatal kill-thread signal)

- Rebased and resolved conflicts with #2250, but I have one open question related to this- a call to `Events#connection_error` was added to just one of the three `ConnectionError` rescue clauses (raised when a client disconnects or times out while the server is reading the request), when there was no logging on any of them previously. This PR refactors those three code paths into a single unified `#client_error` implementation.
  Note a comment in `Reactor` (one of the three `ConnectionError` rescues) explaining why logging isn't recommended in that case: https://github.com/puma/puma/blob/f7b09ddd4696ed77c1d781a366003279b7bbddce/lib/puma/reactor.rb#L229-L231

  The open question is: Should we add a call to `#connection_error` for _all_ client `ConnectionError` exceptions (seems like this isn't recommended), only in certain specific cases (to match the exact behavior introduced by #2250), or none of them (to match the original behavior)?

wjordan:
Two related refactoring thoughts that would go in slightly different directions from this PR:

- If adding a package dependency on [`timers`](https://github.com/socketry/timers) is acceptable, [`Timers::Group`](https://github.com/socketry/timers/blob/master/lib/timers/group.rb) could be leveraged to handle the timeout tracking/sorting logic. See 2530b45cbef126c7711b6e23fd626f9dba903887 for an example of what that would look like on top of this PR.

- I also noticed that [`Async::Reactor`](https://github.com/socketry/async/blob/master/lib/async/reactor.rb) (from the [`async`](https://github.com/socketry/async) gem) overlaps a bit with the purpose of Puma's `Reactor` and integrates `timers` in its own very similar select loop. With a few upstream tweaks it might be possible to use `Async::Reactor` directly and either eliminate this class from Puma entirely or make it a much smaller/simpler wrapper.

wjordan:
It turns out it's quite possible to use `Async::Reactor` here with minimal changes, here's a proof of concept: ce53007ed073332e070b5b762eae788ce29b3909 (code is in [`reactor.rb`](https://github.com/puma/puma/blob/ce53007ed073332e070b5b762eae788ce29b3909/lib/puma/reactor.rb) and [`Server#queue_client`](https://github.com/puma/puma/commit/ce53007ed073332e070b5b762eae788ce29b3909#diff-ef4c610f0ce605eb90f2810856394fd5R243-R269)).

I like how simple and readable the extension code ended up, though the underlying `async` library is a much heavier dependency to take on- probably too drastic a change for now, I thought it was an interesting alternative to check out in any case.

nateberkopec:
> I like how simple and readable the extension code ended up, though the underlying async library is a much heavier dependency to take on

Yes, agreed (sometimes I find myself regretting nio4r, although it doesn't cause any new issues). 

Will come back to review everything else soon

wjordan:
> Will come back to review everything else soon

Let me know if you have a chance to review everything else, and if there's anything else I can do to help the next review pass on this.

wjordan:
> Rebased and resolved conflicts with #2250, but I have one open question related to this- a call to Events#connection_error was added to just one of the three ConnectionError rescue clauses [...]

Based on the #2371 issue reports I'm assuming the answer to this question is that we want to revert to the original consistent behavior on this 😅

nateberkopec:
@wjordan This is good to go, but needs a rebase after #2377.

wjordan:
OK, rebased and ready for another review. Some changes done as part of the rebase:

- After the rebase this PR removes `@shutdown_mutex`, effectively reverting the changes from #2377 since the race-condition is handled by Reactor so the extra synchronization should no longer be necessary in `Server`.
- Simplified [`Client#finish`](https://github.com/puma/puma/pull/2279/files#diff-9faf7093e45c4c686939f8f0c5464aa7R255-R262) and added [`Server#with_force_shutdown`](https://github.com/puma/puma/pull/2279/files#diff-ef4c610f0ce605eb90f2810856394fd5R480-R486), to ensure `client.timeout!` is _always_ called whenever a force-shutdown occurs in blocks where it's allowed. This is related to #2271 and fixes an intermittent test failure in `TestPumaServer#test_force_shutdown` of the sort `Expected 408 to match nil` (see [example](https://github.com/puma/puma/runs/1193795795?check_suite_focus=true#step:8:457)). This could possibly be merged separately but the PR was modifying some of the same code anyway so I figured it would be simpler to fix as part of this.

cjlarose:
@wjordan I found that this branch (tested on de2f108fcb0a0b15076e2caac1ff8d4c32f2df27) actually re-introduces a problem previously fixed by your own #2122: `Unable to add work while shutting down`

```
/usr/src/app/vendor/cache/puma-de2f108fcb0a/lib/puma/thread_pool.rb:186:in `block in <<'
/usr/src/app/vendor/cache/puma-de2f108fcb0a/lib/puma/thread_pool.rb:179:in `synchronize'
/usr/src/app/vendor/cache/puma-de2f108fcb0a/lib/puma/thread_pool.rb:179:in `with_mutex'
/usr/src/app/vendor/cache/puma-de2f108fcb0a/lib/puma/thread_pool.rb:184:in `<<'
/usr/src/app/vendor/cache/puma-de2f108fcb0a/lib/puma/server.rb:290:in `reactor_wakeup'
/usr/src/app/vendor/cache/puma-de2f108fcb0a/lib/puma/reactor.rb:47:in `rescue in add'
/usr/src/app/vendor/cache/puma-de2f108fcb0a/lib/puma/reactor.rb:43:in `add'
/usr/src/app/vendor/cache/puma-de2f108fcb0a/lib/puma/server.rb:416:in `process_client'
/usr/src/app/vendor/cache/puma-de2f108fcb0a/lib/puma/thread_pool.rb:145:in `block in spawn_thread'
2020-10-01 22:51:38 +0000 Read: #<RuntimeError: Unable to add work while shutting down>
curl: (22) The requested URL returned error: 500 Internal Server Error
```

It seems like it's possible now for a thread in the ThreadPool to add a client to the Reactor after the Reactor has started to shutdown. That alone isn't a problem (there's even explicit code to handle this case), but it is possible for `Server#reactor_wakeup` to try to add the client to the ThreadPool after the server has already called `ThreadPool#shutdown`. In that case, the exception `Unable to add work while shutting down` is raised and the client gets a 500 error response. In practice, this affects availability during phased restarts, hot restarts, and shutdowns.

I have a reproducible test case here: https://github.com/cjlarose/puma-phased-restart-errors/tree/reactor_reactor

This uses MRI on Ubuntu in a Docker container. You might have to run it a while in order to produce the failure.

cjlarose:
Unrelated: This branch also fixes flakiness in a few tests like `test_halt_unix` and `test_stop_unix` on TruffleRuby. The old Reactor didn't quite handle reading from the `@ready` pipe reliably on that platform, but the new Reactor basically replaced all of that pipe-reading code. Nice work!

wjordan:
@cjlarose thanks for catching this! It may be a subtly different bug from the one in #2122, since I'm pretty sure there's a still-passing test that covers the issue described in that PR.

I have a couple ideas on how to fix the issue (one of which is to leave the mutex in place), but I'll also spend some time on writing a test that might reliably trigger this bug to prevent future regressions.

cjlarose:
> I'll also spend some time on writing a test that might reliably trigger this bug to prevent future regressions.

That'd be awesome. I've also written a test that does something similar: it just performs a bunch of hot restarts on a single-mode puma server while concurrently performing a bunch of requests. The expectation is that all clients eventually get a successful response.

https://github.com/cjlarose/puma/commit/c203014814b875a76c44fcc7224a8d2aa883f9ec

It doesn't pass on all platforms just yet because of various issues in puma, but I'm working on fixing those problems. If you come up with a way to test the `Unable to add work while shutting down` behavior more precisely, though, that'd be great!

lib/puma/reactor.rb
```diff
@@ -1,394 +1,106 @@
...
+            @timeouts.sort_by!(&:timeout_at)
+          end
         end
+      rescue StandardError => e
+        STDERR.puts "Error in reactor loop escaped: #{e.message} (#{e.class})"
+        STDERR.puts e.backtrace
+        retry
       end
+      # Wakeup all remaining objects on shutdown.
+      @timeouts.each(&@block.method(:call))
```
cjlarose, line 87: I think we can pass the block directly, no?

```ruby
      @timeouts.each(&@block)
```

cjlarose (commented):
The most recent concurrency fixes look good. Just a minor comment.

nateberkopec:
This is really top-notch work. I'm so happy this can be merged.

lib/puma/reactor.rb
wjordan: Yes of course, silly me! Harmless enough since this has been merged, but worth slipping a fix into a future PR.

PR #2280, opened 2020-05-21
Increase query string max length to 64k chars

### LINKED ISSUE #2273: Increase QUERY_STRING max length
I am working an app that uses a GET request with a relatively long query string. Currently, Puma allows `1024 * 10` chars (presumably octets/bytes), s. https://github.com/puma/puma/blob/beab7f01b8c66524d2db68aeccbc34b31264b984/ext/puma_http11/puma_http11.c#L57. In our case that’s not enough. Now, if the above sounds familiar that’s probably due to a issue https://github.com/puma/puma/issues/2134 which was fixed by https://github.com/puma/puma/pull/2168. However, I don’t quite understand how this PR is actually fixing the original issue since the length of path is different from the length of the query params.

Ideally, I’d like to be able to configure the query string max length but that’s probably a lot of work since it’s currently determined at compile time if I understand correctly. Hence I’d like to propose simply increasing the max to 64k characters (or octets as per above). 32k or 16k would also work for us technically but 64k seems more future proof. 

I’ve tried to determine if there’s a max length defined in the RFCs but there doesn’t seem to be one, s.:

- https://tools.ietf.org/html/rfc7230
- https://tools.ietf.org/html/rfc3986#section-3.4

However, there’s some research mentioned in this [StackOverflow answer](https://stackoverflow.com/questions/812925/what-is-the-maximum-possible-length-of-a-query-string#812962). It suggests that that all modern browsers will submit query strings longer than the proposed 64k chars. 

In our case, the alternative to increasing the query string max length would be re-engineering a particular form. While this will work in the particular instance that lead to this feature request overall it rather feels like bandaid as we can’t control how our customers use our product and the lengthy forms they can build with it. In other words, this could happen again and then there’s probably not much that we could do technically.

Would you be interested in/open to a PR?

ur5us:
Closes  #2273


nateberkopec:
Sorry, this is going to be a "no" from me. Increasing the size of this allocation for all users of Puma is not appropriate. I'm OK with a "configure-these-buffers-at-compile-time" feature as described in #2273.

ur5us:
@nateberkopec That’s a bit harsh to close the PR that fast. I was hoping that you’d at least let it sit there for a week or 2 to gather some more feedback.

nateberkopec:
I appreciate your effort in opening up a PR - I am very thankful for every contribution we get. But, I did say on the original issue what kind of solution I would merge. 

The truth is that we can't keep increasing these alloc limits every time someone asks, as everyone will end up with bloated memory allocations and slow parsing. If you need it for your particular use case, though, as long as it doesnt violate any RFC we should allow you to configure it. That's the solution I've proposed.

Of course, the option you have avilable to you today is to maintain your own branch with these values changed. I can promise we're not going to touch them (until we make them configurable OFC), so it would be easy to maintain a patch on top of the current version.

ur5us:
@nateberkopec 

> The truth is that we can't keep increasing these alloc limits every time someone asks

Why not? I’m not try to troll. I’ve linked to multiple RFCs to highlight that current limit was chosen at random and there’s no good explanation why it is the way it is. Maybe I’m missing something though. Moreover though, I’m actually providing a proper fix for https://github.com/puma/puma/issues/2134 as https://github.com/puma/puma/pull/2168 that was used to close it fixed a different length problem which wasn’t even requested.

> end up with bloated memory allocations and slow parsing

If I understand the code correctly then even the current code won’t block “bloated” memory allocations, at least on the C side. The parser will allocate as much memory as it needs. I’m happy to be corrected here but I don’t see a limit in the parser. However, I’m also convinced that this argument won’t even matter for most deployments. Most peeps will deploy Puma behind something like Nginx and then proxy their requests. And it is Nginx that is already limiting the size (length, s. `http2_max_field_size 4k;`). Similar limits can be configured for all sorts of things. Interestingly, header length is allowed to `1024 * (80 + 32)` which is almost twice as large as what I’m proposing for query string max length!

Basically, I’m arguing that Puma is prohibiting a valid use case and I’m not the first one to ask for a solution, s. the mentioned issue https://github.com/puma/puma/issues/2134. Moreover, Nginx et al. will already limit how much data is allowed.

Issue #1839, opened 2019-07-11
Unable to retrieve body from "Transfer-Encoding: chunked" request


dpalumbo:
Does Puma support `Transfer-Encoding: chunked`? Recent PRs make me think it does, but I can't find a clear answer. I'm able to receive the request but the body content is empty.

1) Send a POST request with chunked encoding. 
```
curl -H "Transfer-Encoding: chunked" -H "Content-Type: application/json" -d '{"abc": "123"}' -X POST http://localhost:3000/aendpoint
```

2) Check value of raw body received in a Rails controller: 
```ruby
request.raw_post
""
```

I've seen these PRs: 
https://github.com/puma/puma/pull/1607
https://github.com/puma/puma/pull/1812

But it's not clear to me if they solve the issue of receiving the body of a POST.

### Expected behavior

Body should be populated with data

### Actual behavior

Body is empty

### System configuration

**Ruby version**:  2.5.0
**Rails version**: 5.2.2
**Puma version**: 4.0.0

hahmed:
Hey @nateberkopec 

I looked into this issue, does not look like an issue with puma -- I suspect its a rails bug (managed to repro, just seeing if I can find a fix for it).

1. create rack file and run `bundle exec bin/puma -t 5 test/rackup/chunked-post.ru & PID1=$!`
```
run lambda { |env|
  STDERR.puts env['rack.input'].read # => {"abc": "123"}
  [200, {"Content-Type" => "application/json"}, [""]]
}
```
2. send post chunked request (below) --  the output is correctly coming through via puma
`curl -H "Transfer-Encoding: chunked" -H "Content-Type: application/json" -d '{"abc": "123"}' -X POST http://localhost:9292`

I can see a few tests for post chunked requests, none with `application/json` not sure if we need it? but let me know if we do (we also don't have a form post chunked test too)

hahmed:
Here is a repro of the issue https://github.com/hahmed/chunk-puma-test -- hope this clears up why I think this is a rails issue and not a puma issue

PR #2287, opened 2020-05-29
Set CONTENT_LENGTH for chunked requests

eugeneius:
### Description

Fixes https://github.com/puma/puma/issues/1839.

Chunked requests don't contain a Content-Length header, but Puma buffers the entire request body upfront, which means it can determine the length before dispatching to the application.

The Rack spec doesn't mandate the presence of the CONTENT_LENGTH header, but it does refer to it as a "CGI key" and draws a distinction between it and the HTTP Content-Length header:

https://github.com/rack/rack/blob/v2.2.2/SPEC.rdoc

> The environment must not contain the keys HTTP_CONTENT_TYPE or HTTP_CONTENT_LENGTH (use the versions without HTTP_). The CGI keys (named without a period) must have String values.

RFC 3875, which defines the CGI protocol including CONTENT_LENGTH, says:

https://tools.ietf.org/html/rfc3875#section-4.1.2

> The server MUST set this meta-variable if and only if the request is accompanied by a message-body entity. The CONTENT_LENGTH value must reflect the length of the message-body after the server has removed any transfer-codings or content-codings.

"Removing a transfer-coding" is precisely what Puma is doing when it parses a chunked request.

RFC 7230, the most recent specification of HTTP 1.1, includes a pseudo-code algorithm for decoding chunked requests that roughly matches the behaviour implemented here:

https://tools.ietf.org/html/rfc7230#section-4.1.3

I say "roughly" because we don't update the Transfer-Encoding header or parse trailers.

### DIFF
--- a/History.md
+++ b/History.md
@@ -46,6 +46,7 @@
   * Improvements to `out_of_band` hook (#2234)
   * Prefer the rackup file specified by the CLI (#2225)
   * Fix for spawning subprocesses with fork_worker option (#2267)
+  * Set `CONTENT_LENGTH` for chunked requests (#2287)
 
 * Refactor
   * Remove unused loader argument from Plugin initializer (#2095)
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -420,7 +420,10 @@ module Puma
           raise EOFError
         end
 
-        return true if decode_chunk(chunk)
+        if decode_chunk(chunk)
+          @env[CONTENT_LENGTH] = @chunked_content_length
+          return true
+        end
       end
     end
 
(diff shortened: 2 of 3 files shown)


TheRusskiy:
Hey, is there any chance this could be backported to 4.x version? Seems like a pretty serious omission and 5.x is full of breaking changes.
Thank you.

nateberkopec:
@TheRusskiy We do not have a previous-major-version maintenance policy. I've created a new `4-3-stable` branch, if you'd like to make a pull request against it and backport.

TheRusskiy:
@nateberkopec backporting here: https://github.com/puma/puma/pull/2496

PR #2288, opened 2020-05-31
Dynamically resolve rack.multithread/rack.multiprocess

### LINKED ISSUE #2174: Fix rack.multithread`/`rack.multiprocess
We're currently always setting these to true and false respectively. It should be dependent on your config.

FTLam11:
Closes #2174.

This PR dynamically resolves `rack.multithread` and `rack.multiprocess`.

### DIFF
--- a/lib/puma/binder.rb
+++ b/lib/puma/binder.rb
@@ -20,13 +20,12 @@ module Puma
       @inherited_fds = {}
       @activated_sockets = {}
       @unix_paths = []
-      @conf = conf
 
       @proto_env = {
         "rack.version".freeze => RACK_VERSION,
         "rack.errors".freeze => events.stderr,
-        "rack.multithread".freeze => resolve_option(:multithread, @conf),
-        "rack.multiprocess".freeze => resolve_option(:multiprocess, @conf),
+        "rack.multithread".freeze => conf.options[:max_threads] > 1,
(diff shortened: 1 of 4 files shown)


nateberkopec:
Thanks for your contribution. Ideally, the user should not have to add any config settings. We already know if their app `multithread` if the threadpool max size is > 1, and we know it's multiprocess if `workers` >= 1. So the extra config settings aren't necessary here IMO.

nateberkopec:
Also, the values that need to change in response to config are here: https://github.com/puma/puma/blob/b046d6053aad504b9a03453347c462a0360956b4/lib/puma/binder.rb#L26

The problem is that these values are frozen and constant.

FTLam11:
@nateberkopec Appreciate the explanation/hints, thank you!

lib/puma/binder.rb
```diff
@@ -383,5 +385,14 @@ def loopback_addresses
     def socket_activation_fd(int)
       int + 3 # 3 is the magic number you add to follow the SA protocol
     end
+
+    def resolve_option(key, conf)
```
nateberkopec, line 389: I see what you're going for here, but I'd prefer to just resolve these values on lines 28-29 without a method.

FTLam11: Fair call, thanks for the quick review.

PR #2292, opened 2020-06-06
Run CI using the default version of Bundler [changelog skip]

eregon:
### Description

Run CI using the default version of Bundler
* Which is Bundler on Ruby 2.2 and Bundler 2 on Ruby 2.3+.

This used to fail on TruffleRuby (fixed in https://github.com/oracle/truffleruby/issues/1984) and on Ruby 2.3 (due to https://github.com/ruby/setup-ruby/issues/51, now fixed).

Do not set the --path for bundle install
* This is buggy with Ruby 2.3 + Bundler 2.1.4, see https://github.com/ruby/setup-ruby/issues/51

cc @MSP-Greg, I'm not sure why the `--path` was set but it doesn't seem needed.

### DIFF
--- a/.github/workflows/puma.yml
+++ b/.github/workflows/puma.yml
@@ -48,7 +48,7 @@ jobs:
           if ('${{ matrix.ruby }}' -lt '2.3') {
             gem update --system 2.7.10 --no-document
           }
-          bundle install --jobs 4 --retry 3 --path=.bundle/puma
+          bundle install --jobs 4 --retry 3
 
       - name: compile
         run:  bundle exec rake compile
--- a/.github/workflows/puma.yml
+++ b/.github/workflows/puma.yml
@@ -37,7 +37,6 @@ jobs:
         uses: MSP-Greg/setup-ruby-pkgs@v1
         with:
           ruby-version: ${{ matrix.ruby }}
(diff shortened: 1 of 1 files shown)


nateberkopec:
@MSP-Greg all you on this one

MSP-Greg:
@nateberkopec

Thanks.

@eregon and I spent some of today working on changes to the actions used here and elsewhere, and both of us ran Puma CI in our forks to test them.

There may be some changes tomorrow.  I'm tired of thinking about Actions workflows right now...

dentarg:
Re: `--path`, it was added in https://github.com/puma/puma/pull/2126, not sure in reaction to what. 

In the past, the old build system, Travis, used `--path`: https://github.com/puma/puma/pull/1893#discussion_r386039768

eregon:
Right, `--path` is often used for caching.
Bundler caching is coming in https://github.com/ruby/setup-ruby/pull/56 and setting an explicit value conflicts with that (i.e., does not use the cache), so let's remove it.

MSP-Greg:
I think `--path` is being discouraged in favor of `bundle config`, so probably best to remove.  Caching can be done in another PR.

LGTM

Issue #2299, opened 2020-06-26
Puma MiniSSL Filehandle Leak


JohnPhillips31416:
**Describe the bug**
When running with JRuby and configured with a Puma::MiniSSL::Context with a keystore file path MiniSSL will open 2 filehandles to the keystore file on every SSL REST request.  These filehandles do not get closed until a subsequent Java garbage collection occurs.

**Expected behavior**
Ideally MiniSSL should only open/read the keystore file once, not on each https request.  A simpler fix is just to properly close the 2 FileInputStream objects after use.

https://github.com/puma/puma/blob/477479adce25795389a22cc67430a672727caa39/ext/puma_http11/org/jruby/puma/MiniSSL.java#L148
https://github.com/puma/puma/blob/477479adce25795389a22cc67430a672727caa39/ext/puma_http11/org/jruby/puma/MiniSSL.java#L149

**Desktop (please complete the following information):**
 - OS: Linux
 - Puma Version 4.3.5

nateberkopec:
Nice catch. Want to open a PR with your change?

JohnPhillips31416:
I'll do this when I have time, we should probably refactor so the keystore is only read from disk once.

JohnPhillips31416:
fix is here https://github.com/puma/puma/pull/2302

MSP-Greg:
Closed by #2302

Issue #1125, opened 2016-11-01
MiniSSL JRuby Java ext needs #init? and #shutdown


olleolleolle:
It seems that the MiniSSL C extension has two new methods that are not present in the Java version of the extension.

This commit https://github.com/puma/puma/commit/46416cb49ed2f16614f019cee969bb8f5d0a6146 added the C extension methods.

### Steps to reproduce

1) Run the tests on jruby-9.1.5.0 in Travis, see its output

### Expected behavior

There should be no mention of NoMethodError for MiniSSL::Engine.

### Actual behavior

Noting that the usage of [MiniSSL::Engine#init?](https://github.com/puma/puma/blob/master/lib/puma/minissl.rb#L88) is a NoMethodError on this platform.

Output in tests:

```
  test_form_submit:					Error in reactor loop escaped: undefined method `init?' for #<Puma::MiniSSL::Engine:0x226affa7> (NoMethodError)
/home/travis/build/puma/puma/lib/puma/minissl.rb:88:in `close'
/home/travis/build/puma/puma/lib/puma/client.rb:112:in `close'
/home/travis/build/puma/puma/lib/puma/reactor.rb:54:in `block in run_internal'
org/jruby/RubyArray.java:2784:in `delete_if'
/home/travis/build/puma/puma/lib/puma/reactor.rb:50:in `block in run_internal'
org/jruby/ext/thread/Mutex.java:151:in `synchronize'
/home/travis/build/puma/puma/lib/puma/reactor.rb:44:in `block in run_internal'
org/jruby/RubyArray.java:1734:in `each'
/home/travis/build/puma/puma/lib/puma/reactor.rb:42:in `run_internal'
/home/travis/build/puma/puma/lib/puma/reactor.rb:153:in `block in run_in_thread'
```

[Link to build with this failure](https://travis-ci.org/puma/puma/jobs/172301308#L338)

### System configuration

**Ruby version**: jruby-9.1.5.0

---

hapan:
Hello, any plan to fix this issue? I am getting such error too.

olleolleolle:
@hapan Thanks for asking - I'm not planning on finishing this investigation by fixing the issue.

davidarnold:
@hapan Currently, this is just a cosmetic error and can be ignored.  The error is trapped and the socket is still closed correctly [here](https://github.com/puma/puma/blob/0b3626091e7ca7013312586a54a2f6c150f861b7/lib/puma/minissl.rb#L114-L116)

[...]

PR #2302, opened 2020-07-01
Move initialization of KeyManagerFactory, TrustManagerFactory to server

JohnPhillips31416:
Move initialization of KeyManagerFactory, TrustManagerFactory to server initialization.  This avoids reading the keystore file twice on every ssl request, and also fixes (https://github.com/puma/puma/issues/2299) a filehandle leak from reading the keystore file without closing it properly.

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,6 +1,7 @@
 ### Master
 * Bugfixes
   * Resolve issue with threadpool waiting counter decrement when thread is killed
+  * Fix filehandle leak in MiniSSL (#2299)
 
 ## 5.0.0
 
--- a/ext/puma_http11/org/jruby/puma/MiniSSL.java
+++ b/ext/puma_http11/org/jruby/puma/MiniSSL.java
@@ -22,6 +22,7 @@ import javax.net.ssl.SSLException;
 import javax.net.ssl.SSLPeerUnverifiedException;
 import javax.net.ssl.SSLSession;
 import java.io.FileInputStream;
+import java.io.InputStream;
 import java.io.IOException;
 import java.nio.Buffer;
 import java.nio.ByteBuffer;
@@ -32,6 +33,8 @@ import java.security.NoSuchAlgorithmException;
 import java.security.UnrecoverableKeyException;
 import java.security.cert.CertificateEncodingException;
 import java.security.cert.CertificateException;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.Map;
 
 import static javax.net.ssl.SSLEngineResult.Status;
 import static javax.net.ssl.SSLEngineResult.HandshakeStatus;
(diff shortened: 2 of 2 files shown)


MSP-Greg:
LGTM & thanks.  I cancelled the CI, as truffleruby-head is failing, and freezes on macOS.  All The JRuby passed with this, but I'm not a good choice for reviewing Java code...

JohnPhillips31416:
The pr does 2 things:  I close the filehandles after use, and I move the keystore operations from _initialize_, which is called on each ssl request, to _server_, which is called on _add_ssl_listener_.  The question is whether moving the code in this way is valid for all reasonable/known use cases.

eregon:
@MSP-Greg Could you @-mention me if you see CI issues on TruffleRuby? This particular hang should be solved in the latest nightly.

MSP-Greg:
@eregon

Of course.  It did pass the latest CI: https://github.com/puma/puma/actions/runs/159351533

MSP-Greg:
@eregon

@JohnPhillips31416 has done a good (and appreciated) job with this.

Care to review it?  Everything he's said re Puma makes sense, and you are associated with that Java group (@Oracle Labs ), right?  Being a Windows type, filehandle leaks are a pita.  Thanks.

eregon:
This touches the JRuby native extension, so I'm not really familiar with it.
Maybe @headius could review it?

MSP-Greg:
@JohnPhillips31416

Got a question and a plea for help.  Working on CI, and trying to get the SSL tests working on JRuby, as several are bypassed.  There is a method `Puma::MiniSSL::Engine#init?` which is used in `Puma::MiniSSL::Socket#should_drop_bytes?`, and it's missing in the Java MiniSSL implementation.  The C implementation:

https://github.com/puma/puma/blob/e2b419341153ff2206e1cb84fca1663639d0ef30/ext/puma_http11/mini_ssl.c#L419-L425

The man page for the OpenSSL call is https://www.openssl.org/docs/man1.1.1/man3/SSL_in_init.html, and it's used to determine the handshake status.

You may have also seen logs showing the missing method error.

I noticed several calls to `engine.getHandshakeStatus()` in MiniSSL.java, but that's about as far as I'd like to take it.

If you have time, could you look at adding the method?

JFYI, there are some constants missing re available SSL protocols.  Ideally they would query for whether the protocol is available, but for the time being, I may add them to minissl.rb based on the current protocols available in Java.  With TLSv1.3 and OpenSSL 3, that may be changing in the future...

JohnPhillips31416:
@MSP-Greg this is actually discussed in https://github.com/puma/puma/issues/1125, we're running with the monkey patch described there.  I can try to do a PR which encodes this logic into the minissl ruby code.

MSP-Greg:
@JohnPhillips31416

Thanks.  Not knowing Java very well, nor knowing in what way it processes a handshake, I added the following to minissl.rb in one of the JRuby conditionals:
```ruby
      class Engine
        def init?
          true
        end
      end
```

Since JRuby SSL CI has been bypassed in the past, I've now got it enabled, and the above seems to work.  But. it's logging some errors that I'm not sure should be, all related to clients using protocols that are not accepted by the server.

Any idea if this is a really bad idea?

JohnPhillips31416:
Pretty sure it should return false.  Here https://github.com/puma/puma/blob/e2b419341153ff2206e1cb84fca1663639d0ef30/lib/puma/minissl.rb#L167 it looks like bytes are drained/discarded if init? or shutdown? are true.  Otherwise this seems like an appropriate fix.

JohnPhillips31416:
@MSP-Greg seems reasonable.

headius:
The Java patch looks ok to me. JRuby only supports Java 8 or higher, so you could use the "try-with-resources" syntax to safely handle those FileInputStream:

```java
try (FileInputStream fis = new FileInputStream(...)) {
  // code code code
} // fis will be closed automatically without an explicit finally block
```

JohnPhillips31416:
@headius unfortunately Puma is built with Java 1.5 syntax

headius:
@JohnPhillips31416 So let's change that?

```diff
diff --git a/Rakefile b/Rakefile
index 231be50b..c839b013 100644
--- a/Rakefile
+++ b/Rakefile
@@ -49,6 +49,8 @@ else
   # Java (JRuby)
   Rake::JavaExtensionTask.new("puma_http11", gemspec) do |ext|
     ext.lib_dir = "lib/puma"
+    ext.source_version = "1.7"
+    ext.target_version = "1.7"
   end
 end
```

All supported versions of JRuby require Java 8 or higher. JRuby 9.0, released 5 years ago, required Java 7.

You have to go back to the long-unsupported JRuby 1.7.x to get Java 6 support, and since those versions only support Ruby 1.9.3 features I doubt Puma would even still work.

rake-compiler really should default to 1.8 but I suppose 1.7 is fine for any 9.0.x stragglers.

headius:
PR to update rake-compiler to 1.7: https://github.com/rake-compiler/rake-compiler/pull/172

nateberkopec:
I think with the rake-compiler change merged, this is OK to merge I believe?

MSP-Greg:
PR #2355 updates rake-compiler to `"~> 1.1.1"` in Gemfile.  `1.1.1` includes the fix mentioned above by @headius, so this is ready to merge?

JohnPhillips31416:
LGTM

nateberkopec:
Thanks for all the help everyone!

PR #2303, opened 2020-07-03
Improve MiniSSL for users without OpenSSL available

JuanitoFatas:
### Description

SSL constants removed in https://github.com/puma/puma/commit/3a127f7d407eb2abea94b6e15e35863446899abb. And for people who did not install from OpenSSL extension, this [change](https://github.com/puma/puma/commit/ee8596237fb25cbc3a908586f4100e37ff4d77c6) will end up here. https://github.com/puma/puma/blob/91e57f4e173343746e122e8bb850b0244f508484/ext/puma_http11/mini_ssl.c#L12 which `HAVE_OPENSSL_BIO_H` is false, then arrived at this code path https://github.com/puma/puma/blob/91e57f4e173343746e122e8bb850b0244f508484/ext/puma_http11/mini_ssl.c#L555-L562

resulted in `SSLError`.

Check if the `OPENSSL_VERSION` constant is defined to fix this.

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,6 +1,7 @@
 ### Master
 * Bugfixes
   * Resolve issue with threadpool waiting counter decrement when thread is killed
+  * Fix Puma::MiniSSL when OpenSSL is not available (#2303)
 
 ## 5.0.0
 
@@ -63,7 +64,7 @@
   * JSON parse cluster worker stats instead of regex (#2124)
   * Support parallel tests in verbose progress reporting (#2223)
   * Refactor error handling in server accept loop (#2239)
-  
+
 ## 4.3.4/4.3.5 and 3.12.5/3.12.6 / 2020-05-22
 
 Each patchlevel release contains a separate security fix. We recommend simply upgrading to 4.3.5/3.12.6.
--- a/lib/puma/minissl.rb
+++ b/lib/puma/minissl.rb
@@ -13,7 +13,7 @@ module Puma
 
     # define constant at runtime, as it's easy to determine at built time,
     # but Puma could (it shouldn't) be loaded with an older OpenSSL version
-    HAS_TLS1_3 = !IS_JRUBY &&
+    HAS_TLS1_3 = !IS_JRUBY && defined?(OPENSSL_VERSION) &&
       (OPENSSL_VERSION[/ \d+\.\d+\.\d+/].split('.').map(&:to_i) <=> [1,1,1]) != -1 &&
       (OPENSSL_LIBRARY_VERSION[/ \d+\.\d+\.\d+/].split('.').map(&:to_i) <=> [1,1,1]) !=-1
 


MSP-Greg:
@JuanitoFatas

Thanks.  I started working on this, but got distracted.  Might it be better to not even load the definition of MiniSSL in minissl.rb if Puma is compiled without SSL support?  Same for the ssl methods in Binder and loading of Puma::MiniSSL::ContextBuilder.

Because MiniSSL is compiled in puma_http11, some of the 'requires' need their order changed.  I wanted to add at least one Actions job with SSL support.  Also, I'm not that familiar with the other OpenSSL variants, like libressl.

sj26:
Skipping minissl entirely would also work, but felt like a bigger change. Absolutely do it, if you think that's best! We were opting for the light touch just to fix this loading problem introduced by the 5 beta for demonstration purposes.

sj26:
(To prevent future regressions like, perhaps introducing compile options into the build matrix would work? Not sure if it's worthwhile.)

MSP-Greg:
@sj26

I try to help with SSL issues, and apologies, I never thought about compiling without SSL support.  But, it is a valid need, and should probably be tested.

The other issue that is somewhat mixed with this is, if one compiles with SSL support, then uses Puma without binding to an SSL connection, can it be setup to not load all the SSL related code, including not loading the OpenSSL system libraries.  I kind of stopped when I was trying to see if the code changes could be separated.  A long weekend is a good time, so...

sj26:
I think openssl might be loaded by other aspects of ruby anyway in almost all applications which use puma, we just don't have openssl headers available in production for the bundle to build puma with ssl bindings, so it's probably a pretty uncommon case! We appreciate the attention. Even if this was merged as-is for a fix, and then you considered avoiding minissl entirely for puma without ssl — either way. :-)

(We use unix sockets from puma to another layer which handles ssl.)

MSP-Greg:
> I think openssl might be loaded by other aspects of ruby anyway in almost all applications which use puma

Now that I'm being reminded, I think nio4r always loads Ruby OpenSSL.  Also, not all the tests that use SSL are properly gaurded, so that's more code if we add 'no SSL' to CI.

Let me get the code for fixing building/compiling with `ENV["DISABLE_SSL"]`, and the other issues can be addressed later.

By chance, can you test it?

sj26:
> By chance, can you test it?

Absolutely! We already have a copy of puma running in pre-production using the patch on this branch. But we'd be happy to test anything you'd like to ship here.

I am about to head to bed, but can follow up in the morning [~9am utc+10].

MSP-Greg:
@sj26

Sorry for the delay.  See PR https://github.com/puma/puma/pull/2305.  If you can try that, it would be appreciated.  I haven't added a 'no SSL' build to CI yet.  Soon, hopefully tomorrow.

I was mistaken about NIO.  I believe nio4r 2.5.2 needs OpenSSL for JRuby & Windows, but it doesn't load it.  That was my mistake, and it's been corrected in master, so next release it should be fixed.

JuanitoFatas:
Closing this in favor of #2305. Thanks everyone!

PR #2305, opened 2020-07-06
Allow Puma to compile when built without SSL, load SSL files on demand

MSP-Greg:
### Description

Currently, Puma will not function if built without SSL support.  Also, SSL related files are all loaded regardless, so make them load-on-demand.

#### Changes

1. Allows compiling without OpenSSL & runs 'no ssl' CI on a few jobs
2. SSL test (now `Puma::HAS_SSL` or `Puma.ssl?`) relies on classes that only exist in the compiled extension if properly compiled with OpenSSL.
3. If compiled with 'no ssl', don't load ssl related files in Puma.  Also, don't load Ruby MRI OpenSSL.  JRuby currently  requires it for some reason. (?)

****Commits**** (need to be squashed for bisect)

1. **'Adjust code for compiling without SSL (MRI & JRuby), add ssl detection'** -  Removes MiniSSL.check, replace with Puma::HAS_SSL or Puma.ssl?

2. **'Adjust test files for 'no ssl' compile'**

3. **'Actions - add 'no ssl' workflow, puma-no-ssl.yml'** - three jobs: Ubuntu-20.04 with 2.7 & JRuby, and Windows 2.7

4. **'Update History.md'**

5. **'README.md - add 'SSL Connection Support' section'**

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,6 +1,7 @@
 ## 5.0.0
 
 * Features
+  * Allow compiling without OpenSSL and dynamically load files needed for SSL, add 'no ssl' CI (#2305)
   * EXPERIMENTAL: Add `fork_worker` option and `refork` command for reduced memory usage by forking from a worker process instead of the master process. (#2099)
   * EXPERIMENTAL: Added `wait_for_less_busy_worker` config. This may reduce latency on MRI through inserting a small delay before re-listening on the socket if worker is busy (#2079).
   * EXPERIMENTAL: Added `nakayoshi_fork` option. Reduce memory usage in preloaded cluster-mode apps by GCing before fork and compacting, where available. (#2093, #2256)
--- a/README.md
+++ b/README.md
@@ -29,6 +29,18 @@ $ puma
 Without arguments, puma will look for a rackup (.ru) file in
 working directory called `config.ru`.
 
+## SSL Connection Support
+
+Puma will install/compile with support for ssl sockets, assuming OpenSSL
+development files are installed on the system.
+
+If the system does not have OpenSSL development files installed, Puma will
+install/compile, but it will not allow ssl connections.
+
+If the system has OpenSSL development files installed, but you don't want Puma
+to use ssl connections, set ENV['DISABLE_SSL'] to any value before installing
+Puma.
+
 ## Frameworks
 
 ### Rails
--- a/lib/puma/binder.rb
+++ b/lib/puma/binder.rb
@@ -5,10 +5,16 @@ require 'socket'
 
 require 'puma/const'
 require 'puma/util'
-require 'puma/minissl/context_builder'
 require 'puma/configuration'
 
 module Puma
+
+  if HAS_SSL
+    require 'puma/minissl'
+    require 'puma/minissl/context_builder'
+    require 'puma/accept_nonblock'
+  end
+
   class Binder
     include Puma::Const
 
@@ -155,6 +161,9 @@ module Puma
 
           @listeners << [str, io]
         when "ssl"
+
+          raise "Puma compiled without SSL support" unless HAS_SSL
+
           params = Util.parse_query uri.query
           ctx = MiniSSL::ContextBuilder.new(params, @events).context
 
@@ -245,9 +254,8 @@ module Puma
 
     def add_ssl_listener(host, port, ctx,
                          optimize_for_latency=true, backlog=1024)
-      require 'puma/minissl'
 
-      MiniSSL.check
+      raise "Puma compiled without SSL support" unless HAS_SSL
 
       if host == "localhost"
         loopback_addresses.each do |addr|
@@ -264,7 +272,6 @@ module Puma
       s.setsockopt(Socket::SOL_SOCKET,Socket::SO_REUSEADDR, true)
       s.listen backlog
 
-
       ssl = MiniSSL::Server.new s, ctx
       env = @proto_env.dup
       env[HTTPS_KEY] = HTTPS
@@ -275,8 +282,7 @@ module Puma
     end
 
     def inherit_ssl_listener(fd, ctx)
-      require 'puma/minissl'
-      MiniSSL.check
+      raise "Puma compiled without SSL support" unless HAS_SSL
 
       if fd.kind_of? TCPServer
         s = fd
--- a/lib/puma/minissl.rb
+++ b/lib/puma/minissl.rb
@@ -10,7 +10,6 @@ require 'puma/puma_http11'
 
 module Puma
   module MiniSSL
-
     # define constant at runtime, as it's easy to determine at built time,
     # but Puma could (it shouldn't) be loaded with an older OpenSSL version
     HAS_TLS1_3 = !IS_JRUBY &&
@@ -203,8 +202,6 @@ module Puma
       class SSLError < StandardError
         # Define this for jruby even though it isn't used.
       end
-
-      def self.check; end
     end
 
     class Context
--- a/.github/workflows/puma-no-ssl.yml
+++ b/.github/workflows/puma-no-ssl.yml
@@ -1,0 +1,51 @@
+name: No SSL
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    name: >-
+      ${{ matrix.os }} ${{ matrix.ruby }}
+    env:
+      CI: true
+      TESTOPTS: -v
+      DISABLE_SSL: no_ssl
+
+    runs-on: ${{ matrix.os }}
+    if: |
+      !(   contains(github.event.pull_request.title,  '[ci skip]')
+        || contains(github.event.pull_request.title,  '[skip ci]')
+        || contains(github.event.head_commit.message, '[ci skip]')
+        || contains(github.event.head_commit.message, '[skip ci]'))
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - { os: ubuntu-20.04, ruby: 2.7   }
+          - { os: ubuntu-20.04, ruby: jruby }
+          - { os: windows-2019, ruby: 2.7   }
+
+    steps:
+      - name: repo checkout
+        uses: actions/checkout@v2
+
+      - name: load ruby, ragel
+        uses: MSP-Greg/setup-ruby-pkgs@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+          apt-get: ragel
+          brew: ragel
+          mingw: _upgrade_ openssl ragel
+
+      # won't run on Ruby 2.2, see puma.yml
+      - name:  bundle install
+        shell: pwsh
+        run: bundle install --jobs 4 --retry 3
+
+      - name: compile
+        run:  bundle exec rake compile
+
+      - name: test
+        id: test
+        timeout-minutes: 10
+        run: bundle exec rake test:all
--- a/Rakefile
+++ b/Rakefile
@@ -47,6 +47,29 @@ if !Puma.jruby?
   end
 else
   # Java (JRuby)
+  # ::Rake::JavaExtensionTask.source_files supplies the list of files to
+  # compile.  At present, it only works with a glob prefixed with @ext_dir.
+  # override it so we can select the files
+  class ::Rake::JavaExtensionTask
+    def source_files
+      if ENV["DISABLE_SSL"]
+        # uses no_ssl/PumaHttp11Service.java, removes MiniSSL.java
+        FileList[
+          File.join(@ext_dir, "no_ssl/PumaHttp11Service.java"),
+          File.join(@ext_dir, "org/jruby/puma/Http11.java"),
+          File.join(@ext_dir, "org/jruby/puma/Http11Parser.java")
+        ]
+      else
+        FileList[
+          File.join(@ext_dir, "PumaHttp11Service.java"),
+          File.join(@ext_dir, "org/jruby/puma/Http11.java"),
+          File.join(@ext_dir, "org/jruby/puma/Http11Parser.java"),
+          File.join(@ext_dir, "org/jruby/puma/MiniSSL.java")
+        ]
+      end
+    end
+  end
+
   Rake::JavaExtensionTask.new("puma_http11", gemspec) do |ext|
     ext.lib_dir = "lib/puma"
   end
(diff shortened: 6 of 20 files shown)


lib/puma/binder.rb
```diff
@@ -155,6 +156,14 @@ def parse(binds, logger, log_msg = 'Listening')
 
           @listeners << [str, io]
         when "ssl"
+
+          unless @@minissl_loaded
+            require 'puma/minissl/context_builder'
+            require 'puma/accept_nonblock'
+            require 'puma/minissl'
```
sj26, line 163: These two require lists differ, but use the same guard variable. Should they be equivalent? Is it worth guarding — is requiring an already-required path an expensive enough operation balanced against how often these paths are called?

MSP-Greg: Thank you for reviewing and my apologies.  That is wrong; it was a last minute change for JRuby.

As to the guard, I guess Puma isn't binding all the time, but require will always do a directory search so it loads the first matching file in $LOAD_PATH.

Fixed.

sj26:
Thanks for the quick work on this! Sorry it took me a while to test it, complications. But I just gave it a go and it works great. :+1:

lib/puma/binder.rb
```diff
@@ -14,6 +13,8 @@ class Binder
 
     RACK_VERSION = [1,6].freeze
 
+    @@minissl_loaded = false
```
eregon, line 16: Just my opinion, but could you avoid class variables (and use an `@ivar` on the class instead)?
I consider them kind of deprecated and they're pretty slow on all Ruby implementations.

eregon: Alternatively, would checking with `defined?(MiniSSL)` or some other constant work and not need explicit state?

lib/puma/minissl.rb
```diff
@@ -308,5 +307,5 @@ def close
         @socket.close unless @socket.closed?       # closed? call is for Windows
       end
     end
-  end
+  end if IS_JRUBY || (Puma::MiniSSL.check.nil? rescue nil)
```
eregon, line 310: This should be a better check, silently skipping tests if e.g. there is a typo, seems bad.

lib/puma/binder.rb
MSP-Greg: I thought most of the class variable issues related to their use in 'subclasses', but the constant would work fine.

lib/puma/minissl.rb
MSP-Greg: That check is to not load minissl.rb if MRI Puma wasn't compiled with SSL support.  Similar checks for some of the tests also.  The silent skips will only happen when Puma isn't compiled with SSL support.  I'm working with it in my fork (adding a 'none SSL' job) and locally.  JFYI, working with JRuby head locally.

lib/puma/binder.rb
nateberkopec: Agreed with @eregon

MSP-Greg: Updated to use `load_ssl unless defined? ::Puma::MiniSSL::ContextBuilder`.  Using a constant bothers RuboCop

There are CI failures that seem unrelated to this PR, trying to determine the cause...

nateberkopec:
Hmm, how do we want to do CI for this? Just trigger it with an environment variable?

lib/puma/minissl.rb
nateberkopec: I think we can leave a comment (or add a new method to Puma detect) to explain.

MSP-Greg: Ok.  Let me think about that (MRI/JRuby/TruffleRuby).

MSP-Greg:
> Hmm, how do we want to do CI for this? Just trigger it with an environment variable?

I was looking at that this week, and didn't finish.  extconf.rb works with MRI, but I'm not sure about other platforms?

nateberkopec:
Well MRI is the only platform we're worried about supporting w/o SSL, though, right?

MSP-Greg:
> Well MRI is the only platform we're worried about supporting w/o SSL, though, right?

No, I need to check it with JRuby, I'll also check against TruffleRuby, but I think it should behave similar to MRI... (?)

MSP-Greg:
I believe I've got JRuby compiling without OpenSSL (simple changes).

At present, runtime SSL detection is kind of odd (putting that nicely).  I hate using `rescue` to do so.

I think a better way is checking what classes exist in `puma_http11`.

**Question** - Since `puma_http11` is always used, maybe move the `require` to `Puma`, and add the detection in `puma/detect.rb` so we have a constant/method to check?

A goal is to not need to require Ruby's OpenSSL if Puma is compiled without OpenSSL...

nateberkopec:
> Question - Since puma_http11 is always used, maybe move the require to Puma, and add the runtime OpenSSL detection in puma/detect.rb so we have a constant/method to check?

> A goal is to not need to require Ruby's OpenSSL if Puma is compiled without OpenSSL...

Makes sense!

MSP-Greg:
Two items:

1. There are three classes (I think) that rescue MiniSSL or OpenSSL errors (Client, Reactor, & Server).  If one compiles without SSL, these constants are not defined.  I'm thinking of setting the rescue errors to ['LocalJumpError`](https://msp-greg.github.io/ruby_trunk/Core/LocalJumpError.html), which is defined, and will never be raised?

2. I'll separate lib and test changes, but for this PR to bisect, it needs to be squashed.

MSP-Greg:
This is done.  I added a 'no-ssl' workflow that has three jobs.  One is jruby-head, but (as in the normal workflow) it is allowed to fail, and the only way to check it is by inspection.  Should I change that to a standard jruby job?

MSP-Greg:
I just pushed an update.  Before, the code to determine whether Puma was built with ssl in detect.rb was:
```ruby
HAS_SSL = const_defined?(:MiniSSL, false)
```

Updated code is:
```ruby
# at present, MiniSSL::Engine is only defined in extension code, not in minissl.rb
HAS_SSL = const_defined?(:MiniSSL, false) && MiniSSL.const_defined?(:Engine, false)
```

`Puma::MiniSSL` is defined in minissl.rb (using `module MiniSSL`).  So if minissl.rb was loaded first, the detection would fail.  Currently, `Puma::MiniSSL::Engine` is only defined in 'extension' code, so it's not affected by loading minissl.rb...

lib/puma/binder.rb
```diff
@@ -275,9 +282,6 @@ def add_ssl_listener(host, port, ctx,
     end
 
     def inherit_ssl_listener(fd, ctx)
-      require 'puma/minissl'
```
nateberkopec, line 278: Shouldn't we still have the `raise` here, like in `add_ssl_listener`?

lib/puma/binder.rb
```diff
@@ -5,10 +5,16 @@
 
 require 'puma/const'
 require 'puma/util'
-require 'puma/minissl/context_builder'
 require 'puma/configuration'
 
 module Puma
+
+  if HAS_SSL
```
nateberkopec, line 12: What's the purpose of moving these from the top of the file?

lib/puma/client.rb
```diff
@@ -9,13 +9,13 @@ module WaitReadable
...
 require 'puma/detect'
+require 'puma/puma_http11'
 require 'tempfile'
 require 'forwardable'
 
-if Puma::IS_JRUBY
-  # We have to work around some OpenSSL buffer/io-readiness bugs
-  # so we pull it in regardless of if the user is binding
-  # to an SSL socket
+if Puma::HAS_SSL || Puma::IS_JRUBY
```
nateberkopec, line 16: Hmmm... why are we requiring this now on MRI when we didn't need it before?

lib/puma/client.rb
```diff
@@ -40,6 +40,9 @@ class Client
     # no body share this one object since it has no state.
     EmptyBody = NullIO.new
 
+    # if compiled without ssl, MiniSSL doesn't exist
+    SSLError = HAS_SSL ? ::OpenSSL::SSL::SSLError : ::LocalJumpError
```
nateberkopec, line 44: I don't think LocalJumpError is appropriate, StandardError would be fine.

README.md
```diff
@@ -13,11 +13,19 @@ Puma is a **simple, fast, multi-threaded, and highly concurrent HTTP 1.1 server
 
 ## Built For Speed &amp; Concurrency
 
-Puma processes requests using a C-optimized Ragel extension (inherited from Mongrel) that provides fast, accurate HTTP 1.1 protocol parsing in a portable way. Puma then serves the request using a thread pool. Each request is served in a separate thread, so truly concurrent Ruby implementations (JRuby, Rubinius) will use all available CPU cores.
+Puma processes requests using a C-optimized Ragel extension (inherited from
```
nateberkopec, line 16: Did you change anything in here other than 80-char line breaks?

History.md
```diff
@@ -1,6 +1,7 @@
 ## 5.0.0
 
 * Features
+  * Allow compiling without OpenSSL and dynamically load files needed for SSL, add 'no ssl' CI (#2305)
```
nateberkopec, line 4: It looks like for JRuby, for this to work, you need to have DISABLE_SSL set, but for MRI, you don't. Is that true?

nateberkopec (changes requested):

lib/puma/binder.rb
MSP-Greg: I can add it.  I thought that `inherit_ssl_listener` would never be called unless on a restart, but maybe with systemd?

lib/puma/binder.rb
MSP-Greg: We could move it above and use `Puma::HAS_SSL`.  I put it here for the namespace.

This was an issue I was about to post a message about.

Currently, if Puma successfully compiles with OpenSSL, it loads all the files.  Should we make so that it only loads the files if one binds to an ssl socket?

That is trickier...

lib/puma/client.rb
MSP-Greg: Sorry for the poor explanation,  StandardError would catch everything.  What we want is to change the rescue (which is normally catching an ssl related error) to use an error that will never be raised, and do so only when Puma is not compiled with ssl.

README.md
MSP-Greg: I added a sentence to Quick Start, which maybe should be expanded on?  

"Puma expects to find OpenSSL development files when installed/compiled. If you want to compile it without ssl support, set ENV['DISABLE_SSL']."

I can pull this and do a separate PR.

History.md
MSP-Greg: Re MRI, I know how to uninstall 'openssl dev', and it works locally.

I don't know how to uninstall Java's equivalent of 'openssl dev'.  For MRI systems, it's a separate install, I don't think that's the case with Java?

README.md
MSP-Greg: I removed to format changes and just added an ssl section.

lib/puma/client.rb
dentarg: What about we define our own exception class that fulfill that requirement? We can give that class some good name that explains for the reader what's going on

MSP-Greg: @dentarg

I'll look at that.  The rescue hocus-pocus is only needed in reactor.rb & server.rb (client.rb didn't need it, my mistake), and both involve rescuing MiniSSL::SSLError, so that could work.  Not sure where to put it.  AFK for a bit...

lib/puma/binder.rb
MSP-Greg: I added it.

lib/puma/client.rb
MSP-Greg: A previous version of this relied on MiniSSL for detection, detection is now done with namespace children of MiniSSL.

Hence, `MiniSSL::SSLError` can now be defined in `Puma` when it is built without ssl support.  Hence, no need for 'hocus-pocus'.

MSP-Greg:
Interesting.  The last push (which passed CI here) here also ran in my fork.  The only failure was on
ubuntu-18.04 truffleruby-head, see:

https://github.com/MSP-Greg/puma/runs/1109525323?check_suite_focus=true#step:7:599

Note that the test suite passed, but it just barely timed out, so it shows as a failure.  First time I've seen that, but maybe extend the test step's timeout to 12?  Hate to do that...

MSP-Greg:
I noticed that test_puma_server_ssl.rb would not run as a single test file:
```
rake test TESTOPTS=--verbose TEST=test/test_puma_server_ssl.rb
```

Fixed.

MSP-Greg:
Removed Actions TruffleRuby RuboCop commit

README.md
```diff
@@ -29,6 +29,18 @@ $ puma
...
 
+## SSL Connection Support
+
+Puma will install/compile with support for ssl sockets, assuming OpenSSL
+development files are installed on the system.
+
+If the system does not have OpenSSL development files installed, Puma will
+install/compile, but it will not allow ssl connections.
+
+If the system has OpenSSL development files installed, but you don't want Puma
```
nateberkopec, line 40: I feel like this is a little unclear. If you don't bind Puma to SSL, why should you need to use `DISABLE_SSL`?

As written, this makes it sound like anyone using Puma w/o SSL _must_ set `DISABLE_SSL`.

MSP-Greg: > If you don't bind Puma to SSL, why should you need to use DISABLE_SSL?

You don't have to, but 'If the system has OpenSSL development files', it will compile the SSL functions into `puma_http11.so/jar` and will load the OpenSSL libraries/dlls when using MRI.  It also loads Ruby OpenSSL on MRI. 
 I believe using `DISABLE_SSL` stops all of that.

> this makes it sound like anyone using Puma w/o SSL must set DISABLE_SSL.

Didn't mean to imply that.  Maybe a rephrasing is in order.  I meant to make it clear that it's optional.  But, I didn't say anything about benefits...

BTW, thanks for reviewing.

nateberkopec: Yeah I just ended up removing it. I think it's too confusing. Good feature to have but not one that needs explaining right in README.md. 

Thanks so much for your work on the test suite over the last 2 weeks Greg, it's so much better now.

PR #2306, opened 2020-07-06
Use Tempfile for sockets and state files in tests

### LINKED ISSUE #2150: Open all state/sock files in their own temporary directory, and rm them after receiving stop/kill signal
I was just wondering why my tests were failing locally, then I realized that I had the sock/state files in my test directory because they hadn't been cleaned up properly:

```
/test/test_puma.state
/test/test_server.sock
/test/test_control.sock
```

I used to like not having these in .gitignore because I use `git status` to see if my current state is "clean", but now I have to look for those socket files manually (which is kinda hard because when alphabetically sorted in the test dir they really blend in).

Thoughts anyone?

alexeevit:
### Description
Closes #2150 


test/helpers/integration.rb
```diff
@@ -2,6 +2,7 @@
 
 require "puma/control_cli"
 require "open3"
+#require_relative "../helper"
```
dentarg, line 5: Don't think we need this line? :)

dentarg:
> Closes #2150

I don't think "... and rm them after receiving stop/kill signal" that #2150 begs for, is handled here? (I'm not sure what a good way to handle that looks like). Nonetheless, this looks like a great start!

alexeevit:
> > Closes #2150
> 
> I don't think "... and rm them after receiving stop/kill signal" that #2150 begs for, is handled here? (I'm not sure what a good way to handle that looks like). Nonetheless, this looks like a great start!

Well, I've done one possible implementation of the management of created files. Not sure if it's good design to redefine `#run` but I let you judge 😄

dentarg:
Hehe, creative! :)

I'll let @nateberkopec be the judge 😛

nateberkopec:
Yeah, I don't think `run` is the right place to inject here. I think there are probably other places in Minitest you could take a look at, probably in the signal handling code.

alexeevit:
> Yeah, I don't think `run` is the right place to inject here. I think there are probably other places in Minitest you could take a look at, probably in the signal handling code.

You're right, I found a better place to put my hooks

test/helpers/tmp_path.rb
```diff
@@ -0,0 +1,35 @@
+module TmpPath
+  def capture_exceptions
+    super
+  rescue
+    clean_tmp_paths
+    raise
+  ensure
+    clean_tmp_paths
```
nateberkopec, line 8: ensure is always evaluated, so you don't need the clean_tmp_paths in `rescue`.

nateberkopec (changes requested):
quick fix, otherwise I think it looks brilliant now.

test/helpers/tmp_path.rb
alexeevit: Fixed

PR #2318, opened 2020-07-22
Test fixups, port 0 and timeout handling [changelog skip]

MSP-Greg:
### Description

Before https://github.com/puma/puma/commit/1f40ce031e7df67af4d43f37e295204e824e9b29, each test was wrapped in two threads for timeout guards, one for setup and the test, the other for teardown.  Teardown should be 'bulletproof', so it shouldn't need a timeout thread wrapping it.

The above commit changed it so everything was wrapped in one timeout, but when a test method timed out, that also stopped the teardown code.

The first commit changes it so only the actual test method is wrapped.  There may still be teardown timeout issues, but if so, teardown methods need to be revised so they never timeout.  If they timeout, they're not working correctly, and that may cause issues with subsequent tests.

The second and third commits change test file port handling, and also uses #addr instead of #connected_ports.

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -58,11 +58,27 @@ module TimeoutEveryTestCase
   class TestTookTooLong < Timeout::Error
   end
 
-  def time_it
-    t0 = Minitest.clock_time
-    ::Timeout.timeout(RUBY_ENGINE == 'ruby' ? 60 : 120, TestTookTooLong) { yield }
-  ensure
-    self.time = Minitest.clock_time - t0
+  def run
+    with_info_handler do
+      time_it do
+        capture_exceptions do
+          before_setup; setup; after_setup
+
+          # wrap timeout around test method only
+          ::Timeout.timeout(RUBY_ENGINE == 'ruby' ? 60 : 120, TestTookTooLong) {
+            self.send self.name
+          }
+        end
+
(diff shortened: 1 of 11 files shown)


MSP-Greg:
GitHub is having some issues.  I just pushed an update, and it isn't visible here, nor is it running CI in my fork.  A few changes to test_puma_server_ssl.rb to make it similar to changes in test_puma_server.rb...

nateberkopec:
Cheers Greg, great as usual.

MSP-Greg:
Nate,

Note that I said "Teardown should be 'bulletproof'".  That may not be the case, but I'll keep an eye on it.

JFYI, Puma builds ands tests with current OpenSSL 3.0.0 alpha/master.  And, I hope to be running Ubuntu 20.04 locally in a few weeks.  Feel rather constrained (on Windows) without fork, UNIXSockets, etc...

MSP-Greg:
I forgot to mention that before doing this PR I did some testing re using port 0.  I tried it fast, slow, with threads, with spawn, etc.  Regardless, all OS's on Actions never returned a duplicate port number when running a bunch of calls to find an open port.  Hence, I believe the idea that duplicate port numbers might happen (across threads or spawn's) is unlikely.

PR #2345, opened 2020-08-26
Detect linux properly across all Ruby impls

headius:
JRuby has used "java" for RUBY_PLATFORM for 15 years, so the best way to detect the host OS is to use RbConfig's "host_os" value. This allows JRuby to support the TCP_CORK hack and fix benchmarks of very small request/response cycles.

See jruby/jruby#6366 for a more detailed discussion of TCP_CORK.

### Description

Detect Linux using RbConfig's "host_os" value rather than RUBY_PLATFORM, which is always "java" on JRuby.

### Your checklist for this pull request

- [x] I have reviewed the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.
- [ ] I have added an entry to [History.md](../blob/master/History.md) if this PR fixes a bug or adds a feature. If it doesn't need an entry to HISTORY.md, I have added `[changelog skip]` or `[ci skip]` to the pull request title.
- [ ] I have added appropriate tests if this PR fixes a bug or adds a feature.
- [x] My pull request is 100 lines added/removed or less so that it can be easily reviewed.
- [ ] If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.
- [ ] If this closes any issues, I have added "Closes `#issue`" to the PR description or my commit messages.
- [ ] I have updated the documentation accordingly.
- [ ] All new and existing tests passed, including Rubocop.

PR #2349, opened 2020-08-30
make sure that TCP_CORK is supported

ahorek:
### Description
according to @dentarg feedback https://github.com/puma/puma/pull/2345, we shouldn't use TCP_CORK on JRuby until it's properly supported.

It could also break on very old Linux kernels (< 2.6). Let's just make sure that all Socket constants are usable.

### DIFF
--- a/History.md
+++ b/History.md
@@ -2,6 +2,7 @@
 * Bugfixes
   * Resolve issue with threadpool waiting counter decrement when thread is killed
   * Constrain rake-compiler version to 0.9.4 to fix `ClassNotFound` exception when using MiniSSL with Java8.
+  * Ensure that TCP_CORK is usable
 
 ## 5.0.0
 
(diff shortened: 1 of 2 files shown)


nateberkopec:
Perfect! Thanks for the quick fix.

headius:
@ahorek @nateberkopec It occurs to me that TCP_CORK is independent of TCP_INFO but they are checked together here. Is the TCP_INFO logic in puma necessary for TCP_CORK in any way?

I ask because JRuby currently still does not bind TCP_INFO mostly because the data it carries is a more complex struct than simply a boolean or integer value. If we bind the constant, then we need to be able to support the struct.

If the TCP_INFO logic were handled separately, TCP_CORK could continue to work on JRuby 9.3 (I have merged jruby/jruby#6366) even if we don't have TCP_INFO fully supported...

headius:
FWIW @ahorek if you have some cycles I'd love to support TCP_INFO. Along with mapping the structure, there's another set of constants we'd need to have available in JRuby to navigate it.

Issue #1360, opened 2017-07-09
Allow "use any port" feature


matthijskooijman:
Puma supports systemd socket activation by getting a list of open sockets from systemd and matching that against the sockets it would normally open. This means that the listening parameters (procol, ip, port) must be configured twice (in the systemd .socket file and in the application config) and they must match to make things work.

It would be useful if the application could use a wildcard configuration, meaning "just use whatever systemd passes you", so the listen details only need to be given in systemd, not in the application. I'm not entirely sure how this should work (specify a port number of 0 or -1? A special IP or protocol name? Does this replace the explicitely given list of listen sockets, or is it just one element in the list? Perhaps a completely different interface?), but ideally it would work without changes to applications (which suggests a special IP, port or protocol name would be best?).

nateberkopec:
Binding to the unspecified address won't work for your use case? `puma -b tcp://0.0.0.0 config.ru`

matthijskooijman:
Not sure if you understood my meaning. IIUC that command will still bind to port 80 (or whatever is the default), right? And if I specify a different port in systemd, [this bit of code](https://github.com/puma/puma/blob/master/lib/puma/binder.rb#L92-L106) will not match the systemd port to the requested port 80, and end up trying to open port 80 and then close the port systemd opened.

As for the usecase that promted this: I was running the rails server, which binds to port 3000 by default. I'm using systemd socket activation to let systemd open up port 80 (so rails/puma does not need root to open a port < 1024). However, to let puma actually *use* this port passed in from systemd, I have to pass `--port 80` to rails, so rails requests the same port systemd offers. I was hoping to do something like `--port use-any`, or something like that, so it will just use the systemd socket, regardless of which port/listen address it has.

nateberkopec:
Thanks. I am not a systemd guy, so the real answer is "IDK"! I was just trying to see if any of the behavior we already support re: binding could fill this use case.

FWIW, we tend to support a few more things/weirder scenarios when you use `puma` directly, rather than `rails server`.

dekellum:
Consider that puma offers HTTPS support and that is common when using that support to configure 2 ports: one HTTPS and one HTTP port.  In such a case:

1. There needs to be some means of distinguishing which socket is which on the puma side. (The current implementation does handle that.)
2. Connection parameters like the ssl protocol, key and cert are used only on the puma side.

[...]

PR #2362, opened 2020-09-10
Add option to bind to systemd activated sockets

ekohl:
Currently still a draft, based on https://github.com/puma/puma/issues/1360#issuecomment-689548650. Having it as a PR allows me to refer to this work. That's why the checklist is not filled in yet.

### Description
Systemd can present sockets as file descriptors that are already opened.  By default Puma will use these but only if it was explicitly told to bind to the socket. If not, it will close the activated sockets. This means all configuration is duplicated.

Binds can contain additional configuration, but only SSL config is really relevant since the unix and TCP socket options are ignored.

This means there is a lot of duplicated configuration for no additional value in most setups. This option tells the launcher to bind to all activated sockets, regardless of existing binds.

Note it does not clear configured binds. That means it's still possible to bind to an additional socket not configured as an activated socket.

### DIFF
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -58,6 +58,13 @@ module Puma
 
       @config.load
 
+      if @config.options[:bind_to_activated_sockets]
+        @config.options[:binds] = @binder.synthesize_binds_from_activated_fs(
+          @config.options[:binds],
+          @config.options[:bind_to_activated_sockets] == 'only'
+        )
+      end
+
       @options = @config.options
       @config.clamp
 
--- a/History.md
+++ b/History.md
@@ -6,6 +6,7 @@
   * Integrate with systemd's watchdog and notification features ([#2438])
   * Adds max_fast_inline as a configuration option for the Server object ([#2406])
   * You can now fork workers from worker 0 using SIGURG w/o fork_worker enabled [#2449]
+  * Add option to bind to systemd activated sockets ([#2362])
 
 * Bugfixes
   * Your bugfix goes here <Most recent on the top, like GitHub> (#Github Number)
--- a/docs/systemd.md
+++ b/docs/systemd.md
@@ -129,6 +129,21 @@ Puma will detect the release path socket as different than the one provided by
 systemd and attempt to bind it again, resulting in the exception
  `There is already a server bound to:`.
 
+### Binding
+
+By default you need to configure puma to have binds matching with all
+ListenStream statements. Any mismatched systemd ListenStreams will be closed by
+puma.
+
+To automatically bind to all activated sockets, the option
+`--bind-to-activated-sockets` can be used. This matches the config DSL
+`bind_to_activated_sockets` statement. This will cause puma to create a bind
+automatically for any activated socket. When systemd socket activation is not
+enabled, this option does nothing.
+
+This also accepts an optional argument `only` (DSL: `'only'`) to discard any
+binds that's not socket activated.
+
 ## Usage
 
 Without socket activation, use `systemctl` as root (e.g. via `sudo`) as
--- a/lib/puma/binder.rb
+++ b/lib/puma/binder.rb
@@ -111,6 +111,43 @@ module Puma
       ["LISTEN_FDS", "LISTEN_PID"] # Signal to remove these keys from ENV
     end
 
+    # Synthesize binds from systemd socket activation
+    #
+    # When systemd socket activation is enabled, it can be tedious to keep the
+    # binds in sync. This method can synthesize any binds based on the received
+    # activated sockets. Any existing matching binds will be respected.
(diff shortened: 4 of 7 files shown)


dentarg:
👍 on 699e49510c4f9d734848cb751d9b2df10a2a42ba, I've run into the same thing (perhaps it can be a separate PR?)

MSP-Greg:
@ekohl

Re the patch for minissl.rb, was that because it was compiled without SSL support?  See #2305, which I need to test on JRuby.

ekohl:
Yes, I think it was. This allowed me to continue testing, but then I saw #2305. I suspect that on my Fedora installation I didn't have openssl-devel installed so it probably was compiled without SSL.

MSP-Greg:
@ekohl

Thanks.  I apologize for the lack of 'no ssl' support.  I've helped with updates as newer versions of OpenSSL were released, and didn't consider installations without it.

It's common for Puma to be used with UNIXSockets, so people may want to compile it without OpenSSL.  Re #2305, I  don't work that much with JRuby, and I'd like to check it on that.  I've been busy with other things.

Re your patch to minissl.rb, I can always revert it when I finish #2305.  A lot of apps may load Ruby's OpenSSL for database connections, etc anyway, but I'd like #2305 to **not** load OpenSSL if one chooses to compile Puma without it...

ekohl:
Don't worry about it. It's a development version, not a released version. This PR is still a draft because I only verified it didn't break non-systemd but I have yet to verify actual systemd behavior. It was a trivial workaround that barely held me back.

lib/puma/minissl.rb
```diff
@@ -5,6 +5,8 @@
 rescue LoadError
 end
 
+# Need for OPENSSL_VERSION and OPENSSL_LIBRARY_VERSION
+require 'openssl'
```
nateberkopec, line 9: @MSP-Greg this look right to you?

ekohl:
Since #2303 was merged, I've rebased it on master. That commit was not the focus of this PR.

I'm still in the process of testing it in a systemd env, but some other work came up that requires my attention.

lib/puma/minissl.rb
MSP-Greg: @nateberkopec

I think all the changes in this file should be removed, as #2305 should have fixed issues when compiled without ssl.

Adarsh2204 (commented):
All good👍

ekohl:
I played around with this a bit, but I started to wonder. Ideally you could express something like "clear all binds if you see systemd activated sockets". Right now you can't. Then some other work got in between but I do plan to get back to this.

ekohl:
I have updated this PR and I think that the code itself is good enough to review. It is still lacking tests and docs, but I'd like to know if this is a good approach before writing those.

Locally I tested it by running it directly (`bundle exec bin/rails`) and via a systemd socket. The systemd socket was listening on 0.0.0.0:5000 while configuring the bind on 0.0.0.0:3000. When using the `only` option it showed as:
```
Puma starting in cluster mode...
* Version 5.0.2 (ruby 2.7.1-p83), codename: Spoony Bard
* Environment: production
* Process workers: 2
* Preloading application
* Activated tcp://0.0.0.0:5000
Use Ctrl-C to stop
- Worker 0 (pid: 587260) booted, phase: 0
- Worker 1 (pid: 587265) booted, phase: 0
```

lib/puma/launcher.rb
```diff
@@ -58,6 +58,28 @@ def initialize(conf, launcher_args={})
 
       @config.load
 
+      if @config.options[:bind_to_activated_sockets]
```
nateberkopec, line 61: IMO all of this should live in `Binder`, not in the `Launcher`

nateberkopec:
> It is still lacking tests and docs, but I'd like to know if this is a good approach before writing those.

Approach looks good to me

lib/puma/launcher.rb
ekohl: I just pushed another commit. Is that roughly what you were thinking about?

nateberkopec: Yup!

ekohl:
Rebased and added a bit of text to the docs. Actual code is unchanged.

ekohl:
The failed test timed out but I don't think I can restart it.

nateberkopec:
Don't worry about single non-MRI failures. I can only re-run all, and odds are one of them will timeout randomly again. Greg is working on it and it's better than it used to be, but it's still not perfect.

ekohl:
Rebased to resolve the conflict in History.md. Is there anything else I can do to get this merged?

nateberkopec:
Nope! Will review.

nateberkopec:
Great stuff @ekohl 👍

Issue #2335, opened 2020-08-10
Puma throwing "Transport endpoint is not connected" errors over and over.


vrmerlin:
**Describe the bug**
We are evaluating Puppet Enterprise, which uses Puma. At some random point, puma errors are being generated over and over again until the entire storage space is full. 

Here are the errors we are seeing over and over again:

```
Error in reactor loop escaped: Transport endpoint is not connected - getpeername(2) (Errno::ENOTCONN)
/opt/puppetlabs/server/apps/bolt-server/lib/ruby/gems/puma-4.3.1/lib/puma/minissl.rb:156:in ‘peeraddr’
/opt/puppetlabs/server/apps/bolt-server/lib/ruby/gems/puma-4.3.1/lib/puma/minissl.rb:156:in ‘peeraddr’
/opt/puppetlabs/server/apps/bolt-server/lib/ruby/gems/puma-4.3.1/lib/puma/reactor.rb:239:in ‘rescue in block in run_internal’
/opt/puppetlabs/server/apps/bolt-server/lib/ruby/gems/puma-4.3.1/lib/puma/reactor.rb:218:in ‘block in run_internal’
/opt/puppetlabs/server/apps/bolt-server/lib/ruby/gems/puma-4.3.1/lib/puma/reactor.rb:157:in ‘each’
/opt/puppetlabs/server/apps/bolt-server/lib/ruby/gems/puma-4.3.1/lib/puma/reactor.rb:157:in ‘run_internal’
/opt/puppetlabs/server/apps/bolt-server/lib/ruby/gems/puma-4.3.1/lib/puma/reactor.rb:313:in ‘block in run_in_thread’
/opt/puppetlabs/server/apps/bolt-server/lib/ruby/gems/logging-2.2.2/lib/logging/diagnostic_context.rb:474:in ‘block in create_with_logging_context’
```

Any idea why these errors are being generated? If needed I can work with Puppet, but we are just evaluating their product and don't yet have access to tech support. Since the failures are in puma code I thought I'd try here first.

**Puma config:**

PE uses puma-4.3.1.

nateberkopec:
[...]

PR #2367, opened 2020-09-15
Rescue ENOTCONN

ocowchun:
### Description
Rescue `Errno::ENOTCONN` since it doesn't have socket address

Closes: https://github.com/puma/puma/issues/2335

### DIFF
--- a/History.md
+++ b/History.md
@@ -68,6 +68,7 @@
   * Ensure that TCP_CORK is usable
   * Fix corner case when request body is chunked (#2326)
   * Fix filehandle leak in MiniSSL (#2299)
+  * Rescue ENOTCONN (#2335)
 
 * Refactor
   * Remove unused loader argument from Plugin initializer (#2095)
--- a/lib/puma/minissl.rb
+++ b/lib/puma/minissl.rb
@@ -188,10 +188,14 @@ module Puma
         end
       end
 
-      def peeraddr
+      def peeraddr(swallow: false)
         @socket.peeraddr
-      rescue
-        ["<unknown>"]
+      rescue => e
+        if swallow
+          ["<unknown>"]
+        else
+          raise e
+        end
       end
 
       def peercert
--- a/lib/puma/minissl.rb
+++ b/lib/puma/minissl.rb
@@ -190,6 +190,8 @@ module Puma
 
       def peeraddr
         @socket.peeraddr
+      rescue
+        ["<unknown>"]
       end
 
       def peercert
--- a/lib/puma/reactor.rb
+++ b/lib/puma/reactor.rb
@@ -239,13 +239,7 @@ module Puma
               rescue MiniSSL::SSLError => e
                 @server.lowlevel_error(e, c.env)
                 ssl_socket = c.io
-                begin
-                  addr = ssl_socket.peeraddr.last
-                # EINVAL can happen when browser closes socket w/security exception
-                rescue IOError, Errno::EINVAL, Errno::ENOTCONN
-                  addr = "<unknown>"
-                end
-
+                addr = ssl_socket.peeraddr.last
                 cert = ssl_socket.peercert
 
                 c.close
--- a/lib/puma/reactor.rb
+++ b/lib/puma/reactor.rb
@@ -238,12 +238,11 @@ module Puma
               # SSL handshake failure
               rescue MiniSSL::SSLError => e
                 @server.lowlevel_error(e, c.env)
-
                 ssl_socket = c.io
                 begin
                   addr = ssl_socket.peeraddr.last
                 # EINVAL can happen when browser closes socket w/security exception
-                rescue IOError, Errno::EINVAL
+                rescue IOError, Errno::EINVAL, Errno::ENOTCONN
                   addr = "<unknown>"
                 end
 
--- a/lib/puma/reactor.rb
+++ b/lib/puma/reactor.rb
@@ -239,7 +239,7 @@ module Puma
               rescue MiniSSL::SSLError => e
                 @server.lowlevel_error(e, c.env)
                 ssl_socket = c.io
-                addr = ssl_socket.peeraddr.last
+                addr = ssl_socket.peeraddr(swallow: true).last
                 cert = ssl_socket.peercert
 
                 c.close
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -211,7 +211,7 @@ module Puma
           end
         rescue MiniSSL::SSLError => e
           ssl_socket = client.io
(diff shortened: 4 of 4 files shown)


MSP-Greg:
I looked this earlier, wondering about adding at test for it.  Everything I tried raised a `ConnectionError` instead of an `Puma::MiniSSL::SSLError`.

Also, wondered whether it would be better to replace the following method in `MiniSSL::Socket`:

```ruby
def peeraddr(swallow: false)
  @socket.peeraddr
end
```
with something like:

```ruby
def peeraddr(swallow: false)
  @socket.peeraddr
rescue => e
  if swallow
    ["<unknown>"]
  else
    raise e
  end
end
```

I think it's used four times, three times inside `rescue` code, once outside?

ocowchun:
@MSP-Greg Good point, base on current use of `#peeraddr` I think we can just swallow the error. It doesn't take much effort to modify if somewhere want to handle particular error in the future.

MSP-Greg:
@ocowchun

I noticed that in `Server#normalize_env` we have the following, which calls `Client#peerip`, and sets the value to `LOCALHOST_IP` if `Errno::ENOTCONN` is raised:

https://github.com/puma/puma/blob/18f1810c9cdec3fac50fc49dc35e8b478d637f32/lib/puma/server.rb#L495-L501

Hence, maybe change `Client#peerip` from:

https://github.com/puma/puma/blob/18f1810c9cdec3fac50fc49dc35e8b478d637f32/lib/puma/client.rb#L269-L279

to:

```ruby
def peerip
  @peerip ||= if @remote_addr_header
    (@env[@remote_addr_header] || LOCALHOST_IP).split(/[\s,]/).first
  else
    @io.to_io.peeraddr.last
  end
end
```

Other than cleaning up/shortening the logic, the real change is `@io.peeraddr.last` to `@io.to_io.peeraddr.last`, which allows it work with `MiniSSL::Socket`.

ping @nateberkopec...

nateberkopec:
So, a regular TCPSocket will raise a SocketError as well if it can't get the peer: 

```
TCPSocket.open("www.asdasdasdasdasdkjjkqwlkqwejkljkwqeeq.com", 80) {|sock|
  p sock.peeraddr
}
# SocketError (getaddrinfo: nodename nor servname provided, or not known)

```

So, from my perspective, MiniSSL sockets should have the same contract. That means we should fix this not in the MiniSSL object, but at the location of the callers. Or at least I'm reluctant to.

There's three callsites for peeraddr in Puma rn: Client, Reactor, and Server. Each is doing something slightly different.

I think putting a new abstraction in here would be kind of tough. But we need to think about what happens at each location if peeraddr raises an exception.

MSP-Greg:
Sorry, I should have given a fuller explanation.

In Server & Reactor, the call to `MiniSSL::Socket#peeraddr` happens within a `rescue MiniSSL::SSLError` section, and the value is only used to call `@events.ssl_error`.

In Client, it's only used in `#peerip`, which can be any type of socket.  Hence, adding the `to_io` for an ssl socket, which returns `self` for any type of socket.

Server also contains code to set `env[REMOTE_ADDR]`, which calls `Client#peerip`, but it isn't using the value for Events.

Also, re Events, `peeraddr` is only used for `#ssl_error` calls.

So, I think this is an improvement, but I'm not sure if `MiniSSL::Socket#peercert` will raise and error.  It isn't 'protected' in the current code.

ocowchun:
@MSP-Greg 
>So, I think this is an improvement, but I'm not sure if MiniSSL::Socket#peercert will raise and error. It isn't 'protected' in the current code.

For the `MiniSSL::Socket#peercert` part, 
https://github.com/puma/puma/blob/18f1810c9cdec3fac50fc49dc35e8b478d637f32/lib/puma/minissl.rb#L188-L196

https://github.com/puma/puma/blob/18f1810c9cdec3fac50fc49dc35e8b478d637f32/ext/puma_http11/mini_ssl.c#L427-L464

I have zero knowledge with c, but looks like it will return `nil` when something went wrong. Thus I think `MiniSSL::Socket#peercert` won't raise an error.

MSP-Greg:
@ocowchun

Thanks.  I guess I kind of forgot that many client ssl connections are made **without** a cert.  I'm sure the code is ok.

ocowchun:
@MSP-Greg for the swallow error part, I personally preference maintain the same interface rather than a handy abstraction. Rescue the error in the caller might be a more suitable choice, what do you think?

MSP-Greg:
@ocowchun

I don't have strong feelings about how this is fixed.  But, some thoughts:

1. The original commit you had catches the error in Reactor.  But, the same method is used in Server, and there isn't any rescue for it.  See below.  Can we say for sure that the error can't happen in Server?

2. Puma has several places where exceptions are rescued a bit up the call stack.  That can get tedious to wade through.  Also, handling exceptions for 'attributes' (like `peeraddr`) is different that 'actions', like `socket.close`.

3. At present there are three places where `peeraddr` is called.  Reactor and Server call it only on an MiniSSL::Socket, and Client calls it on a generic socket.  The `MiniSSL::Socket#peeraddr` calls are only to provide a string to the `Events#ssl_error` method.

Given the above, I still prefer to catch in `peeraddr`.  Maybe now or later a comment should be added explaining that `.to_io.peeraddr` should be used if an error needs to be caught upstream.

Regardless, I'll defer to @nateberkopec.  I'm an old guy that has strong opinions, but moving forward is more important.

https://github.com/puma/puma/blob/18f1810c9cdec3fac50fc49dc35e8b478d637f32/lib/puma/server.rb#L211-L218

ocowchun:
@MSP-Greg How about the previous idea you metioned?


```rb
def peeraddr(swallow: false)
  @socket.peeraddr
rescue => e
  if swallow
    ["<unknown>"]
  else
    raise e
  end
end
```

The default behavior of `peeraddr` will follow the interface and we can swallow the error in Reactor and Server, thus several tedious tasks can be saved.

MSP-Greg:
> How about the previous idea you mentioned?

Agreed.

MSP-Greg:
@ocowchun @nateberkopec

I really screwed up.  I'll fix asap.  In #2368, I forgot that I split some attr_* statements to show version info.  Dumb, dumb, dumb...

MSP-Greg:
@ocowchun 

LGTM.

Hate to ask, can you rebase?

ocowchun:
@MSP-Greg Sure.

It's a great experience, thanks you and @nateberkopec to review my PR. 😺

nateberkopec:
Closing in favor of #2375 for a more complete fix. Thanks for getting us started though @ocowchun.

PR #2373, opened 2020-09-19
Consolidate option handling in Server, Server small refactors, doc changes

MSP-Greg:
### Description

In normal use, Server is passed an options hash in its initialize method.  Some code is outside of Server, using attr_accessors, some of which may have been created for CI, etc.  They also display in docs, which isn't needed.

1. server.rb - consolidate option handling, remove `attr_accessor` statements, change other lib files as needed.

2. server.rb - Server had a method defined in `class << self` (`tcp_cork_supported?`), and also a method defined as `self.current`.  Move current into `class << self` block.

3. runner.rb - `remove development?` & `test?`, which are only used to set a value based on options in Server.

4. Change test files to use `Server#instance_variable_get`

5. Add YARD `# @!attribute` tags to methods that are actually attributes (no parameters, getter, setter, or predicate)

For those interested in seeing Puma docs:
current gem https://puma.io/puma
master https://msp-greg.github.io/puma/

- [?] My pull request is 100 lines added/removed or less so that it can be easily reviewed.

### DIFF
--- a/History.md
+++ b/History.md
@@ -6,6 +6,9 @@
 * Bugfixes
   * Your bugfix goes here (#Github Number)
 
+* Refactor
+  * Remove unneeded attr_accessor statements from Server (#2373)
+
 ## 5.0.0
 
 * Features
--- a/lib/puma/runner.rb
+++ b/lib/puma/runner.rb
@@ -54,9 +54,8 @@ module Puma
 
       app = Puma::App::Status.new @launcher, token
 
-      control = Puma::Server.new app, @launcher.events
-      control.min_threads = 0
-      control.max_threads = 1
+      control = Puma::Server.new app, @launcher.events,
+        { min_threads: 0, max_threads: 1 }
 
       control.binder.parse [str], self, 'Starting control server'
 
@@ -69,6 +68,7 @@ module Puma
       @control.binder.close_listeners if @control
     end
 
+    # @!attribute [r] ruby_engine
     def ruby_engine
       if !defined?(RUBY_ENGINE) || RUBY_ENGINE == "ruby"
         "ruby #{RUBY_VERSION}-p#{RUBY_PATCHLEVEL}"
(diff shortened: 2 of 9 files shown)


dentarg:
> 1. remove `attr_accessor` statements

Could be considered breaking changes if someone is working programmatically against `Puma::Server`?

MSP-Greg:
@dentarg

> Could be considered breaking changes

Yes, that is true.  But, given that Puma is essentially a CLI app, I don't think the idea of 'what is the public API' has been given much attention.  Puma is not a young app, and over time, proper encapsulation of objects often leaks.

From my perspective, when I see `attr_accessor` or `attr_writer`, I wonder why write access is given.  In the case of Server, there really isn't any reason for write access to exist.

I suppose I can revert all the attribute changes, but they just clutter up the real methods of the class.  Maybe we should implement 'deprecated' warnings?  I don't know.

I think that the more the API is cleaned up, the easier it is for new contributors to understand how Puma works...

nateberkopec:
I'm OK with removing write access where that write access didn't do anything (e.g. we never recheck the value after server start, so writing to the variable is a no-op).

I don't think we can remove reads without a deprecation policy anymore.

MSP-Greg:
I did check what was passed to the various hooks and the plugin, I'll add back in reader code.

MSP-Greg:
One issue.  In server.rb:83, I changed @max_threads from `16` to `(Puma.mri? ? 5 : 16)`, which aligns it with the new default.

dentarg:
> > 1. remove `attr_accessor` statements
> 
> Could be considered breaking changes if someone is working programmatically against `Puma::Server`?

Sorry, but this time I'm just gonna: https://github.com/puma/puma/issues/2388 😉

MSP-Greg:
I've got an update that adds writers to this commit and marks them deprecated.  Haven't pushed yet.  It was mostly to clean up the following:

```ruby
class Server
  attr_accessor :min_threads, :max_threads
  
  def initialize(app, events=Events.stdio, options={})
    # code
  end
end

class Runner
  def start_server
    server = Server.new(app, events, opts)
    server.min_threads = opts[:min_threads]
    server.max_threads = opts[:max_threads]
    # code
  end
end
```

So, we're passing `opts` to Server.new, then setting two values in `start_server` immediately after using `opts`.  Breaks encapsulation, and adds unneeded code to `start_server`...

Issue #1875, opened 2019-08-01
"Preloader" process for preload + rolling restart


nateberkopec:
https://github.com/puma/puma/pull/1861#issuecomment-517012275

nateberkopec:
@cjlarose kind of reminds me of the puma-wild proposal you made.

cjlarose:
Yeah, I see the connection. In https://github.com/puma/puma/issues/2018#issuecomment-589973247 I described a way to separate the gems of the puma master process from those of the workers in a way that was more robust than the existing `prune_bundler` solution.

The comments in #1861 describe a way to leverage the benefit of `preload_app` (faster worker startup, leverage COW memory) in a system that also supports phased restarts. To start workers, the puma master first forks into what @evanphx calls a "stem-cell" process. The "stem-cell" process preloads the application and then later forks to make the actual worker processes. When it's time to perform a phased restart, the puma master forks a new "stem-cell" process, from which the new workers are forked.

In that new architecture, fixing the kinds of bugs related to #2018 would only require a small change: Whenever the puma master forks to form the "stem-cell" worker-generator process, it must immediately `exec` into a new `ruby` process, removing any references from puma master process's gems. After the `exec`, that process can preload the application and later fork workers. This change can make it possible for CD tools to automatically remove the gems from disk associated with the puma master. It'll also make it possible to upgrade gems like `nio4r` in new versions of the application, without requiring a hard restart, since the gems of the master process are separated from the gems of the worker-generator and worker processes.

The architecture proposed in my comments in #2018 doesn't _require_ the new "stem-cell"/worker-generator process idea, but is definitely compatible with it. This strategy seems like a good way forward.

I might have some time to work on this in the next couple of days if no one's actively working on it already.

nateberkopec:
> no one's actively working on it already.

Just confirming no one in Puma core has spiked this yet.

cjlarose:
I was able to spend some time on this during a company hackathon. Early results are great: I was able to confirm that with the new architecture, it's possible to support both `preload_app` and phased restarts. The cluster integration test suite passes.

The code's not great and it's not ready for review, but it's here: https://github.com/cjlarose/puma/commit/a1f7c1126b1a050f306def22a25030804da1f07e

[...]

PR #2374, opened 2020-09-19
Refactor: extract worker process into separate class [changelog skip]

cjlarose:
### Description

My original motivation for extracting code related to the worker processes is the change described in https://github.com/puma/puma/issues/1875#issuecomment-695143880. I want to be able to create the first worker process (assuming `fork_worker` is on) by first `fork`ing from the master, then immediately `exec`ing into a new process definition with a separate memory space. 

My proof-of-concept in that comment technically _works_ but it's messy because the logic used by the worker process is woven in together with the logic used by the puma master process in `Puma::Cluster`. Worker processes implicitly rely on several instance variables used in `Puma::Cluster`, so for the new process to spin up a new worker, it needs to first spin up a dummy instance of `Puma::Cluster`.

To make it a little bit cleaner, I pulled out the worker logic into a new `Puma::Cluster::Worker` class. The initializer lays out everything that's necessary to construct a new worker process. My hope is that it's clear that this refactor is worth having regardless of the original motivation for it because it makes dependencies of the worker process clear and lowers the cognitive overhead of having worker logic and cluster logic in the same class definition.

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -4,6 +4,7 @@ require 'puma/runner'
 require 'puma/util'
 require 'puma/plugin'
 require 'puma/cluster/worker_handle'
+require 'puma/cluster/worker'
 
 require 'time'
 
@@ -11,10 +12,6 @@ module Puma
   # This class is instantiated by the `Puma::Launcher` and used
   # to boot and serve a Ruby application when puma "workers" are needed
   # i.e. when using multi-processes. For example `$ puma -w 5`
-  #
-  # At the core of this class is running an instance of `Puma::Server` which
-  # gets created via the `start_server` method from the `Puma::Runner` class
-  # that this inherits from.
   #
   # An instance of this class will spawn the number of processes passed in
   # via the `spawn_workers` method call. Each worker will have it's own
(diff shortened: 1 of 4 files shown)


MSP-Greg:
@cjlarose 

Thank you for your work on this.

> cognitive overhead

Nice way of saying...  Having recently spent a few weeks away from thinking about anything Puma, I would agree.

Any thoughts on removing the current `Puma::Cluster::Worker` code from `lib/puma/cluster.rb` and moving it to `lib/puma/cluster/worker.rb` (same namespace of `Puma::Cluster::Worker`) and naming your file `lib/puma/cluster/worker0.rb`, with a namespace of `Puma::Cluster::Worker0`, or something like that?  Just a thought...

Being unable to run fork locally (hope to fix that soon), I can't really test this, but CI is passing.  👍

Lastly, the gem version/update issue would be helped by this.  If `autoload` was used for gems (both external and std-lib),  and maybe Puma files, it might remove issues with "where do we need require statements, and what about CI?", etc

cjlarose:
> Any thoughts on removing the current Puma::Cluster::Worker code from lib/puma/cluster.rb and moving it to lib/puma/cluster/worker.rb (same namespace of Puma::Cluster::Worker) and naming your file lib/puma/cluster/worker0.rb, with a namespace of Puma::Cluster::Worker0, or something like that? Just a thought...

I agree that `Puma::Cluster` is the right namespace for both the new worker class and the existing worker class, but I wasn't crazy about the names `Puma::Cluster::Worker` and `Puma::Cluster::Worker0`, especially since the latter represents _all_ workers, not just the first one. I ended up renaming `Puma::Cluster::Worker` to `Puma::Cluster::WorkerHandle` and making the new class `Puma::Cluster::Worker`. What do you think?

MSP-Greg:
I was mostly concerned with the namespace issue, and the embedded class in Cluster.  As to `WorkerHandle`, I'm not really sure, as it's kind of an different situation.  I can't think of another similar naming issue elsewhere...

JFYI, there are intermittent failures with non MRI jobs in CI, especially on macOS.

AFK.  Thanks, Greg

gingerlime:
I'm a complete outsider to the Puma codebase, but just tried to understand the naming difference you discussed... 

So if I understand this correctly the `WorkerHandle` is the process handle for the worker, which you can then kill, check for liveness etc, and the `Worker` itself contains the actual logic. Right?

What about renaming `Worker` -> `WorkerLogic`? You can then either keep `WorkerHandle` as is? or maybe rename to `WorkerProcHandle` just to be more explicit? because "handle" can feel a little vague without context...

apologies if this doesn't make any sense and I'm butting in to a discussion that's just way over my head :) I know naming is hard, but I hope that my outsider's perspective can be helpful to see things with a fresh outlook.

nateberkopec:
I'm kind of shocked that this clean of an extract was possible. I'm not terribly bothered about the current naming. Need to run through it in more detail but in concept I'm 100% behind this.

lib/puma/cluster.rb
```diff
@@ -248,113 +173,23 @@ def wakeup!
...
-      title += " [#{@options[:tag]}]" if @options[:tag] && !@options[:tag].empty?
-      $0 = title
-
-      Signal.trap "SIGINT", "IGNORE"
-      Signal.trap "SIGCHLD", "DEFAULT"
-
-      fork_worker = @options[:fork_worker] && index == 0
-
       @workers = []
-      if !@options[:fork_worker] || fork_worker
```
nateberkopec, line 261: I never realized this conditional was always true 😞

lib/puma/cluster.rb
```diff
@@ -248,113 +173,23 @@ def wakeup!
...
 
-      server.run.join while restart_server.pop
-
-      # Invoke any worker shutdown hooks so they can prevent the worker
-      # exiting until any background operations are completed
-      @launcher.config.run_hooks :before_worker_shutdown, index, @launcher.events
-    ensure
-      @worker_write << "t#{Process.pid}\n" rescue nil
-      @worker_write.close
+      new_worker = Worker.new index: index,
```
nateberkopec, line 188: Pretty delicious interface here 👍

Issue #2337, opened 2020-08-13
Requests fail during phased restarts, hot restarts, and shutdown


cjlarose:
**Describe the bug**

During a phased restart, some clients do not get successful responses from the server.

Requests fail in a variety of ways. The ones I've observed and that I'm able to reproduce reliably are these:

1. The server closes the connection without first writing a response onto the socket `curl: (52) Empty reply from server`
2. `curl: (56) Recv failure: Connection reset by peer`
3. The exception `Unable to add work while shutting down` is raised, invoking the `lowlevel_error_handler` (by default, this returns a response with response code 500)

**Puma config:**

```ruby
port 3000
workers 2
```

```sh
PUMA_MASTER_PIDFILE=/usr/src/app/puma.pid
bundle exec puma --pidfile "${PUMA_MASTER_PIDFILE}"
```

**To Reproduce**

I've created a repo that reliably reproduces the errors: https://github.com/cjlarose/puma-phased-restart-errors

**Expected behavior**

All requests are passed to the application and the application's responses are all delivered to the requesting clients. The [restart documentation for puma][0] suggests that phased restarts are meant to have zero downtime.

[0]: https://github.com/puma/puma/blob/a24b51b294ff8dd2511f910abe1a0db0d66ed43a/docs/restart.md

**Desktop (please complete the following information):**
 - OS: Linux
 - Puma Version: 4.3.5

MSP-Greg:
@cjlarose 

Thanks for testing Puma.

> During a phased restart, some clients do not get successful responses from the server.

Phased restarts are not meant to be repeatedly done in a short timespan.  There are one or more time settings that affect how long it takes to perform a phased restart.

Do you have unsuccessful responses when doing a single phased restart?

cjlarose:
For the repo that reproduces the problem, I just did a bunch of phased restarts repeatedly to increase the likelihood of failure; I understand these are not normal operating conditions.

But yes, I originally became aware of this problem because it happens to two different applications of mine where we perform phased restarts once every hour or so. Specifically, we encounter the `Unable to add work while shutting down` error in production. From a brief reading of the code, it seems like that error can happen as a result of a request finishing buffering by the `Reactor` after the corresponding thread pool has started its `shutdown` procedure.

MSP-Greg:
> we perform phased restarts once every hour

Ok, that's reasonable.  Using MRI or JRuby?

cjlarose:
MRI 2.6.1 and 2.6.5

[...]

PR #2377, opened 2020-09-22
Prevent connections from entering Reactor after shutdown begins

cjlarose:
### Description

This change addresses one source of connection reset errors related to the shutdown process of `Puma::Sever`. Details about the particular problem are described in https://github.com/puma/puma/issues/2337#issuecomment-694549921 and in the commit messages.

#### Why does this PR not include tests?

The most straightforward way to test this change is to add an integration test that repeatedly fires off requests to a server that is restarting via hot restarts (`SIGUSR2`). The expectation is that no client connection experiences a connection reset error, or better, that every client request gets a successful response.

I've written such an integration test, but unfortunately, there are other sources of connection reset errors that make it so that the test does not pass reliably. I suggest we continue to document and catalog those errors in #2337.

My claim is that change demonstrably fixes one problem that causes connection reset errors and gets us closer to the point that puma reliably handles all client connections that it accepted before shutting down, even if the change doesn't fix every kind of problem. It is possible to use the reproduction steps in #2337 to verify that the error `Error reached top of thread-pool: closed stream (IOError)` is resolved by this PR.

### DIFF
--- a/History.md
+++ b/History.md
@@ -4,7 +4,7 @@
   * Your feature goes here (#Github Number)
 
 * Bugfixes
-  * Your bugfix goes here (#Github Number)
+  * Prevent connections from entering Reactor after shutdown begins (#2377)
 
 ## 5.0.0
 
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -85,6 +85,8 @@ module Puma
       @precheck_closing = true
 
       @requests_count = 0
+
+      @shutdown_mutex = Mutex.new
     end
 
     attr_accessor :binder, :leak_stack_on_error, :early_hints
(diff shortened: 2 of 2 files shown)


wjordan:
It looks like the solution in this PR ended up quite similar to the portion of #2241 involving [`@queue_mutex`](https://github.com/puma/puma/pull/2241/files#diff-ef4c610f0ce605eb90f2810856394fd5R94). I think I triggered this same race condition when running portions of the test suite more quickly.

If I recall, I would guess the refactoring in #2279 may also fix this same race condition by eliminating the bug at the source in the Reactor, rather than in the Server through an extra mutex around its operations. I'm fairly sure that's why I never extracted the `@queue_mutex` code into a separate PR directly, since #2279 makes it unnecessary.

nateberkopec:
>  I recall the refactoring in #2279 may also fix this same race condition at the source in the Reactor, rather than in the Server through an extra mutex around its operations.

That would be great, but that PR is a bigger project. I'll merge this for now, if we can get rid of the extra mutex eventually, great.

Issue #2371, opened 2020-09-17
Puma 5.0.0 HTTP connection error: #<Puma::ConnectionError: Connection error detected during read>


PetrKaleta:
After upgrading to `puma 5.0.0` I am seeing `HTTP connection error: #<Puma::ConnectionError: Connection error detected during read>` error in server logs directly after server starts. This happens just on `Heroku`. Regardless that app is working without any issues...

```
2020-09-17T17:58:33.045281+00:00 app[web.1]: Puma starting in single mode...
2020-09-17T17:58:33.045331+00:00 app[web.1]: * Version 5.0.0 (ruby 2.7.1-p83), codename: Spoony Bard
2020-09-17T17:58:33.045332+00:00 app[web.1]: * Min threads: 10, max threads: 10
2020-09-17T17:58:33.045332+00:00 app[web.1]: * Environment: production
2020-09-17T17:58:33.684588+00:00 app[web.1]: * Listening on http://0.0.0.0:31428
2020-09-17T17:58:33.684657+00:00 app[web.1]: Use Ctrl-C to stop
2020-09-17T17:58:34.163946+00:00 app[web.1]: 2020-09-17 17:58:34 +0000 HTTP connection error: #<Puma::ConnectionError: Connection error detected during read>
2020-09-17T17:58:35.162918+00:00 heroku[web.1]: State changed from starting to up
```

dentarg:
I see this too, on my app I just updated earlier today. It too runs on Heroku. I see many ConnectionError lines, not only directly after server start. Yeah, the app seems to work (it does not have a lot of traffic).

dentarg:
This might be that more logging was added in 5.0.0 and that Puma has always operated in this way.

nateberkopec:
pinging @schneems just because he may have more visibility here, possibly also related to the H13 issue

davejcameron:
I deployed Puma 5 into a production service yesterday and have been noticing increased timeouts. Matches exactly this issue so seems more than just logging.

yosiat:
Getting this error on as well.
With rails development environment straight after upgrading to puma 5.
Doing simple `wrk URL` 

puma 5:
```
Running 10s test @ http://localhost:3000/test/test
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   200.95ms   88.58ms 621.15ms   70.44%
    Req/Sec    21.57     10.24    49.00     61.29%
[...]

PR #2381, opened 2020-09-24
Fix LoadError in CentOS 8 - 'require_relative' should not be necessar…

joergschray:
…y with recent version of rubygems

### Description
The pull request fixes compatibility with CentOS 8: Even for manual installs CentOS 8 separates architecture dependent files from independent ones. Therefore the shared library `puma_http11.so` ends up in a different directory from `puma.rb`, which results in the `require_relative 'puma/puma_http11'` to fail.

Replacing `require_relative` by `require` does not change the behavior in case `puma_http11.so` ends up in the same directory. Apart from that I also consider it best practice in a gem to avoid `require_relative` altogether. Therefore I also replaced the other 
 `require_relative` by `require`.

### Steps to reproduce
Executing the following steps inside a `centos:8` docker container:
```
dnf -y module enable ruby:2.6
dnf -y install ruby ruby-devel gcc make redhat-rpm-config
gem install puma -v5.0.0
puma --version
```
results in
```
	9: from /usr/local/bin/puma:23:in `<main>'
	8: from /usr/local/bin/puma:23:in `load'
	7: from /usr/local/share/gems/gems/puma-5.0.0/bin/puma:6:in `<top (required)>'
	6: from /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:54:in `require'
	5: from /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:54:in `require'
	4: from /usr/local/share/gems/gems/puma-5.0.0/lib/puma/cli.rb:6:in `<top (required)>'
	3: from /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:54:in `require'
	2: from /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:54:in `require'
	1: from /usr/local/share/gems/gems/puma-5.0.0/lib/puma.rb:13:in `<top (required)>'
/usr/local/share/gems/gems/puma-5.0.0/lib/puma.rb:13:in `require_relative': cannot load such file -- /usr/local/share/gems/gems/puma-5.0.0/lib/puma/puma_http11 (LoadError)
```
The paths to the relevant files are
- `/usr/local/share/gems/gems/puma-5.0.0/lib/puma.rb`
- `/usr/local/lib64/gems/ruby/puma-5.0.0/puma/puma_http11.so`

### DIFF
--- a/History.md
+++ b/History.md
@@ -6,6 +6,7 @@
 * Bugfixes
   * Prevent connections from entering Reactor after shutdown begins (#2377)
   * Better error handling during force shutdown (#2271)
+  * Fix LoadError in CentOS 8 (#2381)
 
 * Refactor
   * Change Events#ssl_error signature from (error, peeraddr, peercert) to (error, ssl_socket) (#2375)
--- a/lib/puma.rb
+++ b/lib/puma.rb
@@ -10,8 +10,8 @@ require 'stringio'
 
 require 'thread'
 
-require_relative 'puma/puma_http11'
-require_relative 'puma/detect'
+require 'puma/puma_http11'
+require 'puma/detect'
 
 module Puma
   autoload :Const, 'puma/const'


nateberkopec:
This is a bugfix, so needs a changelog entry. Obviously we don't test on CentOS, but I want to make sure the tests still pass on every other platform anyway.

joergschray:
Added a changelog entry as requested.

MSP-Greg:
@joergschray

Interesting.

> I also consider it best practice in a gem to avoid require_relative altogether

Sometimes there is a desire to bypass $LOAD_PATH, especially with vendored files.

Re CentOS, is it generally true that an `*.rb` will be able to be loaded with `require_relative`?

joergschray:
Pure ruby files could be loaded via `require_relative` in CentOS.

However in a gem you usually open up your own namespace. So everything starting with `puma/` can always be loaded via `require`. Of course, a monkey patch will break the `require`, but that is what a monkey patch is supposed to do.

So unless you put files in a foreign namespace, I do not see a need for using `require_relative` in a gem.

nateberkopec:
Merits of require/require_relative aside, I certainly want CentOS to work. Merging. Thanks for the patch!

ehelms:
Thanks for fixing this. I've hit the same testing on CentOS 7 as well so it is a wider deployment issue. A 5.0.1 with this fix would be greatly appreciated.

nateberkopec:
Yeah I want to get a fix in for #2371, then we'll do 5.0.1.

PR #2395, opened 2020-09-30
Actions: two workflows, MRI & non MRI, separate README badges [changelog skip]

MSP-Greg:
### Description

Currently two Actions workflows exist, one for with SSL, another without SSL.  Note that these could be combined.

But, non MRI jobs are often failing, and we all hate red badges.  Also, things like Actions step timeouts, etc may vary between MRI and non-MRI jobs.

PR creates two new workflows that separate jobs into MRI and non-MRI.  Adjust README badges as such.


dentarg:
I wonder if there's another phrase to use instead of "non MRI" (I can't think of one right now). This came up because I noticed that the filtering over at https://github.com/puma/puma/actions?query=workflow%3AMRI does not work as expected, "MRI" catches both. Oh well :)

MSP-Greg:
That's why I changed the links.  Let me see if a non-breaking space works...

cjlarose:
Is there any value in keeping JRuby and TruffleRuby in the same action? Seems like we'd avoid the name "non MRI" by splitting them.

I guess then we'd get more red X's that way, but at least it'd still be easy to tell if MRI tests are passing or not.

MSP-Greg:
I used 'non_MRI', and filtering etc works.  Non-breaking space did not.

PR #2401, opened 2020-10-01
Test Suite - convert client writes to syswrite [changelog skip]

MSP-Greg:
### Description
Recently I was working with Puma, testing with both client sockets created in Ruby, and a browser.  The Ruby client socket blocked on a `#read` for several seconds, but the browser showed it instantly.  I changed the `#read` call to `#sysread`, and the Ruby client behaved as the browser.  I have also seen similar issues before with reads & writes.

In general, using `#sysread` can be messy, because so many apps need to use other, buffered read methods.  OTOH, using `#syswrite` is easier.

So, I decided to convert all `#write`'s in the test suite to `syswrite`'s, and see if it helped, escpecially with non MRI jobs.

This PR implements that. I think I converted most of the `#write`'s, but I may have missed some.  It does seem to improve non MRI CI, and jobs seem to complete, rather than time out.  There are still intermittent failures/errors.

Lastly, Puma uses both `#write`/`<<` and `#syswrite`.  If this helps stabilize CI, maybe all 'writes' should be converted to `#syswrite`?

### DIFF
--- a/test/helpers/integration.rb
+++ b/test/helpers/integration.rb
@@ -50,7 +50,7 @@ class TestIntegration < Minitest::Test
   def cli_server(argv, unix: false, config: nil)
     if config
       config_file = Tempfile.new(%w(config .rb))
-      config_file.write config
+      config_file.syswrite config
       config_file.close
       config = "-C #{config_file.path}"
     end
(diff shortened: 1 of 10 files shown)


nateberkopec:
Looks like the non-mri failures are the same.

I'd probably want to see a big/obvious improvement in return for having to add "please use syswrite" to my PR review mental checklist...

MSP-Greg:
It seemed that the change resulted in fewer individual test timeouts for non MRI, but that wasn't something I was doing a lot of A/B testing on.

I've been working with it a bit more, and the bottom line is that there are small differences in how Windows and Ubuntu/macOS handle sockets.  For instance, I changed test_puma_server over to sysread, and all Window CI jobs pass, but almost all Ubuntu/macOS jobs fail.

Hence, closing.

PR #2405, opened 2020-10-02
Consolidate fast_write calls in Server, extract early_hints assembly

MSP-Greg:
### Description

`Server#handle_request` is a very long method, and also has many `#fast_write` calls.  `#fast_write` calls, which involve `rescue` clauses, when done with a reasonably defined string in terms of length, should be minimized.

Also, extract code to assemble the 'early hints' string into `#str_early_hints`.

### DIFF
--- a/History.md
+++ b/History.md
@@ -7,6 +7,7 @@
   * Cleanup daemonization in rc.d script (#2409)
 
 * Refactor
+  * Consolidate fast_write calls in Server, extract early_hints assembly (#2405)
   * Remove upstart from docs (#2408)
   * Consolidate option handling in Server, Server small refactors, doc changes (#2389)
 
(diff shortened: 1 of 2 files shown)


MSP-Greg:
I named the method containing the extracted code `str_early_hints`.  Wondering about something like `assemble_response_early_hints` or `resp_early_hints`, etc?  Reason for asking is is I'd like to extract code for the headers, and the method would be named similarly.

Two items: a) intent on extracting/refactoring without moving exception handling code, and b) get some type of coverage working before adding/changing tests to specifically run against the new methods.  Once things are extracted, there will be less of a need to use sockets to test the particulars of the code.

Re coverage, I need to investigate more, but it looks like coveralls is supporting GH Actions.  Not sure about CodeClimate...

PR #2406, opened 2020-10-02
Adds configuration option to set max_fast_inline

### LINKED ISSUE #2361: Request: Make MAX_FAST_INLINE a configuration variable

At the company I work at, we're using puma for an internally-consumed service.  This service is currently only consumed by a single client--another internal service.  We are currently on puma version `3.12.1`.  In puma version `3.12.2` a change was made to prevent one client from spamming requests and hogging the service [I think].
https://github.com/puma/puma/compare/v3.12.1...v3.12.2

This change increases our average request duration by about 25% in a load testing environment--regarding load testing, we've integrated k6 load testing into our ci workflow.  As a result of the increased request duration, we have not yet upgraded from version `3.12.1`.

Our load testing is roughly similar [or attempts to be roughly similar] to behaviour we've experienced under peak load.

Expose `MAX_FAST_INLINE` as a configuration variable.  Then, in use cases such as mine, `MAX_FAST_INLINE` can be set to `Float::INFINITY` to maintain the pre version `3.12.2` behaviour.

`MAX_FAST_INLINE = 10` could continue to be the default value if `MAX_FAST_INLINE` is not set in the configuration, etc.

For the moment, we're holding off on upgrading.  It's not an ideal, long-term solution however.

**Additional context**

This was first described in a separate issue:
https://github.com/puma/puma/issues/1981#issuecomment-686903570

fdel15:
### Description
Closes #2361

Adds a config option to set `max_fast_inline` and override the `MAX_FAST_INLINE` default. You can set this variable to `Float::Infinity` to address use cases that wish to maintain the pre version 3.12.2 behavior (as discussed in the issue).

~Adds an ENV option to set MAX_FAST_INLINE constant. You can set this variable to zero to address use cases that wish to maintain the pre version 3.12.2 behavior (as discussed in the issue).~

~I didn't add any tests because it is not clear to me where to add them. This change affects the `Puma::Server#process_client` method, and it doesn't look like we have any tests set up at the moment. I also did not see any tests written when this change was introduced in the commits referenced in the issue https://github.com/puma/puma/compare/v3.12.1...v3.12.2~


History.md
```diff
@@ -9,6 +9,11 @@
 * Refactor
   * Consolidate option handling in Server, Server small refactors, doc chang (#2389)
 
+## 5.0.3 / 2020-10-02
```
nateberkopec, line 12: Please add any changelog entries to the "master" entry, above this one

lib/puma/const.rb
```diff
@@ -124,7 +124,7 @@ module Const
     # we amortize the cost of going back to the reactor for a well behaved
     # but very "greedy" client across 10 requests. This prevents a not
     # well behaved client from monopolizing the thread forever.
-    MAX_FAST_INLINE = 10
+    MAX_FAST_INLINE = ENV.fetch('MAX_FAST_INLINE', 10).to_i
```
nateberkopec, line 127: Rather than have a constant set by an ENV variable, I think we should add this to the configuration object and then read it from there.

lib/puma/server.rb
```diff
@@ -426,7 +426,7 @@ def process_client(client, buffer)
 
             check_for_more_data = @status == :run
 
-            if requests >= MAX_FAST_INLINE
+            if requests >= MAX_FAST_INLINE && MAX_FAST_INLINE.positive?
```
nateberkopec, line 429: e.g. server should have `@max_fast_inline` set by the options, and then that is read here. Don't check here to see if it's positive, check when you're setting it.

lib/puma/const.rb
nateberkopec, line 127: I think it would be cleaner to make this an attribute of the Server object, which is read from configuration. 

We don't have any env variables setting constants right now, probably shouldn't start.

History.md
```diff
@@ -9,6 +9,11 @@
...
+
+* Features
```
nateberkopec, line 14: Please add changelog entries to the "master" entry above this one

nateberkopec (changes requested):

nateberkopec:
Cheers @fdel15! Thanks for your contribution!

Let's move this into the configuration object, like the other attributes on Server which get set as "options".

fdel15:
Thanks for the feedback Nate. I should have time this weekend to make the requested changes.

fdel15:
>  You can set this variable to zero to address use cases that wish to maintain the pre version 3.12.2 behavior

I originally had it working this way because an ENV variable is a string, so the INFINITY option that the issue suggested would not work. Since we are no longer using an ENV variable, we can allow users to set configure `max_fast_inline` to INFINITY to maintain the pre version 3.12.2 behavior.

dentarg:
@fdel15 can you update the title and PR description here to reflect what is actually happening now?

Could be discussed elsewhere, but I'll raise it now because ([the `History.md` diff here looks a bit funny](https://github.com/puma/puma/pull/2406/files#diff-e1bbd4f15e3b63427b4261e05b948ea8)). @nateberkopec @MSP-Greg should we remove the requirement that contributors update the changelog? now that @MSP-Greg has some tooling to do it

dentarg:
Also: I don't think CI should be skipped for this PR

fdel15:
> can you update the title and PR description here to reflect what is actually happening now?

Updated.

> Also: I don't think CI should be skipped for this PR

I removed the CI skip. I added it because I didn't add any tests in the PR. I feel like now that I am doing more than just adding an ENV variable I should be adding tests, but I am not sure how to test this feature. Any guidance on this would be greatly appreciated.

MSP-Greg:
@fdel15

Thanks for the PR.  This jumps around, so I won't inline it.  In server.rb
```ruby
attr_reader :max_fast_inline
```
and:
```ruby
if requests >= max_fast_inline
```

One would normally prefer that the API only exposes methods and attributes that are required externally. `max_fast_inline`  is currently not used anywhere other than the `Server` object, so it shouldn't have a reader.  The conditional can just use the instance variable set in the `initialize` method.

Re History.md, you may need to rebase for that.  Also, the code I've got for it only adds the link info.  At present it doesn't add items to the log, as that would require exact tagging for what 'group' an entry is listed in.

nateberkopec:
Agree with @MSP-Greg's feedback that we want to keep the interface private.

MSP-Greg:
This needs to be added to `puma/dsl.rb`?

fdel15:
> One would normally prefer that the API only exposes methods and attributes that are required externally.

Feedback addressed. Good call about not exposing an internal attribute.

> This needs to be added to puma/dsl.rb?

I think this is a good call as well. I added a `max_fast_inline` method and a test in `test/config`

MSP-Greg:
@fedl15

Do you see the message 'This branch cannot be rebased due to conflicts'?   You may need to remove the commits in your PR branch that are duplicated, update your forks' master branch, then rebase on master in your PR branch.  I think that made sense.  I can help with more info if needed.

You'll notice all the files listed as changed above.  It makes it hard to see what's changed when the master branch's commits are mixed in...

fdel15:
> Do you see the message 'This branch cannot be rebased due to conflicts'?

I thought I fixed the conflicts when I rebased. I must've done something silly though that caused all the files to be listed as changed. Sorry for not catching that earlier. I cherry picked all my commits for the branch and force pushed them up. This should be good for review now.

History.md
```diff
@@ -1,7 +1,18 @@
 ## Master
 
-* Features
-  * Your feature goes here <Most recent on the top, like GitHub> (#Github Number)
```
dentarg, line 4: I think we want to keep this line :)

History.md
```diff
@@ -1,7 +1,18 @@
...
+* Bugfixes
+  * Cleanup daemonization in rc.d script (#2409)
+
+* Refactor
+  * Extract req/resp methods to new request.rb from server.rb (#2419)
+  * Refactor Reactor and Client request buffering (#2279)
+  * client.rb - remove JRuby specific 'finish' code (#2412)
+  * Consolidate fast_write calls in Server, extract early_hints assembly (#2405)
+  * Remove upstart from docs (#2408)
+  * Consolidate option handling in Server, Server small refactors, doc changes (#2389)
```
dentarg, line 15: These lines should not be here, they are already in `History.md` (https://github.com/puma/puma/pull/2406/commits/72ed95944cd7f848413f021684be5c0fa114ed9b adds them for some reason)

fdel15: This should be good now

lib/puma/dsl.rb
```diff
@@ -810,5 +810,12 @@ def fork_worker(after_requests=1000)
     def nakayoshi_fork(enabled=true)
       @options[:nakayoshi_fork] = enabled
     end
+
+    # The number of requests to attempt inline before sending a client back to
+    # the reactor to be subject to normal ordering.
+    #
+    def max_fast_inline(num_of_requests)
+      @options[:max_fast_inline] = Float(num_of_requests)
```
nateberkopec, line 818: I don't think it really matters, but why a float here? We can't have a non-integer number of requests, can we?

nateberkopec:
Solid. Thanks for seeing this one through!

nateberkopec:
Whoops. I kinda forgot I wanted the next release to be a patchlevel. I may revert and then quickly re-revert this on Monday when we release.

lib/puma/dsl.rb
fdel15: To handle the use case mentioned in the issue to be able to set the value to Infinity. Using Float seemed like the simplest solution to me.

nateberkopec: Got it!

PR #2125, opened 2020-02-22
Remove master dependency on nio4r

cjlarose:
### Description

Fixes #2018 by removing `nio4r` as a dependency of the puma master process when using `prune_bundler`.

Just opening this PR for now to get some discussion on this. If it seems like a good idea, I can clean it up and add tests and everything.

### Your checklist for this pull request

- [x] I have reviewed the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.
- [ ] I have added an entry to [History.md](../blob/master/History.md) if this PR fixes a bug or adds a feature. If it doesn't need an entry to HISTORY.md, I have added `[changelog skip]` the pull request title.
- [ ] I have added appropriate tests if this PR fixes a bug or adds a feature.
- [x] My pull request is 100 lines added/removed or less so that it can be easily reviewed.
- [ ] If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.
- [x] If this closes any issues, I have added "Closes `#issue`" to the PR description or my commit messages.
- [ ] I have updated the documentation accordingly.
- [ ] All new and existing tests passed, including Rubocop.

PR #2407, opened 2020-10-02
Create new worker processes in separate memory spaces from master process

### LINKED ISSUE #2018: phased-restart: Could not find nio4r-2.5.1 in any of the sources (Bundler::GemNotFound)

cjlarose:
### Description

Fixes #2018

A fairly common and reasonable deployment strategy has been broken since the introduction of `nio4r` in puma 4.0. Imagine that a user is using puma in cluster mode. In order to deploy a new version of an application, a user's deployment tool does this:

1. Add new version of the application's source to the server in a new directory
2. Update the "current release" symlink to point to the new release directory (the symlink is configured as puma's `directory` in the config file)
3. Issue a phased restart
4. After verifying that the release was successful, delete the old version of the application from disk

During the next deployment, workers will fail to start. They'll encounter the error `Could not find nio4r-2.5.1 in any of the sources (Bundler::GemNotFound)`. This is, in part, due to a quirk in Bundler where Bundler will raise this error if you're trying to activate a gem whose native extensions are no longer present on disk (even if the gem is already loaded into memory). However, this exception reveals a deeper problem related to the intimate relationship between the gems used by the puma master process and those used by the puma worker processes.

Users can, in part, separate the gems used in workers from those used in the puma master by turning on `prune_bundler`. That option will "eject" the puma master process from the full Bundler context, re-activating only the small number of gems used by the puma master process, including `puma`, `nio4r`, and gems added to `extra_runtime_dependencies`. The `GemNotFound` issue, however reveals that even `prune_bundler` isn't always sufficient in severing the relationship between gems used by the master and worker processes. In addition to being unable to delete gems from old releases, the existing `prune_bundler`-based solution doesn't allow a user to upgrade the version of `nio4r` or any gem in `extra_runtime_dependencies` with a phased restart.

The solution proposed in this PR is to place worker processes in a completely separate memory space. The proposed change uses `spawn` to spin up new workers (this is essentially a `fork` followed by an `exec` on Unix-like systems). Since workers use a separate memory space, the gems they load can be totally different than those loaded in the puma master process. The drawback, of course, is that this `spawn`-based stategy doesn't allow one to take advantage of COW memory between the puma master and the puma workers. For this reason, we still use the old `fork` method in two cases:

1) If `preload_app` is enabled, since if `preload_app` is enabled, phased restarts aren't supported anyway. This option behaves as it did before: it forgoes the ability to perform phased restarts in exchange for minimizing the amount of memory used by fully leveraging the benefit of COW memory.
2) When workers fork from other workers (when `fork_worker` is enabled). Similar to `preload_app`, the main benefit of this option is to reduce memory by leveraging COW memory. The new spawn-based strategy is used only when creating `worker 0`, the process from which other workers fork. Subsequent workers are `forked` in the traditional way from `worker 0`.

What does this mean for `prune_bundler`? Before this change, it was necessary to use `prune_bundler` if you wanted phased restarts and the ability to change gem versions between different releases of your application. After this change, `prune_bundler` won't be necessary in this case. `prune_bundler` will essentially just be a memory optimization for the puma master process since it'll shed some gems from the puma master.

### DIFF
--- a/History.md
+++ b/History.md
@@ -4,6 +4,7 @@
   * Your feature goes here <Most recent on the top, like GitHub> (#Github Number)
 
 * Bugfixes
+  * Fix errors related to phased restarts when using deployment strategies that delete gems for old releases ([#2407])
   * Cleanup daemonization in rc.d script (#2409)
 
 * Refactor
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -103,11 +103,6 @@ is unavailable would be nice as well. Here's how to do it:
 1. Don't use `preload!`. This dirties the master process and means it will have
 to shutdown all the workers and re-exec itself to get your new code. It is not compatible with phased-restart and `prune_bundler` as well.
 
-1. Use `prune_bundler`. This makes it so that the cluster master will detach itself
-from a Bundler context on start. This allows the cluster workers to load your app
-and start a brand new Bundler context within the worker only. This means your
-master remains pristine and can live on between new releases of your code.
-
 1. Use phased-restart (`SIGUSR1` or `pumactl phased-restart`). This tells the master
 to kill off one worker at a time and restart them in your new code. This minimizes
 downtime and staggers the restart nicely. **WARNING** This means that both your
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -199,7 +199,7 @@ module Puma
     end
 
     def phased_restart
-      return false if @options[:preload_app]
+      return false unless supports_phased_restart?
 
       @phased_restart = true
       wakeup!
@@ -459,6 +459,10 @@ module Puma
     end
 
     private
+
+    def supports_phased_restart?
+      !@options[:preload_app]
+    end
 
     # loops thru @workers, removing workers that exited, and calling
     # `#term` if needed
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -93,7 +93,42 @@ module Puma
     def spawn_worker(idx, master)
       @launcher.config.run_hooks :before_worker_fork, idx, @launcher.events
 
-      pid = fork { worker(idx, master) }
+      pid = if supports_phased_restart?
+              pipes = { check_pipe: @check_pipe.fileno, worker_write: @worker_write.fileno }
+              if @options[:fork_worker]
+                pipes[:fork_pipe] = @fork_pipe.fileno
+                pipes[:wakeup] = @wakeup.fileno
+              end
+
+              script = <<-RUBY
+                original_argv = #{@launcher.original_argv.inspect}
+                worker_index = #{idx}
+                master_pid = #{master}
+
+                pipes = Hash[#{pipes}.map { |k, v| [k, IO.for_fd(v)] }]
+                pipes[:worker_write].sync = true
+                pipes[:wakeup].sync = true if pipes[:wakeup]
+
+                require 'bundler/setup'
+                require 'puma/cli'
+                require 'puma/cluster/worker'
+
+                cli = Puma::CLI.new original_argv
+                worker = Puma::Cluster::Worker.new index: worker_index,
+                                                   master: master_pid,
+                                                   launcher: cli.launcher,
+                                                   pipes: pipes
+                cli.launcher.binder.parse cli.launcher.options[:binds], worker
+                worker.run
+              RUBY
+
+              env = @launcher.binder.redirects_for_restart_env
+              pipe_map = Hash[pipes.map { |_, v| [v, v] }]
+              binds_map = @launcher.binder.redirects_for_restart
+              spawn env, Gem.ruby, '-e', script, pipe_map.merge(binds_map)
+            else
+              fork { worker(idx, master) }
+            end
       if !pid
         log "! Complete inability to spawn new workers detected"
         log "! Seppuku is the only choice."
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -118,6 +118,21 @@ module Puma
                                                    master: master_pid,
                                                    launcher: cli.launcher,
                                                    pipes: pipes
+
+                begin
+                  unless Puma.jruby? # INFO in use by JVM already
+                    Signal.trap "SIGINFO" do
+                      cli.launcher.thread_status do |name, backtrace|
+                        worker.log name
+                        worker.log backtrace.map { |bt| "  \#{bt}" }
+                      end
+                    end
+                  end
+                rescue Exception
+                  # Not going to log this one, as SIGINFO is *BSD only and would be pretty annoying
+                  # to see this constantly on Linux.
+                end
+
                 cli.launcher.binder.parse cli.launcher.options[:binds], worker
                 worker.run
               RUBY
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -583,18 +583,12 @@ module Puma
       @options[:lowlevel_error_handler] = obj
     end
 
-    # This option is used to allow your app and its gems to be
-    # properly reloaded when not using preload.
+    # This option is a memory optimization for use in cluster mode.
     #
-    # When set, if Puma detects that it's been invoked in the
-    # context of Bundler, it will cleanup the environment and
-    # re-run itself outside the Bundler environment, but directly
-    # using the files that Bundler has setup.
-    #
-    # This means that Puma is now decoupled from your Bundler
-    # context and when each worker loads, it will be loading a
-    # new Bundler context and thus can float around as the release
-    # dictates.
+    # When set, if the Puma master process detects that it's been invoked in
+    # the context of Bundler, it will cleanup the environment and re-run itself
+    # outside the Bundler environment, but directly using the files that
+    # Bundler has setup. This has no effect on worker processes.
     #
     # @see extra_runtime_dependencies
     #
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -89,7 +89,7 @@ module Puma
       @status = :run
     end
 
-    attr_reader :binder, :events, :config, :options, :restart_dir
+    attr_reader :binder, :events, :config, :options, :restart_dir, :original_argv
 
     # Return stats about the server
     def stats
--- a/test/config/cluster_stats.rb
+++ b/test/config/cluster_stats.rb
@@ -1,0 +1,3 @@
+app -> {[200, {}, ['']]}
+workers 1
+clear_binds!
--- a/test/config/prune_bundler_with_deps.rb
+++ b/test/config/prune_bundler_with_deps.rb
@@ -1,2 +1,5 @@
 prune_bundler true
 extra_runtime_dependencies ["rdoc"]
+before_fork do
+  puts "Last LOAD_PATH: #{$LOAD_PATH[-1]}"
+end
--- a/test/helpers/integration.rb
+++ b/test/helpers/integration.rb
@@ -54,11 +54,12 @@ class TestIntegration < Minitest::Test
       config_file.close
       config = "-C #{config_file.path}"
     end
+    puma_path = File.expand_path '../../../bin/puma', __FILE__
     if unix
-      cmd = "#{BASE} bin/puma #{config} -b unix://#{@bind_path} #{argv}"
+      cmd = "#{BASE} #{puma_path} #{config} -b unix://#{@bind_path} #{argv}"
     else
       @tcp_port = UniquePort.call
-      cmd = "#{BASE} bin/puma #{config} -b tcp://#{HOST}:#{@tcp_port} #{argv}"
+      cmd = "#{BASE} #{puma_path} #{config} -b tcp://#{HOST}:#{@tcp_port} #{argv}"
     end
     @server = IO.popen(cmd, "r")
     wait_for_server_to_boot
--- a/test/rackup/hello-last-load-path.ru
+++ b/test/rackup/hello-last-load-path.ru
@@ -1,3 +1,0 @@
-run lambda { |env|
-  [200, {}, [$LOAD_PATH[-1]]]
-}
--- a/test/test_integration_cluster.rb
+++ b/test/test_integration_cluster.rb
@@ -229,10 +229,10 @@ RUBY
   end
 
   def test_load_path_includes_extra_deps
-    cli_server "-w #{WORKERS} -C test/config/prune_bundler_with_deps.rb test/rackup/hello-last-load-path.ru"
-    last_load_path = read_body(connect)
+    cli_server "-w #{WORKERS} -C test/config/prune_bundler_with_deps.rb test/rackup/hello.ru"
 
-    assert_match(%r{gems/rdoc-[\d.]+/lib$}, last_load_path)
+    true while (line = @server.gets) !~ /^Last LOAD_PATH/
+    assert_match(%r{gems/rdoc-[\d.]+/lib$}, line)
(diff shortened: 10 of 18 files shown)


MSP-Greg:
Re last CI run, the only MRI failure (ubuntu-18.04 2.6) was due to incorrect code, but never noticed before.  See https://github.com/puma/puma/commit/b17579010ab.  Due to timing & luck, it's passed quite a bit previously...

cjlarose:
@MSP-Greg Thanks for that! I rebased to include your commit.

nateberkopec:
> prune_bundler will essentially just be a memory optimization for the puma master process since it'll shed some gems from the puma master.

Is that even worth it? The amount of code/headache from that feature is considerable. We could eject this feature.

nateberkopec:
What effects will this have on how signal handling currently works in Puma? I'm particularly thinking around sending signals to the master process.

cjlarose:
> > prune_bundler will essentially just be a memory optimization for the puma master process since it'll shed some gems from the puma master.
> 
> Is that even worth it? The amount of code/headache from that feature is considerable. We could eject this feature.

Yeah I think the removal of `prune_bundler` is worth considering. We can do some more testing and measuring separately on whether or not it really is much of a memory optimization, but my guess is that the cost of maintaining it would not longer be worth the benefit it provides.

Importantly, though, I _think_ this PR doesn't introduce any breaking changes as far as semver goes. Removing `prune_bundler` would, so I think we'd have to wait at least until 6.0 to do it.

cjlarose:
> What effects will this have on how signal handling currently works in Puma? I'm particularly thinking around sending signals to the master process.

I'll look more into this.

schneems:
It feels like this is a downstream fix to an upstream problem. Is there not some bundler or Ruby API that we can use to tell our process that 

> This is, in part, due to a quirk in Bundler where Bundler will raise this error if you're trying to activate a gem whose native extensions are no longer present on disk (even if the gem is already loaded into memory)

Could you reproduce this behavior outside of puma using only bundler? If so then we could report it upstream to bundler, and maybe we can also brainstorm on some workarounds here. 

My initial gut reaction is: I want shared memory whenever possible. Processes are expensive (from a memory standpoint) so I want to benefit from CoW whenever possible.

cjlarose:
> Could you reproduce this behavior outside of puma using only bundler? If so then we could report it upstream to bundler, and maybe we can also brainstorm on some workarounds here.

Yeah I can create a small reproducible script for this and report it to Bundler.

> My initial gut reaction is: I want shared memory whenever possible. Processes are expensive (from a memory standpoint) so I want to benefit from CoW whenever possible.

I think you might be right. This change is essentially a trade-off that'll cost puma more memory, but gives us the following:

* The ability for users to upgrade gems in worker processes like `nio4r` without downtime (using a phased restart on cluster-mode puma)
* Related: prevent some kinds of issues related to the application using different versions of gems from the puma master (especially the `json` gem, see #2206)
* The potential to remove code related to `prune_bundler` and relieve ourselves of the maintenance burden of supporting it. This can simplify configuration for users, especially those that rely on `extra_runtime_dependencies`

Fixing the bug in Bundler (or finding an appropriate workaround in puma) wouldn't give us these other benefits, but it would help make it possible for folks who are stuck on 4.x to upgrade. Also, I should mention that it's also totally possible to fix the original problem described in #2018 by just not loading the `nio4r` gem in the master at all (workers need the gem, but the master never needs it). I proposed this in #2125 and it was initially rejected, but I think it's probably worth reconsidering since it'll help real people migrate their apps to puma 5.x basically immediately.

cjlarose:
Opened rubygems/rubygems#4004 for the Bundler bug. We can see what the maintainers say there and decide how to move forward based on that.

nateberkopec:
Just in general, removing `fork` from Puma's base behavior is going to have to clear a very high bar to make me not nervous. It's up there with "we parse HTTP w/the old mongrel extension".

nateberkopec:
@cjlarose I think you're right about re-opening https://github.com/puma/puma/pull/2125. What do you think - close this, fix the bundler bug (or see where that goes), and reopen/merge 2125?

cjlarose:
That sounds good. Let's close this and I'll take another stab at #2125 (I'll open a separate PR I think and fix the tests and rebase and everything). If the bundler bug gets fixed, too, that's great, but we don't need to make that our blocker.

S3od22 (commented):
Sss

rmacklin:
> Let's close this and I'll take another stab at #2125 (I'll open a separate PR I think and fix the tests and rebase and everything).

For reference, that newer PR is https://github.com/puma/puma/pull/2427

PR #2408, opened 2020-10-03
[ci skip] Remove upstart from docs

fliiiix:
upstart is dead!
upstart as system init - last release - 14.04 LTS, public support ending April 2019
upstart as user init - last release - 16.04 LTS, public support ending April 2021

I don't think anyone is still using upstart. So I would recommend to remove it. 

If you disagree to just close this PR.


nateberkopec:
Cheers! Thanks for all your fixes this morning. The deployment docs are really just a collection of community wisdom, so they depend on people like you keepign them up to date 🙇

Issue #2137, opened 2020-02-28
Need more test coverage around Rack compliance


nateberkopec:
Double check https://github.com/rack/rack/blob/master/SPEC.rdoc#the-headers- for test coverage, and add where necessary.

PR #2413, opened 2020-10-06
Add test case for response header [changelog skip]

kuei0221:
### Description
This PR tends to solve issue #2137 , adding more test for the response header, and follow the requirement mentioned in the rack doc.

Test include:
1. The header keys must be Strings. 
2. Special headers starting “rack.” are for communicating with the server, and must not be sent back to the client. 
3. The header must not contain a Status key. 
4. The header must conform to RFC7230 token specification, i.e. cannot contain non-printable ASCII, DQUOTE or “(),/:;<=>?@[]{}”. 
5. The values of the header must be Strings, consisting of lines (for multiple header values, e.g. multiple Set-Cookie values) separated by “\n”. The lines must not contain characters below 037.

If my test case is correct, I have found most features mentioned above did not cover yet (so the most tests should fail anyway) and would like to do so after some feedback. I am a newbie btw, so let me know anything if I can make things better. Thanks! 

> Double check https://github.com/rack/rack/blob/master/SPEC.rdoc#the-headers- for test coverage, and add where necessary.

### DIFF
--- a/test/test_response_header.rb
+++ b/test/test_response_header.rb
@@ -1,0 +1,140 @@
+require_relative "helper"
+require "puma/events"
+require "net/http"
+
+class TestResponseHeader < Minitest::Test
+  parallelize_me!
+
+  def setup
+    @host = "127.0.0.1"
+
+    @ios = []
+
+    @app = ->(env) { [200, {}, [env['rack.url_scheme']]] }
+
+    @events = Puma::Events.strings
+    @server = Puma::Server.new @app, @events
+  end
+
+  def teardown
+    @server.stop(true)
+    @ios.each { |io| io.close if io && !io.closed? }
+  end
+
+  def server_run(app: @app, early_hints: false)
+    @server.app = app
+    @port = (@server.add_tcp_listener @host, 0).addr[1]
+    @server.early_hints = true if early_hints
+    @server.run
+  end
+
+  def send_http_and_read(req)
+    send_http(req).read
+  end
+
+  def send_http(req)
+    new_connection << req
+  end
+
+  def new_connection
+    TCPSocket.new(@host, @port).tap {|sock| @ios << sock}
+  end
+
+  # The header keys must be Strings
+  def test_integer_key
+    server_run app: ->(env) { [200, { 1 => 'Boo'}, []] }
+    data = send_http_and_read "GET / HTTP/1.0\r\n\r\n"
+
+    assert_match(/HTTP\/1.1 500 Internal Server Error/, data)
+  end
+
+  # The header must respond to each
+  def test_nil_header
+    server_run app: ->(env) { [200, nil, []] }
+    data = send_http_and_read "GET / HTTP/1.0\r\n\r\n"
+
+    assert_match(/HTTP\/1.1 500 Internal Server Error/, data)
+  end
+
+  # The values of the header must be Strings
+  def test_integer_value
+    server_run app: ->(env) { [200, {'Content-Length' => 500}, []] }
+    data = send_http_and_read "GET / HTTP/1.0\r\n\r\n"
+
+    assert_match(/HTTP\/1.0 200 OK\r\nContent-Length: 500\r\n\r\n/, data)
+  end
+
+  def assert_ignore_header(name, value, opts={})
+    header = { name => value }
+
+    if opts[:early_hints]
+      app = ->(env) do
+        env['rack.early_hints'].call(header)
+        [200, {}, ['Hello']]
+      end
+    else
+      app = -> (env) { [200, header, ['hello']]}
+    end
+
+    server_run(app: app, early_hints: opts[:early_hints])
+    data = send_http_and_read "GET / HTTP/1.0\r\n\r\n"
+
+    refute_match("#{name}: #{value}", data)
+  end
+
+  # The header must not contain a Status key.
+  def test_status_key
+    skip 'implement later'
+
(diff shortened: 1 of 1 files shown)


nateberkopec:
Wow! So glad someone's looking at #2137. Thank you so much.

test/test_puma_server.rb
```diff
@@ -1087,4 +1087,80 @@ def test_accept_econnaborted
...
+  end
+
+  def test_response_header_value_is_integer
+    server_run app: ->(env) { [200, {'Content-Length' => 500}, []] }
+    data = send_http_and_read "GET / HTTP/1.0\r\n\r\n"
+
+    assert_match(/HTTP\/1.0 200 OK\r\nContent-Length: 500\r\n\r\n/, data)
+  end
+
+  #Special headers starting “rack.” are for communicating with the server, and must not be sent back to the client.
```
nateberkopec, line 1118: Not 100% necessary, but it would be cool if you added a comment like this to every test

test/test_puma_server.rb
```diff
@@ -1087,4 +1087,80 @@ def test_accept_econnaborted
...
+  def test_response_header_name_include_rack
+    server_run app: ->(env) { [200, {'rack.commad_to_server_only' => 'work'}, []] }
```
nateberkopec, line 1120: typo: command

test/test_puma_server.rb
```diff
@@ -1087,4 +1087,80 @@ def test_accept_econnaborted
...
+    data = send_http_and_read "GET / HTTP/1.0\r\n\r\n"
+
+    refute_match(/rack/, data)
+  end
+
+  # testing header value must conform rfc token specification
+  # i.e. cannot contain non-printable ASCII, DQUOTE or “(),/:;<=>?@[]{}”.
+  # besides, header value can be consisted by line separted by \n. Line should must not contain characters below 037.
+  #
+  #There are three different tests because there are three ways to set header
```
nateberkopec, line 1130: missing space

test/test_puma_server.rb
```diff
@@ -1087,4 +1087,80 @@ def test_accept_econnaborted
...
+  def test_response_header_name_include_rack
+    server_run app: ->(env) { [200, {'rack.commad_to_server_only' => 'work'}, []] }
+    data = send_http_and_read "GET / HTTP/1.0\r\n\r\n"
+
+    refute_match(/rack/, data)
+  end
+
+  # testing header value must conform rfc token specification
+  # i.e. cannot contain non-printable ASCII, DQUOTE or “(),/:;<=>?@[]{}”.
+  # besides, header value can be consisted by line separted by \n. Line should must not contain characters below 037.
```
nateberkopec, line 1128: separated

test/test_puma_server.rb
```diff
@@ -1087,4 +1087,80 @@ def test_accept_econnaborted
...
+  #
+  #There are three different tests because there are three ways to set header
+  # content in Puma. Regular (rack env), early hints, and a special case for
+  # overriding content-length. (refernce form http ins)
+
+  def assert_follow_token_specification(app, opts={})
+    server_run(early_hints: opts[:early_hints], app: app)
+
+    data = send_http_and_read "GET / HTTP/1.0\r\n\r\n"
+    assert_match(/ABC/,data)
```
nateberkopec, line 1138: missing space here

nateberkopec:
This PR is amazing!!! I think the tests are perfect.

If you want to fix these problems, you can do that. Or, instead you can `skip` the failures and then let someone else fix them or come back and fix them yourself in a separate PR.

MSP-Greg:
Maybe move all these tests (and possibly others to a new test file?

Also, I hope to extract the code that processes the response headers received from `@app.call(env)` in `Server#handle_request`, so these tests could be done without sending an http client request (or starting the Server instance)...

kuei0221:
Thank you @nateberkopec for checking! I will skip the test now, and solve them in a separate PR as you suggest.

BTW, I have also try to do what @MSP-Greg suggests, and I totally agree that it will be great to extract the process from `Server#handle_request`. But considering the amount of feature and code includes in that method, I will need more understanding of codebase to do that job.

Thank you both!

nateberkopec:
No problem. I think this can be merged after you move it to it's own (new) file, and then skip all tests that fail.

MSP-Greg:
@kuei0221

Thanks for your work on this.  The request code and the server object are connected, but I extracted all the request code to another file, and am including it in server.  I also pulled the header loop out, so it can be fed a header hash and IOBuffer.  It modifies the IOBuffer 'in place'.  I think.

One thing I'll leave to @nateberkopec is what headers should be just ignored (not included in the response) and which should throw errors.

Anyway, it will be easy to convert code over later.  I've already got the 'rake.' headers excluded.  Not sure how many conditionals it would need, that could be another method...

kuei0221:
I have found some errors with my works about RFC specification, and all of them should be fixed,  please have a look.

1. The header must conform to RFC7230 token specification

- Token specification implemented on the header key, not header value.

2. The values of the header must be Strings, consisting of lines (for multiple header values, e.g. multiple Set-Cookie values) separated by “\n”. The lines must not contain characters below 037.

- Header value should only contain characters below 037, but DQUOTE or “(),/:;<=>?@[]{}” is permitted.

Besides, the test case now expects header with illegal characters (depends on key or value) will be ignored.

nateberkopec:
Brilliant! Amazing first contribution @kuei0221 👏

PR #2419, opened 2020-10-08
Extract req/resp methods to new request.rb from server.rb

MSP-Greg:
### Description

server.rb is a very long file with several very long methods.

1. Extract `default_server_port`, `fast_write`, `fetch_status_code`, `handle_request`, `normalize_env`, `possible_header_injection?`, and `str_early_hints` from server.rb to request.rb

2. Extract `req_env_post_parse` and `str_headers` from `handle_request`

3. Updated comments/documentation

See https://msp-greg.github.io/puma/Puma/Request.html for the doc of the PR.

### DIFF
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -1,0 +1,438 @@
+# frozen_string_literal: true
+
+module Puma
+
+  # The methods here are included in Server, but are separated into this file.
+  # All the methods here pertain to passing the request to the app, then
+  # writing the response back to the client.
+  #
+  # None of the methods here are called externally, with the exception of
+  # #handle_request, which is called in Server#process_client.
+  # @version 5.0.3
+  #
+  module Request
+
+    include Puma::Const
+
+    # Takes the request contained in +client+, invokes the Rack application to construct
+    # the response and writes it back to +client.io+.
+    #
+    # It'll return +false+ when the connection is closed, this doesn't mean
+    # that the response wasn't successful.
+    #
+    # It'll return +:async+ if the connection remains open but will be handled
+    # elsewhere, i.e. the connection has been hijacked by the Rack application.
+    #
+    # Finally, it'll return +true+ on keep-alive connections.
+    # @param client [Puma::Client]
+    # @param lines [Puma::IOBuffer]
+    # @return [Boolean,:async]
+    #
+    def handle_request(client, lines)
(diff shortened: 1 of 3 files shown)


lib/puma/request.rb
```diff
@@ -0,0 +1,438 @@
+# frozen_string_literal: true
+
+module Puma
```
nateberkopec, line 3: If this is going to be a composed object extraction, this file should be `Puma::Server::Request` and then live in `lib/puma/server/request.rb`.

lib/puma/request.rb
```diff
@@ -0,0 +1,438 @@
...
+
```
nateberkopec, line 4: It looks like you've found a seam but haven't quite cut through all the way yet. I think you can go all the way to a composed object:

```ruby
while true 
  case Request.new(client,buffer).handle
```

... and then in `lib/puma/server/request.rb`, everything except `handle_request` can be marked private. 

This would greatly reduce the API surface of Server. Right now this moves methods around in different files, but with a composed object we could really make Server a lot smaller.

```

**Review by nateberkopec — CHANGES_REQUESTED** [review]

File: lib/puma/request.rb (hunk 2 above)
Comment (MSP-Greg) [reply lib/puma/request.rb:4]: I did think about this.  Two items:

1. It would be a breaking API change.  `handle_request` and `normalize_env` are currently public methods of Server.

2. Coupling - the following Server instance variables are used in Request code:
```
@app              available
@early_hints      available
@events           available
@queue_requests   hidden
@thread_pool      hidden
```

We could add a couple of readers and pass the server instance into the class init?  Also, re class init, I'm not sure if the buffer should/can be contained within Request, rather than be passed in.

I thought that the current layout would maintain the API, but make it easy to create a class when v6.0.0 was being considered.

But, re API changes, the earlier issue with the API changes I made affected parts of the API that could be used for Puma startup/setup.  These changes are more 'runtime'.  Possibly less likely to cause issues...

nateberkopec: OK, got it. I'll re-review this with a mind to "this is a temporary stopping point to an extraction"

nateberkopec:
Ok, LGTM. We can revisit this for an extracted object when 6.0 rolls around.

nateberkopec:
Wrong button, lol

Issue #1735, opened 2019-03-06
Lack of SSL max_version can cause excessive slow negotation, 502 on AWS w/ELB


amichal:
### Steps to reproduce

1) rails app running in `puma (Version 3.12.0 (ruby 2.5.3-p105` serving a simple text file out of `/public`

2) puma running as a systemd service
```
[Service]
Type=forking
WorkingDirectory=/var/www/test-app/current
EnvironmentFile=/var/www/test-app/shared/.env.production
ExecStart=/bin/bash -lc 'cd /var/www/test-app/current && /usr/share/rvm/bin/rvm default do bundle exec puma --daemon --pidfile /var/www/test-app/shared/tmp/pids/puma.pid --bind "ssl://0.0.0.0:3000?key=/var/www/test-app/shared/ssl-cert-snakeoil.key&cert=/var/www/test-app/shared/ssl-cert-snakeoil.pem" --config /var/www/test-app/current/config/puma.rb --environment production --redirect-stdout /var/www/test-app/shared/log/puma.log --redirect-stderr /var/www/test-app/shared/log/puma.error.log'
PIDFile=/var/www/test-app/shared/tmp/pids/puma.pid
```
3) Configure a AWS Application load balancer to accept https traffic on (443) and forward to the puma instance above via https (port 3000)

3) Make lots of requests via a browser. Also can be automated. e.g.:  
`siege -b --no-parser -t1m -c30 https://lblb......eu-central-1.elb.amazonaws.com/robots.txt | grep 502`

### Expected behavior
100% 200 OK

### Actual behavior
A small percentage ( 5-10 requests from 4000 completed in 1 minute) for the above test fails with a 502 Bad Gateway.

### System configuration
ruby 2.5.3-p105
Rails 5.2.2
puma version 3.12.0

AWS ELB logs for failed requests show the request arriving -- taking <1ms to be read by AWS and <0.5s to be processed in total. They show puma connection closed without a reason and consistently processing 216 bytes from puma. Puma stdout shows nothing, neither does stderr. I've added x-aws-trace-id headers to my rails request logs to confirm that the rails app never gets a chance to log the failing requests.

#### ENV

RAILS_ENV = producton
RAILS_MAX_THREADS = 8
WEB_CONCURRENCY = 6
on a  t2.xlarge with nothing else running on it

puma.rb
```
# Puma can serve each request in a thread from an internal thread pool.
# The `threads` method setting takes two numbers: a minimum and maximum.
[...]

PR #2426, opened 2020-10-13
Add no_tlsv1_3 to Puma::DSL#ssl_bindings and Puma::MiniSSL::Context

MSP-Greg:
### Description

Disables TLSv1.3 when front-ends don't support it

Also:

1. Remove 'no_tls' false query parameters in bind string
2. Update minissl.c for no_tlsv1_3, whitespace
3. Add test_tls_v1_3_rejection to test_puma_server_ssl.rb
4. Add three protocol tests to test_config.rb
5. extconf.rb - 1.1.0 -> 1.1.1 change (macro to func)

See #1735

### DIFF
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -369,6 +369,12 @@ module Puma
     # Instead of `bind 'ssl://127.0.0.1:9292?key=key_path&cert=cert_path'` you
     # can also use the this method.
     #
+    # Protocols can be disabled by the use of `no_tlsv1`, `no_tlsv1_1`, and
+    # `no_tlsv1_3`.  `no_tlsv1`, and `no_tlsv1_1` set the minimum protocol to
+    # TLSv1.1 and TLSv1.2 respectively.  `no_tlsv1_3` should only be used when a
+    # front-end does not support TLSv1.3.
+    # The default value for all three is `false`, set them to `true` to enable.
+    #
     # @example
     #   ssl_bind '127.0.0.1', '9292', {
     #     cert: path_to_cert,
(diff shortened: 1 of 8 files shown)


lib/puma/dsl.rb
```diff
@@ -387,16 +393,26 @@ def threads(min, max)
...
+      if opts[:no_tlsv1_3]
+        tls_str = "#{tls_str || ''}&no_tlsv1_3=true"
+      end
+
       ca_additions = "&ca=#{opts[:ca]}" if ['peer', 'force_peer'].include?(verify)
 
       if defined?(JRUBY_VERSION)
         keystore_additions = "keystore=#{opts[:keystore]}&keystore-pass=#{opts[:keystore_pass]}"
-        bind "ssl://#{host}:#{port}?cert=#{opts[:cert]}&key=#{opts[:key]}&#{keystore_additions}&verify_mode=#{verify}&no_tlsv1=#{no_tlsv1}&no_tlsv1_1=#{no_tlsv1_1}#{ca_additions}"
+        bind "ssl://#{host}:#{port}?cert=#{opts[:cert]}&key=#{opts[:key]}&#{keystore_additions}&verify_mode=#{verify}#{tls_str}#{ca_additions}"
```
nateberkopec, line 412: I've just realized how silly it is that we pass around arguments to `bind` internally as a string rather than as arguments. File that one away for the future...

lib/puma/minissl/context_builder.rb
```diff
@@ -45,8 +45,9 @@ def context
           ctx.ssl_cipher_filter = params['ssl_cipher_filter'] if params['ssl_cipher_filter']
         end
 
-        ctx.no_tlsv1 = true if params['no_tlsv1'] == 'true'
+        ctx.no_tlsv1   = true if params['no_tlsv1']   == 'true'
         ctx.no_tlsv1_1 = true if params['no_tlsv1_1'] == 'true'
+        ctx.no_tlsv1_3 = true if params['no_tlsv1_3'] == 'true'
```
nateberkopec, line 50: At what point to we decide there's a more elegant interface here than just disabling individual TLS minor versions? I seem to remember something was discussed a long ways back...

MSP-Greg: When we drop support for OpenSSL 1.0.2?  1.1.1 has min/max functions, with 1.0.2 you have to supply a list of allowed protocols.

JFYI, TLSv1.2 came out in 2008, TLSv1.3 in 2018.  What we have now will probably work for a few years...

MSP-Greg:
Issue #1735 pertained to a problem with load balancers (and hence, a high speed connection to Puma), and the question of whether Puma' downward negotiation from TLSv1.3 to TLSv1.2 slowed things down.

While working on #2519, and also testing locally with WSL2/Ubuntu, I checked the timing of using TLSv1.2 clients with Puma.  Timing both with and without the server restricted to TLSv1.2, there was no significant difference.

Hence, closing.

Issue #2390, opened 2020-09-29
lowlevel_error_handler is unexpectedly invoked when an HTTP/1.1 client closes its connection first


cjlarose:
**Describe the bug**

If a client opens a TCP connection, writes a HTTP/1.1 request (without explicitly specifying a `Connection: close` header or anything like that), reads the response, then closes the TCP connection, then puma unexpectedly invokes the `lowlevel_error_handler`. I think this is a regression introduced in 5.0.1, possibly by #2384.

**Puma config:**

```ruby
port ENV['PORT'] if ENV['PORT']

app do |env|
  [200, {}, ["Hello World"]]
end

lowlevel_error_handler do |err|
  STDERR.puts err
  STDERR.puts err.backtrace
  [500, {}, ["error page"]]
end
```

```sh
bundle exec ruby -Ilib bin/puma -C test/config/lowlevel_test.rb
```

**To Reproduce**

```
$ curl -vvv http://localhost:9292
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 9292 (#0)
> GET / HTTP/1.1
> Host: localhost:9292
> User-Agent: curl/7.64.1
> Accept: */*
>
< HTTP/1.1 200 OK
< Content-Length: 11
<
* Connection #0 to host localhost left intact
Hello World* Closing connection 0
```

**Expected behavior**

The lowlevel handler isn't invoked, consistent with the behavior of puma v5.0.0

**Desktop (please complete the following information):**
 - OS: macOS, Linux
 - Puma Version 5.0.2

cjlarose:
The error reported to the `lowlevel_error_handler`

```
EOFError
/Users/chris.larose/dev/puma/lib/puma/client.rb:170:in `try_to_finish'
/Users/chris.larose/dev/puma/lib/puma/reactor.rb:224:in `block in run_internal'
/Users/chris.larose/dev/puma/lib/puma/reactor.rb:157:in `each'
/Users/chris.larose/dev/puma/lib/puma/reactor.rb:157:in `run_internal'
/Users/chris.larose/dev/puma/lib/puma/reactor.rb:307:in `block in run_in_thread'
```

cjlarose:
@schneems I think the problem is that the `Puma::Reactor` used to rely on  getting `ConnectionError` in this case when calling `Puma::Client#try_to_finish`, but now it gets an `EOFError` instead and doesn't explicitly handle that error.

[...]

PR #2429, opened 2020-10-14
Add test_client_quick_close_no_lowlevel_error_handler_call [changelog skip]

MSP-Greg:
### Description

See https://github.com/puma/puma/issues/2390

Fixed by https://github.com/puma/puma/pull/2279

Fails in 5.0.2, showing EOFError being thrown.

### DIFF
--- a/test/test_puma_server.rb
+++ b/test/test_puma_server.rb
@@ -1103,4 +1103,41 @@ EOF
     stub_accept_nonblock(error)
     assert_empty @events.stderr.string
   end
+
+  # see      https://github.com/puma/puma/issues/2390
+  # fixed by https://github.com/puma/puma/pull/2279
+  #
+  def test_client_quick_close_no_lowlevel_error_handler_call
+    handler = ->(err, env, status) {
+      @events.stdout.write "LLEH #{err.message}"
+      [500, {"Content-Type" => "application/json"}, ["{}\n"]]
+    }
+
+    @server = Puma::Server.new @app, @events, {:lowlevel_error_handler => handler}
+
+    server_run app: ->(env) { [200, {}, ['Hello World']] }
+
+    # valid req & read, close
+    sock = TCPSocket.new @host, @port
+    sock.syswrite "GET / HTTP/1.0\r\n\r\n"
+    resp = sock.sysread 256
+    sock.close
+    assert_match 'Hello World', resp
+    sleep 0.5
+    assert_empty @events.stdout.string
+
+    # valid req, close
+    sock = TCPSocket.new @host, @port
+    sock.syswrite "GET / HTTP/1.0\r\n\r\n"
+    sock.close
+    sleep 0.5
+    assert_empty @events.stdout.string
+
+    # invalid req, close
+    sock = TCPSocket.new @host, @port
+    sock.syswrite "GET / HTTP"
+    sock.close
+    sleep 0.5
+    assert_empty @events.stdout.string
+  end
 end


test/test_puma_server.rb
```diff
@@ -1103,4 +1103,26 @@ def test_accept_econnaborted
...
+      [500, {"Content-Type" => "application/json"}, ["{}\n"]]
+    }
+
+    @server = Puma::Server.new @app, @events, {:lowlevel_error_handler => handler}
+
+    server_run app: ->(env) { [200, {}, ['Hello World']] }
+
+    sock = TCPSocket.new @host, @port
+    # normal request doesn't generate EOFError
+    sock.syswrite "GET / HTTP"
```
cjlarose, line 1122: Since this test isn't specifically testing malformed HTTP requests, it'd be nice to make sure that this one is well-formed.

```ruby
sock.syswrite "GET / HTTP/1.1\r\n\r\n"
```

test/test_puma_server.rb
```diff
@@ -1103,4 +1103,26 @@ def test_accept_econnaborted
...
+    sock.close
+
+    sleep 0.7 # need some delay for the lleh to be called
+    assert_empty @events.stdout.string
```
cjlarose, line 1126: How do you feel about making an additional assertion that the client a successful response?

test/test_puma_server.rb
MSP-Greg: Let me check, but I don't think that threw the error in 5.0.2.  I started with a normal request and a read, but nothing was raised.  I think.

cjlarose: Oh I think I understand your comment `normal request doesn't generate EOFError` then. 

That's interesting. That means that this test isn't reproducing exactly the same kinda thing we see with `curl` as in #2390, but instead found a different way to cause the `EOFError` reliably.

MSP-Greg: > Oh I think I understand your comment normal request doesn't generate EOFError then.

But just for you I checked things.

As long as the request 'write' is valid, no error, with or without a read.  So, with 5.0.2, as the test is, the failure shows
```
Expected "LLEH EOFError" to be empty.
```

So 'EOFError' is the err.message, and the class is the same.  I had to add the delay, as it passed without it...

test/test_puma_server.rb
cjlarose: In the other thread, @MSP-Greg demonstrated that the request has to be malformed, so the client won't get a successful response. Doesn't make sense to test for it in this case.

test/test_puma_server.rb
cjlarose: Thanks for checking! That makes sense.

As I understand it, this test does not test the exact same regression described in #2390 , but does test for another, closely-related regression that was introduced in 5.0.1.

The test as written ensures that if a client closes the connection before writing a complete HTTP request, that puma doesn't invoke the low-level error handler. That's a good test to have.

A test that would more closely simulate what's described in #2390 would have the client write a complete HTTP request to the socket before closing it, but if we modify the test as written to do that, it doesn't reliably reproduce the problem for some reason. If we figure out how to do that in the future, it'd be great to have, but for now this test is a good substitute.

cjlarose:
LGTM. Thanks for adding this!

test/test_puma_server.rb
MSP-Greg: Maybe add another socket that writes a normal requests reads the response and closes immediately?  Or maybe two more, such that:

Socket 1 - writes a valid request, reads it, closes immediately
Socket 2 - writes a valid request, closes immediately
Socket 3 - writes an invalid request, closes immediately

At the end of that, @events.stdout.string should still be empty?

cjlarose: Yeah, all of those three cases are cases where we'd expect that low-level error handler to not be invoked, but for all three cases, it actually was invoked on puma 5.0.1. I could see each of those three being valuable test cases.

But so far we're only able to write a test that reliably demonstrates the failure for the socket 3 case.

MSP-Greg: > But so far we're only able to write a test that reliably demonstrates the failure for the socket 3 case

Correct, at least in 5.0.2.  I updated to use three cases.  Given that all should pass, might as well.  Having been bitten by 'that will never break, so why write a test?', no harm in adding them...

PR #2431, opened 2020-10-16
Fire on_booted after server starts

### LINKED ISSUE #2212: TestCLI#test_control failing with "pool_capacity": 0
**Describe the bug**
> A clear and concise description of what the bug is.

`TestCLI#test_control` (`test/test_cli.rb:175`)sometimes fails with the result `"pool_capacity": 0` with following message.
Do you have any idea to pass the test such as injecting sleeping time before the assertion, and etc?

```
  5) Failure:
TestCLI#test_control [/builddir/build/BUILD/puma-4.3.3/usr/share/gems/gems/puma-4.3.3/test/test_cli.rb:175]:
Expected /{ "started_at": "\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z", "backlog": 0, "running": 0, "pool_capacity": 16, "max_threads": 16 }/ to match # encoding: ASCII-8BIT
#    valid: true
"{ \"started_at\": \"2020-03-30T18:59:23Z\", \"backlog\": 0, \"running\": 0, \"pool_capacity\": 0, \"max_threads\": 16 }".
```

**Puma config:**

> Please copy-paste your Puma config AND your command line options here.

It's not the target.

**To Reproduce**
> Please add reproduction steps here.

This issue sometimes happened for CPU architectures ppc64le or s390x on Fedora project's build system. Note this system does not provide the internet to outside in it. But the loop back network is available.

> Your issue will be solved very quickly if you can reproduce it with a "hello world" rack application. To do this, copy this into a file called `hello.ru`:

I can not reproduce it with the "hello world" rack application.

> There is also a Dockerfile available for reproducing Linux-specific issues. To use:

I have a willing to work for that, if you need it. Maybe I will try to reproduce it with [Travis multple CPU feature](https://docs.travis-ci.com/user/multi-cpu-architectures/) `arch: ppc64le` and `arch: s390x`.

**Expected behavior**
> A clear and concise description of what you expected to happen.

`TestCLI#test_control` passes.

 - OS: Fedora 33 (= rawhide = development version) Linux
 - Puma Version: 4.3.3
 - Ruby Version: 2.7.0

wjordan:
### Description

This PR slightly changes the timing behavior of the `on_booted` event, ensuring that the event is only fired after the server is running (that is, after `Server#run` has been called), so server stats are guaranteed to be available when this event fires.

In cluster mode, this is when the first stats 'ping' has been received from all worker processes. The timing of the 'ping' has also been updated to send the first one immediately, instead of after `Const::WORKER_CHECK_INTERVAL` (default 5) seconds have passed.

Fixes #2212.

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -395,9 +395,9 @@ module Puma
         stop
       end
 
-      @launcher.events.fire_on_booted!
-
       begin
+        booted = false
+
         while @status == :run
           begin
             if @phased_restart
@@ -438,6 +438,10 @@ module Puma
                 when "p"
                   w.ping!(result.sub(/^\d+/,'').chomp)
                   @launcher.events.fire(:ping!, w)
+                  if !booted && @workers.none? {|worker| worker.last_status.empty?}
+                    @launcher.events.fire_on_booted!
+                    booted = true
+                  end
                 end
               else
                 log "! Out-of-sync worker list, no #{pid} worker"
--- a/History.md
+++ b/History.md
@@ -5,6 +5,7 @@
 
 * Bugfixes
   * Cleanup daemonization in rc.d script (#2409)
+  * Fire `on_booted` after server starts
 
 * Refactor
   * Extract req/resp methods to new request.rb from server.rb (#2419)
--- a/lib/puma/cluster/worker.rb
+++ b/lib/puma/cluster/worker.rb
@@ -103,22 +103,24 @@ module Puma
           return
         end
 
-        Thread.new(@worker_write) do |io|
-          Puma.set_thread_name "stat payload"
+        while restart_server.pop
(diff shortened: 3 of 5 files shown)


lib/puma/cluster.rb
```diff
@@ -438,6 +438,10 @@ def run
                 when "p"
                   w.ping!(result.sub(/^\d+/,'').chomp)
                   @launcher.events.fire(:ping!, w)
+                  if !booted && @workers.none? {|worker| worker.last_status.empty?}
```
nateberkopec, line 441: this conditional kind of hurts my head but I don't have a better suggestion

dentarg: Extract logic to helper method with clever name?

PR #2432, opened 2020-10-16
Add Client#io_ok?, check before Reactor#register

MSP-Greg:
### Description

The issue - how to make Puma respond gracefully when a io connection can misbehave/fail at any point in its handling?

This is a difficult thing to test in CI, especially with multiple VM's running on CI runners.

The are infrequent errors in Puma CI, one example below:
```
Error in reactor loop escaped: mode not supported for this object: r (ArgumentError)
org/nio4r/Selector.java:124:in `register'
/Users/runner/work/puma/puma/lib/puma/reactor.rb:93:in `register'
/Users/runner/work/puma/puma/lib/puma/reactor.rb:77:in `select_loop'
/Users/runner/work/puma/puma/lib/puma/reactor.rb:34:in `block in run'
```

The error `mode not supported for this object: r` implies that the socket is no longer valid in some way.

Does it represent a CI specific issue, or is it something that could occur in real use?

At present, there is a great deal of 'rescue handling' re sockets, and where and how exceptions are raised/caught can get messy.

So, is there a way to conditionally test the socket?  This PR is just a guess, using an `io_ok?` method in Client.

I believe @schneems has mentioned something similar somewhere.

Anyway, if people think it's valid, maybe a method needs to also be added somewhere that takes a socket as a parameter,  which `Client#io_ok?` could call.  I also haven't thought about logging, etc.

### DIFF
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -85,6 +85,12 @@ module Puma
 
     def_delegators :@io, :closed?
 
+    # Test to see if io meets a bare minimum of functioning, @to_io needs to be
+    # used for MiniSSL::Socket
+    def io_ok?
+      @to_io.is_a?(::BasicSocket) && !closed?
+    end
+
     # @!attribute [r] inspect
     def inspect
       "#<Puma::Client:0x#{object_id.to_s(16)} @ready=#{@ready.inspect}>"
--- a/History.md
+++ b/History.md
@@ -5,9 +5,11 @@
   *  Adds max_fast_inline as a configuration option for the Server object (#2406)
 
 * Bugfixes
+  * Fix hang on shutdown in refork (#2442)
+  * Fix `Bundler::GemNotFound` errors for `nio4r` gem during phased restarts (#2427, #2018)
+  * Server run thread safety fix (#2435) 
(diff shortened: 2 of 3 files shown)


MSP-Greg:
Two recent macOS JRuby failures:
https://github.com/puma/puma/runs/1266457331?check_suite_focus=true#step:8:68
https://github.com/puma/puma/runs/1260828085?check_suite_focus=true#step:8:93

lib/puma/client.rb
```diff
@@ -85,6 +85,12 @@ def initialize(io, env=nil)
 
     def_delegators :@io, :closed?
 
+    # Test to see if io meets a bare minimum of functioning, @to_io needs to be
+    # used for MiniSSL::Socket
+    def io_ok?
+      ::BasicSocket === @to_io && !closed?
```
nateberkopec, line 91: personally I'd prefer `@to_io.is_a?(BasicSocket)`

nateberkopec:
Seems valid to me.

lib/puma/client.rb
MSP-Greg: Sure.  I'm rebasing & since this should probably have a History.md entry, I'll update it.  I'll do the links later, as it's now two command lines...

PR #2434, opened 2020-10-18
Updates for non MRI CI stablity [changelog skip]

MSP-Greg:
### Description

This PR contains various updates to (hopefully) increase the stability of the non MRI CI.

I ran this fourteen times, and only two jobs failed (excepting jruby-head). TruffleRuby failed once in `bundle install`, and a JRuby job failed once in the manner shown in PR https://github.com/puma/puma/pull/2432.

#### Actions changes

1. Big Sur is available on Actions, added a single job in both MRI & non-MRI workflows.

2. Set all JRuby jobs to use JDK11.  Since the matrix is sparse, changed it to an 'include' only matrix.

#### Test Changes

3. `test/helper.rb

    * changed the `run` method prepended into `Minitest::Test`.  `setup` and the test method now run in the same `Timeout` thread.
    
    * Only run one `Timeout.timeout` for teardown methods.

4. `TestIntegrationSingle#test_int_refuse` appears to be the test that was intermittently locking up the JRuby CI.  Now skipped on JRuby.

5. `TestLauncher#test_puma_stats` timed out and failed intermittently, but only on jruby-head.  It still times out, but often passes on retry.

6. 'add small sleep after `Server#run`' is added in three files for JRuby only.  Possibly helps with timeout issues.  Again, these are intermittent.

5 & 6 are largely jruby-head issues, and since they are intermittent, it's hard to determine a true solution.  Or, the last four commits I could remove or squash ('jruby-head band-aids').  I think.

### DIFF
--- a/.github/workflows/mri.yml
+++ b/.github/workflows/mri.yml
@@ -26,6 +26,7 @@ jobs:
           - { os: windows-2019 , ruby: mingw }
           - { os: ubuntu-20.04 , ruby: 2.7   , no-ssl: ' no SSL' }
           - { os: windows-2019 , ruby: 2.7   , no-ssl: ' no SSL' }
+          - { os: macos-11.0   , ruby: 2.7  }
         exclude:
           - { os: ubuntu-20.04 , ruby: 2.2  }
           - { os: ubuntu-20.04 , ruby: 2.3  }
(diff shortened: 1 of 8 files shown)


nateberkopec:
> Big Sur is available on Actions

Wow TIL, that's great. Good job GitHub!

> add small sleep

Obviously not ideal, but I much much prefer a test suite that we can trust over one that's always red. So this is an improvement that we can fix up later.

Feels so good to see that bright green check.

PR #2435, opened 2020-10-19
Server run thread safety fix [changelog skip]

wjordan:
### Description
This PR fixes a race-condition creating the `@check, @notify` pipe in `Server`, which caused various intermittent test failures on JRuby.

The issue can be triggered by a call to `#run` (which calls `#handle_servers` in a background-thread) followed immediately by `#stop` (which calls `#notify_safely`). This can result in two separate `@check, @notify` pipes getting created concurrently, causing the signal sent by `#stop` to not get picked up by the listen-loop in `#handle_servers`, and therefore fail to exit as expected. This PR adds a test `test_run_stop_thread_safety` demonstrating the race condition that consistently fails on JRuby without this fix (see [this test run](https://github.com/wjordan/puma/actions/runs/314657885)).

The fix moves pipe-creation into `Server#run` (before the background thread is spawned), making the behavior more consistent that the server will stop as long as the call to `#stop` comes after the call to `#run` returns.

This PR includes a related timing change in `Single`, moving the `Use Ctrl-C to stop` log message after the call to `Server#run`. This is because the `wait_for_server_to_boot` integration-test helper uses this log message to determine when the server is 'booted', and some integration tests (e.g., `test_int_refuse`) send an `INT` signal (which calls `stop`) immediately after this log message is printed, causing intermittent failures if the `stop` call arrived before the `run` call finished. The change ensures that `stop` always occurs after `run`, avoiding intermittent failures in this test.

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -248,6 +248,8 @@ module Puma
         @thread_pool.auto_trim!(@auto_trim_time)
       end
 
+      @check, @notify = Puma::Util.pipe unless @notify
+
       @events.fire :state, :running
 
       if background
@@ -300,7 +302,6 @@ module Puma
     end
 
     def handle_servers
-      @check, @notify = Puma::Util.pipe unless @notify
       begin
         check = @check
         sockets = [check] + @binder.ios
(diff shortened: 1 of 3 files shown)


test/test_puma_server.rb
```diff
@@ -1141,4 +1141,12 @@ def test_client_quick_close_no_lowlevel_error_handler_call
     sleep 0.5
     assert_empty @events.stdout.string
   end
+
+  def test_run_stop_thread_safety
+    100.times do
```
nateberkopec, line 1146: How long does this take to run?

lib/puma/server.rb
```diff
@@ -640,19 +641,16 @@ def graceful_shutdown
     end
 
     def notify_safely(message)
-      @check, @notify = Puma::Util.pipe unless @notify
-      begin
-        @notify << message
-      rescue IOError, NoMethodError, Errno::EPIPE
-         # The server, in another thread, is shutting down
+      @notify << message
```
nateberkopec, line 644: Nice cleanup here.

test/test_puma_server.rb
MSP-Greg: I believe 0.3 sec on JRuby, less on MRI.  I should have reviewed this earlier, but...

I suspect I can revert the band-aids for JRuby (the sleep after run).  This is the same issue of `Server#run` getting messed up.

wjordan: [0.02s](https://github.com/puma/puma/runs/1276840378#step:8:356) on MRI, [0.28s](https://github.com/puma/puma/runs/1276840522#step:9:422) on JRuby

MSP-Greg (approved):

Issue #1367, opened 2017-07-12
systemd socket activation fails on jruby


dekellum:
In #976 a similar problem was reported but without reporter followup. I've finally gotten around to testing on jruby using most recent versions of everything. With later jruby the exception is now somewhat different. At minimum, we could document the incompatibly in /docs/systemd and/or warn/raise a better error message if an attempt is made to use socket activation with jruby. 

### Steps to reproduce

1. Follow /docs/systemd recipe and configure socket activation for a simple test app using latest puma and a recent MRI ruby (2.2.7).  Confirm that this works correctly (e.g. app works through activated socket)

2. Now switch the application to use jruby (9.1.12.0) and restart via systemctl

### Expected behavior

Socket activation works on jruby the same as it does on MRI ruby.

### Actual behavior

Works fine on MRI ruby, but with jruby it fails on startup:

~~~
IOError: closed stream
                    for_fd at org/jruby/ext/socket/RubyBasicSocket.java:109
  block in import_from_env at /home/david/.gem/jruby/2.3.0/gems/puma-3.9.1-java/lib/puma/binder.rb:65
                     times at org/jruby/RubyFixnum.java:305
  block in import_from_env at /home/david/.gem/jruby/2.3.0/gems/puma-3.9.1-java/lib/puma/binder.rb:63
                      each at org/jruby/RubyHash.java:1343
           import_from_env at /home/david/.gem/jruby/2.3.0/gems/puma-3.9.1-java/lib/puma/binder.rb:57
                initialize at /home/david/.gem/jruby/2.3.0/gems/puma-3.9.1-java/lib/puma/launcher.rb:52
                initialize at /home/david/.gem/jruby/2.3.0/gems/puma-3.9.1-java/lib/puma/cli.rb:68
                    <main> at /home/david/.gem/jruby/2.3.0/gems/puma-3.9.1-java/bin/puma:8
                      load at org/jruby/RubyKernel.java:979
                    <main> at /home/david/src/guppy_puma/sbin/puma:17
~~~

### System configuration

**Puma**: 3.9.1
**Ruby version**: jruby 9.1.12.0 (2.3.3) 2017-06-15 33c6439 OpenJDK 64-Bit Server VM 25.131-b11 on 1.8.0_131-b11 +jit [linux-x86_64]
**Rails version**: N/A

dekellum:
[...]

PR #2438, opened 2020-10-20
Add systemd notify and watchdog support

ekohl:
### Description
This is a rebased version of https://github.com/puma/puma/pull/2260 with additional changes. I squashed all commits from the other PR into a single one to keep credits for @acmh while also making it easy to rebase. All my changes are on top of it.

It is more than 169 lines change, but splitting it up might be hard with credit. If desired, I can first submit the additional events as a separate patch.


nateberkopec:
> It is more than 169 lines change, but splitting it up might be hard with credit. 

That's fine. I appreciate the effort and desire to share the credit :)

ekohl:
Added a `skip_on :jruby` to the tests since that doesn't work. At least a WIP test passed so let's see the full set now.

nateberkopec:
Hmmm... any idea why it doesn't work with JRuby? It _should_ in theory, right?

ekohl:
It gave an error on bind with the unix socket, I'm not sure why. https://github.com/puma/puma/blob/master/docs/systemd.md#socket-activation does state:
> Note: Socket activation doesn't currently work on JRuby. This is tracked in #1367.

Following the trail of issues it looks like this is still an issue on JRuby.

lib/puma/systemd.rb
```diff
@@ -0,0 +1,47 @@
...
+          sleep ping_f
+          SdNotify.watchdog
+        end
+      end
+    end
+
+    private
+
+    def watchdog_sleep_time
+      return unless SdNotify.watchdog?
```
ehelms, line 33: I think it would improve the understanding here to move this up to just after the `def start_watchdog` line.

ekohl: Agreed, moved.

nateberkopec:
LGTM but would appreciate review from others. We'll merge this in after 5.0.3 releases.

acmh:
Congratz @ekohl , thank you for the credit! Good work!

ekohl:
Thanks for the initial work @acmh! It was a very solid base to improve on.

rromanchuk:
I think i have a weird configuration issue using type=notify, everything starts up fine, serving request and then systemd kills at the WatchdogSec interval. I'll come back with better details after i poke around some more. Maybe i missed something with sd_notify 

```
puma.service: start operation timed out. Terminating.
puma[19721]: [19721] - Gracefully shutting down workers...
puma[20181]: On worker shutdown...
puma[20182]: On worker shutdown...
systemd[1]: puma.service: Failed with result 'timeout'.
```

lib/puma/systemd.rb
```diff
@@ -0,0 +1,46 @@
...
+      @events.on_restart { SdNotify.reloading }
+    end
+
+    def start_watchdog
+      return unless SdNotify.watchdog?
+
+      ping_f = watchdog_sleep_time
+
+      log "Pinging systemd watchdog every #{ping_f.round(1)} sec"
+      Thread.new do
```
schuetzm, line 23: Hmm... doesn't it defeat the purpose if you start the watchdog as a separate thread? How would that ever fail? It should either be integrated into the main event loop or at least somehow check whether the service is actually still working.

ekohl: I think that's a fair point but I wouldn't know how to do it better. Perhaps this is a good enough implementation for now and it allows for a better implementation in the future. Maybe the thread should perform a request to see if it's served? The question then becomes: which one? Should it be configurable?

ekohl:
> I think i have a weird configuration issue using type=notify, everything starts up fine, serving request and then systemd kills at the WatchdogSec interval. I'll come back with better details after i poke around some more. Maybe i missed something with sd_notify

Did you figure it out? Perhaps you should start without the watchdog to see if that does work.

Dantemss:
@rromanchuk I just ran into a similar issue with a different daemon... if your `ExecStart` uses a shell, make sure you `exec` the puma command rather than just calling it so that systemd gets the correct process as the main process (otherwise you have both a shell and puma processes, systemd thinks the shell is the main process and for whatever reason that makes `SdNotify.watchdog?` return false)

mperham:
With Sidekiq I decided to vendor the code for the `sd_notify` gem because it's only ~50 lines of stable code. I didn't want to require the user to add the gem to their gemfile manually, that shouldn't be something they need to care about IMO.

PR #2441, opened 2020-10-21
Fix-up test_refork, hot_restart_does_not_drop_connections [changelog skip]

MSP-Greg:
### Description

**test_refork** - since refork is unlikely to be used with less than 3 workers, set workers to 3, previous value was 2.  Test sends 1k+ sockets, and will error if all do not receive a response.  Removed 'speed of light' client loop, as they can cause issues with CI.

**hot_restart_does_not_drop_connections** - fixed Windows error.  Removed 'speed of light' client loop.

**test/integration.rb** - added `fast_connect` which opens and sends a request without a read.  Uses `fast_write` (`syswrite`).  Moved `fast_write` to below `fast_connect`.

### DIFF
--- a/test/helpers/integration.rb
+++ b/test/helpers/integration.rb
@@ -116,6 +116,35 @@ class TestIntegration < Minitest::Test
     s
   end
 
+  # use only if all socket writes are fast
+  # does not wait for a read
+  def fast_connect(path = nil, unix: false)
+    s = unix ? UNIXSocket.new(@bind_path) : TCPSocket.new(HOST, @tcp_port)
+    @ios_to_close << s
+    fast_write s, "GET /#{path} HTTP/1.1\r\n\r\n"
+    s
+  end
+
+  def fast_write(io, str)
+    n = 0
+    while true
+      begin
+        n = io.syswrite str
+      rescue Errno::EAGAIN, Errno::EWOULDBLOCK => e
+        if !IO.select(nil, [io], nil, 5)
+          raise e
+        end
+
+        retry
+      rescue Errno::EPIPE, SystemCallError, IOError => e
+        raise e
+      end
+
+      return if n == str.bytesize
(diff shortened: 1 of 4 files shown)


MSP-Greg:
@wjordan @cjlarose

`TestIntegrationCluster@test_refork` has been intermittently failing.  Deos the following look ok:
https://github.com/puma/puma/blob/dce4b60c902ea798317fd2e23f14fddf83f11271/test/test_integration_cluster.rb#L193-L217

wjordan:
> TestIntegrationCluster@test_refork has been intermittently failing. Deos the following look ok:

@MSP-Greg I just pushed #2442 which have fixes the underlying issues that were causing `test_refork` to fail.

MSP-Greg:
@wjordan

> just pushed #2442 which have fixes the underlying issues

Thanks.  The above test change to `test_refork` was about a more real world test.

The issue you've fixed is part of the larger issue of "Puma changes (various restart types/forks, etc) can be triggered internally and externally.  Do they play well together, especially if they overlap before Puma has reached a 'steady-state'?"

wjordan:
> The above test change to test_refork was about a more real world test.

Ah, I see- yes, in that case I think the changes look OK for making the tests a little more realistic :+1:

I think it can also be helpful to have a handful of _unrealistic_ tests as well, to expose and cover edge-cases that might be lurking in a public API and to help clarify behavior around those edges that might otherwise pop up unexpectedly on users. Ideally there is a good mix and I think Puma's test coverage does a pretty good job here overall.

MSP-Greg:
Given the choice of 'unrealistic' or 'edge-case', the 2nd might be better.  Regardless, I agree.

I think there could be more (or better) phased-restart/SIGUSR1 tests, but haven't looked at them yet.  Also, updating the API (v6) may allow some testing to be moved out the integration/popen tests into faster, 'narrower' tests...

PR #2443, opened 2020-10-22
[2440] Better byebug integration

### LINKED ISSUE #2440: Better byebug integration
[Cluster-mode Puma and byebug (default gem in Rails) play poorly together.](https://stackoverflow.com/questions/40998929/puma-rails-5-binding-pry-only-available-for-60-seconds-before-timeout)

Not sure what a good course of action is here.

juanmanuelramallo:
# Description
Proposal to solve #2440

I tested it locally with a Rails app and it worked as expected. So if this approach is in the right path, I'd need pointers on how to write tests for it.

I haven't added an entry to the HISTORY.md file because I don't know if this is correct. I'd like to first verify that. 

- [ ] I have added an entry to [History.md](../blob/master/History.md) if this PR fixes a bug or adds a feature. If it doesn't need an entry to HISTORY.md, I have added `[changelog skip]` or `[ci skip]` to the pull request title.

Closes #2440

### DIFF
--- a/lib/puma/configuration.rb
+++ b/lib/puma/configuration.rb
@@ -12,6 +12,7 @@ module Puma
     DefaultTCPHost = "0.0.0.0"
     DefaultTCPPort = 9292
     DefaultWorkerTimeout = 60
+    DefaultDevelopmentWorkerTimeout = 3600
     DefaultWorkerShutdownTimeout = 30
   end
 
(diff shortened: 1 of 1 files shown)


juanmanuelramallo:
Hey @nateberkopec, is this something similar to what you had in mind?

nateberkopec:
I'm not sure this will work. Many people may be (accidentally, true) deployed into a "development" environment, so this PR makes me nervous.

nateberkopec:
I think you may have misread my comment on #2440. I meant maybe we should open a PR _to Rails_ (the default puma config, I mean), and increase worker_timeout in dev mode there.

Certainly as written, this change would have to wait until 6.0.

What do you think?

juanmanuelramallo:
Fair enough! 
Yep it makes total sense to actually update the default puma config from Rails itself rather than doing it here. 

Closing the PR then.

PR #2448, opened 2020-10-23
Remove accept_nonblock.rb, add test_integration_ssl.rb

MSP-Greg:
### Description

`accept_nonblock.rb` has existed since Puma's SSL implementation used Ruby OpenSSL classes and objects.  It has since removed those, and only uses the OpenSSL libraries.

Remove the file, as it is no longer needed.  It loaded `openssl`, and that is also unneeded on MRI Rubies.

Add a new 'ssl' test file that starts a Puma server using `IO.popen`, which decouples the CI environment and what it has loaded.  The test is run on all platforms.


lib/puma/binder.rb
```diff
@@ -12,7 +12,15 @@ module Puma
   if HAS_SSL
     require 'puma/minissl'
     require 'puma/minissl/context_builder'
-    require 'puma/accept_nonblock'
+
+    # odd bug in 'pure Ruby' nio4r only with 2.5.2.  Fixed in newer version,
```
dentarg, line 16:
```suggestion
    # odd bug in 'pure Ruby' nio4r v2.5.2 (fixed in newer versions),
```

To make it a bit more clear what software 2.5.2 refers to.

I assume https://github.com/socketry/nio4r/issues/238 is the issue you are referring to, should we include that URL in the comment?

MSP-Greg:
@dentarg - how about:
```
    # Odd bug in 'pure Ruby' nio4r verion 2.5.2, which installs with Ruby 2.3.
    # NIO doesn't create any OpenSSL objects, but it rescues an OpenSSL error.
    # The bug was that it did not require openssl.
    # @todo remove when Ruby 2.3 support is dropped
    #
```

I suspect the reason the bug wasn't noticed was because their test suite uses net/http ssl clients, and it loads Ruby OpenSSL.  Which is the reason I added a separate test in this PR.

lib/puma/binder.rb
MSP-Greg: @dentarg - how about:
```
    # Odd bug in 'pure Ruby' nio4r verion 2.5.2, which installs with Ruby 2.3.
    # NIO doesn't create any OpenSSL objects, but it rescues an OpenSSL error.
    # The bug was that it did not require openssl.
    # @todo remove when Ruby 2.3 support is dropped
    #
```

I suspect the reason the bug wasn't noticed was because their test suite uses net/http ssl clients, and it loads Ruby OpenSSL.  Which is the reason I added a separate test in this PR.

**EDIT:** Since this is only an issue with Windows Ruby 2.3, and has been fixed in nio4r, I don't  know how much to explain.  It feels odd linking to my own issues/PR's in other repos...

dentarg: @MSP-Greg LGTM! 👍

MSP-Greg: Updated per discussion

PR #2472, opened 2020-10-29
Log the configuration if an env variable is provided

### LINKED ISSUE #2414: Feature idea: `puma config` command

One of the things that made https://github.com/puma/puma/issues/2371 hard to debug was that Puma's behavior comes from a mix of defaults, config file defined settings, and config flags. It can be tricky to see what the user's settings "really" are for debugging purposes.

I'm proposing we add a `puma config` command that would:

1) Boot puma, including loading the config/puma.rb file
2) Output config
3) Exit

Alternatively, we could add some env var like `PUMA_DEBUG_CONFIG` that could output puma configuration in the boot process. 

**Additional context**

One area to be concerned with is that running puma via `rails s` provides more config as rails has their own default values that they pass in. To get a "true" representation of the actual config, it might be more accurate to go with the alternative method of outputting config in the web bootup process directly. Maybe both options would be helpful.

ccverak:
### Description
This PR adds support for printing the puma config if an env variable is provided as agreed in #2414 
I'm aware am missing several things including the tests but I just want early feedback if possible, this is my first contribution to the project :) 
Closes #2414 

Here is an example of the output:
image

### DIFF
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -87,6 +87,8 @@ module Puma
       Puma.stats_object = @runner
 
       @status = :run
+
+      log_config if ENV['PUMA_LOG_CONFIG']
     end
 
     attr_reader :binder, :events, :config, :options, :restart_dir
@@ -520,6 +522,15 @@ module Puma
       else
         Bundler.with_unbundled_env { yield }
       end
+    end
+
+    def log_config
+      log "Configuration:"
+
+      @config.final_options
+        .each { |config_key, value| log "- #{config_key}: #{value}" }
+
+      log "\n"
     end
   end
 end
--- a/History.md
+++ b/History.md
@@ -2,6 +2,7 @@
 
 * Features
   * Your feature goes here <Most recent on the top, like GitHub> (#Github Number)
+  * Prints the loaded configuration if the environment variable `PUMA_LOG_CONFIG` is present ([#2472])
   * Integrate with systemd's watchdog and notification features ([#2438])
   * Adds max_fast_inline as a configuration option for the Server object ([#2406])
   * You can now fork workers from worker 0 using SIGURG w/o fork_worker enabled [#2449]
--- a/README.md
+++ b/README.md
@@ -85,6 +85,9 @@ Puma provides numerous options. Consult `puma -h` (or `puma --help`) for a full 
 You can also find several configuration examples as part of the
 [test](https://github.com/puma/puma/tree/master/test/config) suite.
 
+For debugging purposes, you can set the environment variable `PUMA_LOG_CONFIG` with a value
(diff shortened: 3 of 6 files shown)


lib/puma/launcher.rb
```diff
@@ -521,5 +523,21 @@ def with_unbundled_env
         Bundler.with_unbundled_env { yield }
       end
     end
+
+    def log_config
+      log "Configuration:"
+
+      @config.options.default_options
```
nateberkopec, line 530: Hmmm. In practice, I'd rather that this method just called something like `@config.final_options` and then iterate through those. Then, we also use _that_ method when calculating the final config. That method may not exist yet, you may have to refactor it out and use it in both places.

Normally I wouldn't be so eager to DRY up 2 places, but in this case it's critical for correctness.

nateberkopec: OK, I see now why that doesn't exist, because this would be the first place in Puma we try to iterate through every option.

So I think we need a new method on UserFileDefaultOptions that basically does what you do here.

nateberkopec (changes requested):

lib/puma/launcher.rb
MSP-Greg: I agree.  This issue has been on my list, as our use of a Server object in CI (and also by capybara) means that the Server object needs to have a bunch of default values in it's initialize method, which is a PITA and prone to error, as they exist in `Puma::Configuration#puma_default_options`...

ccverak:
Feedback addressed :) I'd appreciate some help with the naming, especially with `LOG_CONFIG`, I was also considering `DUMP_CONFIG` or `PRINT_CONFIG` but wasn't too convinced of any... which one do you prefer?

/cc @nateberkopec and @MSP-Greg

MSP-Greg:
I'll leave it to @nateberkopec, but we might consider whether to add PUMA to the ENV name, it can help with 'what does this affect?' questions...

lib/puma/launcher.rb
```diff
@@ -87,6 +87,8 @@ def initialize(conf, launcher_args={})
       Puma.stats_object = @runner
 
       @status = :run
+
+      log_config if ENV['LOG_CONFIG']
```
nateberkopec, line 91: @MSP-Greg is right - any new ENV variables should be prefixed with "PUMA_"

nateberkopec (changes requested):

lib/puma/launcher.rb
ccverak: Makes sense 👍

ccverak:
Feedback addressed

nateberkopec:
Brilliant, great first feature.

ccverak:
Thanks you so much @nateberkopec and @MSP-Greg for the support 😄

Issue #2471, opened 2020-10-29
phased-restart errors after upgrading from Puma 3.12.6 to 5.0.4


gingerlime:
**Describe the bug**

This seems to happen when doing phased-restarts using versioned deploy folders with symlinks.

We use a folder structure similar to this:

```
/app/app.d/{version (int)}
/app/app -> /app/app.d/{version} # symlink
/app/prev -> /app/app.d/{version-1} # symlink
```

When we deploy, we increment the version, run bundle etc, then switch the symlink and do a phased-restart of puma.

All seems to work fine, but after a few deploys, we also start pruning older versions, and then we see this error in the puma.log

```
/usr/lib/ruby/gems/2.6.0/gems/bundler-2.1.4/lib/bundler/spec_set.rb:86:in `block in materialize': Could not find json-2.3.1 in any of the sources (Bundler::GemNotFound)
        from /usr/lib/ruby/gems/2.6.0/gems/bundler-2.1.4/lib/bundler/spec_set.rb:80:in `map!'
        from /usr/lib/ruby/gems/2.6.0/gems/bundler-2.1.4/lib/bundler/spec_set.rb:80:in `materialize'
        from /usr/lib/ruby/gems/2.6.0/gems/bundler-2.1.4/lib/bundler/definition.rb:170:in `specs'
        from /usr/lib/ruby/gems/2.6.0/gems/bundler-2.1.4/lib/bundler/definition.rb:237:in `specs_for'
        from /usr/lib/ruby/gems/2.6.0/gems/bundler-2.1.4/lib/bundler/definition.rb:226:in `requested_specs'
        from /usr/lib/ruby/gems/2.6.0/gems/bundler-2.1.4/lib/bundler/runtime.rb:101:in `block in definition_method'
        from /usr/lib/ruby/gems/2.6.0/gems/bundler-2.1.4/lib/bundler/runtime.rb:20:in `setup'
        from /usr/lib/ruby/gems/2.6.0/gems/bundler-2.1.4/lib/bundler.rb:149:in `setup'
        from /usr/lib/ruby/gems/2.6.0/gems/bundler-2.1.4/lib/bundler/setup.rb:20:in `block in <top (required)>'
        from /usr/lib/ruby/gems/2.6.0/gems/bundler-2.1.4/lib/bundler/ui/shell.rb:136:in `with_level'
[...]

PR #2473, opened 2020-10-31
Remove use of json gem from WorkerHandle#ping!

cjlarose:
### Description

This change is the first in a series meant to address phased restart errors related to the `json` gem described in #2471. The changes for this specific PR are described in a comment in the same issue: https://github.com/puma/puma/issues/2471#issuecomment-719834731

This PR essentially reverts the changes to the handling of puma worker stats introduced by #2124 in a way that intentionally avoids the use of the `json` gem.

I added an integration test to ensure that users can upgrade/downgrade the JSON gem as part of phased restart (it reliably reproduces the error reported in #2471). While I was in there, I added a similar integration test for upgrading/downgrading the nio4r gem. That test was effectively fixed by #2427.

### DIFF
--- a/History.md
+++ b/History.md
@@ -10,6 +10,7 @@
   * Your bugfix goes here <Most recent on the top, like GitHub> (#Github Number)
   * Ignore illegal (by Rack spec) response header ([#2439])
   * Close idle connections immediately on shutdown ([#2460])
+  * Fix some instances of phased restart errors related to the `json` gem (#2473)
 
 ## 5.0.4 / 2020-10-27
 
--- a/lib/puma/cluster/worker.rb
+++ b/lib/puma/cluster/worker.rb
@@ -108,11 +108,17 @@ module Puma
           server_thread = server.run
           stat_thread ||= Thread.new(@worker_write) do |io|
             Puma.set_thread_name "stat payload"
+            base_payload = "p#{Process.pid}"
 
             while true
               begin
-                require 'json'
-                io << "p#{Process.pid}#{server.stats.to_json}\n"
+                b = server.backlog || 0
+                r = server.running || 0
+                t = server.pool_capacity || 0
+                m = server.max_threads || 0
+                rc = server.requests_count || 0
+                payload = %Q!#{base_payload}{ "backlog":#{b}, "running":#{r}, "pool_capacity":#{t}, "max_threads": #{m}, "requests_count": #{rc} }\n!
+                io << payload
               rescue IOError
                 Thread.current.purge_interrupt_queue if Thread.current.respond_to? :purge_interrupt_queue
                 break
--- a/lib/puma/cluster/worker_handle.rb
+++ b/lib/puma/cluster/worker_handle.rb
@@ -42,8 +42,11 @@ module Puma
 
       def ping!(status)
         @last_checkin = Time.now
-        require 'json'
-        @last_status = JSON.parse(status, symbolize_names: true)
+        captures = status.match(/{ "backlog":(?<backlog>\d*), "running":(?<running>\d*), "pool_capacity":(?<pool_capacity>\d*), "max_threads": (?<max_threads>\d*), "requests_count": (?<requests_count>\d*) }/)
+        @last_status = captures.names.inject({}) do |hash, key|
+          hash[key.to_sym] = captures[key].to_i
+          hash
+        end
       end
 
       # @see Puma::Cluster#check_workers
--- a/test/helpers/integration.rb
+++ b/test/helpers/integration.rb
@@ -53,11 +53,12 @@ class TestIntegration < Minitest::Test
       config_file.close
       config = "-C #{config_file.path}"
     end
+    puma_path = File.expand_path '../../../bin/puma', __FILE__
     if unix
-      cmd = "#{BASE} bin/puma #{config} -b unix://#{@bind_path} #{argv}"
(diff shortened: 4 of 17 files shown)


nateberkopec:
Really bad-ass tests here. Nice work.

MSP-Greg:
@cjlarose

I was also looking at this the weekend, but added another phased restart, as the some people posting with the issue implied that the problem happened after more than a single phased restart.

So, I added the following to the bottom of `TestWorkerGemIndependence#change_gem_version_during_phased_restart`:
```ruby
    # 2nd phased restart, already installed
    set_release_symlink File.expand_path(old_app_dir, __dir__)
    start_phased_restart

    connection = connect
    initial_reply = read_body(connection)
    assert_equal old_version, initial_reply
```

That failed.  Also, I believe that $LOAD_PATH & ENV['BUNDLER_GEMFILE'] always use absolute/realdirpath entries, so that may also affect things.  I haven't gotten much further.

When debugging, things got messy.  I was looking at a lot of things, including $LOAD_PATH, which, if there are a lot entries, may add a lot of disk hits/lookups when code calls `require`.  When multiple phased restart's are done with a symlinked directory, there may be alot of entries in $LOAD_PATH that point to deleted folders?

And lastly, the tests currently run under a Bundler context, so maybe pass a block to `bundle_install` so `cli_server '--prune-bundler -w 1'` can run within `with_unbundled_env`?

MSP-Greg:
@cjlarose

What I said above could be interpreted a few ways.  I was working on it intermittently, but my 'gut feeling' was that this is really messy.  Hence, I'm asking for your opinion.  I'm looking at a generic issue something like:

Puma may have dependencies on external gems and/or Ruby default gems.  These may also be shared by the application.  Given an indeterminate number of phased restarts, will changes to those gems work correctly in both Puma and the app?

cjlarose:
> When multiple phased restart's are done with a symlinked directory, there may be alot of entries in $LOAD_PATH that point to deleted folders?

I haven't observed this. What I normally see (assuming `prune_bundler` is on) is that the `LOAD_PATH` will contain the normal default Ruby stuff and entries for `puma` itself (and for `json`) from the `directory` in which the puma master process started. I haven't seen `LOAD_PATH` on the master process gain new entries after phased restarts.

> And lastly, the tests currently run under a Bundler context, so maybe pass a block to bundle_install so cli_server '--prune-bundler -w 1' can run within with_unbundled_env?

Running `cli_server` in the unbundled env seems like the way to go. I'll open a PR for this.

> Puma may have dependencies on external gems and/or Ruby default gems. These may also be shared by the application. Given an indeterminate number of phased restarts, will changes to those gems work correctly in both Puma and the app?

If there are any external gems on which the Puma master process relies, application developers will be unable to upgrade/downgrade that gem with a phased restart. If any of those gems have native extensions, operators will be unable to delete those gems on disk for old releases. Given this, I think our best path forward is to make it so that the Puma master process has _no_ dependencies on _any_ external gems. The good news is that we're almost there. #2427 removed the dependency on `nio4r` in the master process and soon we'll get rid of `json`. I think that's the last one.

MSP-Greg:
@cjlarose

> I haven't observed this.

At present, not being able to use fork locally (Windows), I always assumed that the $LOAD_PATH in fork'd code was the same as the master process $LOAD_PATH.  Never output in CI, so I don't know.

By chance, have you tried adding an additional phased-restart?

Thanks, Greg

cjlarose:
> By chance, have you tried adding an additional phased-restart?

Doing an additional phased restart passes just fine, as long as you change the `start_phased_restart` helper to expect the new phase number

```diff
diff --git a/test/test_worker_gem_independence.rb b/test/test_worker_gem_independence.rb
index 05385884..9ce63b57 100644
--- a/test/test_worker_gem_independence.rb
+++ b/test/test_worker_gem_independence.rb
@@ -56,6 +56,14 @@ class TestWorkerGemIndependence < TestIntegration
     connection = connect
     new_reply = read_body(connection)
     assert_equal new_version, new_reply
+
+    # back to old version
+    set_release_symlink File.expand_path(old_app_dir, __dir__)
+    start_phased_restart expected_phase: '2'
+
+    connection = connect
+    initial_reply = read_body(connection)
+    assert_equal old_version, initial_reply
   end
 
   def current_release_symlink
@@ -67,10 +75,10 @@ class TestWorkerGemIndependence < TestIntegration
     FileUtils.symlink target_dir, current_release_symlink, force: true
   end
 
-  def start_phased_restart
+  def start_phased_restart(expected_phase: '1')
     Process.kill :USR1, @pid
 
-    true while @server.gets !~ /booted, phase: 1/
+    true while @server.gets !~ /booted, phase: #{expected_phase}/
   end
 
   def with_unbundled_env
```

cjlarose:
> At present, not being able to use fork locally (Windows), I always assumed that the $LOAD_PATH in fork'd code was the same as the master process $LOAD_PATH. Never output in CI, so I don't know.

Yeah, `$LOAD_PATH` is the same after a fork. In a puma worker process, `$LOAD_PATH` will be the same as the puma master process up until it executes `require 'bundler/setup` (or the application modifies its `$LOAD_PATH` manually or via `gem`).

MSP-Greg:
> long as you change the

How many stupid mistakes do I get a month?

Re $LOAD_PATH, makes sense.  I should be running Ubuntu sometime soon...

PR #1801, opened 2019-05-20
Fix: Invalid JSON on gc-stats

Jesus:
Credits of this patch should be for @jdsundberg, who reported the issue
but never sent a patch. Link to his bug report is below.

Closes #1687.

PR #2479, opened 2020-11-03
Remove use of json gem

### LINKED ISSUE #2471: phased-restart errors after upgrading from Puma 3.12.6 to 5.0.4

cjlarose:
### Description

Fixes #2471. The changes proposed in this PR in particular are described in a [comment](https://github.com/puma/puma/issues/2471#issuecomment-719834731) on the same issue. This PR introduces a new `Puma::JSON` module, which is meant to handle the specific places where Puma needs to perform JSON serialization. 

- [X] I have added an entry to [History.md](../blob/master/History.md) if this PR fixes a bug or adds a feature. If it doesn't need an entry to HISTORY.md, I have added `[changelog skip]` or `[ci skip]` to the pull request title.

### DIFF
--- a/History.md
+++ b/History.md
@@ -13,6 +13,7 @@
   * Ignore illegal (by Rack spec) response header ([#2439])
   * Close idle connections immediately on shutdown ([#2460])
   * Fix some instances of phased restart errors related to the `json` gem (#2473)
+  * Remove use of `json` gem to fix phased restart errors (#2479)
 
 ## 5.0.4 / 2020-10-27
 
--- a/lib/puma/json.rb
+++ b/lib/puma/json.rb
@@ -42,7 +42,7 @@ module Puma
           output << '}'
         when String
           serialize_string output, value
-        when Integer
+        when Integer, Float
           output << value.to_s
         else
           raise SerializationError, "Unexpected value of type #{value.class}"
--- a/lib/puma/json.rb
+++ b/lib/puma/json.rb
@@ -27,6 +27,10 @@ module Puma
             serialize_value output, member
           end
           output << ']'
+        when String
+          output << '"'
+          output << value
+          output << '"'
         when Integer
           output << value.to_s
         end
--- a/lib/puma/json.rb
+++ b/lib/puma/json.rb
@@ -29,7 +29,7 @@ module Puma
           output << ']'
         when String
           output << '"'
-          output << value
+          output << value.gsub('"', '\"')
           output << '"'
         when Integer
           output << value.to_s
--- a/lib/puma/json.rb
+++ b/lib/puma/json.rb
@@ -80,10 +80,8 @@ module Puma
 
       def serialize_object_key(output, value)
         case value
-        when Symbol
+        when Symbol, String
           serialize_string output, value.to_s
-        when String
-          serialize_string output, value
         else
           raise SerializationError, "Could not serialize object of type #{value.class} as object key"
         end
--- a/lib/puma/json.rb
+++ b/lib/puma/json.rb
@@ -54,6 +54,8 @@ module Puma
           output << 'true'
         when false
           output << 'false'
+        when nil
+          output << 'null'
         else
           raise SerializationError, "Unexpected value of type #{value.class}"
         end
--- a/lib/puma/json.rb
+++ b/lib/puma/json.rb
@@ -51,6 +51,8 @@ module Puma
 
       def serialize_object_key(output, value)
         case value
+        when Symbol
+          serialize_string output, value.to_s
         when String
           serialize_string output, value
         else
--- a/lib/puma/json.rb
+++ b/lib/puma/json.rb
@@ -17,6 +17,11 @@ module Puma
   # where Puma relies on JSON serialization internally.
 
   module JSON
+    QUOTE = /"/
+    BACKSLASH = /\\/
+    CONTROL_CHAR_TO_ESCAPE = /[\x00-\x1F]/ # As required by ECMA-404
+    CHAR_TO_ESCAPE = Regexp.union QUOTE, BACKSLASH, CONTROL_CHAR_TO_ESCAPE
+
     class SerializationError < StandardError; end
 
     class << self
@@ -63,13 +68,13 @@ module Puma
 
       def serialize_string(output, value)
         output << '"'
-        output << value.gsub(/[\\"\x00-\x1F]/) do |character|
+        output << value.gsub(CHAR_TO_ESCAPE) do |character|
           case character
-          when '\\'
+          when BACKSLASH
             '\\\\'
-          when '"'
+          when QUOTE
             '\\"'
-          when /[\x00-\x1F]/
+          when CONTROL_CHAR_TO_ESCAPE
             '\u%.4X' % character.ord
           else
             character
--- a/lib/puma/json.rb
+++ b/lib/puma/json.rb
@@ -21,14 +21,9 @@ module Puma
 
     class << self
       def generate(value)
-        case value
-        when Array, Hash
-          parts = []
-          serialize_value parts, value
-          parts.join ''
-        else
-          raise SerializationError, "Could not serialize object of type #{value.class}"
-        end
+        parts = []
+        serialize_value parts, value
+        parts.join ''
       end
 
       private
--- a/lib/puma/json.rb
+++ b/lib/puma/json.rb
@@ -20,6 +20,10 @@ module Puma
 
       def serialize_value(output, value)
         case value
+        when true
+          output << 'true'
+        when false
+          output << 'false'
         when Array
           output << '['
           value.each_with_index do |member, index|
--- a/lib/puma/json.rb
+++ b/lib/puma/json.rb
@@ -63,7 +63,16 @@ module Puma
 
       def serialize_string(output, value)
         output << '"'
-        output << value.gsub(/[\\"]/, '\\' => '\\\\', '"' => '\\"')
+        output << value.gsub(/[\\"]/) do |character|
+          case character
+          when '\\'
+            '\\\\'
+          when '"'
+            '\\"'
+          else
+            character
+          end
+        end
         output << '"'
       end
 
--- a/lib/puma/json.rb
+++ b/lib/puma/json.rb
@@ -40,6 +40,8 @@ module Puma
           serialize_string output, value
         when Integer
           output << value.to_s
+        else
+          raise SerializationError, "Unexpected value of type #{value.class}"
         end
       end
 
--- a/lib/puma/json.rb
+++ b/lib/puma/json.rb
@@ -29,7 +29,7 @@ module Puma
           output << ']'
         when String
           output << '"'
-          output << value.gsub('"', '\"')
+          output << value.gsub(/[\\"]/, '\\' => '\\\\', '"' => '\\"')
           output << '"'
         when Integer
           output << value.to_s
--- a/lib/puma/json.rb
+++ b/lib/puma/json.rb
@@ -76,8 +76,6 @@ module Puma
             '\\"'
           when CONTROL_CHAR_TO_ESCAPE
             '\u%.4X' % character.ord
-          else
-            character
           end
         end
         output << '"'
--- a/lib/puma/json.rb
+++ b/lib/puma/json.rb
@@ -1,6 +1,21 @@
 # frozen_string_literal: true
 
 module Puma
+
+  # Puma deliberately avoids the use of the json gem and instead performs JSON
+  # serialization without any external dependencies. In a puma cluster, loading
+  # any gem into the puma master process means that operators cannot use a
+  # phased restart to upgrade their application if the new version of that
+  # application uses a different version of that gem. The json gem in
+  # particular is additionally problematic because it leverages native
+  # extensions. If the puma master process relies on a gem with native
+  # extensions and operators remove gems from disk related to old releases,
+  # subsequent phased restarts can fail.
+  #
+  # The implementation of JSON serialization in this module is not designed to
+  # be particularly full-featured or fast. It just has to handle the few places
+  # where Puma relies on JSON serialization internally.
+
   module JSON
     class SerializationError < StandardError; end
 
--- a/lib/puma/json.rb
+++ b/lib/puma/json.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+module Puma
+  module JSON
+    class SerializationError < StandardError; end
+
+    class << self
+      def generate(value)
+        case value
+        when Array
+          parts = []
+          serialize_value parts, value
+          parts.join ''
+        else
+          raise SerializationError, "Could not serialize object of type #{value.class}"
+        end
+      end
+
+      private
+
+      def serialize_value(output, value)
+        case value
+        when Array
+          output << '['
+          value.each_with_index do |member, index|
+            output << ',' if index != 0
+            serialize_value output, member
+          end
+          output << ']'
+        when Integer
+          output << value.to_s
+        end
+      end
+    end
+  end
+end
--- a/lib/puma/json.rb
+++ b/lib/puma/json.rb
@@ -63,12 +63,14 @@ module Puma
 
       def serialize_string(output, value)
         output << '"'
-        output << value.gsub(/[\\"]/) do |character|
+        output << value.gsub(/[\\"\x00-\x1F]/) do |character|
           case character
           when '\\'
             '\\\\'
           when '"'
             '\\"'
+          when /[\x00-\x1F]/
+            '\u%.4X' % character.ord
           else
             character
           end
--- a/lib/puma/json.rb
+++ b/lib/puma/json.rb
@@ -35,17 +35,6 @@ module Puma
 
       def serialize_value(output, value)
         case value
-        when true
-          output << 'true'
-        when false
-          output << 'false'
-        when Array
-          output << '['
-          value.each_with_index do |member, index|
-            output << ',' if index != 0
-            serialize_value output, member
-          end
-          output << ']'
         when Hash
           output << '{'
           value.each_with_index do |(k, v), index|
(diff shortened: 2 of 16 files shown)


lib/puma/json.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+module Puma
```
nateberkopec, line 3: You really went above and beyond here. Better than what we had in 3.x by a mile!

nateberkopec:
100% for the approach, will do a line by line review soon

History.md
```diff
@@ -13,6 +13,7 @@
   * Ignore illegal (by Rack spec) response header ([#2439])
   * Close idle connections immediately on shutdown ([#2460])
   * Fix some instances of phased restart errors related to the `json` gem (#2473)
+  * Remove use of `json` gem to fixed phased restart errors (#2479)
```
rmacklin, line 16: Minor typo here:
```suggestion
  * Remove use of `json` gem to fix phased restart errors (#2479)
```

rmacklin (commented):
This is great!

The integration tests you added should prevent the specific scenarios that currently make it impossible to upgrade/downgrade `json` during a phased restart. If I understand correctly, though, the test suite could still pass even if someone were to introduce a new usage of `json` in some other code path, right?

I'm not sure there's much we can do about that (unless we want to ban the usage of `require 'json'` in general...) but just wanted to confirm that we'd still want to be on the lookout for new usages of `json` that could cause problems with phased restarts.

test/test_json.rb
```diff
@@ -54,4 +54,14 @@ def test_json_generates_string_for_hash_with_symbol_keys
     value = { key: 'value' }
     assert_equal '{"key":"value"}', Puma::JSON.generate(value)
   end
+
+  def test_json_generates_string_for_true
+    value = [true]
+    assert_equal '[true]', Puma::JSON.generate(value)
+  end
+
+  def test_json_generates_string_for_true
```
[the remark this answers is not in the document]
cjlarose: Commited your suggestion in  2bef0c4

History.md
cjlarose: Fixed up in 0bd5531

lib/puma/json.rb
```diff
@@ -40,6 +40,8 @@ def serialize_value(output, value)
           serialize_string output, value
         when Integer
           output << value.to_s
+        else
+          raise SerializationError, "Unexpected value of type #{value.class}"
```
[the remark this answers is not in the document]
cjlarose: Nice catch! 90f5141

cjlarose:
> The integration tests you added should prevent the specific scenarios that currently make it impossible to upgrade/downgrade `json` during a phased restart. If I understand correctly, though, the test suite could still pass even if someone were to introduce a new usage of `json` in some other code path, right?
> 
> I'm not sure there's much we can do about that (unless we want to ban the usage of `require 'json'` in general...) but just wanted to confirm that we'd still want to be on the lookout for new usages of `json` that could cause problems with phased restarts.

This is true. The main value of the test suite changes for phased restarts in this PR is to prevent regressions like those introduced by #1801 and #2124. There's additionally [a test that ensures that the `JSON` module isn't loaded when the puma master process starts up](https://github.com/puma/puma/blob/19e1584a134be3ce50b8852b57876fbcc4c7fe7d/test/test_integration_cluster.rb#L279-L284). But ultimately, you're right in that if someone else comes along and adds a "dynamic" `require 'json'` to the puma master process, that could introduce a new regression. It's not clear what the best path forward would be if we wanted to ensure that the `json` gem was _never_ loaded into the puma master process to prevent all regressions of this type. My hope is that the introduction of the `Puma::JSON` module is documentation enough to make it clear to contributors and to maintainers that we shouldn't `require 'json'` again.

lib/puma/json.rb
```diff
@@ -0,0 +1,85 @@
...
+  # be particularly full-featured or fast. It just has to handle the few places
+  # where Puma relies on JSON serialization internally.
+
+  module JSON
+    class SerializationError < StandardError; end
+
+    class << self
+      def generate(value)
+        case value
+        when Array, Hash
```
ekohl, line 25: Why is this needed? `serialize_value` already handles all types or raises a `SerializationError` if needed.

lib/puma/json.rb
```diff
@@ -0,0 +1,85 @@
...
+          parts = []
```
ekohl, line 26: Rather than an array, should it use [StringIO](https://ruby-doc.org/stdlib-2.5.1/libdoc/stringio/rdoc/StringIO.html)? Not sure if there is a (performance) benefit for one or the other.

lib/puma/json.rb
```diff
@@ -0,0 +1,85 @@
...
+  #
+  # The implementation of JSON serialization in this module is not designed to
+  # be particularly full-featured or fast. It just has to handle the few places
+  # where Puma relies on JSON serialization internally.
+
+  module JSON
+    class SerializationError < StandardError; end
+
+    class << self
+      def generate(value)
```
ekohl, line 23: Does it make sense to attempt using the native `JSON` module for those who don't use phased restarts? It'd be faster.

lib/puma/json.rb
```diff
@@ -0,0 +1,85 @@
...
+        output << value.gsub(/[\\"]/, '\\' => '\\\\', '"' => '\\"')
+        output << '"'
+      end
+
+      def serialize_object_key(output, value)
+        case value
+        when Symbol
+          serialize_string output, value.to_s
+        when String
+          serialize_string output, value
```
MSP-Greg, line 78: In the past I used inspect for strings and symbols:
```ruby
when Symbol, String
  output << value.to_s.inspect
```
or 
```ruby
when Symbol
  output << value.to_s.inspect
when String
  output << value.inspect
```
I think it does all the needed processing?

lib/puma/json.rb
cjlarose: I think my implementation was based on an understanding of JSON according to RFC 4627, which requires that a valid JSON text is either an object or any array, but it looks like that requirement was dropped in RFC 7158. I'll update the implementation.

lib/puma/json.rb
cjlarose: I considered using `stringio` since it seems like a nice way to construct a string incrementally without an explosion in extra space. I went with just a normal `Array` because even `stringio` isn't in Ruby Core and, just like `json`, users can technically change the version of `stringio` they're using in a phased restart. That said `puma` _already_ uses `stringio` internally, so I could go either way on this.

I can't imagine the performance improvement would be too significant, especially since again, this module isn't really on a hot path.

nateberkopec:
Thank you so much for the reviews @MSP-Greg @ekohl and @rmacklin 🙇

lib/puma/json.rb
cjlarose: I'll dig deeper into this. I realized I'm not handling all of the different escape sequences correctly. Newline characters, for example, are written literally instead of `\n`.

I'll have to double check that everything that `.inspect` emits is valid JSON. If it is, I'll use it, otherwise, I'll handle the special characters in a different way.

lib/puma/json.rb
cjlarose: The reason I'd switch to `stringio` is if it would improve readability/maintainability. I'll play with it and see what it looks like.

lib/puma/json.rb
MSP-Greg: Thanks.  Your comment said
```
# The implementation of JSON serialization in this module is not designed to
# be particularly full-featured or fast. It just has to handle the few places
# where Puma relies on JSON serialization internally.
```
I suspect you know better than any of us what Puma sends to `JSON.generate` (having just reviewed it), so I'll trust your judgement as to what edge cases it should handle.

lib/puma/json.rb
cjlarose: Fixed in 8d05a2fc832e0a1b8e142fc0d98b3075dc1f11c0

lib/puma/json.rb
cjlarose: Switch to StringIO in 513aab1ea343bbc3c9e2adcc3acb8645aff64c74.

I ended up doing this because I think using the `Array` approach I was doing before stood out as unusual after thinking about it. If someone was trying to build a String incrementally while keep allocations to a minimum, a reader of the code would expect the author to have used `stringio`. Deviating from that expectation would suggest that the distinction was significant, but in this case, it isn't.

lib/puma/json.rb
cjlarose: Added 98b3408b4de75652aca8cfc66b08a6deee6da794 to handle control characters correctly and added 28e9d8ffe5ce7e51bced113e8c266182e58f01aa to simplify the handling of `String` and `Symbol`.

AFAICT, there's no guarantee that `.inspect` produces valid JSON (or will always produce valid JSON into the future), and adding support for control characters explicitly ended up not being so bad. I think the cost of adding the cases for the different characters is small, and it makes the implementation more closely match the language of the specification. I prefer that to having to worry about some edge case down the road where `.inspect` produces something that JSON parsers don't like.

test/test_json.rb
```diff
@@ -0,0 +1,92 @@
+require_relative "helper"
+require "puma/json"
+
+class TestJSON < Minitest::Test
```
ekohl, line 4: For all of these, does it make sense to assert `JSON.parse(Puma::JSON.generate(value)) == value`?

cjlarose: Ensuring that `Puma::JSON.generate` returns strings that can by parsed by `::JSON.parse` seems useful, for sure.

Updated in b905c34c264e847abed3b36d0711c95265deea2e

test/test_json.rb
```diff
@@ -89,4 +90,17 @@ def test_generate_raises_error_for_unexpected_value_type
     end
     assert_equal 'Unexpected value of type Regexp', ex.message
   end
+
+  private
+
+  def assert_puma_json_generates_string(expected_output, value_to_serialize, expected_roundtrip: value_to_serialize)
```
ekohl, line 96: I've never seen this before. Does `value_to_serialize` refer to the value that was passed in as the second argument?

cjlarose: It seems to work that way, but I couldn't find that behavior being explicitly supported in any official Ruby documentation, so I made it a little more clear in 0bb2df3

cjlarose:
I think I've addressed all of the feedback so far. Lemme know if there's anything else we should do!

ekohl (approved):
It's a complex PR so I may have missed something, but overall I think it's a good approach.

Issue #1906, opened 2019-08-12
4.0.1->4.1.0, Rails logs flushing incorrectly


pauloddr:
### Steps to reproduce

I can't seem to reproduce the issue locally, so I'll try my best to explain what happened.

1) Deployed an existing Rails app on Heroku after upgrading Puma from 4.0.1 to 4.1.0.

2) Made some requests. App works fine. Checked the logs.

3) Request logs were absent. (Heroku Router logs were present.)

4) Rebooted the application and all previous request logs were dumped at once with the wrong times (app shutdown time).

### Expected behavior

Request logs should appear right after they happen.

### Actual behavior

Request logs are dumped after the application exits instead, preventing real-time logging and breaking log times.

Reverting to Puma 4.0.1 restores intended behavior.

### System configuration

**Ruby version**: 2.6.1
**Rails version**: 6.0.0.rc2
**Puma version**: 4.1.0

### Log Dump

```
Aug 12 01:38:13 app app/web.1:  => Booting Puma 
Aug 12 01:38:13 app app/web.1:  => Rails 6.0.0.rc2 application starting in production  
Aug 12 01:38:13 app app/web.1:  => Run `rails server --help` for more startup options 
Aug 12 01:38:15 app app/web.1:  ** [NewRelic][2019-08-12 08:38:12 +0000 web.1 (4)] INFO : Starting the New Relic agent version 6.5.0.357 in "production" environment. 

... more NewRelic logs ...

Aug 12 01:38:15 app app/web.1:  Puma starting in single mode... 
Aug 12 01:38:15 app app/web.1:  * Version 4.1.0 (ruby 2.6.1-p33), codename: Fourth and One 
Aug 12 01:38:15 app app/web.1:  * Min threads: 2, max threads: 2 
Aug 12 01:38:15 app app/web.1:  * Environment: production 
Aug 12 01:38:15 app app/web.1:  * Listening on tcp://0.0.0.0:56371 
Aug 12 01:38:15 app app/web.1:  Use Ctrl-C to stop 
Aug 12 01:38:15 app heroku/web.1:  State changed from starting to up 

<missing request logs>

Aug 12 01:38:43 app heroku/router:  at=info method=GET path="/songs" host=...

<missing request logs>

Aug 12 01:38:43 app heroku/router:  at=info method=GET path="/stats/current" host=...

<missing request logs>

Aug 12 01:38:43 app heroku/router:  at=info method=GET path="/session" host=...

<missing request logs>

[...]

PR #2486, opened 2020-11-15
Avoid mutating $stdout and $stderr

### LINKED ISSUE #1948: Don't mutate global STDOUT/STDERR
...but don't break everyone's logging at the same time. https://github.com/puma/puma/pull/1946

ccverak:
### Description

Closes #1948 

Uses `flush` after writing to stdout and stderr so we don't need to use `sync=true`

### DIFF
--- a/lib/puma/events.rb
+++ b/lib/puma/events.rb
@@ -30,9 +30,6 @@ module Puma
       @stdout = stdout
       @stderr = stderr
 
-      @stdout.sync = true
-      @stderr.sync = true
-
       @debug = ENV.key? 'PUMA_DEBUG'
       @error_logger = ErrorLogger.new(@stderr)
 
@@ -66,6 +63,8 @@ module Puma
     #
     def log(str)
       @stdout.puts format(str) if @stdout.respond_to? :puts
+
+      @stdout.flush unless @stdout.sync
     rescue Errno::EPIPE
     end
 
--- a/lib/puma/runner.rb
+++ b/lib/puma/runner.rb
@@ -106,8 +106,8 @@ module Puma
         end
 
         STDOUT.reopen stdout, (append ? "a" : "w")
-        STDOUT.sync = true
         STDOUT.puts "=== puma startup: #{Time.now} ==="
+        STDOUT.flush unless STDOUT.sync
       end
 
       if stderr
@@ -116,8 +116,8 @@ module Puma
         end
 
         STDERR.reopen stderr, (append ? "a" : "w")
-        STDERR.sync = true
         STDERR.puts "=== puma startup: #{Time.now} ==="
+        STDERR.flush unless STDERR.sync
       end
     end
 
--- a/History.md
+++ b/History.md
@@ -2,6 +2,7 @@
 
 * Features
   * Your feature goes here <Most recent on the top, like GitHub> (#Github Number)
+  * Uses `flush` after writing messages to avoid mutating $stdout and $stderr using `sync=true` ([#2486])
   * Prints the loaded configuration if the environment variable `PUMA_LOG_CONFIG` is present ([#2472])
   * Integrate with systemd's watchdog and notification features ([#2438])
   * Adds max_fast_inline as a configuration option for the Server object ([#2406])
--- a/lib/puma/control_cli.rb
+++ b/lib/puma/control_cli.rb
@@ -237,6 +237,7 @@ module Puma
 
         if sig.nil?
           @stdout.puts "'#{@command}' not available via pid only"
+          @stdout.flush unless @stdout.sync
           return
         elsif sig.start_with? 'SIG'
           Process.kill sig, @pid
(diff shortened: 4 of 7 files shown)


lib/puma/events.rb
```diff
@@ -30,9 +30,6 @@ def initialize(stdout, stderr)
       @stdout = stdout
       @stderr = stderr
 
-      @stdout.sync = true
-      @stderr.sync = true
```
dentarg, line 34: I think this still needs to happen otherwise we will have a change in behaviour? These aren't the global STDOUT/STDERR we want to avoid mutating

ccverak: I'm trying to simulate the sync behavior using `flush` what do you think? See https://github.com/puma/puma/blob/89fc1e3ed6f950e65baf4463b41f94871bb3ce5b/lib/puma/events.rb#L67

lib/puma/runner.rb
```diff
@@ -116,8 +116,8 @@ def redirect_io
         end
 
         STDERR.reopen stderr, (append ? "a" : "w")
-        STDERR.sync = true
         STDERR.puts "=== puma startup: #{Time.now} ==="
+        STDERR.flush unless STDERR.sync
```
dentarg, line 120: Likewise here, we still want the buffer flushed, but we don't want to be modifying STDOUT/STDERR

lib/puma/events.rb
ccverak: Will take a look later to do it only where needed, but if `flush` does the trick I would prefer it over using `sync=true`

lib/puma/runner.rb
ccverak: Sorry, I don't get it, I think I'm missing something I don't sync that `STDERR.sync` does nothing it is just a getter :man_shrugging:

dentarg: Yes, `flush` probably does what we want, but `#sync` is false by default, hence the change in behaviour if it is never set to true

```
$ ruby -e 'p STDOUT.sync'
false
```

dentarg: Oops, I read the code as `STDERR.flush if STDERR.sync`, sorry! I understand my comments are confusing now.

lib/puma/events.rb
dentarg: Okay, I understand your approach, hehe, but it is confusing. From https://ruby-doc.org/core-2.7.1/IO.html#sync-method

>When sync mode is true, all output is immediately flushed to the underlying operating system and is not buffered by Ruby internally.

I think we need to respect this logic :)

dentarg: We want to set sync to true, but we don't want to do it on `STDOUT`/`STDERR` because people might use these constants to initialise their loggers, where they want a different behaviour.

ccverak: Oh! I was assuming that `flush` and `sync=true` had the same logic and that they both work only at Ruby level. What about `fsync` or `fdatasync` then?  

https://ruby-doc.org/core-2.7.2/IO.html#method-i-fdatasync
https://ruby-doc.org/core-2.7.2/IO.html#method-i-fsync 

Both seem to also work at the operating system level

Do you think that exploring this option worths the effort?

dentarg: Not sure we are dealing with files that needs to be written to disk here?

ccverak: I'll research a bit further but at this point, I think that flushing after every message is just the same as using `sync=true`. If I confirm it would that be enough to move forward?

nateberkopec: I'm ok with that

dentarg: Hmm, I was confused yesterday, sorry, I get your approach. Should hopefully work.

nateberkopec:
May want to rebase against master as Greg just fixed an issue where we were using puts instead of stdout

ccverak:
I have tried with different apps locally and on Heroku and everything works fine. Maybe someone has a clear case that should be tested with this change? Ideas?

ccverak:
Also, I have mixed feelings now because I introduced some duplication, do you think it's a good idea to encapsulate the new writing logic for the code outside `events` and `error_logger`?

nateberkopec:
Re: correctness, anything from #1906 could be extracted as a test

nateberkopec:
You can leave the refactor/DRYness for another PR if you want, don't let it block this feature.

nateberkopec:
Ideas for test re: comments in that thread:

> output is no longer being flushed before Process.daemon calls fork

ccverak:
I've been checking the current integration tests and it seems that https://github.com/puma/puma/blob/master/test/test_integration_single.rb#L126 already covers #1906 I'm I missing something? I thought that maybe the same tests but with workers could be helpful to cover the issue we had with Process.daemon forks, HDYT?

nateberkopec:
Took another look, I think this is good as written

PR #2494, opened 2020-11-25
Fail build if compiling extensions raises warnings (#1953)

### LINKED ISSUE #1953: Fail build if compiling extensions raises warnings

rubenochiavone:
### Description
Added `append_cflags config_string 'WERRORFLAG'` into extconf to instruct compiler to consider warnings as errors. Also fixed all warnings reported by GCC with default configuration.

Closes #1953.

### DIFF
--- a/.github/workflows/mri.yml
+++ b/.github/workflows/mri.yml
@@ -60,6 +60,10 @@ jobs:
         shell: bash
         run: echo 'DISABLE_SSL=true' >> $GITHUB_ENV
 
+      - name: set WERRORFLAG
+        shell: bash
+        run: echo 'MAKE_WARNINGS_INTO_ERRORS=true' >> $GITHUB_ENV
+
       - name: compile
         run:  bundle exec rake compile
 
--- a/.github/workflows/non_mri.yml
+++ b/.github/workflows/non_mri.yml
@@ -54,6 +54,10 @@ jobs:
         shell: bash
         run: echo 'DISABLE_SSL=true' >> $GITHUB_ENV
 
+      - name: set WERRORFLAG
+        shell: bash
+        run: echo 'MAKE_WARNINGS_INTO_ERRORS=true' >> $GITHUB_ENV
+
       - name: compile
         run:  bundle exec rake compile
 
--- a/History.md
+++ b/History.md
@@ -2,9 +2,11 @@
 
 * Features
   * Your feature goes here <Most recent on the top, like GitHub> (#Github Number)
+  * Fail build if compiling extensions raises warnings ([#1953])
 
 * Bugfixes
   * Your bugfix goes here <Most recent on the top, like GitHub> (#Github Number)
+  * Fix all compiler warnings reported in mini_ssl source code ([#1953])
 
 ## 5.1.0 / 2020-11-30
 
--- a/History.md
+++ b/History.md
@@ -2,11 +2,12 @@
 
 * Features
   * Your feature goes here <Most recent on the top, like GitHub> (#Github Number)
-  * Fail build if compiling extensions raises warnings ([#1953])
+  * Fail build if compiling extensions raises warnings on GH Actions ([#1953])
+  * Add MAKE_WARNINGS_INTO_ERRORS environment variable to toggle whether a build should treat all warnings into errors or not ([#1953])
 
 * Bugfixes
   * Your bugfix goes here <Most recent on the top, like GitHub> (#Github Number)
-  * Fix all compiler warnings reported in mini_ssl source code ([#1953])
+  * Fix compiler warnings, but skipped warnings related to ragel state machine generated code ([#1953])
 
 ## 5.1.0 / 2020-11-30
 
--- a/ext/puma_http11/extconf.rb
+++ b/ext/puma_http11/extconf.rb
@@ -27,10 +27,12 @@ end
 
 if ENV["MAKE_WARNINGS_INTO_ERRORS"]
   # Make all warnings into errors
+  # Except `implicit-fallthrough` since most failures comes from ragel state machine generated code
   if respond_to? :append_cflags
     append_cflags config_string 'WERRORFLAG'
+    append_cflags '-Wno-implicit-fallthrough'
   else
-    $CFLAGS += ' ' << (config_string 'WERRORFLAG')
+    $CFLAGS += ' ' << (config_string 'WERRORFLAG') << ' -Wno-implicit-fallthrough'
   end
 end
 
--- a/ext/puma_http11/extconf.rb
+++ b/ext/puma_http11/extconf.rb
@@ -3,7 +3,11 @@ require 'mkmf'
 dir_config("puma_http11")
 
 # Make all warnings into errors
-append_cflags config_string 'WERRORFLAG'
+if respond_to? :append_cflags
+  append_cflags config_string 'WERRORFLAG'
+else
+  $CFLAGS += ' ' << (config_string 'WERRORFLAG')
+end
 
 if $mingw && RUBY_VERSION >= '2.4'
   append_cflags  '-fstack-protector-strong -D_FORTIFY_SOURCE=2'
--- a/ext/puma_http11/extconf.rb
+++ b/ext/puma_http11/extconf.rb
@@ -2,13 +2,6 @@ require 'mkmf'
 
 dir_config("puma_http11")
 
-# Make all warnings into errors
-if respond_to? :append_cflags
-  append_cflags config_string 'WERRORFLAG'
-else
-  $CFLAGS += ' ' << (config_string 'WERRORFLAG')
-end
-
 if $mingw && RUBY_VERSION >= '2.4'
   append_cflags  '-fstack-protector-strong -D_FORTIFY_SOURCE=2'
   append_ldflags '-fstack-protector-strong -l:libssp.a'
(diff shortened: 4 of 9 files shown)


MSP-Greg:
> Also fixed all warnings reported by GCC with default configuration.

The problem is what is 'default configuration'...

Other people have inquired about similar issues.  Given the various OS's that testing is done on, writing code that will not generate warnings is difficult.

Do we want to write code such that warnings don't occur with the most recent build tools?  Not sure about that.  Also, not sure if it pertains to Puma, but there are 32 bit platforms, which may make it more difficult.

At a minimum, any PR's submitted need to pass the compile stage...

rubenochiavone:
> The problem is what is 'default configuration'...

Indeed. This was my machine configuration...

After you comment I was thinking that instead of always fail build on warnings a flag could be added to toggle that behavior. Perhaps an environment variable that is checked before adding `'WERRORFLAG'` to cflags on `ext/puma_http11/extconf.rb`. And then raise this flag during checks to make sure that at least on those machines no warnings are reported.

> Other people have inquired about similar issues. Given the various OS's that testing is done on, writing code that will not generate warnings is difficult.
> 
> Do we want to write code such that warnings don't occur with the most recent build tools? Not sure about that. Also, not sure if it pertains to Puma, but there are 32 bit platforms, which may make it more difficult.
> 
> At a minimum, any PR's submitted need to pass the compile stage on all jobs...

Some of the reported failures don't seem so difficult to fix. Do you think it is worth trying to fix them?

MSP-Greg:
> Do you think it is worth trying to fix them?

I suppose we could fix for Ubuntu 20.04?  I think it's using gcc 9.3.0, Windows is using 10.2.0.

A few of the warnings are due to a function/macro change between OpenSSL 1.1.0 & 1.1.1.  I've got a fix for that (extconf.rb).  I think we can assume almost all users are using either 1.1.1 or 1.0.2...

rubenochiavone:
> I suppose we could fix for Ubuntu 20.04?

For some reason that I don't know adding `-Werror` to cflags before looking for OpenSSL methods prevents mkmf from finding `DTLS_method` and `TLS_server_method` methods:

```
checking for whether -Werror is accepted as CFLAGS... yes
...
checking for DTLS_method() in openssl/ssl.h... no
checking for TLS_server_method() in openssl/ssl.h... no
checking for SSL_CTX_set_min_proto_version in openssl/ssl.h... yes
```

See https://github.com/puma/puma/pull/2494/checks?check_run_id=1451645306#step:6:1.

But If I add `-Werror` to cflags after mkmf verifies OpenSSL headers everything worked as expected:

```
...
checking for DTLS_method() in openssl/ssl.h... yes
checking for TLS_server_method() in openssl/ssl.h... yes
checking for SSL_CTX_set_min_proto_version in openssl/ssl.h... yes
checking for whether -Werror is accepted as CFLAGS... yes
```

See https://github.com/rubenochiavone/puma/runs/1470714817?check_suite_focus=true#step:6:1.

This error only happens on Ubuntu 20.04 with Ruby MRI 2.5 (2.5.8) builds. Since the goal is make it work for Ubuntu 20.04 I'm going to move `-Werror` addition right after looking for OpenSSL methods.

Any thoughts on that?

MSP-Greg:
I think it's difficult to optimize the warnings, given all the Ruby versions, OS versions. etc.  I added a change to extconf.rb that fixed one of the OpenSSL issues (PR https://github.com/puma/puma/pull/2499), see:

https://github.com/puma/puma/actions/runs/388957884

With flags differing (as they're pulled from the Ruby versions), and the fact that there's several gcc versions, trying to minimize the warning could result in a lot of spaghetti code that would be messy to maintain.

Also, there's platforms we can't easily run in CI that may use other compiler versions.

Simply put, I'm not sure if the benefits are worth the costs... What do you think, after looking into it more?

nateberkopec:
> Also, there's platforms we can't easily run in CI that may use other compiler versions.

I'm OK with this. The "promise" we're making here is that Puma runs on CI without compiler warnings, not that it will run _on your machine_ without compiler warnings.

> With flags differing (as they're pulled from the Ruby versions), and the fact that there's several gcc versions, trying to minimize the warning could result in a lot of spaghetti code that would be messy to maintain.

This is a good point, but is this concern hypothetical?

nateberkopec:
In short:

@rubenochiavone did the hard work here of making the changes required to make Puma CI without compiler warnings. Is the resulting code in this pull request _better_ or _worse_ than it was before?

MSP-Greg:
@nateberkopec

Sorry, I phrased that poorly.

> Is the resulting code in this pull request better or worse than it was before?

Better.

MSP-Greg (approved):

nateberkopec:
@rubenochiavone If you want to fix the macos failure and then rebase, we'll merge 👍

rubenochiavone:
Thanks @MSP-Greg and @nateberkopec for sharing your thoughts.

First I would like to clarify what I said

> For some reason that I don't know adding -Werror to cflags before looking for OpenSSL methods prevents mkmf from finding DTLS_method and TLS_server_method methods:

My best guess here is that mkmf passes flags set in extconf.rb to small programs built during configure to check methods. When looking for methods mkmf attempts to compile code using the target method and if the program results in failure the checking is false. On OpenSSL (between 1.0.2h-1.1.0) `__owur` directive appends a GCC attribute to the function that tells the compiler if the developer forgot to hold the returned value from a function a warning should be raised, then if we enable `WERRORFLAG` mkmf passes `WERRORFLAG` to this checking program and the build fails because possibly it ignores returned value and no warning is accepted. That's why I think that both `DTLS_method` and `TLS_server_method` were not found [here](https://github.com/puma/puma/pull/2494/checks?check_run_id=1451645306#step:6:1) and postponing cflags addition worked [here](https://github.com/rubenochiavone/puma/runs/1470714817?check_suite_focus=true#step:6:1).

Maybe this is a bug in MRI Ruby 2.5.8? Other versions don't have this behavior.

Second, I think that puma should be a friendly project for anyone and enforcing no warnings policy by default could be frustrating because people could be prevented from running puma because of a compiler warning. So, I would like to take a step back in that sense. Like I mentioned I think that a flag that toggles this behavior could be interesting.

Third, I also believe that removing all warnings when possible is not only a good practice but could result in better compiled code. For example, the NORETURN flag optimize compiled code:

> noreturn
> A few standard library functions, such as abort and exit, cannot return. GCC knows this automatically. Some programs define their own functions that never return. You can declare them noreturn to tell the compiler this fact. For example,
> 
> void fatal () __attribute__ ((noreturn));
>
> void
> fatal (/* … */)
> {
>   /* … */ /* Print error message. */ /* … */
>   exit (1);
> }
>
> The noreturn keyword tells the compiler to assume that fatal cannot return. It can then optimize without regard to what would happen if fatal ever did return. This makes slightly better code. More importantly, it helps avoid spurious warnings of uninitialized variables.
> 
> The noreturn keyword does not affect the exceptional path when that applies: a noreturn-marked function may still return to the caller by throwing an exception or calling longjmp. 
> ...
> 

https://gcc.gnu.org/onlinedocs/gcc-10.2.0/gcc/Common-Function-Attributes.html#Common-Function-Attributes.

Additionally, future compilers versions can turn some of those warnings into errors and keeping the code free of warnings can be an advantage. Even further, as compilers improve warnings reports it can even prevents bugs:

> Enhancements to existing warnings:
> 
>     -Warray-bounds detects more out-of-bounds accesses to member arrays as well as accesses to elements of zero-length arrays.
>     -Wformat-overflow makes full use of string length information computed by the strlen optimization pass.
>     -Wrestrict detects overlapping accesses to dynamically allocated objects.
>     -Wreturn-local-addr diagnoses more instances of return statements returning addresses of automatic variables.
>     -Wstringop-overflow detects more out-of-bounds stores to member arrays including zero-length arrays, dynamically allocated objects and variable length arrays, as well as more instances of reads of unterminated character arrays by string built-in functions. The warning also detects out-of-bounds accesses by calls to user-defined functions declared with the new attribute access.
>     -Warith-conversion re-enables warnings from -Wconversion, -Wfloat-conversion, and -Wsign-conversion that are now off by default for an expression where the result of an arithmetic operation will not fit in the target type due to promotion, but the operands of the expression do fit in the target type.

https://gcc.gnu.org/gcc-10/changes.html

Conversely, I also believe that fixing some of those warnings don't represent any code improvement instead it adds unnecessary maintenance burden, for example:

```diff
+++ b/ext/puma_http11/mini_ssl.c
@@ -88,6 +88,10 @@ DH *get_dh1024() {
   static unsigned char dh1024_g[] = { 0x02 };
 
   DH *dh;
+#if !(OPENSSL_VERSION_NUMBER < 0x10100005L || defined(LIBRESSL_VERSION_NUMBER))
+  BIGNUM *p, *g;
+#endif
+
   dh = DH_new();
 
 #if OPENSSL_VERSION_NUMBER < 0x10100005L || defined(LIBRESSL_VERSION_NUMBER)
@@ -99,7 +103,6 @@ DH *get_dh1024() {
     return NULL;
   }
 #else
-  BIGNUM *p, *g;
   p = BN_bin2bn(dh1024_p, sizeof(dh1024_p), NULL);
   g = BN_bin2bn(dh1024_g, sizeof(dh1024_g), NULL);
```

Finally, here is what I propose:

- Toggle `WERRORFLAG` on extconf.rb using an environment variable (could be `MAKE_WARNINGS_INTO_ERRORS`)
- Enable `WERRORFLAG` on all GH Actions
- Make sure PRs are free of warnings on all GH Actions

MSP-Greg:
@rubenochiavone

Nice.  Both the above comment and the code update.  I think we're on the 'same page', and setting it up so it just affects the CI here is a good implementation.

MSP-Greg:
@rubenochiavone

I hate to ask, but I'm running `macos-11.0 head` in my fork (working on updating the test framework), and it's failing compile with the following log.  Note that all the other 'head' builds compile.  I don't know what caused this, but it was compiling in earlier runs in my fork...

```
/usr/bin/make
compiling ../../../../ext/puma_http11/http11_parser.c
error: unknown warning option '-Wmisleading-indentation'; did you mean '-Wbinding-in-condition'? [-Werror,-Wunknown-warning-option]
make: *** [http11_parser.o] Error 1
rake aborted!
```

Note that CI here is running the following, and they all compiled:
```
macos-10.15 2.7
macos-10.15 head
macos-11.0 2.7
ubuntu 18.04 head
ubuntu 20.04 head
mingw head
```

rubenochiavone:
@MSP-Greg maybe this could help: https://github.com/libexpat/libexpat/issues/245#issuecomment-457842237.

PR #2515, opened 2020-12-15
[ci skip] K8s docs

### LINKED ISSUE #2343: Kubernetes downscale result in 5xx for the pods that have been killed
**Describe the bug**
We are running an API using Puma in a Kubernetes cluster that uses a Horizontal Pod Autoscaler (HPA). We noticed that whenever the HPA scales down, the downstream services consuming our API get a 5xx error response, one for each of the pods that got killed in the process. 

We've added puma-metrics and see that the running number of threads is well below the maximum configured, whenever the errors occur. Requests are farily quick (<200ms), so it's not that the graceful shutdown timeout is exceeded. 

It looks like requests are thrown away/not fully processed after a graceful shutdown triggered by Kubernetes.

We don't know where to dig further, so I was curious to learn more about this, and maybe about the Puma internals, so we can spot the culprit.

**Puma config:**

It reproduces with the default configuration.

**To Reproduce**

Mentioned above.

**Expected behavior**

Requests shouldn't get discarded and 5xx errors should occur. On graceful shutdown, Puma should stop accepting new requests and wait until the ongoing ones finish.

 - OS: Alpine 3.11 in Docker, ran in Kubernetes
 - Puma Version: 4.3.5

WJWH:
### Description
Adds documentation for running in Kubernetes, including the rolling deploy/graceful shutdown interaction and its workaround mentioned in #2343. Closes #2343 @nateberkopec

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,6 +283,7 @@ Some platforms do not support all Puma features.
 
   * **JRuby**, **Windows**: server sockets are not seamless on restart, they must be closed and reopened. These platforms have no way to pass descriptors into a new process that is exposed to Ruby. Also, cluster mode is not supported due to a lack of fork(2).
   * **Windows**: Cluster mode is not supported due to a lack of fork(2).
+  * **Kubernetes**: The way Kubernetes handles pod shutdowns interacts poorly with server processes implementing graceful shutdown, like Puma. See the [kubernetes section of the documentation](docs/kubernetes.md) for more details.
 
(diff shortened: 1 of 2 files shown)


nateberkopec:
Thanks!!!

acmh:
Nice work! Know I know why I got 5XX range during a rolling deployment to a new version. Thanks for this contribution!

PR #2522, opened 2021-01-05
Allow specifying the desired IO selector backend for nio4r

jcmfernandes:
### Description

nio4r will soon have experimental support for io_uring via libev. The brave might want to try it out :) regardless of that, allowing users to explicitly set the IO selector backend seems like a good idea to me.

### DIFF
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -866,9 +866,18 @@ module Puma
       @options[:max_fast_inline] = Float(num_of_requests)
     end
 
-    # Change the backend for the IO selector.
+    # Specify the backend for the IO selector.
     #
-    # The default is +:auto+ which will let nio4r choose the backend.
+    # Provided values will be passed directly to +NIO::Selector.new+, with the
+    # exception of +:auto+ which will let nio4r choose the backend.
+    #
+    # Check the documentation of +NIO::Selector.backends+ for the list of valid
+    # options. The available options on your system will depend on the operating
+    # system.
+    #
+    # The default is +:auto+.
+    #
+    # @see https://github.com/socketry/nio4r/blob/master/lib/nio/selector.rb
     #
     def io_selector_backend(backend)
       @options[:io_selector_backend] = backend.to_sym
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -872,8 +872,10 @@ module Puma
     # exception of +:auto+ which will let nio4r choose the backend.
     #
     # Check the documentation of +NIO::Selector.backends+ for the list of valid
-    # options. The available options on your system will depend on the operating
-    # system.
+    # options. Note that the available options on your system will depend on the
+    # operating system. If you want to use the pure Ruby backend (not
+    # recommended due to its comparatively low performance), set environment
+    # variable +NIO4R_PURE+ to +true+.
     #
     # The default is +:auto+.
     #
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -865,5 +865,13 @@ module Puma
     def max_fast_inline(num_of_requests)
       @options[:max_fast_inline] = Float(num_of_requests)
     end
+
+    # Change the backend for the IO selector.
+    #
+    # The default is +:auto+ which will let nio4r choose the backend.
+    #
+    def io_selector_backend(backend)
+      @options[:io_selector_backend] = backend.to_sym
+    end
   end
 end
--- a/lib/puma/reactor.rb
+++ b/lib/puma/reactor.rb
@@ -18,9 +18,9 @@ module Puma
     # Create a new Reactor to monitor IO objects added by #add.
     # The provided block will be invoked when an IO has data available to read,
     # its timeout elapses, or when the Reactor shuts down.
-    def initialize(&block)
+    def initialize(backend, &block)
       require 'nio'
-      @selector = NIO::Selector.new
+      @selector = backend == :auto ? NIO::Selector.new : NIO::Selector.new(backend)
       @input = Queue.new
       @timeouts = []
       @block = block
--- a/lib/puma/reactor.rb
+++ b/lib/puma/reactor.rb
@@ -3,6 +3,8 @@
 require 'puma/queue_close' unless ::Queue.instance_methods.include? :close
 
 module Puma
+  class UnsupportedBackend < StandardError; end
+
   # Monitors a collection of IO objects, calling a block whenever
   # any monitored object either receives data or times out, or when the Reactor shuts down.
   #
@@ -20,6 +22,9 @@ module Puma
     # its timeout elapses, or when the Reactor shuts down.
     def initialize(backend, &block)
       require 'nio'
+      unless backend == :auto || NIO::Selector.backends.include?(backend)
+        raise "unsupported IO selector backend: #{backend} (available backends: #{NIO::Selector.backends.join(', ')})"
+      end
       @selector = backend == :auto ? NIO::Selector.new : NIO::Selector.new(backend)
       @input = Queue.new
       @timeouts = []
--- a/test/test_config.rb
+++ b/test/test_config.rb
@@ -274,6 +274,14 @@ class TestConfigFile < TestConfigFileBase
     assert_equal Float::INFINITY, conf.options[:max_fast_inline]
   end
 
+  def test_config_files_with_symbol_convert
+    conf = Puma::Configuration.new(config_files: ['test/config/with_symbol_convert.rb']) do
+    end
+    conf.load
+
+    assert_equal :ruby, conf.options[:io_selector_backend]
+  end
+
   def test_config_raise_exception_on_sigterm
     conf = Puma::Configuration.new do |c|
       c.raise_exception_on_sigterm false
--- a/History.md
+++ b/History.md
@@ -16,9 +16,9 @@
 
 ## 5.1.1 / 2020-12-10
 
-* Bugfixes
+* Bugfixes 
   * Fix over eager matching against banned header names ([#2510])
-
+  
 ## 5.1.0 / 2020-11-30
 
 * Features
--- a/History.md
+++ b/History.md
@@ -2,6 +2,7 @@
 
 * Features
   * Your feature goes here <Most recent on the top, like GitHub> (#Github Number)
+  * Add option to specify the desired IO selector backend for libev ([#pending])
   * Add ability to set OpenSSL verification flags (MRI only) ([#2490])
   * Uses `flush` after writing messages to avoid mutating $stdout and $stderr using `sync=true` ([#2486])
   * Fail build if compiling extensions raises warnings on GH Actions ([#1953])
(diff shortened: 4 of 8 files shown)


lib/puma/dsl.rb
```diff
@@ -865,5 +865,13 @@ def nakayoshi_fork(enabled=true)
     def max_fast_inline(num_of_requests)
       @options[:max_fast_inline] = Float(num_of_requests)
     end
+
+    # Change the backend for the IO selector.
+    #
+    # The default is +:auto+ which will let nio4r choose the backend.
+    #
```
nateberkopec, line 872: Let's mention here that the option is passed directly to nio4r, and that valid options are here: https://github.com/socketry/nio4r/blob/4111b1cee87ce8857558900b8df76a5bd6c3e985/lib/nio/selector.rb#L27

lib/puma/reactor.rb
```diff
@@ -18,9 +18,9 @@ class Reactor
     # Create a new Reactor to monitor IO objects added by #add.
     # The provided block will be invoked when an IO has data available to read,
     # its timeout elapses, or when the Reactor shuts down.
-    def initialize(&block)
+    def initialize(backend, &block)
       require 'nio'
-      @selector = NIO::Selector.new
+      @selector = backend == :auto ? NIO::Selector.new : NIO::Selector.new(backend)
```
nateberkopec, line 23: It appears the default option is `:ruby`, not `:auto`. https://github.com/socketry/nio4r/blob/4111b1cee87ce8857558900b8df76a5bd6c3e985/lib/nio/selector.rb#L16

nateberkopec:
Thanks! Left some comments, lmk what you think.

lib/puma/reactor.rb
cjlarose: Curious on @jcmfernandes 's thoughts here, too, but I kinda like having an explicit default of `:auto`, which means to use whatever `NIO`'s default backend is.

Otherwise, if NIO ever changes their default backend, we end up in a weird situation where puma's default is always `:ruby`, regardless of NIO's actual default.

MSP-Greg: At present, `epoll` is the default backend on Ubuntu.  Windows uses `ruby`.

lib/puma/dsl.rb
jcmfernandes: WIll do! :+1:

lib/puma/reactor.rb
jcmfernandes: @nateberkopec @cjlarose as stated in https://github.com/socketry/nio4r/pull/258 passing `nil` as a way to say "please choose the best" only works when using libev (well, my PR changes that but it's not even part of a release at this point :smile_cat:).

Hence, either we pass any value set in the puma configuration as is to `NIO::Selector.new` (except for `nil`, which we don't want to pass given I wrote above), or we introduce a value that serves as the "choose the best please". It can be `nil`, but I thought about having something more explicit, hence the `:auto`. Happy to change it though!

cjlarose: `:auto` seems fine by me.

Another option to consider is instead of making it a symbol, we just let the user specify a block that would return the instance the user actually wants:

```ruby
# config.rb
nio_selector { NIO::Selector.new(:ruby) }
# or
nio_selector { NIO::Selector.new }
```

This gives the user the ultimately flexibility, without imposing our own defaults or conventions. Using blocks in the config file is already well-established by `on_worker_boot`, `on_worker_shutdown`, etc.

jcmfernandes:
> Thanks! Left some comments, lmk what you think.

Thank you! I updated the docs.

lib/puma/dsl.rb
```diff
@@ -865,5 +865,22 @@ def nakayoshi_fork(enabled=true)
...
+    # Provided values will be passed directly to +NIO::Selector.new+, with the
+    # exception of +:auto+ which will let nio4r choose the backend.
+    #
+    # Check the documentation of +NIO::Selector.backends+ for the list of valid
+    # options. The available options on your system will depend on the operating
+    # system.
+    #
+    # The default is +:auto+.
+    #
+    # @see https://github.com/socketry/nio4r/blob/master/lib/nio/selector.rb
```
dentarg, line 880: Should we also write something about the environment variable [`NIO4R_PURE`](https://github.com/socketry/nio4r/blob/v2.5.4/lib/nio.rb) here?

This was confusing to me before I learned about it:

```ruby
$ bundle e pry -rnio
[1] pry(main)> NIO::VERSION
=> "2.5.4"
[2] pry(main)> NIO::Selector.new(:ruby)
ArgumentError: unsupported backend: :ruby
from (pry):1:in `initialize'
```

```ruby
$ bundle e pry -rnio
[1] pry(main)> RUBY_PLATFORM
=> "x86_64-darwin18"
[2] pry(main)> NIO::VERSION
=> "2.5.4"
[3] pry(main)> [nil, :ruby, :java, :epoll, :poll, :kqueue, :select, :port, :unknown].map { |backend| NIO::Selector.new(backend).backend rescue nil }
=> [:kqueue, nil, nil, nil, :poll, :kqueue, :select, nil, nil]
```

```ruby
$ NIO4R_PURE=true bundle e pry -rnio
[1] pry(main)> RUBY_PLATFORM
=> "x86_64-darwin18"
[2] pry(main)> NIO::VERSION
=> "2.5.4"
[3] pry(main)> [nil, :ruby, :java, :epoll, :poll, :kqueue, :select, :port, :unknown].map { |backend| NIO::Selector.new(backend).backend rescue nil }
=> [nil, :ruby, nil, nil, nil, nil, nil, nil, nil]
```

dentarg (commented):
Should we rescue the `ArgumentError` coming from nio4r and raise a better one?

It is not that clear that it is coming from nio4r:

```
$ grep io_selector_backend test/config/app.rb
io_selector_backend :ruby

* Puma version: 5.1.1 (ruby 2.7.1-p83) ("At Your Service")
*  Min threads: 0
*  Max threads: 5
*  Environment: development
*          PID: 15846
* Listening on http://0.0.0.0:57461
bundler: failed to load command: bin/puma (bin/puma)
  23: from /Users/dentarg/.gem/ruby/2.7.1/bin/bundle:23:in `<main>'
  22: from /Users/dentarg/.gem/ruby/2.7.1/bin/bundle:23:in `load'
  21: from /Users/dentarg/.rubies/ruby-2.7.1/lib/ruby/gems/2.7.0/gems/bundler-2.2.4/exe/bundle:37:in `<top (required)>'
  20: from /Users/dentarg/.rubies/ruby-2.7.1/lib/ruby/site_ruby/2.7.0/bundler/friendly_errors.rb:130:in `with_friendly_errors'
  19: from /Users/dentarg/.rubies/ruby-2.7.1/lib/ruby/gems/2.7.0/gems/bundler-2.2.4/exe/bundle:49:in `block in <top (required)>'
  18: from /Users/dentarg/.rubies/ruby-2.7.1/lib/ruby/site_ruby/2.7.0/bundler/cli.rb:24:in `start'
  17: from /Users/dentarg/.rubies/ruby-2.7.1/lib/ruby/site_ruby/2.7.0/bundler/vendor/thor/lib/thor/base.rb:485:in `start'
  16: from /Users/dentarg/.rubies/ruby-2.7.1/lib/ruby/site_ruby/2.7.0/bundler/cli.rb:30:in `dispatch'
  15: from /Users/dentarg/.rubies/ruby-2.7.1/lib/ruby/site_ruby/2.7.0/bundler/vendor/thor/lib/thor.rb:392:in `dispatch'
  14: from /Users/dentarg/.rubies/ruby-2.7.1/lib/ruby/site_ruby/2.7.0/bundler/vendor/thor/lib/thor/invocation.rb:127:in `invoke_command'
  13: from /Users/dentarg/.rubies/ruby-2.7.1/lib/ruby/site_ruby/2.7.0/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
  12: from /Users/dentarg/.rubies/ruby-2.7.1/lib/ruby/site_ruby/2.7.0/bundler/cli.rb:494:in `exec'
  11: from /Users/dentarg/.rubies/ruby-2.7.1/lib/ruby/site_ruby/2.7.0/bundler/cli/exec.rb:28:in `run'
  10: from /Users/dentarg/.rubies/ruby-2.7.1/lib/ruby/site_ruby/2.7.0/bundler/cli/exec.rb:63:in `kernel_load'
   9: from /Users/dentarg/.rubies/ruby-2.7.1/lib/ruby/site_ruby/2.7.0/bundler/cli/exec.rb:63:in `load'
   8: from bin/puma:10:in `<top (required)>'
   7: from /Users/dentarg/src/puma/lib/puma/cli.rb:80:in `run'
   6: from /Users/dentarg/src/puma/lib/puma/launcher.rb:181:in `run'
   5: from /Users/dentarg/src/puma/lib/puma/single.rb:53:in `run'
   4: from /Users/dentarg/src/puma/lib/puma/server.rb:241:in `run'
   3: from /Users/dentarg/src/puma/lib/puma/server.rb:241:in `new'
   2: from /Users/dentarg/src/puma/lib/puma/reactor.rb:23:in `initialize'
   1: from /Users/dentarg/src/puma/lib/puma/reactor.rb:23:in `new'
/Users/dentarg/src/puma/lib/puma/reactor.rb:23:in `initialize': unsupported backend: :ruby (ArgumentError)
```

test/test_config.rb
```diff
@@ -274,6 +274,14 @@ def test_config_files_with_float_convert
     assert_equal Float::INFINITY, conf.options[:max_fast_inline]
   end
 
+  def test_config_files_with_symbol_convert
+    conf = Puma::Configuration.new(config_files: ['test/config/with_symbol_convert.rb']) do
+    end
+    conf.load
+
+    assert_equal :ruby, conf.options[:io_selector_backend]
+  end
```
dentarg, line 283: Should we add a test that starts Puma with a different backend than the default (auto)?

lib/puma/dsl.rb
jcmfernandes: I agree. I might actually try to change that in nio4r. I'll document it.

test/test_config.rb
jcmfernandes: Sure, let me add a test for that, thanks for the suggestion! I, unfortunately, can't test for a specific backend since the available ones will vary from platform to platform. Unless I enforce `NIO4R_PURE=true` which I'm not very fond of.

jcmfernandes:
I hope I addressed all feedback, thank you! Let me know what you think.

lib/puma/reactor.rb
cjlarose: @nateberkopec Do you have a preference for using `:auto` to mean "pick the best IO selector" vs using the block for configuration? Or many a different idea?

nateberkopec: In my head, whatever I provide to `nio_selector` should be passed as an arg to Nio::Selector.new(). So, if you specify `nio_selector :auto`, I expect that to be passed to `NIO::Selector.new`. 

IMO I think the easiest to understand interface here is that the default is nil rather than this new `:auto` symbol, but due to https://github.com/socketry/nio4r/pull/258 I think the best backwards/forwards compatibile option is thew new :auto symbol.

So I'm OK with it.

nateberkopec:
Great work from everyone - @jcmfernandes and reviewers 👍

jcmfernandes:
Thank you @nateberkopec and all the reviewers! :raised_hands:

PR #2525, opened 2021-01-10
Docs for stats

jjb:
As far as I can tell there aren't any docs for a) accessing stats b) what individual stats mean

maybe we can flesh that out a bit in this ticket, and then i'll make a PR with a new document in /docs ?

i threw together this very rough doc - please give comments, or point me to existing docs/discussion where i can draw from

### DIFF
--- a/docs/stats.md
+++ b/docs/stats.md
@@ -25,6 +25,6 @@ stats can be accessed via
 * last_status: a hash of info about the worker's state handling requests
   * backlog: requests that are waiting for an available thread to be available. if this is above 0, you need more capacity [always true?]
   * running: how many threads are running
-  * pool_capacity: ??? (i've observed this number stay at 3 even when threads cool back down to 0, so i dont' know what it means)
+  * pool_capacity: the number of requests that the server is capable of taking right now. For example if the number is 5 then it means there are 5 threads sitting idle ready to take a request. If one request comes in, then the value would be 4 until it finishes processing.
   * max_threads: the maximum number of threads puma is configured to spool up per worker 
   * requests_count: the number of requests this worker has served since starting
--- a/docs/stats.md
+++ b/docs/stats.md
@@ -6,7 +6,7 @@ Stats can be accessed in two ways:
 
 `$ pumactl stats` or `GET /stats`
 
-Read more about `pumactl` and the control server [here](https://github.com/puma/puma#controlstatus-server).
+[Read more about `pumactl` and the control server in the README.](https://github.com/puma/puma#controlstatus-server).
 
 ### Puma.stats
 
--- a/docs/stats.md
+++ b/docs/stats.md
@@ -1,0 +1,30 @@
+## accessing stats
+
+stats can be accessed via 
+1. pumactr
+2. Puma.stats when in non-worker mode
+3. Puma.stats when in worker mode (should not be invoked from worker process, right?)
+
+
+## meaning of stats
+
+* started_at: when puma was started
+* phase: ??
+* workers: ??
+* booted_workers: how many workers currently running?
+* old_workers: ??
+* worker_status: array of hashes of info for each worker (see below)
+
+## meaning of worker stats
+
+* started_at: when the worker was started
+* pid: the process id of the worker process
+* index: each worker gets a number. if puma is configured to have 3 workers, then this will be 0, 1, or 2
+* booted: if it's done booting [?]
+* last_checkin: ??
+* last_status: a hash of info about the worker's state handling requests
+  * backlog: requests that are waiting for an available thread to be available. if this is above 0, you need more capacity [always true?]
+  * running: how many threads are running
+  * pool_capacity: ??? (i've observed this number stay at 3 even when threads cool back down to 0, so i dont' know what it means)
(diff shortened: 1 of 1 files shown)


docs/stats.md
```diff
@@ -0,0 +1,30 @@
+## accessing stats
+
+stats can be accessed via 
+1. pumactr
```
dentarg, line 4:
```suggestion
1. `pumactl`
```

docs/stats.md
```diff
@@ -0,0 +1,30 @@
...
+
+* started_at: when the worker was started
+* pid: the process id of the worker process
+* index: each worker gets a number. if puma is configured to have 3 workers, then this will be 0, 1, or 2
+* booted: if it's done booting [?]
+* last_checkin: ??
+* last_status: a hash of info about the worker's state handling requests
+  * backlog: requests that are waiting for an available thread to be available. if this is above 0, you need more capacity [always true?]
+  * running: how many threads are running
+  * pool_capacity: ??? (i've observed this number stay at 3 even when threads cool back down to 0, so i dont' know what it means)
```
dentarg, line 28: https://github.com/puma/puma/blob/ee3341d6708d8cc729417cb9d749ec4af0d5b6f0/lib/puma/server.rb#L204-L214

docs/stats.md
```diff
@@ -0,0 +1,30 @@
+## accessing stats
+
+stats can be accessed via 
+1. `pumactl`
```
cjlarose, line 4: I think these are the ways you can access stats:

```
1. `pumactl stats`, assuming puma was started with the use of the control/status server
2. `GET /stats` request to the control/status server directly
3. `Puma.stats` from within the Ruby process
```

docs/stats.md
```diff
@@ -0,0 +1,30 @@
...
+2. Puma.stats when in non-worker mode
+3. Puma.stats when in worker mode (should not be invoked from worker process, right?)
```
cjlarose, line 6: I think officially these modes are called "single mode" and "cluster mode". The ways you can access stats are the same, they just have different formats depending on the mode.

docs/stats.md
```diff
@@ -0,0 +1,30 @@
...
+
+
+## meaning of stats
+
+* started_at: when puma was started
+* phase: ??
```
cjlarose, line 12: Following how phased restarts work in `cluster.rb` is probably the best resource for this. Off the top of my head, I think the cluster itself has a `phase`, and then each individual worker has a phase as well. Phase starts at 0 for both, but when you issue a phased restart (described in the [restart docs](https://github.com/puma/puma/blob/e2a15675a37bd25184fa7ad8343d23577464a5a0/docs/restart.md)), the phase of the cluster is incremented. Then, workers started in the old phase die off while new workers are spun up, recording their phase as the cluster's current phase. Monitoring the phases of the workers can be used, for example, to see if a phased restart operation was successful or not.

cjlarose (commented):
This is a great idea! Left some comments for the stuff I know about off the top of my head

docs/stats.md
nateberkopec: yeah I'd prefer to see `$ pumactl stats` as the way we refer to that, just to make it clear its a CLI command

docs/stats.md
nateberkopec: yup let's keep the terminology consistent w/single+cluster.

docs/stats.md
```diff
@@ -0,0 +1,30 @@
...
+* old_workers: ??
+* worker_status: array of hashes of info for each worker (see below)
+
+## meaning of worker stats
+
+* started_at: when the worker was started
+* pid: the process id of the worker process
+* index: each worker gets a number. if puma is configured to have 3 workers, then this will be 0, 1, or 2
+* booted: if it's done booting [?]
+* last_checkin: ??
```
nateberkopec, line 24: Last time the worker responded to the master process' heartbeat check.

docs/stats.md
jjb: Looking at the code I wasn't able to confidently infer what the different phases were and how to describe them. Maybe someone who knows could flesh that out in https://github.com/puma/puma/blob/master/docs/restart.md and then we can link to it in the stats doc

jjb:
Thanks for the feedback - I think I addressed everything, see latest version: https://github.com/jjb/puma/blob/patch-2/docs/stats.md

docs/stats.md
```diff
@@ -0,0 +1,142 @@
+## accessing stats
+
+Stats can be accessed in two ways:
+
+### control server
+
+`$ pumactl stats` or `GET /stats`
+
+Read more about `pumactl` and the control server [here](https://github.com/puma/puma#controlstatus-server).
```
dentarg, line 9: Descriptive link text [for accessibility reasons](https://accessibility.oregonstate.edu/descriptivelinks):

```suggestion
[Read more about `pumactl` and the control server in the README.](https://github.com/puma/puma#controlstatus-server).
```

nateberkopec:
Great start, we can add more in with other PRs.

jjb:
Awesome! I was going to squish those ugly commits together 😅 😭

PR #2530, opened 2021-01-18
replace `while true` with `loop do`

jjb:
Nicer code, and I'm guessing it's a tiny bit more performant

### DIFF
--- a/examples/plugins/redis_stop_puma.rb
+++ b/examples/plugins/redis_stop_puma.rb
@@ -32,7 +32,7 @@ Puma::Plugin.create do
     return unless redis.ping == 'PONG'
 
     in_background do
-      while true
+      loop do
         sleep 2
         if message = redis.get("puma::restart::#{hostname}")
           redis.del("puma::restart::#{hostname}")
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -356,7 +356,7 @@ module Puma
     end
 
     def read_chunked_body
-      while true
+      loop do
         begin
           chunk = @io.read_nonblock(4096)
         rescue IO::WaitReadable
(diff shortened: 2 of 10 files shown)


nateberkopec:
Is there a RuboCop rule we can add for this?

nateberkopec:
benchmarks/wrk/ssl_realistic_response.sh shows no real difference between this and master but I'm OK with changing it anyway as I do like the look of it better + it may matter in the future so I'm ok with merging with a Rubocop rule

jjb:
🚨 loop is actually slower

i thought it might be because maybe loop is implemented in ruby and not C, but [that's not the case](https://github.com/ruby/ruby/blob/master/vm_eval.c#L1328) 🤷 . 

```ruby
require 'benchmark'

TIMES = 100_000_000

def with_while
  i = 0
  while true
    i+=1
    return if i==TIMES
  end
end

def with_while_with_condition
  i = 0
  while 1==1
    i+=1
    return if i==TIMES
  end
end


def with_loop
  i = 0
  loop do
    i+=1
    return if i==TIMES
  end
end

puts "warming up\n"
with_while
with_while_with_condition
with_loop

Benchmark.bm(10) do |x|
  x.report("while") do
    with_while
  end
  x.report("while_with_condition") do
    with_while_with_condition
  end
  x.report("loop") do
    with_loop
  end
end
```

```
while        1.731089   0.012813   1.743902 (  1.889690)
while_with_condition  2.686485   0.010185   2.696670 (  2.752244)
loop         5.964884   0.045229   6.010113 (  6.293856)
```

dentarg:
Seems like the use of `while true` has been intentional, example: https://github.com/puma/puma/commit/cf44fa7e3267d6b90a6187b1477fcdd3235517d5

jjb:
@dentarg good find! 

spock saying 'fascinating'

PR #2533, opened 2021-01-22
Binder - add/fix ssl 0.0.0.0/localhost logging, use ternary operator

MSP-Greg:
### Description

Binder - add/fix ssl 0.0.0.0/localhost logging, use tertiary

test_binder.rb - remove skips for:
  test_correct_zero_port_ssl
  test_logs_all_localhost_bindings_ssl

### DIFF
--- a/lib/puma/binder.rb
+++ b/lib/puma/binder.rb
@@ -160,6 +160,7 @@ module Puma
             io = inherit_tcp_listener uri.host, uri.port, sock
             logger.log "* Activated #{str}"
           else
+            ios_len = @ios.length
             params = Util.parse_query uri.query
 
             opt = params.key?('low_latency')
@@ -167,14 +168,8 @@ module Puma
 
             io = add_tcp_listener uri.host, uri.port, opt, bak
 
-            @ios.each do |i|
-              next unless TCPServer === i
-              addr = if i.local_address.ipv6?
-                "[#{i.local_address.ip_unpack[0]}]:#{i.local_address.ip_unpack[1]}"
-              else
-                i.local_address.ip_unpack.join(':')
-              end
-
+            @ios[ios_len..-1].each do |i|
+              addr = loc_addr_str i
               logger.log "* #{log_msg} on http://#{addr}"
             end
           end
@@ -229,8 +224,13 @@ module Puma
             io = inherit_ssl_listener sock, ctx
             logger.log "* Activated #{str}"
           else
+            ios_len = @ios.length
             io = add_ssl_listener uri.host, uri.port, ctx
-            logger.log "* Listening on #{str}"
+
+            @ios[ios_len..-1].each do |i|
+              addr = loc_addr_str i
+              logger.log "* #{log_msg} on ssl://#{addr}?#{uri.query}"
+            end
           end
 
           @listeners << [str, io] if io
@@ -297,11 +297,7 @@ module Puma
     end
 
     def inherit_tcp_listener(host, port, fd)
-      if fd.kind_of? TCPServer
-        s = fd
-      else
-        s = TCPServer.for_fd(fd)
-      end
+      s = fd.kind_of?(::TCPServer) ? fd : ::TCPServer.for_fd(fd)
 
       @ios << s
       s
@@ -339,11 +335,8 @@ module Puma
     def inherit_ssl_listener(fd, ctx)
       raise "Puma compiled without SSL support" unless HAS_SSL
 
-      if fd.kind_of? TCPServer
-        s = fd
-      else
-        s = TCPServer.for_fd(fd)
-      end
+      s = fd.kind_of?(::TCPServer) ? fd : ::TCPServer.for_fd(fd)
+
       ssl = MiniSSL::Server.new(s, ctx)
 
       env = @proto_env.dup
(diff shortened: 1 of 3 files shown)


MSP-Greg:
I believe that current code may show binds multiple times if there are multiple binds.  Just pushed an update that should fix that...

lib/puma/binder.rb
```diff
@@ -229,8 +225,15 @@ def parse(binds, logger, log_msg = 'Listening')
             io = inherit_ssl_listener sock, ctx
             logger.log "* Activated #{str}"
           else
+            ios_len = @ios.length
             io = add_ssl_listener uri.host, uri.port, ctx
-            logger.log "* Listening on #{str}"
+
+            @ios[ios_len..-1].each do |i|
```
nateberkopec, line 231: Is the index necessary? are you just trying to reverse it? just use `reverse_each`

nateberkopec:
That iterator looks weird otherwise seems fine

lib/puma/binder.rb
MSP-Greg: No, not reverse.  If there are multiple binds, I believe the current loop will print multiple entries?

cjlarose: I think I'm missing some context. What problem does this PR solve?

MSP-Greg: See two tests that had skips removed.  Only pertains to '0.0.0.0' or 'localhost' binds

MSP-Greg:
Insert the following line at `test_binder.rb:100`:
```
STDOUT.puts '', @events.stdout.string
```

Run with the current PR, output is below, showing three IPv4 and 3 IPv6 addresses:
```
* Listening on http://127.0.0.1:35425
* Listening on http://[::1]:35425
* Listening on http://127.0.0.1:33875
* Listening on http://[::1]:33875
* Listening on http://127.0.0.1:34505
* Listening on http://[::1]:34505
```

Now, remove the 'array' part from `binder.rb:171`, and the output is below, showing the duplication:
```
* Listening on http://127.0.0.1:41103
* Listening on http://[::1]:41103
* Listening on http://127.0.0.1:41103
* Listening on http://[::1]:41103
* Listening on http://127.0.0.1:36667
* Listening on http://[::1]:36667
* Listening on http://127.0.0.1:41103
* Listening on http://[::1]:41103
* Listening on http://127.0.0.1:36667
* Listening on http://[::1]:36667
* Listening on http://127.0.0.1:41087
* Listening on http://[::1]:41087
```

lib/puma/binder.rb
cjlarose: I think I'm following this now. In this specific case, we want to print that we're listening on all newly-added entries to `@ios`.

In that case, I think it might be more straightforward for `add_ssl_listener` to return all instances of `Puma::MiniSSL::Server` it created.

```ruby
ssl_servers = add_ssl_listener uri.host, uri.port, ctx
ssl_servers.each do |server|
  addr = loc_addr_str i
  ssl_qry = str[/\?.+\z/]
  logger.log "* #{log_msg} on ssl://#{addr}#{ssl_qry}"
end
```

cjlarose: We can do a similar thing to `add_tcp_listener` to make the code a little easier to read.

MSP-Greg: Either works, but the logging for inherited and/or activated listeners is in the parse methods...

cjlarose: I think that's fine. I'm not suggesting we move the logging code (that can stay in `parse`). I'm just suggesting that when `parse` calls `add_ssl_listener`, it gets back the new instances of `Puma::MiniSSL::Server`. That way `parse` can iterate over them easily and log each one.

MSP-Greg: IDK.

I did just notice that `ssl_qry = str[/\?.+\z/]` can be moved out of the loop.

Note that `add_ssl_listener` currently returns the TCP server 'behind' the SSL server.  The SSL server is added to `@ios`.  That may not be needed, as I think it can also change if MiniSSL::Server does a better job of barking like a TCP server.  I think I added MiniSSL::Server#closed? for that purpose in another PR.

A while ago, someone mentioned refactoring/rewriting BInder, so I haven't really looked at it.  I've also got code for abstract unix sockets, and it runs with restart ok...

So, maybe wait on more serious Binder changes with Puma 6?  Just a thought...

cjlarose: Yeah, happy to defer a refactor to a future PR 👍

MSP-Greg: I moved the `ssl_qry` statement and also removed the `next unless *Server === i` statements, as with the @ios length handling, they're no longer needed...

cjlarose: Both changes look good to me 👍

MSP-Greg: @cjlarose Thanks for looking at it, @nateberkopec also...

lib/puma/binder.rb
```diff
@@ -229,8 +224,14 @@ def parse(binds, logger, log_msg = 'Listening')
...
             logger.log "* Activated #{str}"
           else
+            ios_len = @ios.length
             io = add_ssl_listener uri.host, uri.port, ctx
-            logger.log "* Listening on #{str}"
+
+            ssl_qry = str[/\?.+\z/]
+            @ios[ios_len..-1].each do |i|
+              addr = loc_addr_str i
+              logger.log "* #{log_msg} on ssl://#{addr}#{ssl_qry}"
```
dentarg, line 233: Can't we use the URI object to get the query instead of the regexp?

```suggestion
              logger.log "* #{log_msg} on ssl://#{addr}?#{uri.query}"
```

dentarg:
> use tertiary

Heh, I didn't get it before I looked at the diff, I'm more used to the term "ternary operator"

lib/puma/binder.rb
MSP-Greg: Good idea.  Changed...

PR #2536, opened 2021-01-27
Document workers and threads environment variables [changelog skip]

dentarg:
Related to https://github.com/puma/puma/issues/2393, #2143


nateberkopec:
Good clarification 👍

PR #2537, opened 2021-01-27
Bump version to 5.2.0 [changelog skip]

dentarg:
🍰


alksl:
We're never leaving this version :cake:  :smile:

Issue #2352, opened 2020-09-04
Solution: Puma/Rails locks up when performing an HTTP request to self in Rails development mode


earksiinni:
This issue is a follow up to https://github.com/puma/puma/issues/1433.

As established in https://github.com/puma/puma/issues/1433, the root issue indeed isn't Puma. The problem is with [Rails' autoload interlock,](https://guides.rubyonrails.org/threading_and_code_execution.html#load-interlock) which prevents a thread from using Rails' autoloading mechanism to load constants when application code is running inside of another thread.

What happens when you do a "loopback" request (i.e., Rails executes a request to a server that in turn executes a request back to Rails before the first request can be completed) is something like this:

1. Rails fires its request (request 1, thread 1) to external service using synchronous HTTP, probably using some lib like Net::HTTP, HTTParty or Typhoeus. As long as the request is pending, Rails counts it as running application code and activates the autoload interlock.
2. The external service makes a request back to Rails (request 2) while request 2 is pending.
3. Puma spawns Rails on a new thread (thread 2) to respond to request 2.
4. However, thread 2 is blocked by the autoload interlock initiated by thread 1. The Rails environment can’t load and request 2 can't be handled, it just remains in limbo until a timeout. (You can confirm this, by the way, with the [`ActionDispatch::DebugLocks` middleware](https://guides.rubyonrails.org/threading_and_code_execution.html#actiondispatch-debuglocks).)
5. So, Rails thread 1 is waiting for a response to request 1, the external service is waiting for a response to request 2 before responding to request 1, and thread 2 is waiting for thread 1 to finish execution and lift the lock before it responds to request 2. Deadlock achieved 💥.

The solution is to [wrap request 1 inside `ActiveSupport::Dependencies.interlock.permit_concurrent_loads`](https://guides.rubyonrails.org/threading_and_code_execution.html#permit-concurrent-loads). Anything wrapped inside that block won't trigger the interlock, allowing new threads to access the Rails environment and boot properly.

So something like:

```
response = ActiveSupport::Dependencies.interlock.permit_concurrent_loads do
  # Your HTTP request code here. For example:
  Faraday.post url, data: 'foo'
end

do_something_with response
```

[...]

PR #2540, opened 2021-01-27
[ci skip] Add documentation on interlock issue in Rails development mode

wildmaples:
### Description

Hello, it's my first time here! 👋  

Closes https://github.com/puma/puma/issues/2352 

The documentation is a new file `docs/rails_dev_mode.md` as suggested and pretty much a paraphrase/copy pasta of @earksiinni's analysis and observation. 

- [x] All new and existing tests passed, including Rubocop. (I mean... it probs should cause this is just a documentation change)

### DIFF
--- a/docs/rails_dev_mode.md
+++ b/docs/rails_dev_mode.md
@@ -1,0 +1,29 @@
+# Running Puma in Rails Development Mode
+
+## "Loopback requests" 
+
+Be cautious of "loopback requests", where a Rails application executes a request to a server that in turn, results in another request back to the same Rails application before the first request is completed. Having a loopback request will trigger [Rails' load interlock](https://guides.rubyonrails.org/threading_and_code_execution.html#load-interlock) mechanism. The load interlock mechanism prevents a thread from using Rails autoloading mechanism to load constants while the application code is still running inside another thread.
+
+This issue only occurs in the development environment as Rails' load interlock is not used in production environments. 
+
+### Solutions
+
+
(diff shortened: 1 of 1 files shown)


cjlarose:
This is good to document. Just a few notes worth investigating:

* This problem isn't unique to puma. Any multithreaded web server you use to serve a Rails application will have the same problem.
* I haven't verified this, but this is only a problem with the "classic" autoloader. I suspect this isn't a problem with zeitwerk.
* Another workaround is probably to just remove the `ActionDispatch::Reloader` from the Rack middleware stack. That's the thing that prevents concurrent requests from autoloading.

docs/rails_dev_mode.md
```diff
@@ -0,0 +1,29 @@
...
+      # Your HTTP request code here. For example:
+      Faraday.post url, data: 'foo'
+    end
+
+    do_something_with response
+    ```
+
+####  2. Use multiple processes on Puma
+
+  Alternatively, you may also enable multiple single-threaded workers on Puma. By doing so, you are sidestepping the problem by creating multiple processes rather than new threads. However, this workaround is not ideal because debugging tools such as [byebug](https://github.com/deivid-rodriguez/byebug/issues/487) and [pry](https://github.com/pry/pry/issues/2153), work poorly with any multi-process web server.
```
nateberkopec, line 29: This workaround only works if these multiple processes are only using 1 thread. If you use cluster mode with multiple threads, the lock can still occur.

nateberkopec (changes requested):

PR #2541, opened 2021-01-29
Move ssl detection to puma.rb [changelog skip]

MSP-Greg:
### Description

`puma.rb` loads `puma/puma_http11`, which contains the compiled c code for ssl.  Move ssl detection from `puma/detect.rb` to `puma.rb`

Fix `Puma::IS_WINDOWS` and `Puma.windows?` to work with JRuby.

No breaking API changes.

JFYI, quick benchmark of predicate methods (`Puma.ssl?`) vs constants (`Puma::HAS_SSL`), showed constants were about twice as fast.  But, both are 'fast enough'.  Haven't looked thru the code to change...

### DIFF
--- a/lib/puma.rb
+++ b/lib/puma.rb
@@ -19,6 +19,24 @@ module Puma
   autoload :Server, 'puma/server'
   autoload :Launcher, 'puma/launcher'
 
+  # at present, MiniSSL::Engine is only defined in extension code (puma_http11),
+  # not in minissl.rb
+  HAS_SSL = const_defined?(:MiniSSL, false) && MiniSSL.const_defined?(:Engine, false)
+
+  if HAS_SSL
+    require 'puma/minissl'
+  else
+    module MiniSSL
+      # this class is defined so that it exists when Puma is compiled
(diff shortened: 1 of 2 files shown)


nateberkopec:
Hmmm... why are we moving it? It looks like everything is still loaded in the same order, right? It feels cleaner to keep it in `detect`.

MSP-Greg:
> why are we moving it?

Since `HAS_SSL` is defined at loadtime, (and hence `Puma.ssl?`), `puma/puma_http11` has to be loaded first.  Currently I think it's loaded in `puma.rb` and one other file.  I may have had an issue running a single test file, but also `control_cli.rb` uses methods from `detect.rb`.   It doesn't load `puma.rb`, but it can if a 'start' command is issued.

So if `detect.rb` is loaded first, `HAS_SSL` is false, even though it may be true.  By moving it to `puma.rb`, where the file that determines it is required, we avoid issues forever.  That's why I added the comment that (I hope) states that only env/os related settings should be contained in `detect.rb`...  I think that made sense...

Issue #2542, opened 2021-01-30
Puma::MiniSSL::SSLError: System error: Undefined error: 0 - 0 with 5.2.0


mscrivo:
**Describe the bug**

Since upgrading our app to use 5.2.0 (from 5.1.1), our devs have been getting the following error on particular endpoints:

`SSL error, peer: 127.0.0.1, peer cert: : #<Puma::MiniSSL::SSLError: System error: Undefined error: 0 - 0`

The problem is 100% reproducible on some endpoints, but not other in the same app. I'm still trying to figure out what the difference is, but haven't been able to yet. 

Going back to 5.1.1 results in no SSL errors.

```
Puma starting in single mode...
* Puma version: 5.2.0 (ruby 2.7.1-p83) ("Fettisdagsbulle")
*  Min threads: 0
*  Max threads: 5
*  Environment: development
*          PID: 40350
* Listening on ssl://0.0.0.0:446?key=key&cert=crt
Use Ctrl-C to stop
127.0.0.1 - - [29/Jan/2021:21:34:03 -0500] "GET /feature-gating-new HTTP/1.1" 200 29552 0.2847
2021-01-29 21:34:03 -0500 SSL error, peer: 127.0.0.1, peer cert: : #<Puma::MiniSSL::SSLError: System error: Undefined error: 0 - 0>
127.0.0.1 - - [29/Jan/2021:21:34:05 -0500] "GET /api/pages/home HTTP/1.1" 200 495 0.0305
127.0.0.1 - - [29/Jan/2021:21:34:07 -0500] "GET /api/pages/feature-gating-new HTTP/1.1" 200 20666 0.1790
2021-01-29 21:34:07 -0500 SSL error, peer: 127.0.0.1, peer cert: : #<Puma::MiniSSL::SSLError: System error: Undefined error: 0 - 0>
```

**Puma config:**

command line args:
`bundle exec  puma -v -b "ssl://0.0.0.0:446?key=key&cert=cert" -b "tcp://0.0.0.0:83" app/config.ru`

The cert is a valid wildcard cert from a CA.

config.ru has no special configuration for dev, just loads our sinatra app.

**To Reproduce**

Unfortunately, I cannot reproduce with a simple Hello world app, which indicates it's probably something in the payload of the endpoint?  Is there a way I can enable more verbosity to give us a clue of where the error is coming from?

I did want to bring this up though, in case there's an implementation issue with the SSL changes that went into 5.2

**Expected behavior**
No SSL errors

[...]

PR #2519, opened 2020-12-31
Fix SSL Context - minissl.c, minissl.rb, extconf.rb

MSP-Greg:
### Description

While working on an update to the test suite, client connections were slow with ssl tests.  

Currently, a new context is created for each accepted client connection.  A context also reads two files for the cert and key.

Refactored to use one context created in MiniSSL::Server.new.

Code is included in a `misc` folder that reports on the times needed to open the socket and write a simple request.  It logs the time of each client's GET request, each having about a 100k body, then sorts the times, and outputs percentile times.  See the socket_times.md in the misc folder for info.

Times for this PR (see 'test ssl connection times` step):

https://github.com/MSP-Greg/puma/actions/runs/453868127

Times for current master:

https://github.com/MSP-Greg/puma/actions/runs/453894649

The CI for the PR has an additional 'test tcp connection times' step, which helps tell the two apart.

Times for this PR are approx two to three times faster than master on both Ubuntu & macOS.

But, there are a few odd things:

1. Checking it locally with Windows WSL2/Ubuntu, I noticed a larger speed increase, but that may be specific to that implementation.  Oddly, the tcp times were almost the same as Actions CI.

2. The longest macOS times are really high with the PR, compared to master.  Not sure about what could be causing the issue.

I'd appreciate if anyone running *nix or macOS natively can check the results.

To run locally, compile Puma, then (from the Puma folder) run `misc/run_socket_times.sh`.  Save the misc folder in master, and the tests will also run there.  You may need to run the following:
```
chmod +x ./misc/run_ssl_socket_times.sh
```

For more info on running `misc/run_socket_times.sh`, see [`socket_times.md`](https://github.com/puma/puma/blob/4aa4d05a41ad/misc/socket_times.md)

Note that the last three commits may be removed for the PR...

Thanks, and Happy New Year, Greg

### Your checklist for this pull request

- [x] I have reviewed the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.
- [x] I have added an entry to [History.md](../blob/master/History.md) if this PR fixes a bug or adds a feature. If it doesn't need an entry to HISTORY.md, I have added `[changelog skip]` or `[ci skip]` to the pull request title.
- [x] I have added appropriate tests if this PR fixes a bug or adds a feature.
- [x] My pull request is 100 lines added/removed or less so that it can be easily reviewed.
- [ ] If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.
[...]

Issue #2425, opened 2020-10-13
Puma 4.3.6 - Rails 6.0.3.4 - TLS brings in memory leak


Melchizedech:
**Describe the bug**
Creating a "hello world" rack application and configuring puma to run using TLS, excessive memory is used by the application.

**Puma config:**

```
threads 1, 1
workers 1
ssl_bind '0.0.0.0', 3000, {
    key: "certificates/certificate.key",
    cert: "certificates/certificate.crt",
    verify_mode: 'none',
    no_tlsv1: true,
    no_tlsv1_1: true
  }

bind 'tcp://0.0.0.0:3500'

# Specifies the `pidfile` that Puma will use.
pidfile ENV.fetch("PIDFILE") { "server.pid" }
```

CLI : docker run --rm -p 3000:3000 -p 3500:3500 puma_tls_memory_leak

**To Reproduce**

I've created a repo and a dockerfile to be able to reproduce, using your hello world rack : 
https://github.com/Melchizedech/rails_puma_tls_memory_leak

Running `while :; do curl -k --location 'https://localhost:3000' -v ; done` will increase the RAM used by the application from 25MB (RSS) to 60MB (RSS). Then every minute or so, another extra MB (RSS) will be consumed.

If we instead run ``while :; do curl --location 'http://localhost:3000/' -v ; done``, the RAM used by the app increase from 25 to 32MB (RSS) and eventually, the app will start consuming extra MB of RAM.

I definitely think there is an issue on Puma with TLS.

Here is the graph of one application that we use, it is a small Rails app that receives close to _no_ user traffic. The "real" traffic it receives comes from the liveness probe. 

Cluster A : 
![App in a production cluster](https://user-images.githubusercontent.com/1233671/95807310-951cb580-0d55-11eb-892b-b18a06a8a490.png)

Cluster B :
![App in another production cluster](https://user-images.githubusercontent.com/1233671/95807312-964de280-0d55-11eb-8651-9aaee369b78f.png)

Here are some screenshots from the ram consumption of a brand new rails app only running the API mode, while we hit a non-existing route, behind a similar configuration.

Just after boot : 
![App after having booted in production mode](https://user-images.githubusercontent.com/1233671/95807377-c301fa00-0d55-11eb-88e5-583ec80b2f76.png)

After 6 min of constant cURL
[...]

PR #2543, opened 2021-01-30
Bug Fix - MiniSSL::Socket#write - use data.byteslice(wrote..-1)

MSP-Greg:
### Description

MiniSSL::Socket#write performs some string 'dice & chop', and was using both byte and character methods.  Convert all to byte methods.

Closes https://github.com/puma/puma/issues/2542

Note that this is only noticeable when the response headers/body are such that x.length != x.bytesize

I will write a test for it, but did the checking locally with a 600kB body.  Was using `curl` and `git diff`, need to find a more Ruby based solution...

### DIFF
--- a/lib/puma/minissl.rb
+++ b/lib/puma/minissl.rb
@@ -133,7 +133,7 @@ module Puma
 
           return data_size if need == 0
 
-          data = data[wrote..-1]
+          data = data.byteslice(wrote..-1)
         end
       end
 
--- a/History.md
+++ b/History.md
@@ -1,3 +1,8 @@
+## 5.2.1 / 2021-01-
+
+* Bugfixes
+  * MiniSSL::Socket#write - use data.byteslice(wrote..-1) ([#2543])
+
 ## 5.2.0 / 2021-01-27
 
 * Features


lib/puma/minissl.rb
```diff
@@ -133,7 +133,7 @@ def write(data)
 
           return data_size if need == 0
 
-          data = data[wrote..-1]
+          data = data.byteslice(wrote..-1)
```
dentarg, line 136: @MSP-Greg Do you understand why 5.1.1 doesn't _seem_ to have the problem stated in #2542? This line of code wasn't changed in #2519 (but surrounding code was: https://github.com/puma/puma/pull/2519/files#diff-5f9e58f45420759185f3cff5203a523482b915cc3e331bb0f060977c2078d903)

MSP-Greg: I will look at 5.1.1 and 5.2.0.  I believe the bug requires a response with multi-byte characters, longer than 16kB, and ssl.  And, since some html parsers can be rather forgiving...

nateberkopec: If I'm reading the code right, I think the bug was in changing line 133 from `data.bytesize` to `data`, no?

MSP-Greg: Confused.  Do you mean changing `data.bytesize` to `data_size`, which is set to the value of `data.bytesize`?

Anyway, I'm working thru it.  The code in 5.1.1 was writing 512 byte chunks to the TCPSocket, which seemed not optimized.  But, it also appears that it wrote all the data with `@engine.write`, where the current code is writing it in 16kB chunks.  Because 5.1.1 was writing everything with `@engine.write`, maybe the issues with how it was looping weren't apparent?

Trying to figure out why the difference in write size is happening...  Regardless, I believe the current code is a lot faster?

MSP-Greg: I had to review the code changes in minissl.c, but I now recall.  It's involved, but I'll write it up.

MSP-Greg:
First, the question of why did [`MiniSSL::Socket#write`](https://github.com/puma/puma/blob/v5.1.1/lib/puma/minissl.rb#L117-L137) work in 5.1.1.  Note that the code has two while loops.  With the SSLContext options used in 5.1.1, the outer loop never looped, as all of ‘data’ was written in `@engine.write data`.  Using the 600kB body, the code wrote the whole string with `@engine.write`, extracted it in 512 byte chunks, and wrote each to the TCPSocket.

Secondly, regarding the recent changes in Puma’s SSL for 5.2.0, I decided something was wrong while working on the new test framework, as it logged connection times.  There was quite a large difference between TCP/Unix sockets and SSL sockets.

Also, I was aware of ['Puma 4.3.6 - Rails 6.0.3.4 - TLS brings in memory leak'](https://github.com/puma/puma/issues/2425).  That wasn’t ever really pinned down, but I thought it possible that a large number of open/idle ssl clients could use a fair bit of memory.

Hence, the changes in 5.2.0.  I changed the code to share an SSLContext, instead of creating a new one for each accepted connection, made some other adjustments to the Context, and re-wrote `MiniSSL::Socket.write`.  Now, with the 600kB body, it’s written in 16kB chunks, extracted in 4kB chunks, which are assembled into a 16kb string, which is then written to the TCPSocket.

I thought this would be a reasonable implementation to both improve the speed and lower the memory.  Testing locally, times have dropped for responses with both 10kB and 200kB bodies.  The options chosen should reduce the memory required for open connections (keep-alive).  I haven’t referenced OpenSSL man pages, but I (re-)read quite a few for the update…

nateberkopec:
> the outer loop never looped

OK, now I get it

MSP-Greg:
> OK, now I get it

Thanks for looking at it.  I hope the explanation was reasonable, it's kind of messy...

PR #2548, opened 2021-02-03
Remove macOS 11 builds, try to stabilize CI [changelog skip]

dentarg:
Recent examples (these were all yellow but I cancelled them)

- https://github.com/puma/puma/runs/1826040463?check_suite_focus=true
- https://github.com/puma/puma/runs/1825682787?check_suite_focus=true
- https://github.com/puma/puma/runs/1825683423?check_suite_focus=true
- https://github.com/puma/puma/runs/1825636952?check_suite_focus=true
- https://github.com/puma/puma/runs/1825636286?check_suite_focus=true

GitHub Actions is out of capacity for macOS: https://github.com/actions/virtual-environments/issues/2486

Perhaps better to add macos-11.0 back when the above issue is resolved.

### DIFF
--- a/.github/workflows/mri.yml
+++ b/.github/workflows/mri.yml
@@ -26,7 +26,6 @@ jobs:
           - { os: windows-2019 , ruby: mingw }
           - { os: ubuntu-20.04 , ruby: 2.7   , no-ssl: ' no SSL' }
           - { os: windows-2019 , ruby: 2.7   , no-ssl: ' no SSL' }
-          - { os: macos-11.0   , ruby: 2.7  }
         exclude:
           - { os: ubuntu-20.04 , ruby: 2.2  }
           - { os: ubuntu-20.04 , ruby: 2.3  }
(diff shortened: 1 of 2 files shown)


dentarg:
Both PR 2550 and PR 2549 had macos-11.0 jobs failing (but not all of them). If we merge this, we can have an issue tracking the upstream issue and add the jobs back when the upstream issue is resolved. Marking the jobs with `allow-failure` doesn't work as the problem is that the jobs doesn't even start.

https://github.com/actions/virtual-environments/issues/2486#issuecomment-773103761 says you should get a clear error message if you don't have access to the "private preview" for macOS 11, we don't see that message, so we do have access, but it hasn't been stable. Might have to do with the jobs running in forks (that doesn't have access)? 🤔 

WDYT? Maybe there's a risk we loose our "private preview" access if we remove the jobs?

nateberkopec:
I'm fine with leaving them out. Adding in macos 11 builds didn't catch any bugs and the risk of adding new ones is very low.

PR #2550, opened 2021-02-04
server.rb - properly cork & uncork ssl clients

MSP-Greg:
### Description

`Server##cork_socket` & `Server##uncork_socket` work on TCPSockets, but when using an ssl client, a MiniSSL::Socket is passed to them.

Use 'to_io', which is available with all 'socket objects' used in Puma.

I haven't worked with wireshark for a while, and that's probably the only way to see a difference?

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,6 +1,8 @@
 ## 5.2.1 / 2021-01-
 
 * Bugfixes
+  * Fix TCP cork/uncork operations to work with ssl clients (#2550)
+  * Require rack/common_logger explicitly if :verbose is true ([#2547])
   * MiniSSL::Socket#write - use data.byteslice(wrote..-1) ([#2543])
 
 ## 5.2.0 / 2021-01-27
@@ -19,7 +21,6 @@
   * Fix phased restart errors related to nio4r gem when using the Puma control server ([#2516])
   * Add `#string` method to `Puma::NullIO` ([#2520])
   * Fix binding via Rack handler to IPv6 addresses ([#2521])
-  * Require rack/common_logger explicitly if :verbose is true ([#2547])
 
 * Refactor
   * Refactor MiniSSL::Context on MRI, fix MiniSSL::Socket#write ([#2519])
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -120,17 +120,13 @@ module Puma
       # :nodoc:
       # @version 5.0.0
       def tcp_cork_supported?
-        RbConfig::CONFIG['host_os'] =~ /linux/ &&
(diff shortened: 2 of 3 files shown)


lib/puma/server.rb
```diff
@@ -120,15 +120,15 @@ def current
...
+        RbConfig::CONFIG['host_os'].include?('linux') &&
           Socket.const_defined?(:IPPROTO_TCP) &&
           Socket.const_defined?(:TCP_CORK)
       end
 
       # :nodoc:
       # @version 5.0.0
       def closed_socket_supported?
-        RbConfig::CONFIG['host_os'] =~ /linux/ &&
+        RbConfig::CONFIG['host_os'].include?('linux') &&
```
nateberkopec, line 131: What's the motivation for the change here?

MSP-Greg: General code improvement.

1. If you need a boolean, choose a method that returns one.  I think that was the reason `String#match?` was added in Ruby 2.4.
2. Use string methods if a RegExp test isn't needed.  If and when macOS supports 'cork' (and hence `/linux|darwin/`), it can be a string method.

**Off topic:** A long time ago, my first 'significant' contribution to 'OS Ruby' was working on building & testing Windows/MinGW Ruby.  Since then, I've always kept an eye on the commits to ruby/ruby.  There are a lot of commits there that include small code changes like the change above.

I can revert...

MSP-Greg: @nateberkopec

I looked at this a bit more, and `SOCKET::TCP_CORK` is not defined in Windows and macOS.  So, a better solution might be to add `HAS_TCP_CORK` to detect.rb and add a conditional to the method calls? So:
```ruby
cork_socket io
uncork_socket io
```

would change to:
```ruby
cork_socket(io) if HAS_TCP_SOCKET
uncork_socket(io) if HAS_TCP_SOCKET
```

MSP-Greg: Ran `ruby -rsocket -e 'puts Socket.constants.sort'` in CI, and updated the code based on that.  The needed constants are not defined in Windows or macOS, so removed the linux detection...

History.md
```diff
@@ -19,7 +21,6 @@
   * Fix phased restart errors related to nio4r gem when using the Puma control server ([#2516])
   * Add `#string` method to `Puma::NullIO` ([#2520])
   * Fix binding via Rack handler to IPv6 addresses ([#2521])
-  * Require rack/common_logger explicitly if :verbose is true ([#2547])
```
nateberkopec, line 22: whoops good catch

PR #2557, opened 2021-02-16
Restore sync=true on global stdout/stderr streams

### LINKED ISSUE #2545: Puma 5.2 log not flushed
Since I upgrade to Puma 5.2, the logs are not flushed right away.

```
#

# Specifies the `worker_timeout` threshold that Puma will use to wait before
# terminating a worker in development environments.
#
worker_timeout 3600 if ENV.fetch("RAILS_ENV", "development") == "development"

#
port ENV.fetch("PORT") { 3000 }

#

# processes).
#

#
# preload_app!

plugin :tmp_restart
```

In Puma 5.1.1, the logs where flush right away on stdout.

 - OS: Ubuntu 18.04
 - Docker version 20.10.3, build 48d30b5
 - Ruby 2.7.2 or 3.0
 - Puma Version 5.2

cjlarose:
### Description

Fixes #2545. In Puma 5.2, Puma stopped mutating the `sync` flag on the global `STDOUT` and `STDERR` objects. This can cause user application logs to be not flushed on every write if users were depending on this behavior. This PR restores `sync` behavior for those streams to the pre-5.2 behavior. We may re-introduce the change again in 6.0 with better documentation.

To verify this change, I used a Rack app that writes to stdout on every request

```ruby
#stdout_test.ru
app = lambda do |env|
  $stdout.write "hello\n"
  [200, {"Content-Type" => "text/plain"}, ["Hello World"]]
end

run app
```

Start puma, piping its stdout to something that's not a TTY:

```sh
bundle exec ruby -Ilib bin/puma stdout_test.ru | tee puma.out
```

Then, send a bunch of requests

```sh
seq 100 | xargs -n 1 -I{} curl http://localhost:9292
```

On Puma 5.2, you can observe that no output is visible from Puma until many more requests complete.

This change introduces a config option to the config DSL, so that users can opt into making it so that puma doesn't set `sync` flag on the stdout and stderr streams. The default behavior is consistent with Puma's behavior prior to 5.2. In a future release (maybe 6.0), we may change the default or remove the option altogether.

```ruby
mutate_stdout_and_stderr_to_sync_on_write false
```

### DIFF
--- a/History.md
+++ b/History.md
@@ -2,6 +2,7 @@
 
 * Bugfixes
   * Add `#flush` and `#sync` methods to `Puma::NullIO`  ([#2553])
+  * Restore `sync=true` on `STDOUT` and `STDERR` streams #2557
 
 ## 5.2.1 / 2021-02-05
 
--- a/lib/puma/configuration.rb
+++ b/lib/puma/configuration.rb
@@ -206,7 +206,8 @@ module Puma
         :first_data_timeout => Const::FIRST_DATA_TIMEOUT,
         :raise_exception_on_sigterm => true,
         :max_fast_inline => Const::MAX_FAST_INLINE,
-        :io_selector_backend => :auto
+        :io_selector_backend => :auto,
+        :mutate_stdout_and_stderr_to_sync_on_write => true,
       }
     end
 
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -888,5 +888,9 @@ module Puma
     def io_selector_backend(backend)
       @options[:io_selector_backend] = backend.to_sym
     end
+
+    def mutate_stdout_and_stderr_to_sync_on_write(enabled=true)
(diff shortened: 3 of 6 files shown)


nateberkopec:
Thanks. I started on a similar PR with some more stuff about deprecation/logging a warning, but I can just layer that on top of this.

MSP-Greg:
The result of this I'm in favor of.

Re the name `mutate_global_stdout_and_stderr_to_flush_on_write`, when I see 'global', I think `$stdout` & `$stderr`.  Also, consider removing `mutate` and add `sync` just to clarify the settings?

Maybe I'm getting too sensitive about naming, as I often choose poor names myself...

cjlarose:
> Re the name `mutate_global_stdout_and_stderr_to_flush_on_write`, when I see 'global', I think `$stdout` & `$stderr`. Also, consider removing `mutate` and add `sync` just to clarify the settings?

Good point! I renamed the setting to `mutate_stdout_and_stderr_to_sync_on_write`.

PR #2564, opened 2021-03-03
Add support for Linux's abstract sockets

### LINKED ISSUE #2526: Support for Linux's abstract sockets
Linux supports listening on an abstract socket that is not represented by a path on the filesystem (https://man7.org/linux/man-pages/man7/unix.7.html). I think ruby's `UNIXSocket` support abstract sockets, but some of the `File.exist?` type checks in puma fail because abstract sockets start with a NUL byte `"\0"`. Thoughts on supporting abstract sockets?

MSP-Greg:
### Description

Closes #2526

Tests for abstract UNIXSockets exist in the updated test framework. Will open a PR for it after all the library file changes have been accepted.  Tested both server & control binding...

### DIFF
--- a/lib/puma.rb
+++ b/lib/puma.rb
@@ -39,6 +39,20 @@ module Puma
     HAS_SSL
   end
 
+  def self.abstract_unix_socket?
+    @abstract_unix ||=
+      if HAS_UNIX_SOCKET
+        begin
+          ::UNIXServer.new("\0puma.temp.unix").close
+          true
+        rescue ArgumentError  # darwin
+          false
+        end
+      else
+        false
+      end
+  end
+
   # @!attribute [rw] stats_object=
   def self.stats_object=(val)
     @get_stats = val
--- a/lib/puma/binder.rb
+++ b/lib/puma/binder.rb
@@ -177,11 +177,19 @@ module Puma
           @listeners << [str, io] if io
         when "unix"
           path = "#{uri.host}#{uri.path}".gsub("%20", " ")
+          abstract = false
+          if str.start_with? 'unix://@'
+            raise "OS does not support abstract UNIXSockets" unless Puma.abstract_unix_socket?
+            abstract = true
+            path = "@#{path}"
+          end
 
           if fd = @inherited_fds.delete(str)
+            @unix_paths << path unless abstract
             io = inherit_unix_listener path, fd
             logger.log "* Inherited #{str}"
           elsif sock = @activated_sockets.delete([ :unix, path ])
+            @unix_paths << path unless abstract || File.exist?(path)
             io = inherit_unix_listener path, sock
             logger.log "* Activated #{str}"
           else
@@ -205,6 +213,7 @@ module Puma
               end
             end
 
+            @unix_paths << path unless abstract || File.exist?(path)
             io = add_unix_listener path, umask, mode, backlog
             logger.log "* #{log_msg} on #{str}"
           end
@@ -355,8 +364,6 @@ module Puma
     # Tell the server to listen on +path+ as a UNIX domain socket.
     #
     def add_unix_listener(path, umask=nil, mode=nil, backlog=1024)
-      @unix_paths << path unless File.exist? path
-
       # Let anyone connect by default
       umask ||= 0
 
@@ -373,8 +380,7 @@ module Puma
             raise "There is already a server bound to: #{path}"
           end
         end
-
-        s = UNIXServer.new(path)
+        s = UNIXServer.new path.sub(/\A@/, "\0") # check for abstract UNIXSocket
         s.listen backlog
         @ios << s
       ensure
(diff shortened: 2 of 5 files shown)


MSP-Greg:
Rebased and added two simple tests.  The new test framework has a test or two creating hundreds of abstract unix clients...

MSP-Greg:
Thought I'd rebase this.  Also, curl can be used to verify the PR.  Compile the Puma repo, then

```bash
# assumes tcp 127.0.0.1:40000 is avaiable, change as needed
bundle exec ruby -Ilib bin/puma -q -w2 -t5:5 -b tcp://127.0.0.1:40000 test/rackup/realistic_response.ru
curl -kso /dev/null -w '%{size_download}' http://127.0.0.1:40000

bundle exec ruby -Ilib bin/puma -q -w2 -t5:5 -b unix://@puma.unix test/rackup/realistic_response.ru
curl -kso /dev/null -w '%{size_download}' --abstract-unix-socket puma.unix http:

```
The file realistic_response.ru actually returns a 200,000 byte body.  Both curl commands should output 200000, which is the body byte size...

lib/puma.rb
```diff
@@ -39,6 +39,20 @@ def self.ssl?
     HAS_SSL
   end
 
+  def self.abstract_unix_socket?
```
nateberkopec, line 42: Should this be in `detect`?

lib/puma/binder.rb
```diff
@@ -407,24 +412,24 @@ def inherit_unix_listener(path, fd)
...
-      redirects = listeners.map { |a| [a[1].to_i, a[1].to_i] }.to_h
+      redirects = @listeners.map { |a| [a[1].to_i, a[1].to_i] }.to_h
       redirects[:close_others] = true
       redirects
     end
 
     # @version 5.0.0
     def redirects_for_restart_env
-      listeners.each_with_object({}).with_index do |(listen, memo), i|
+      @listeners.each_with_object({}).with_index do |(listen, memo), i|
```
nateberkopec, line 432: Why do we have to work around the accessor method?

lib/puma/control_cli.rb
```diff
@@ -176,7 +176,8 @@ def send_request
         when 'tcp'
           TCPSocket.new uri.host, uri.port
         when 'unix'
-          UNIXSocket.new "#{uri.host}#{uri.path}"
+          UNIXSocket.new(@control_url.start_with?('unix://@') ?
```
nateberkopec, line 179: Everywhere that we check for a socket starting with `@`, I think we should have a comment explaining this a check for an abstract socket.

lib/puma/binder.rb
```diff
@@ -374,7 +381,7 @@ def add_unix_listener(path, umask=nil, mode=nil, backlog=1024)
           end
         end
 
-        s = UNIXServer.new(path)
+        s = UNIXServer.new path.sub(/\A@/, "\0")
```
nateberkopec, line 384: Everywhere that we check for a socket starting with `@`, I think we should have a comment explaining this a check for an abstract socket.

lib/puma/binder.rb
```diff
@@ -177,11 +177,19 @@ def parse(binds, logger, log_msg = 'Listening')
           @listeners << [str, io] if io
         when "unix"
           path = "#{uri.host}#{uri.path}".gsub("%20", " ")
+          abstract = false
+          if str.start_with? 'unix://@'
```
nateberkopec, line 181: (this one is explained by the next line so no comment required here)

nateberkopec:
Minor style concerns, otherwise LGTM

lib/puma/binder.rb
MSP-Greg: This is something that is somewhat personal preference, as I doubt there's any performance difference in the general case.

First, `@listeners` is initialized.  Secondly, I haven't looked at it for a while (Puma 6 is not around the corner), but there are a lot of public methods/accessors that could be removed.  Changing our code to allow that is helpful.

### General

1. Removing accessors may break tests, but conversely, accessors added just to make testing simple may not be a good idea.  Tests can use `#instance_variable_get` if needed.
2. Reducing the public API often shows where encapsulation has been broken, often where an accessor should be replaced  with one or more 'action' methods.
3. I'll never forget breaking Capybara.
4. I'm not sure if I changed all the code in Binder to use `@listeners`.

lib/puma.rb
MSP-Greg: I could make it more clear, but I thought `detect` would be for OS dependent constants/methods, and Puma was used for constants/methods that could only be determined after something was required.

To check 'ssl', 'puma/puma_http11' needs to be required.  To check `UNIXSocket` (and abstract), we need to require 'socket'?

lib/puma/binder.rb
MSP-Greg: Done

lib/puma/control_cli.rb
MSP-Greg: Done

PR #2565, opened 2021-03-06
Print warning when running one-worker cluster

### LINKED ISSUE #2534: Log warning when max worker count is 1
I'm struggling to think of a good reason to use cluster w/1 worker vs single mode.

CGA1123:
Running Puma in cluster-mode is likely a misconfiguration in most
scenarios.

Cluster mode has some overhead of running an addtional 'control' process
in order to manage the cluster. If only running a single worker it is
likely not worth paying that overhead vs running in single mode with
additional threads instead.

There are some scenarios where running cluster mode with a single worker
may still be warranted and valid under certain deployment scenarios, see
the linked issue for details.

From experience at work, we were able to migrate our Rails applications
from single worker cluster to single-mode and saw a reduction of RAM
usage of around ~15%.

Closes #2534

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -332,6 +332,13 @@ module Puma
       # This is aligned with the output from Runner, see Runner#output_header
       log "*      Workers: #{@options[:workers]}"
 
+      if @options[:workers] == 1
+        log "! WARNING: Detected running cluster mode with 1 worker:"
+        log "! Running Puma in cluster mode with a single worker is often a misconfiguration."
+        log "! Consider running Puma in single-mode in order to reduce memory overhead."
+        log "! See: https://github.com/puma/puma/issues/2534"
+      end
+
       # Threads explicitly marked as fork safe will be ignored.
       # Used in Rails, but may be used by anyone.
       before = Thread.list.reject { |t| t.thread_variable_get(:fork_safe) }
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -332,12 +332,6 @@ module Puma
       # This is aligned with the output from Runner, see Runner#output_header
       log "*      Workers: #{@options[:workers]}"
 
-      if @options[:workers] == 1
-        log "! WARNING: Detected running cluster mode with 1 worker:"
-        log "! Running Puma in cluster mode with a single worker is often a misconfiguration."
-        log "! Consider running Puma in single-mode in order to reduce memory overhead."
-      end
-
       # Threads explicitly marked as fork safe will be ignored.
       # Used in Rails, but may be used by anyone.
       before = Thread.list.reject { |t| t.thread_variable_get(:fork_safe) }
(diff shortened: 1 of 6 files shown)


lib/puma/cluster.rb
```diff
@@ -332,6 +332,13 @@ def run
       # This is aligned with the output from Runner, see Runner#output_header
       log "*      Workers: #{@options[:workers]}"
 
+      if @options[:workers] == 1
+        log "! WARNING: Detected running cluster mode with 1 worker:"
+        log "! Running Puma in cluster mode with a single worker is often a misconfiguration."
+        log "! Consider running Puma in single-mode in order to reduce memory overhead."
+        log "! See: https://github.com/puma/puma/issues/2534"
```
dentarg, line 339: I don't think we need to link the issue in the warning message

dentarg (commented):
I think it would be good to be able to silence the warning, as outlined in #2534, there might be users who really want to run Puma this way.

dentarg:
A test would also be good, so future refactoring can be done with confidence. Looks like [`test_run_hooks_and_exception`](https://github.com/puma/puma/blob/503b1cde5347ce9e78f81cc3b6b1724508c8ba1a/test/test_config.rb#L320-L332) could provide some inspiration for that.

CGA1123:
@dentarg Thanks for taking a look!

What would be the preferred way to suppress this warning? Via an environment variable? A new configuration option?

dentarg:
I would go for environment variable

nateberkopec:
Env variable seems like the wrong choice @dentarg - wouldnt you want this warning to be supressed "for all time"? Env variables mean that other devs in dev environments, for example, would have to also have the env variable set.

dentarg:
@nateberkopec you're right, that makes sense. I was thinking of other types of software where ENVs is often used to hide warnings.

I guess an ENV is good if you're in a situation where you can't change the config for some reason but you have control over the environment. (Might not be a common thing with this case)

CGA1123:
In that case then should this be implemented as a new method in `Puma::DSL` or as a new keyword argument to `Puma::DSL#workers`? or something?

One potential issue with adding it as a new keyword argument to `#workers` is that many people may be relying on the `WEB_CONCURRENCY` environment variable to manage worker counts.
So it may be more appropriate to add it as a new `DSL` method? `silence_single_worker_warning` or something like that?

dentarg:
> So it may be more appropriate to add it as a new `DSL` method? `silence_single_worker_warning` or something like that?

Sounds good to me 👍 The warning could inform about this setting

CGA1123:
I'm not sure that the failing test is related to these changes, seems like a flake!

lib/puma/cluster.rb
```diff
@@ -470,6 +472,15 @@ def run
 
     private
 
+    def single_worker_warning
+      return if @options[:workers] != 1 || @options[:silence_single_worker_warning]
+
+      log "! WARNING: Detected running cluster mode with 1 worker."
+      log "! Running Puma in cluster mode with a single worker is often a misconfiguration."
+      log "! Consider running Puma in single-mode in order to reduce memory overhead."
```
nateberkopec, line 480: Let's say "in single-mode (workers = 0)"

PR #2568, opened 2021-03-09
Scatter/gather IO with writev

robotmay:
### Description

:wave: Apologies in advance for the half-finished PR. I'm not sure if this is the right way to implement this and I'm very open to feedback on how to improve it/collaborating with others :smile: 

This adds scatter/gather IO support using `writev`, which supports writing arrays directly to the output since Ruby 2.5 (this'll need a version switch somewhere): https://bugs.ruby-lang.org/issues/9323. This means we can avoid generating big strings before writing the output, which will save memory and GC effort.

I'd expect to see a pretty reasonable performance improvement from this but it will require support further down the stack too. I've got [some investigation details](https://gitlab.com/gitlab-org/gitlab/-/issues/301025) of where support is currently lacking from the stack I'm working with, but there's most likely others that do/don't work elsewhere.

I _think_ my current implementation works but it would be nice to see if it works for others too. It's written quite verbosely due to my inexperience with the Puma codebase :sweat_smile: 

- [ ] I have added an entry to [History.md](../blob/master/History.md) if this PR fixes a bug or adds a feature. If it doesn't need an entry to HISTORY.md, I have added `[changelog skip]` or `[ci skip]` to the pull request title.

### DIFF
--- a/lib/puma/io_buffer.rb
+++ b/lib/puma/io_buffer.rb
@@ -1,11 +1,13 @@
 # frozen_string_literal: true
 
 module Puma
-  class IOBuffer < String
-    def append(*args)
-      args.each { |a| concat(a) }
+  class IOBuffer < Array
+    def bytesize
+      sum(&:bytesize)
     end
 
     alias reset clear
+    alias size bytesize
+    alias to_s join
   end
 end
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative "writer"
+
 module Puma
 
   # The methods here are included in Server, but are separated into this file.
@@ -123,7 +125,7 @@ module Puma
           end
 
           lines << LINE_END
-          fast_write io, lines.to_s
+          fast_write io, lines
           return res_info[:keep_alive]
         end
 
@@ -137,7 +139,7 @@ module Puma
 
         lines << line_ending
 
-        fast_write io, lines.to_s
+        fast_write io, lines
 
         if response_hijack
           response_hijack.call io
@@ -194,23 +196,7 @@ module Puma
     # @raise [ConnectionError]
     #
     def fast_write(io, str)
-      n = 0
-      while true
-        begin
-          n = io.syswrite str
-        rescue Errno::EAGAIN, Errno::EWOULDBLOCK
-          if !IO.select(nil, [io], nil, WRITE_TIMEOUT)
-            raise ConnectionError, "Socket timeout writing data"
-          end
-
-          retry
-        rescue  Errno::EPIPE, SystemCallError, IOError
-          raise ConnectionError, "Socket timeout writing data"
-        end
-
-        return if n == str.bytesize
-        str = str.byteslice(n..-1)
-      end
+      Writer.write(io, str)
     end
     private :fast_write
 
--- a/lib/puma/writer.rb
+++ b/lib/puma/writer.rb
@@ -1,0 +1,77 @@
+# frozen_string_literal: true
+
+module Puma
+  class Writer
+    class << self
+      def write(io, data)
+        case data.class
+        when Array
+          write_array(io, data)
+        when String
+          write_str(io, data)
+        else
+          write_str(io, data.to_s)
+        end
+      end
+
+      def write_str(io, str)
+        n = 0
+
+        while true
+          handle_errors(io) do
+            n = io.write str
+          end
+
+          return if n == str.bytesize
+
+          str = str.byteslice(n..-1)
+        end
+      end
+
+      def write_array(io, array)
+        n = 0
+
+        while true
+          handle_errors(io) do
+            n = io.write(*array)
+          end
+
+          return if n == array_bytesize(array)
+
+          array = array_byteslice_at_byte(array, n)
+        end
+      end
+
+      private
+
+      def handle_errors(io)
+        yield
+      rescue Errno::EAGAIN, Errno::EWOULDBLOCK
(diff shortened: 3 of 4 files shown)


nateberkopec:
Paging @MSP-Greg, interested in your opinion

MSP-Greg:
@nateberkopec

Sorry, I did see this earlier, but busy with etc.  Haven't looked at the CI.

At present, this is just affecting the headers of the response, as the body is written one array element at a time (if it's an array of strings).

Also, minissl.rb may need a patch for the write, not sure.

When I was looking at #2425 with current master, using 200kB response bodies, Ruby did seem to have an issue with releasing the strings, but that was a one element array, so I'm not sure if this will affect that.

I wanted to test this later, maybe use a body array length of 20, with 10kB elements?  Put your Rails hat on, what could we consider the form of the typical Rails' Rack response body?

@robotmay - Thanks.  Nice idea.  I'll try to see what improvements it makes...

robotmay:
Thanks both!

Doh I thought I was getting the body :sweat_smile:  I'll try to reread the code and look at where the body is being sent. I was originally hoping to get a working prototype with a basic Rails app going to try out some benchmarks but I ended up too busy at the time to try rejigging Rails' buffer setup. 

I wasn't too sure about adding the `Array`/`String` conditional in the writer and whether that would affect the performance or not, but I couldn't think of a clever way to match up the `byteslice` implementations :thinking: I definitely think there's a better way of writing this.

I'm really curious to see what sorta results you get - I don't really have a rough guess as to what difference it'll make at the moment.

MSP-Greg:
@robotmay

Sorry, I should have linked/listed to the body code, starts at [`request.rb:L148`](https://github.com/puma/puma/blob/714ce75a1cb/lib/puma/request.rb#L148-L157).
```ruby
res_body.each do |part|
  next if part.bytesize.zero?
  if chunked
    str = part.bytesize.to_s(16) << line_ending << part << line_ending
    fast_write io, str
  else
    fast_write io, part
  end
  io.flush
end
```
It calls fast_write for each part, so it never concatenates or creates another array.  Also, not sure about the differences between multiple `#syswrite <str>` calls (and the begin..rescue..end block) and one `#write <array>` call.

I looked at your link for 'some investigation details', and some of those could be operating on a large number of strings.  As mentioned above, I'm not sure about how many strings rack responses have...

robotmay:
Aah interesting, thanks! That'll be interesting to see whether passing the whole body to the writer would be faster than iterating it :thinking: 

I did find that we can't mix `syswrite` and `write` calls (I did see this documented somewhere) so I'm not sure if there are any other differences there to be aware of.

robotmay:
At GitLab some of our response bodies are...quite large, I suspect :sweat_smile: 

I was really just trying to experiment with `writev` support in the full Ruby stack as it's something [Elixir/Phoenix have](https://www.evanmiller.org/elixir-ram-and-the-template-of-doom.html) and it seemed like something that might have a positive performance impact in Ruby too. So there wasn't a specific problem with running Puma, just that it was the next step in the list of things to modify to support it :smile:

nateberkopec:
Checking in here - do you need any help with this @robotmay?

scarroll32:
This looks like a great extension to Puma @robotmay !

robotmay:
I somewhat forgot about this 😅 
I'd quite like to get back onto this and rework it. I assume nobody else has revisited this in the past year?

scarroll32:
@nateberkopec does Puma have anything like a feature flags or configuration options that would make this change optional?

We are looking at this in the context of https://gitlab.com/gitlab-org/gitlab/-/issues/301025

MSP-Greg:
@robotmay

I am currently working on a benchmark suite, and the first one I'll be opening a PR for measures response time for a set of wrk runs, allowing one to vary the response body type and size. It works with 'single' string bodies, File/IO, chunked, and array.  One can specify the array sizes.

Hence, it would work very well for experimenting with this.  Currently, Puma does not 'join' the array into one large string, it writes each element thru the fast write method.

Also, once I finish the above, I've got a PR that improves response times.  Part of the change is to buffer the actual socket writes, such that they're only written to the socket when the buffer reaches a certain size.

nateberkopec:
@scarroll32 If I could make a request, since this is on Gitlab's radar, is a benchmark which models a real-world situation you have and shows a clear improvement with this change (or a similar change). That would certainly bump it up everyone's priority list.

MSP-Greg:
@scarroll32 I suspect you may have lots of benchmark code written for internal use.  If anyone has time, wondering if you see a performance improvement with PR #2896.

It does use some IO.copy_stream, but I found that writev was still slower that using a string buffer (Puma::IOBuffer/StringIO) for limited string assembly followed by a socket#write...

robotmay:
Doh sorry @MSP-Greg & @nateberkopec, I forgot to reply earlier in the month 😓 

We do have pretty comprehensive benchmarks! I think the results are all public actually 🤔 The problem with the writev stuff is that it involves making a bunch of tweaks at other levels too, like the Rails template rendering stack, which makes testing stuff out a bit of a pain. However if there's something like #2896 that we can test out without making those changes, then I'm more than happy to give that a go.

I'll go check with our performance testing team and see if we can set up a test run based off a merge request - I'd be really curious to see what impact that change has 😄

MSP-Greg:
@robotmay

> see if we can set up a test run based off a merge request

Thanks for looking into it.

Wondering if any info, ie 'couldn't test' or 'no significant change', etc...

nateberkopec:
Closing as inactive, please bump once you are able to make the necessary changes 😄

PR #2575, opened 2021-03-13
Add between_bytes_timeout to fix timing out when we get a slow request

### LINKED ISSUE #2574: Timeout during long file upload
**Not really a bug, more clarification**

We use a little file upload server that is a simple rack application and uses Puma as the server. I upgraded it from 4.3.3 to 5.2.2 and now we see timeouts after 30 seconds (`first_data_timeout`) when uploading files to it where the upload takes longer than 30 seconds. I can increase `first_data_timeout` and the timeout happens at the appropriate time.

An example using curl of an upload is pretty simple `curl --data-binary @file.zip server:5000/upload?some=params`

The application runs in our Kubernetes cluster behind Traefik which sits behind an AWS ALB/WAF. All that doesn't seem to matter though as using port-forward to talk directly to the pod gives the same result.

Looking in some code it looks like `first_data_timeout` is deprecated in v6? Or is that just the `attr_writer` for it?

Just wanting some guidance as to what I should be looking at to solve this problem. I could just revert back to 4.3.3 (which I'll probably do for now), but I'd like to keep upgrading and running the newer version for security, performance, etc., reasons.

I'm slightly confused as to why it's happening in the first place, since I don't know where the data is buffering for it to be waiting and causing puma to timeout. Or is the name `first_data_timeout` just not accurate in this case and it is getting data, but it's timing out for another reason?

Do I need some other puma configuration? Should I be running nginx in a sidebar to sit in front of puma?

**Puma config**

```
workers 1
threads 1, 16

preload_app!

rackup      DefaultRackup
port        5000
environment 'production'
```

darkhelmet:
### Description
Closes #2574 by adding a `between_bytes_timeout` config option to handle slow requests instead of treating `first_data_timeout` as a global request timeout.

### Some notes

- I didn't set a default in `const.rb` for this since it defaults to whatever is provided for `:first_data_timeout`
- I added the `assert_proper_timeout` method in the server test file because I saw another assertion added to that file, there was only one, and I didn't see a rhyme or reason as to where I should add that assertion. Let me know if there's a better spot for it.

### DIFF
--- a/History.md
+++ b/History.md
@@ -4,6 +4,7 @@
   * Your feature goes here <Most recent on the top, like GitHub> (#Github Number)
   * Warn when running Cluster mode with a single worker (#2565)
   * Add reason to worker time out and startup time when worked boots ([#2528])
+  * Add :between_bytes_timeout to prevent timeout during slow incoming requests (#2575)
 
 * Bugfixes
   * Your bugfix goes here <Most recent on the top, like GitHub> (#Github Number)
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -207,7 +207,7 @@ module Puma
 
     def finish(first_data_timeout, between_bytes_timeout)
       return if @ready
-      timeout = @parsed_bytes > 0 ? first_data_timeout : between_bytes_timeout
+      timeout = @parsed_bytes > 0 ? between_bytes_timeout : first_data_timeout
       IO.select([@to_io], nil, nil, timeout) || timeout! until try_to_finish
     end
 
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -205,8 +205,9 @@ module Puma
       try_to_finish
     end
 
-    def finish(timeout)
+    def finish(first_data_timeout, between_bytes_timeout)
       return if @ready
+      timeout = @parsed_bytes > 0 ? first_data_timeout : between_bytes_timeout
       IO.select([@to_io], nil, nil, timeout) || timeout! until try_to_finish
     end
 
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -272,6 +272,12 @@ module Puma
       @options[:first_data_timeout] = Integer(seconds)
     end
 
+    # Define how long the tcp socket stays open, once data has been received.
+    # @see Puma::Server.new
+    def between_bytes_timeout(seconds)
+      @options[:between_bytes_timeout] = Integer(seconds)
+    end
+
     # Work around leaky apps that leave garbage in Thread locals
     # across requests.
     def clean_thread_locals(which=true)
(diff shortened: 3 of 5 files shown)


wjordan:
This looks like a great start, thanks for putting this PR together to follow up with this issue!

In order to distinguish between 'first-byte' and 'between-byte' timeouts, not all between-byte timeouts will be set from the reactor-wakeup, there are a couple of other edge cases to consider:

- If `queue_requests` is `false`, the connection is never added to the Reactor and will use the loop in `Client#finish` to read the request fully
- If part of the request is readable on the connection before it's initially accepted (so it's partially read before it ever gets added to the Reactor), the connection will be added to the Reactor with its timeout set from `process_client` even though it's already past the first byte. (I think this case is what may be causing some of the tests to intermittently fail)

One possible way of handling these cases would be to accept a second `between_bytes_timeout` parameter for both `Client#set_timeout` and `Client#finish` methods and use the second timeout if `@parsed_bytes == 0`.

That said, given the slightly greater amount of changes needed, it might be worth splitting it into two separate PRs, one to narrowly fix the regression on `first_data_timeout` behavior and another to add the `between_bytes_timeout` option. @nateberkopec any thoughts on direction to go with this?

darkhelmet:
I got a test and change for when `queue_requests = false` but I'm not sure how to write a test for your other point.

Yeah, we can definitely split this up if you want, or do whatever other refactorings. Just let me know what direction I should go with it.

darkhelmet:
Well, now I'm just breaking things. I'll wait for some guidance :)

nateberkopec:
Sorry @darkhelmet, I've been busy. I or someone else will be around to help soon.

dchandekstark:
Just commenting to bump the issue.  Thanks to @darkhelmet for reporting and working on a solution. We got bit by this using version 5.0.3 exactly.  Will likely downgrade to 5.0.2 until this PR is sorted out and released.

wjordan:
I made some tweaks to the PR in 6050b4f22901436c01c850a1d674a53054d23aa6, which fixes up the logic on some of the edge cases so all the tests pass. It also updates `test_no_timeout_after_data_received` to match the case described in the issue making it a proper 'regression test' where it fails without this PR and passes with it, to make sure we don't unintentionally reintroduce the same bug again later on.

@darkhelmet please take a look! If the extra commit looks good to you feel free merge the commit into your branch to update this PR.

nateberkopec:
@darkhelmet Looks you need to rebase on top of master

nateberkopec:
Code LGTM

nateberkopec:
@darkhelmet Where exactly do you need further help or direction here?

darkhelmet:
@nateberkopec I haven't had time to look at this further. With #2606 the main problem is fixed as well, so the need isn't as great. If you want to get this PR out of site, we can close it and I can revisit in the future if I find the time.

nateberkopec:
Cool! Happy to help when you've got the time. 👍

PR #2578, opened 2021-03-17
Update link to rubinius

sandstrom:
The rubinius website doesn't seem to be related to Ruby anymore. Now it's simply says:

image

I guess that link used to be relevant, but it no longer is.


README.md
```diff
@@ -16,7 +16,7 @@ Puma is a **simple, fast, multi-threaded, and highly concurrent HTTP 1.1 server
 
 Puma processes requests using a C-optimized Ragel extension (inherited from Mongrel) that provides fast, accurate HTTP 1.1 protocol parsing in a portable way. Puma then serves the request using a thread pool. Each request is served in a separate thread, so truly concurrent Ruby implementations (JRuby, Rubinius) will use all available CPU cores.
 
-Puma was designed to be the go-to server for [Rubinius](https://rubinius.com), but also works well with JRuby and MRI.
```
dentarg, line 19: Does Puma still work well with Rubinius? 🤔

IMHO we can keep the historical info (how Puma started), drop the link and state that Puma works with Ruby and JRuby.

dentarg: Or just change the link to https://github.com/rubinius/rubinius

sandstrom:
@dentarg Good point, I've updated!

PR #2586, opened 2021-03-26
Add rack_url_scheme to Puma::DSL, allows setting of rack.url_scheme header

### LINKED ISSUE #2569: Add config option to manually define Rack url scheme
We need to run Puma behind a service that is unfortunately not passing down `HTTP_X_FORWARDED` headers. We also need to run Puma on a port other than port 443 and not in SSL mode, while the frontend service handles the SSL.

This has caused a problem within Puma where redirects generated by Rails (notably on session create) use `http` rather than `https` and this is confusing clients and the frontend service.

The only way I can find in Puma itself to set Rack's url scheme is with this line in `lib/puma/request.rb`:

`env[RACK_URL_SCHEME] = default_server_port(env) == PORT_443 ? HTTPS : HTTP`

Following the thread, `default_server_port` checks if Puma is running in SSL mode, then checks if the `X_FORWARDED` headers are there, returning `PORT_443` in this case, otherwise `PORT_80`. To work around this I have monkey patched `default_server_port` to return `PORT_443` in our environment regardless.

It would be great to be able to more explicitly control the Rack url scheme, and I suggest that the port assumptions here could do with refactoring for environments where Puma is running on a custom port (neither 443 or 80).

If there's some other way to achieve what I'm trying to achieve or if I've missed some configuration option, please let me know.

MSP-Greg:
### Description

Closes #2569.

### DIFF
--- a/lib/puma/binder.rb
+++ b/lib/puma/binder.rb
@@ -41,6 +41,7 @@ module Puma
         "rack.multithread".freeze => conf.options[:max_threads] > 1,
         "rack.multiprocess".freeze => conf.options[:workers] >= 1,
         "rack.run_once".freeze => false,
+        RACK_URL_SCHEME => conf.options[:rack_url_scheme],
         "SCRIPT_NAME".freeze => ENV['SCRIPT_NAME'] || "",
 
         # I'd like to set a default CONTENT_TYPE here but some things
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -381,6 +381,12 @@ module Puma
       @options[:rackup] ||= path.to_s
     end
 
+    # Allows setting `env['rack.url_scheme']`, which may need to be set due
+    # to frontend configuration.  Normal values are 'http' or 'https'.
+    def rack_url_scheme(scheme=nil)
+      @options[:rack_url_scheme] = scheme
+    end
+
     def early_hints(answer=true)
       @options[:early_hints] = answer
     end
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -381,8 +381,9 @@ module Puma
       @options[:rackup] ||= path.to_s
     end
 
-    # Allows setting `env['rack.url_scheme']`, which may need to be set due
-    # to frontend configuration.  Normal values are 'http' or 'https'.
+    # Allows setting `env['rack.url_scheme']`.
+    # Only necessary if X-Forwarded-Proto is not being set by your proxy
+    # Normal values are 'http' or 'https'.
     def rack_url_scheme(scheme=nil)
       @options[:rack_url_scheme] = scheme
     end
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -51,7 +51,7 @@ module Puma
       head = env[REQUEST_METHOD] == HEAD
 
       env[RACK_INPUT] = body
-      env[RACK_URL_SCHEME] = default_server_port(env) == PORT_443 ? HTTPS : HTTP
+      env[RACK_URL_SCHEME] ||= default_server_port(env) == PORT_443 ? HTTPS : HTTP
 
       if @early_hints
         env[EARLY_HINTS] = lambda { |headers|
--- a/test/test_puma_server.rb
+++ b/test/test_puma_server.rb
@@ -304,7 +304,6 @@ EOF
     assert_match(/HTTP\/1.0 500 Internal Server Error/, data)
   end
 
-
   def test_eof_on_connection_close_is_not_logged_as_an_error
     server_run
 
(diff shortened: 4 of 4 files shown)


nateberkopec:
I would've appreciated the test refactoring as a separate (or several) separate PRs 😫

test/helpers/svr_in_proc.rb
```diff
@@ -0,0 +1,140 @@
...
+      end
+
+      @server = Puma::Server.new app, @events, @conf.options
+      @server.inherit_binder Puma::Binder.new(@events, @conf)
+      @server.binder.parse @conf.options[:binds], @events
+      @server.run
+      sleep 0.01
+    end
+
+    def rack_env_to_body
```
nateberkopec, line 101: This is a test fixture, and as such should be defined in the test that uses it, IMO

test/test_rack_env.rb
```diff
@@ -0,0 +1,39 @@
...
+    skip_unless :aunix
+    bind_type :aunix
+    rack_url_scheme 'http'
+  end
+
+  private
+
+  def rack_url_scheme(val)
+    start_server rack_env_to_body
+    assert_includes fast_connect_get_body.lines, "rack.url_scheme: #{val}\n"
```
nateberkopec, line 37: We should have a higher-level assert:

```
assert_body_includes "rack.url_scheme"
```

...so test writers don't have to understand `fast_connect_get_body.lines`

test/test_rack_env.rb
```diff
@@ -0,0 +1,39 @@
...
+    bind_type :ssl
+    rack_url_scheme 'https'
+  end
+
+  def test_rack_url_scheme_user
+    bind_type config: "rack_url_scheme 'user'"
+    rack_url_scheme 'user'
+  end
+
+  def test_rack_url_scheme_aunix_http
```
nateberkopec, line 26: What's the reasoning behind testing this with abstract unix sockets? I don't see why this test is necessary given the other three.

test/helpers/svr_in_proc.rb
```diff
@@ -0,0 +1,140 @@
...
+      lambda { |env|
+        body = ''.dup
+        env.sort.each { |a| body << "#{a.first}: #{a[1]}\n" }
+        [200, {}, [body]]
+      }
+    end
+
+    def ci_test
```
nateberkopec, line 109: Looks like this is not used in this pull request?

test/helpers/sockets.rb
```diff
@@ -0,0 +1,435 @@
...
+        msg << "\n read   %4d" % replies[:reqs_good_rd].reduce(:+)
+        replies[:reqs_good_rd].each { |i| msg << "  %4d" % i }
+        msg << "\n\n"
+      end
+      msg
+    end
+
+    # Generates the request/response time distribution string
+    #
+    def time_info(threads, clients_per_thread, time_ary, req_per_client = 1)
```
nateberkopec, line 381: Not used in the PR?

nateberkopec:
If you could remove everything from this PR that's not actually used in this PR, that would help me a lot to understand what's going on.

The whole approach and changes are 💯 but I can't follow everything when there's so much to sort through.

MSP-Greg:
All the files in test/helpers are intended to be test fixture files.  These are part of the test rewrite I've been using for while, and I've also mentioned also.

This PR includes some of them, and uses them to for the new test file 'test_rack_env.rb'.  PR #2595 'Improve client response code, chunked bodies', also uses `sockets.rb` in the benchmark files and some new rackup files.

Re the helper files:

`sockets.rb` is a fixture for creating client sockets, whether a single client or a stream of them.  It shares variables with `svr_in_proc.rb`, which is used to create in-process server instances.

So, these fixtures can be used to simplify test_puma_server, test_puma_server_sll, etc.  I've got a few test files I've converted to the new system, but those files are not in this PR.

A file that isn't in this PR is `svr_popen.rb`, which creates instances of Puma via `IO.popen`.  I do have the integration tests converted to the new system.

So, rather than a huge PR with a a whole new test system, I thought adding the fixtures operating on one test file would be a good first step.  I expect they will evolve, have more comments added, etc.  I think I started a 'Test Fixture Overview` markdown file also.

test/test_rack_env.rb
MSP-Greg: I can remove, I added it to show how simple it is to change from one socket type to another.  It is also skipped because when I did the PR, abstract UNIXSockets weren't supported.  I'll remove.

test/helpers/svr_in_proc.rb
MSP-Greg: `svr_in_proc.rb` is a generic fixture file. See below.

test/helpers/svr_in_proc.rb
MSP-Greg: See below.  After these are in the repo, other test files can be rewritten (and simplified) by using these fixtures.

test/test_rack_env.rb
MSP-Greg: Done

MSP-Greg:
@nateberkopec

#2620 is a new PR with only the fixtures in it.  If you're okay with merging it, I'll remove them from this PR?

MSP-Greg:
Removed the additional test files, added two tests to test_puma_server.rb...

PR #2591, opened 2021-04-06
systemd - fix event firing

### LINKED ISSUE #2572: [phased-restart] add support for systemd type notify
The PR #2438 added support for systemd services of type `notify`.

```ruby
# lib/puma/systemd.rb
module Puma
  class Systemd
    ...
    
    def hook_events
      @events.on_booted { SdNotify.ready }
      @events.on_stopped { SdNotify.stopping }
      @events.on_restart { SdNotify.reloading }
    end
    
    ...
  end
end
```

**The Problem:**  
The systemd status gets updated as expected when performing a `hot restart`.  
But the status stays `active` when performing a `phased restart`.  
In the current implementation, `SdNotify.reloading` is never called during a `phased restart`.

**A Common Scenario:**  
If we perform a `phased restart`, Puma terminates an old worker and starts a new one.
And if Puma is not able to boot up a new worker for some reason, it keeps trying.

In this case, the systemd status shouldn't stay active.  
During a `phased restart` the status should be reloading.

**Additional Context:**  
We monitor the status of all our systemd processes.  
If the status of a process is not active any more (for a certain period of time), we get notified.

It would be awesome to keep this monitoring approach for Puma as well.

MSP-Greg:
### Description

Fixes event firings needed for systemd integration.

Closes #2572.

Note that `@events.fire_on_stopped!` was not called when in cluster mode, this also fixes that.

These tests take some time, I tested locally with two restarts, it passed.

### DIFF
--- a/test/test_integration_systemd.rb
+++ b/test/test_integration_systemd.rb
@@ -33,16 +33,16 @@ class TestIntegrationSystemd < TestIntegration
 
   def test_systemd_notify_usr1_phased_restart_cluster
     skip_unless :fork
-    restart :USR1
+    assert_restarts_with_systemd :USR1
   end
 
   def test_systemd_notify_usr2_hot_restart_cluster
     skip_unless :fork
-    restart :USR2
+    assert_restarts_with_systemd :USR2
   end
 
   def test_systemd_notify_usr2_hot_restart_single
-      restart :USR2, workers: 0
+    assert_restarts_with_systemd :USR2, workers: 0
   end
 
   def test_systemd_watchdog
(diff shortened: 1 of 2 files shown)


MSP-Greg:
Note to self - cluster requires fork... 2nd CI Arjun should pass...

test/test_integration_systemd.rb
```diff
@@ -59,4 +56,28 @@ def test_systemd_watchdog
     stop_server
     assert_match(socket_message, "STOPPING=1")
   end
+
+  private
+
+  def restart(signal, workers: 2)
```
nateberkopec, line 62: Stylistically, I think every test should end with `assert` or, rarely, a cleanup step or two. This method is basically a set of assertions, so I think it would be appropriate to call it `assert_restarts_with_systemd` or similar.

PR #2600, opened 2021-04-17
Refactor drain on shutdown

wjordan:
### Description
This refactoring PR moves the `drain_on_shutdown` implementation out of `#graceful_shutdown` to reduce duplication with the main server accept loop, including a test covering the basic behavior of this option to prevent regressions.

There might be some slight change in timing around when exactly the `:state` event gets fired (before or after connections are all drained), but this doesn't seem important enough to worry about (there was zero test coverage around it).

- dad2708 refactors `TestPumaServer#server_run`, in order to allow arbitrary option arguments to be passed. This made writing the added test with a custom option a bit easier and more readable, and allowed me to clean up a bit of boilerplate in the rest of the test suite.

### DIFF
--- a/test/test_puma_server.rb
+++ b/test/test_puma_server.rb
@@ -22,10 +22,9 @@ class TestPumaServer < Minitest::Test
     @ios.each { |io| io.close if io && !io.closed? }
   end
 
-  def server_run(app: @app, early_hints: false)
-    @server.app = app
+  def server_run(**options, &block)
+    @server = Puma::Server.new block || @app, @events, options
     @port = (@server.add_tcp_listener @host, 0).addr[1]
-    @server.early_hints = true if early_hints
     @server.run
     sleep 0.15 if Puma.jruby?
   end
@@ -54,7 +53,7 @@ class TestPumaServer < Minitest::Test
   end
 
   def test_normalize_host_header_missing
-    server_run app: ->(env) do
+    server_run do |env|
       [200, {}, [env["SERVER_NAME"], "\n", env["SERVER_PORT"]]]
     end
 
@@ -63,7 +62,7 @@ class TestPumaServer < Minitest::Test
   end
 
   def test_normalize_host_header_hostname
-    server_run app: ->(env) do
+    server_run do |env|
       [200, {}, [env["SERVER_NAME"], "\n", env["SERVER_PORT"]]]
     end
 
@@ -75,7 +74,7 @@ class TestPumaServer < Minitest::Test
   end
 
   def test_normalize_host_header_ipv4
-    server_run app: ->(env) do
+    server_run do |env|
       [200, {}, [env["SERVER_NAME"], "\n", env["SERVER_PORT"]]]
     end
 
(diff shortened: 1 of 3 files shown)


test/test_puma_server.rb
```diff
@@ -146,7 +145,7 @@ def test_default_server_port
   end
 
   def test_default_server_port_respects_x_forwarded_proto
-    server_run app: ->(env) do
+    server_run do |env|
```
MSP-Greg, line 148: Missed one (I've done similar)? Maybe:
```ruby
server_run { [200, {}, [env['SERVER_PORT']]] }
```

MSP-Greg (commented):
Any idea about the Ruby 2.2 fails?  Otherwise, LGTM.

MSP-Greg:
@wjordan

Maybe change the test to the following?  Passed in my fork...

```ruby
def test_drain_on_shutdown
  server_run(drain_on_shutdown: true, max_threads: 1) do
    sleep 0.005
    [200, {}, ["DONE"]]
  end
  connections = Array.new(20) { send_http "GET / HTTP/1.0\r\n\r\n" }
  @server.stop
  @thread.join
  connections.each do |s|
    assert_match 'DONE', s.read
  end
end

def test_not_drain_on_shutdown
  server_run(max_threads: 1) do
    sleep 0.005
    [200, {}, ["DONE"]]
  end
  connections = Array.new(20) { send_http "GET / HTTP/1.0\r\n\r\n" }
  @server.stop
  @thread.join
  bad = 0
  connections.each do |s|
    begin
      assert_match 'DONE', s.read
    rescue Errno::ECONNRESET
      bad += 1
    end
  end
  assert_operator bad, :>, 5
end
```

wjordan:
> Any idea about the Ruby 2.2 fails?

Oops! `Queue#close` was added in Ruby 2.3, I had earlier written a simple backport patch (`Puma::QueueClose`) but the simple patch did not un-block already-waiting threads when close was called as expected. I added a few extra lines to the patch which fixes the test on Ruby 2.2. Thanks for taking a look!

I also added a `test_not_drain_on_shutdown` based on your patch, thanks!

MSP-Greg:
Please have a look at the errors in the 2.2 test logs.  I added the `sleep 0.005` as it only adds 100 mS to each test, and pretty much forces the behavior we want, regardless of issues with test runners, etc, etc...  Or, I think the `test_drain_on_shutdown` test in the patch should work in CI.

Something is amiss with my Windows Ruby 2.2 (build system) on the new desktop, I can't build Puma, so I just used CI...

wjordan:
> Please have a look at the errors in the 2.2 test logs.

Oops again- I messed up the `QueueClose#pop` implementation (returns `nil` even if there are items in the closed queue), fixed. (Hope Ruby 2.2 support can be dropped soon...)

While looking at the error log for this, I may have spotted a separate race condition in `Reactor#select_loop`- I'll work on that issue separately, but I'm glad this detour led to that discovery!

PR #2601, opened 2021-04-17
Support for non-blocking fibers

wjordan:
### Description
This PR adds support for [Non-blocking Fibers](https://docs.ruby-lang.org/en/3.0.0/Fiber.html#class-Fiber-label-Non-blocking+Fibers) to the `Server` class.
`Puma::FiberPool` uses non-blocking fibers instead of threads to process connections.
A new `fiber_scheduler` configuration option sets a custom fiber scheduler and enables the use of the FiberPool.
For debugging/testing, use the `SCHEDULER=1` env variable which enables the `libev_scheduler` backend.

Some of the code is still pretty rough around the edges, but it's working well enough for basic benchmarks and passes most of the server tests at this point (some of the force-shutdown-related behavior is not quite working perfectly yet). Sharing this as an early draft PR so anyone else interested can take a look and start testing/experimenting with it!

(Resolves #2517)

### Performance

- `benchmarks/wrk/hello.sh` (Single-process, 4 concurrent keepalive connections):
<details><summary>Before: <strong>7362.54</strong> req/sec</summary>
<pre><code>$ benchmarks/wrk/hello.sh
Puma starting in single mode...
* Puma version: 5.2.2 (ruby 3.0.0-p0) ("Fettisdagsbulle")
*  Min threads: 4
*  Max threads: 4
*  Environment: development
*          PID: 2089795
* Listening on http://0.0.0.0:9292
Use Ctrl-C to stop
Running 30s test @ http://localhost:9292
  2 threads and 4 connections
connection 0: 55213 requests completed
connection 1: 55628 requests completed
connection 0: 55170 requests completed
connection 1: 55606 requests completed
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   554.64us  289.19us   7.26ms   77.99%
    Req/Sec     3.70k   306.21     5.30k    87.87%
  Latency Distribution
     50%  511.00us
     75%  677.00us
     90%    0.90ms
     99%    1.47ms
  221617 requests in 30.10s, 16.06MB read
Requests/sec:   7362.54
Transfer/sec:    546.44KB</code></pre></details>

<details><summary>After: <strong>24432.62</strong> req/sec (3.3x faster)</summary>
<pre><code>$ SCHEDULER=1 benchmarks/wrk/hello.sh
Puma starting in single mode...
* Puma version: 5.2.2 (ruby 3.0.0-p0) ("Fettisdagsbulle")
*  Min threads: 4
*  Max threads: 4
*  Environment: development
[...]

PR #2595, opened 2021-04-09
Improve client response code, chunked bodies

MSP-Greg:
### Description

When clients start writing a request, Puma does several things.

1. Determines which client to process next
2. Read and parse the request
3. Send the request to the app
4. Process the response

Step 3 is trivial, so improving Puma involves optimizing the other steps.  Many of the benchmarks use a minimal response with few headers and a small, single string body.  This drives the time needed for step 4 to zero, and hence measures steps 1 and 2.

This PR looks at the performance of step 4, using more typical responses, especially bodies that are not simple, small string arrays.

What this PR changes:

1. `Puma::IOBuffer` is subclassed from `StringIO`.  The basic API is not changed, and because it is more stream-like than a `String`, it isn't affected by encoding issues (see #2583).  It is now used to assemble both the headers and the body.

2. `Puma::Request` - added `fast_write_body` method.  This is optimized for increasing the speed of writing the request, both for when body is an enum, and also if it is a File object.

3. Current code wrote the headers, then the body.  Updated code writes when the next response addition would result in over 128kB in the response buffer.  This limit can be increased, I just used it as it was the largest chunk that my system seemed to read in the client code...

4. Current code uses `syswrite` for most client socket operations.  After a lot of benchmarking, and reviewing code elsewhere (like Unicorn), I'm not sure if there's a clear case for either `syswrite` or `write`.  The code in the PR has been changed to `write_nonblock`.

5. Several files have been included that allow benchmarking request/response performance.  Scripts for using `test/helpers/sockets.rb` to summarize total response times and 'wrk' are included.  See [`benchmarks/request_reponse_time_benchmarks.md`](https://github.com/puma/puma/blob/88d6566d0b72429e792553bd405f767d34155d8a/benchmarks/request_reponse_time_benchmarks.md) for a doc regarding these additions.

Performance using included files is below.  Some of the differences are insignificant, but the speed increase in 'Chunked Body' - '10kB Body' and especially '100kB Body' are considerable.  The 'Chunked Body' rack app returns a 1kB byte string, so the '100kB Body' has 100 steps in the enumeration.

Data is included for this PR, PR 2597 (`write_nonblock`), and master.  'RPS' is requests per second, the timing info is in mS, timed from the start of the request write to receiving the full body.  Note that when using sockets.rb / create_clients, one can hit a server hard enough that the request/response time will begin to increase and the RPS number will only show small increases.

<details>
[...]

PR #2602, opened 2021-04-17
Defer checking if socket is closed

calvinxiao:
Avoid making `getsockopt` system call for every request

### Description

In function `handle_request`, we can check if the socket is closed when rescuing IOError, saving one `getsockopt` system call for every request.

In the current implementation, the socket may be closed by the client after we check the socket status using `closed_socket?` and before finishing writing the data.

Writing to a closed socket will raise `IOError`. I am not sure which of the following `Error` I should use, my commit uses "Socket timeout writing data"

```
raise ConnectionError, "Connection error detected during write"

raise ConnectionError, "Socket timeout writing data"
```

Analysis detail:

1. Run `hello.ru` benchmark for 30 seconds, 
2. While running benchmark, run `sudo perf stat -e 'syscalls:*' -p $(pidof bundle) sleep 10`, this will use perf to collect all syscalls made by puma process for 10 seconds

perf stat result:

```
 Performance counter stats for process id '1732':

             42009      syscalls:sys_enter_getpeername
             42009      syscalls:sys_exit_getpeername
             42009      syscalls:sys_enter_recvfrom
             42010      syscalls:sys_exit_recvfrom
             84019      syscalls:sys_enter_setsockopt
             84021      syscalls:sys_exit_setsockopt
             42010      syscalls:sys_enter_getsockopt
             42010      syscalls:sys_exit_getsockopt
              8215      syscalls:sys_enter_epoll_ctl
              8215      syscalls:sys_exit_epoll_ctl
              8156      syscalls:sys_enter_epoll_wait
              8156      syscalls:sys_exit_epoll_wait
             37811      syscalls:sys_enter_select
             37811      syscalls:sys_exit_select
                11      syscalls:sys_enter_ppoll
                11      syscalls:sys_exit_ppoll
              6235      syscalls:sys_enter_read
              6235      syscalls:sys_exit_read
             88227      syscalls:sys_enter_write
             88226      syscalls:sys_exit_write
                 1      syscalls:sys_enter_madvise
                 1      syscalls:sys_exit_madvise
                21      syscalls:sys_enter_mprotect
                21      syscalls:sys_exit_mprotect
            274663      syscalls:sys_enter_futex
            274659      syscalls:sys_exit_futex
                 1      syscalls:sys_enter_sched_yield
                 1      syscalls:sys_exit_sched_yield
```

Benchmark RPS is around 4k to 5k, here is my basic understanding of what each syscall is used for:

- `getpeername`, for getting the peer IP address
- `recvfrom`, ?
- `setsockopt`, `cork_socket` and `uncork_socket`, this will be removed in https://github.com/puma/puma/pull/2595
-  `getsockopt`, check socket status to see if the socket is closed
- `epoll_*`, epoll functions used in `nio4r`
- `select`, `IO.select` will get the socket that is ready for reading or writing
- `ppoll`, probably some blocking IO functions
- `read` and `write`, for reading request and writing response data
- `madvise` and `mprotect`, GC stuffs
- `futex`, puma uses threads, `mutex` and `conditional variable`  would cause lot's of `futex` usage, looking forward to getting rid of them(ALL) in https://github.com/puma/puma/pull/2601
- `sched_yield`, process management, https://linux.die.net/man/2/sched_yield

### DIFF
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -159,12 +159,12 @@ module Puma
             fast_write io, CLOSE_CHUNKED
             io.flush
           end
-        rescue SystemCallError, IOError
+        rescue SystemCallError
           raise ConnectionError, "Connection error detected during write"
         end
-      rescue IOError => e
+      rescue IOError
         return false if closed_socket?
-        raise e
+        raise ConnectionError, "Socket timeout writing data"
       ensure
         uncork_socket io
 
@@ -205,7 +205,7 @@ module Puma
           end
 
           retry
-        rescue  Errno::EPIPE, SystemCallError, IOError
+        rescue  Errno::EPIPE, SystemCallError
           raise ConnectionError, "Socket timeout writing data"
         end
 
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -32,8 +32,6 @@ module Puma
       env = client.env
       io  = client.io   # io may be a MiniSSL::Socket
 
-      return false if closed_socket?(io)
-
       normalize_env env, client
 
       env[PUMA_SOCKET] = io
@@ -164,7 +162,9 @@ module Puma
         rescue SystemCallError, IOError
           raise ConnectionError, "Connection error detected during write"
         end
-
+      rescue IOError => e
+        return false if closed_socket?
+        raise e
       ensure
         uncork_socket io
 


calvinxiao:
I didn't find any performance improvement, would be nice to save one system call though.

Just find a slide explaining ruby and system calls, https://speakerdeck.com/franckverrot/rubykaigi-2016-hijacking-syscalls-with-ruby?slide=12 

Also, `getsockopt` is very fast:

```
          getsockopt   119.930k i/100ms
          getsockopt      1.202M (± 3.1%) i/s -      6.116M in   5.094363s
```

calvinxiao:
For the `recvfrom` syscall, it's used in `read_nonblock` to check `IO::WaitReadable` etc.

- https://github.com/ruby/ruby/blob/78f188524f551c97b1a7a44ae13514729f1a21c7/ext/socket/init.c#L341
- https://linux.die.net/man/2/recvfrom

> MSG_DONTWAIT (since Linux 2.2)
Enables nonblocking operation; if the operation would block, the call fails with the error EAGAIN or EWOULDBLOCK (this can also be enabled using the O_NONBLOCK flag with the F_SETFL fcntl(2)).

MSP-Greg:
@calvinxiao

Isn't `return false if closed_socket?(io)` used to prevent calling the app if the client/socket is closed?

calvinxiao:
@MSP-Greg 

Since `closed_socket?(io)` is very fast, calling the app when the socket is closed will cost much more.

I just tested that client can send data then close, the server is able to read the data but cannot write, we should close this PR in case of some DDOS attack.

Learned a lot when perfing puma. 😎

MSP-Greg:
> Learned a lot when perfing puma.

Me also.  While working on #2595, I was reviewing Rails & Rack code, more than I've ever done before.  I started with every optimization I could see, tried all the Rack responses I could think up, sat on the code for a bit, then reverted the 'micro' optimizations that made the code more complex but yielded very small improvements.

> we should close this PR in case of some DDOS attack.

Probably right...  Thanks, Greg

PR #2605, opened 2021-04-20
Improve parsing of HTTP_HOST header

### LINKED ISSUE #2584: Simple method of parsing HTTP_HOST in normalize_env breaks on IPv6 addresses.
**Describe the bug**

Currently, `normalize_env` (in `lib/puma/request.rb` line `231`) splits `HTTP_HOST` into `SERVER_NAME` and `SERVER_PORT` by finding the colon using `String.index` (if there is one.) This works fine for IPv4 addresses, but with IPv6 addresses this results in improperly split values being stored in `SERVER_NAME` and `SERVER_PORT`.

For example, an `HTTP_HOST` value of `[::1]:9292` would result in `SERVER_NAME` being set to `[` and `SERVER_PORT` being set to `:1]:9292`. Further up the stack, `Rack` catches this and panics.

**Puma config:**

A "hello world" `config.ru`:

```ruby
run lambda { |env| [200, {"content-type" => "text/plain"}, [""]] }
```

...then just run `rackup`.

**To Reproduce**

Open a browser and navigate to the application using the IPv6 loopback address ([::1]:9292). I'd create a link, but GitHub seems to choke on the bracketing.

Alternatively, use `curl`:

```sh
curl 'http://[::1]:9292/'
```

I won't dump the entire HTML result in here, but you should receive a backtrace from `Rack` starting with:

```
Rack::Lint::LintError at /
env[SERVER_PORT] is not an Integer
```

**Expected behavior**

The "hello world" application above returns an empty response, so you should see nothing at all. Just a blank page. You can confirm this by replacing `lib/puma/request.rb` line `233`:

```ruby
        if colon = host.index(":")
```

...with:

```ruby
        if colon = host.index(/:\d+$/)
```

 - OS: macOS Big Sur (11.2.3)
 - Puma Version: 5.2.2, though I've confirmed the offending line still exists in HEAD.

pascalbetz:
IPV6 Host was not properly parsed.

With the insanely detailed bug report his was fixed in no time:  👍

Closes #2584

### Description

### DIFF
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -231,7 +231,7 @@ module Puma
     #
     def normalize_env(env, client)
       if host = env[HTTP_HOST]
-        if colon = host.index(/:\d+$/)
+        if colon = host.index(COLON_INDEX_REGEX)
           env[SERVER_NAME] = host[0, colon]
           env[SERVER_PORT] = host[colon+1, host.bytesize]
         else
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -231,7 +231,11 @@ module Puma
     #
     def normalize_env(env, client)
       if host = env[HTTP_HOST]
-        if colon = host.index(COLON_INDEX_REGEX)
+        # host can be a hostname, ipv4 or bracketed ipv6. Followed by an optional port.
+        if colon = host.rindex("]:") # IPV6 with port
+          env[SERVER_NAME] = host[0, colon+1]
+          env[SERVER_PORT] = host[colon+2, host.bytesize]
+        elsif !host.start_with?("[") && colon = host.index(":") # not hostname or IPV4 with port
           env[SERVER_NAME] = host[0, colon]
           env[SERVER_PORT] = host[colon+1, host.bytesize]
         else
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -231,7 +231,7 @@ module Puma
     #
     def normalize_env(env, client)
       if host = env[HTTP_HOST]
-        if colon = host.index(":")
+        if colon = host.index(/:\d+$/)
           env[SERVER_NAME] = host[0, colon]
           env[SERVER_PORT] = host[colon+1, host.bytesize]
         else
--- a/test/test_puma_server.rb
+++ b/test/test_puma_server.rb
@@ -53,6 +53,36 @@ class TestPumaServer < Minitest::Test
     TCPSocket.new(@host, @port).tap {|sock| @ios << sock}
   end
 
+  def test_normalize_host_header_missing
+    server_run app: ->(env) do
+      [200, {}, [env["SERVER_NAME"], "\n", env["SERVER_PORT"]]]
+    end
+
+    data = send_http_and_read "GET / HTTP/1.0\r\n\r\n"
+    assert_equal "localhost\n80", data.split("\r\n").last
+  end
+
+  def test_normalize_host_header_ipv4
+    server_run app: ->(env) do
+      [200, {}, [env["SERVER_NAME"], "\n", env["SERVER_PORT"]]]
+    end
+
+    data = send_http_and_read "GET / HTTP/1.0\r\nHost: 123.123.123.123:456\r\n\r\n"
+    assert_equal "123.123.123.123\n456", data.split("\r\n").last
+  end
+
+  def test_normalize_host_header_ipv6
+    server_run app: ->(env) do
+      [200, {}, [env["SERVER_NAME"], "\n", env["SERVER_PORT"]]]
+    end
+
+    data = send_http_and_read "GET / HTTP/1.0\r\nHost: ::ffff:127.0.0.1:9292\r\n\r\n"
+    assert_equal "::ffff:127.0.0.1\n9292", data.split("\r\n").last
+
+    data = send_http_and_read "GET / HTTP/1.0\r\nHost: ::1:9292\r\n\r\n"
+    assert_equal "::1\n9292", data.split("\r\n").last
+  end
+
   def test_proper_stringio_body
     data = nil
 
--- a/test/test_puma_server.rb
+++ b/test/test_puma_server.rb
@@ -62,6 +62,18 @@ class TestPumaServer < Minitest::Test
     assert_equal "localhost\n80", data.split("\r\n").last
   end
 
+  def test_normalize_host_header_hostname
+    server_run app: ->(env) do
+      [200, {}, [env["SERVER_NAME"], "\n", env["SERVER_PORT"]]]
+    end
+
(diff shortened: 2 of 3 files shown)


lib/puma/request.rb
```diff
@@ -231,7 +231,7 @@ def fetch_status_code(status)
     #
     def normalize_env(env, client)
       if host = env[HTTP_HOST]
-        if colon = host.index(":")
+        if colon = host.index(/:\d+$/)
```
pascalbetz, line 234: Does it make sense to extract this to a constant?

test/test_puma_server.rb
```diff
@@ -53,6 +53,36 @@ def new_connection
     TCPSocket.new(@host, @port).tap {|sock| @ios << sock}
   end
 
+  def test_normalize_host_header_missing
```
pascalbetz, line 56: There is probably a smarter way to check those headers?

lib/puma/request.rb
kuei0221: I have run the benchmark `hello.sh` with your changes, and it seems that extracting to a constant will make it faster. You can give it a try if that makes sense to you as well 😃

No constant
```
❯ benchmarks/wrk/hello.sh
* Puma version: 5.2.2 (ruby 2.7.0-p0) ("Fettisdagsbulle")
*  Min threads: 4
*  Max threads: 4
*  Environment: development
*          PID: 23551
Use Ctrl-C to stop
  2 threads and 4 connections
    Latency   699.53us  301.57us   4.14ms   81.63%
    Req/Sec     2.90k   622.92     3.97k    66.45%
  Latency Distribution
     50%  612.00us
     75%  813.00us
     90%    1.07ms
     99%    1.81ms
  173587 requests in 30.10s, 10.59MB read
Requests/sec:   5766.76
Transfer/sec:    360.42KB
```

Extract to constant
```
❯ benchmarks/wrk/hello.sh
* Puma version: 5.2.2 (ruby 2.7.0-p0) ("Fettisdagsbulle")
*  Min threads: 4
*  Max threads: 4
*  Environment: development
*          PID: 23609
Use Ctrl-C to stop
  2 threads and 4 connections
    Latency   623.64us  271.01us  15.81ms   87.05%
    Req/Sec     3.24k   405.07     4.22k    74.09%
  Latency Distribution
     50%  570.00us
     75%  688.00us
     90%    0.89ms
     99%    1.49ms
  194270 requests in 30.10s, 11.86MB read
Requests/sec:   6454.13
Transfer/sec:    403.38KB
```

pascalbetz: @kuei0221 

Thanks for the benchmark results. I tried to reproduce the results and could not find such a difference between requests/s of inline regexp and constant regexp. There was a small difference between the old (`index(":")`) and new version).

I'll try with different ruby version later today.

kuei0221: You can ignore the benchmark if that doesn't make sense to you ( it defer from machine to machine, and I believe my machine is not very good 😂 )
I think it is a good idea to make it a constant in `const.rb`, so we don't need to allocated this value everytime it is called.
The following comment in that file give us a good reason to do so:
https://github.com/puma/puma/blob/a717b27354011c9f0877e614177b5fba5fc08004/lib/puma/const.rb#L90-L100

MSP-Greg:
A few things:

1.  The comment about '3%' was committed fourteen years ago.
2.  If this is changed to a constant, it could also be placed in the file that calls it.
3.  Using `rindex` instead of `index` would also solve the bug that this is fixing.
4.  Generally, how to handle constants that are only called by one file/class/module is a matter of style.  Often lookup values are treated differently than objects used for code control (branching, conditionals, etc).

pascalbetz:
Thanks for the feedback, I'll look into it.
I just realized that my testdata is faulty (missing brackets for ipv6).

I don't think that `rindex` allone would work though. The port is optional in the _host_ header, so an IPV6 without port would return a faulty index.

MSP-Greg:
> The port is optional in the host header

Sorry, didn't know/check that...

MSP-Greg:
@pascalbetz

Not that familiar with IPv6, couldn't an address without a port also satisfy `/:\d+$/`?

pascalbetz:
@MSP-Greg I think my test data for IPV6 is wrong and the Regexp is faulty.

I need to look up the specs to be sure but IMHO the Host header can be

```
<hostname | IPV4 | ("[" IPV6 "]")>
```
along with optional

```
<colon><port>
```
So IPV6 would be

`[3ffe:2a00:100:7031::1]`

and if there is a port

`[3ffe:2a00:100:7031::1]:1234`

If this is correct, then I can write a better RE or do something around `rindex("]:")`

MSP-Greg:
I'd vote for `rindex ']:'`

pascalbetz:
@MSP-Greg thanks for the feedback. You make it easy for me to contribute.


I added `rindex("]:")` to detect IPV6 with port and added a "not IPV6 check" to the existing logic.

MSP-Greg (approved):

lib/puma/request.rb
```diff
@@ -231,7 +231,11 @@ def fetch_status_code(status)
     #
     def normalize_env(env, client)
       if host = env[HTTP_HOST]
-        if colon = host.index(":")
+        # host can be a hostname, ipv4 or bracketed ipv6. Followed by an optional port.
```
nateberkopec, line 234: Good comments 👍

pascalbetz:
Hey Puma Team

To me it's easy to contribute to puma. Which is hard work on your end. 

Thanks a lot 🤩

nateberkopec:
Thank you for your contribution Pascal!

kaorihinata:
I apologize. I was too busy to create the pull request myself at the time. Had I seen this one I would have actually suggested:

```ruby
if host.include?("]:")
  env[SERVER_NAME], _, env[SERVER_PORT] = host.rpartition(":")
elsif not host.include?("[") and host.include?(":")
  env[SERVER_NAME], env[SERVER_PORT] = host.split(":")
else
  env[SERVER_NAME], env[SERVER_PORT] = host, default_server_port(env)
end
```

No offset juggling makes it look less like an awk script, and it's almost identical in performance. On my system it edges out the new code, but only very slightly (hanging in the 458k to 464k range, instead of the 448k to 461k range), but at that point you're averaging `wrk` runs.

I was also going to note that the `hello.sh` script being hard coded to bind to `http://0.0.0.0:9292` is interpreted on some platforms (for example, macOS) to mean "all IPv4 addresses", and `localhost` is always resolved to the IPv4 loopback address, so `wrk` never hits anything beyond the IPv4 address:port scenario.

pascalbetz:
oh, did not know `rpartition`

kaorihinata:
> oh, did not know `rpartition`

Ruby does have a truly staggering number of methods on some objects.

PR #2606, opened 2021-04-20
Fix timing out requests too early

### LINKED ISSUE #2574: Timeout during long file upload

wjordan:
### Description

Fixes #2574 (a regression released in v5.0.3), with an added test to prevent future regressions.

#### See also

#2575 extends this fix by adding a `between_bytes_timeout` configuration option that allows for different timeout values for the initial data received by the client, and subsequent chunks of data. This extra option may eventually be useful but this PR is intended to more narrowly fix the regression first.

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -295,6 +295,9 @@ module Puma
         @thread_pool << client
       elsif shutdown || client.timeout == 0
         client.timeout!
+      else
+        client.set_timeout(@first_data_timeout)
+        false
       end
     rescue StandardError => e
       client_error(e, client)
--- a/test/test_puma_server.rb
+++ b/test/test_puma_server.rb
@@ -398,6 +398,30 @@ EOF
     test_timeout_in_data_phase
   end
 
+  # https://github.com/puma/puma/issues/2574
+  def test_no_timeout_after_data_received
+    @server.first_data_timeout = 1
+    server_run
+
+    sock = send_http "POST / HTTP/1.1\r\nHost: test.com\r\nContent-Type: text/plain\r\nContent-Length: 11\r\n\r\n"
+    sleep 0.5
+
+    sock << "hello"
+    sleep 0.5
+    sock << "world"
+    sleep 0.5
+    sock << "!"
+
+    data = sock.gets
+
+    assert_equal "HTTP/1.1 200 OK\r\n", data
+  end
+
+  def test_no_timeout_after_data_received_no_queue
+    @server = Puma::Server.new @app, @events, queue_requests: false
+    test_no_timeout_after_data_received
+  end
+
   def test_http_11_keep_alive_with_body
     server_run app: ->(env) { [200, {"Content-Type" => "plain/text"}, ["hello\n"]] }
 
--- a/test/test_puma_server.rb
+++ b/test/test_puma_server.rb
@@ -383,10 +383,12 @@ EOF
   end
 
   def test_timeout_in_data_phase
-    @server.first_data_timeout = 2
+    @server.first_data_timeout = 1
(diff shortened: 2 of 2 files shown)


MSP-Greg:
@wjordan 

LGTM.  Maybe change `test_timeout_in_data_phase` (line 385 in the PR) to the following?  Using `1.1` worked locally, added 0.05 for CI:
```ruby
def test_timeout_in_data_phase
  @server.first_data_timeout = 1
  server_run

  sock = send_http "POST / HTTP/1.1\r\nHost: test.com\r\nContent-Type: text/plain\r\nContent-Length: 5\r\n\r\n"
  sleep 1.15
  sock << "Hello"

  data = sock.gets

  assert_equal "HTTP/1.1 408 Request Timeout\r\n", data
end
```

MSP-Greg:
@wjordan

So the 'Windows' guy (me) now does 95% of his Puma work with WSL2/Ubuntu20.04.  Sorry for that.

Not sure what I think of the 'Errno::ECONNABORTED' being raised so quickly, what's your opinion?

MSP-Greg:
I saw you setting the timing to zero, I'm running the following in my fork right now...
```ruby
def test_timeout_in_data_phase
  @server.first_data_timeout = 1
  server_run

  sock = send_http "POST / HTTP/1.1\r\nHost: test.com\r\nContent-Type: text/plain\r\nContent-Length: 5\r\n\r\n"
  sleep 1.15
  sock << "Hello"

  if Puma::IS_WINDOWS
    assert_raises(Errno::ECONNABORTED) { sock.gets }
  else
    data = sock.gets
    assert_equal "HTTP/1.1 408 Request Timeout\r\n", data
  end
end
```

MSP-Greg:
Looks like both versions are passing.  I also had a nonrelated failure...

MSP-Greg:
I made the change with `assert_raises` since it was the `gets` that was causing the failure, not the `<<`/`write`.

wjordan:
It looks like a Windows `TCPSocket` raises `ECONNABORTED` on the next read, if a socket is written to after being closed at the other end- so skipping the write if the response is already returned (`unless IO.select()`) seems to work and still tests to make sure the timeout is sent in time.

MSP-Greg:
> tests to make sure the timeout is sent in time.

Works for me.

PR #2607, opened 2021-04-21
Handle fatal error that has no backtrace

### LINKED ISSUE #2552: Child workers becoming zombies/unreachable after exception
**Describe the bug**
Every once in a while (it's happening aprox. monthly), we are experiencing downtime due to puma freezing completely.

When I ssh into the server, I see all 25 workers are there, sitting idle with 0 CPU usage.

Investigating the logs, puma.stderr and puma.stdout give me nothing; but in `production.log`, I can always trace back the last request served, and this is what I get in Honeybadger:

```
fatal: machine stack overflow in critical region

Backtrace: middleware_tracing.rb  > 99 >  call(...)

97 | events.notify(:before_call, env) if first_middleware
-- | --
98 |  
99 | result = (target == self) ? traced_call(env) : target.call(env)
100 |  
101 | if first_middleware
```

It seems that this exception _might_ be caused by `WickedPDF` (and `remotipart` together?) - see https://github.com/mileszs/wicked_pdf/issues/891 and https://github.com/mileszs/wicked_pdf/issues/810.

However, despite being a long shot, I'm opening this issue here in Puma because maybe something could be done by Puma to prevent the entire worker pool halting to a stall.

A simple `sudo systemctl restart puma.service` puts us back online.

After the incident, I can never reproduce the problem manually, not even hitting the same URL that triggered the "machine stack overflow in critical region". 

**Puma config:**

**Puma version 5.0.4**

```
[Unit]
Description=Puma {{ param_app_name }} application server
After=network.target

# Socket activation
Requires=puma-{{ param_app_name}}.socket

[Service]
Type=simple

User={{ param_user_name }}

WorkingDirectory=/var/www/{{ param_app_name }}/application

ExecStart=/home/{{ param_user_name }}/.rbenv/shims/bundle exec --keep-file-descriptors puma -C /var/www/{{ param_app_name }}/shared/config/puma.rb

Restart=always

[Install]
WantedBy=multi-user.target
```

```
[Unit]
Description=Puma HTTP Server Accept Sockets for {{ param_app_name }}

[Socket]
SocketUser={{ param_user_name }}
SocketGroup={{ param_user_name }}
ListenStream=/var/www/{{ param_app_name }}/shared/tmp/sockets/puma-{{ param_app_name }}.sock

# Socket options matching Puma defaults
NoDelay=true
ReusePort=true
Backlog=1024

[Install]
WantedBy=sockets.target
```

```
shared_dir = "/var/www/{{ param_app_name }}/shared"

FileUtils.mkdir_p("#{shared_dir}/tmp/pids")

workers 24
threads 1, 1

bind "unix://#{shared_dir}/tmp/sockets/puma-{{ param_app_name }}.sock"

environment "production"

stdout_redirect "#{shared_dir}/log/puma.stdout.log", "#{shared_dir}/log/puma.stderr.log", true

pidfile "#{shared_dir}/tmp/pids/puma.pid"

before_fork do
  ActiveRecord::Base.connection_pool.disconnect! if defined?(ActiveRecord)
end

on_worker_boot do

  if defined?(MessageBus)
    MessageBus.after_fork
  end

end

lowlevel_error_handler do |e|
  Honeybadger.notify(e) if defined?(Honeybadger)
  [500, {}, ["Error\n"]]
end
```

**To Reproduce**
Unfortunately, it happens completely at random. 

Ubuntu 18.04.LTS
Puma version 5.0.4

calvinxiao:
### Description
Handle error that `e.backtrace` is nil.

Current implementation assumes `e.backtrace` is an Array, and will raise `NoMethodError` to the top of thread-pool, this will output an empty response body.

Closes #2552

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -528,7 +528,8 @@ module Puma
       end
 
       if @leak_stack_on_error
-        [status, {}, ["Puma caught this error: #{e.message} (#{e.class})\n#{e.backtrace.join("\n")}"]]
+        backtrace = e.backtrace.nil? ? '<no backtrace available>' :  e.backtrace.join("\n")
+        [status, {}, ["Puma caught this error: #{e.message} (#{e.class})\n#{backtrace}"]]
       else
         [status, {}, ["An unhandled lowlevel error occurred. The application logs may have details.\n"]]
       end
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -528,7 +528,7 @@ module Puma
       end
 
       if @leak_stack_on_error
-        backtrace = e.backtrace.nil? ? '<no backtrace available>' :  e.backtrace.join("\n")
+        backtrace = e.backtrace.nil? ? '<no backtrace available>' : e.backtrace.join("\n")
         [status, {}, ["Puma caught this error: #{e.message} (#{e.class})\n#{backtrace}"]]
       else
         [status, {}, ["An unhandled lowlevel error occurred. The application logs may have details.\n"]]
--- a/test/test_puma_server.rb
+++ b/test/test_puma_server.rb
@@ -285,7 +285,7 @@ EOF
     assert_match(/{}\n$/, data)
   end
 
-  def test_low_level_error_message
+  def test_lowlevel_error_message
     skip_if :windows
     @server = Puma::Server.new @app, @events, {:force_shutdown_after => 2}
 
--- a/test/test_puma_server.rb
+++ b/test/test_puma_server.rb
@@ -285,6 +285,26 @@ EOF
     assert_match(/{}\n$/, data)
   end
 
+  def test_low_level_error_message
+    @server = Puma::Server.new @app, @events
+
+    server_run app: ->(env) do
+      require 'json'
+
+      # will raise fatal: machine stack overflow in critical region
+      obj = {}
+      obj['cycle'] = obj
+      ::JSON.dump(obj)
+    end
+
+    data = send_http_and_read "GET / HTTP/1.0\r\n\r\n"
+
+    assert_match(/HTTP\/1.0 500 Internal Server Error/, data)
+    assert (data.size > 0), "Expected response message to be not empty"
+  end
+
+
+
   def test_force_shutdown_error_default
     @server = Puma::Server.new @app, @events, {:force_shutdown_after => 2}
 
--- a/test/test_puma_server.rb
+++ b/test/test_puma_server.rb
@@ -304,8 +304,6 @@ EOF
     assert (data.size > 0), "Expected response message to be not empty"
   end
 
-
-
   def test_force_shutdown_error_default
     @server = Puma::Server.new @app, @events, {:force_shutdown_after => 2}
 
--- a/test/test_puma_server.rb
+++ b/test/test_puma_server.rb
@@ -286,7 +286,7 @@ EOF
   end
 
   def test_low_level_error_message
-    @server = Puma::Server.new @app, @events
+    @server = Puma::Server.new @app, @events, {:force_shutdown_after => 2}
 
     server_run app: ->(env) do
       require 'json'
--- a/test/test_puma_server.rb
+++ b/test/test_puma_server.rb
@@ -286,6 +286,7 @@ EOF
   end
 
   def test_low_level_error_message
+    skip_if :windows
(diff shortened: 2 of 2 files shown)


calvinxiao:
@MSP-Greg 

1. I added test in `test_puma_server.rb`, is it appropriate?
2. This might need backport.

MSP-Greg:
> I added test in test_puma_server.rb, is it appropriate?

Yes.

Maybe remove the two extra lines after the test?

calvinxiao:
> > I added test in test_puma_server.rb, is it appropriate?
> 
> Yes.
> 
> Maybe remove the two extra lines after the test?

Good catch, pushed a new commit.

nateberkopec:
@calvinxiao Backports are optional on Puma. Mostly we just don't have the time to do them. We do them for security issues but for bugfixes, no one has taken up the job of "backport person". 

That's 100% something you or someone else can do.

nateberkopec:
@calvinxiao Backports are optional on Puma. Mostly we just don't have the time to do them. We do them for security issues but for bugfixes, no one has taken up the job of "backport person". 

That's 100% something you or someone else can do.

lib/puma/server.rb
```diff
@@ -528,7 +528,8 @@ def lowlevel_error(e, env, status=500)
       end
 
       if @leak_stack_on_error
-        [status, {}, ["Puma caught this error: #{e.message} (#{e.class})\n#{e.backtrace.join("\n")}"]]
+        backtrace = e.backtrace.nil? ? '<no backtrace available>' : e.backtrace.join("\n")
```
nateberkopec, line 531: Style nitpick: in this case, the ternary doesn't make a lot of sense. Say `e.backtrace` is `false` (i mean, it can't be, but say that it is): then we're going to call `false.join("\n")`. What you really care about is the ability to `respond_to?` the `join` method, so you should check for that instead.

nateberkopec:
Why is the test windows only?

lib/puma/server.rb
MSP-Greg: Current Ruby docs for [`Exception#backtrace`](https://msp-greg.github.io/ruby_trunk/Core/Exception.html#backtrace-instance_method) state:

'In the case no backtrace has been set, nil is returned'

That seems imply that the method always exists?

MSP-Greg:
> Why is the test marked as skip on Windows?

Appears to be a rather odd Windows bug, not sure about other OS's...

lib/puma/server.rb
nateberkopec: that's why I said it's a style nitpick and not a correctness nitpick 😄 IMO anything passing the test in the ternary should work

MSP-Greg: I guess we could check whether the backtrace returns an array?

Sometimes with stuff like this it may better to keep the scope to what Ruby docs state, so if the Ruby API changes, we **might** see it in tests?  Believe it or not, I've seen things like that happen working with Ruby master and even OpenSSL.  IOW, keep the test very narrow, and edges cases appear.

JFYI, I have no style...

test/test_puma_server.rb
```diff
@@ -285,6 +285,25 @@ def test_force_shutdown_custom_error_message
...
+    skip_if :windows
+    @server = Puma::Server.new @app, @events, {:force_shutdown_after => 2}
+
+    server_run app: ->(env) do
+      require 'json'
+
+      # will raise fatal: machine stack overflow in critical region
+      obj = {}
+      obj['cycle'] = obj
+      ::JSON.dump(obj)
```
eregon, line 298: I think this is not a reliable test for this issue because:
* Not all Ruby implementations might expose no backtrace for a stack overflow (that's pretty unhelpful, isn't it?)
* a stack overflow is a regular exception but given it messes the VM badly you can consider the VM corrupted after it, so maybe stack overflows should become fatal in the future, since it's very rarely safe to rescue it (you would need no `ensure` and no `finally` in the VM, and no second stack overflow while handling the stack overflow very close to the limit, and probably other things which cannot be guaranteed)
* More details on https://github.com/puma/puma/pull/2811

Maybe we can just test creating an exception and using `set_backtrace(nil)`?

dentarg: 👍 @eregon I created https://github.com/puma/puma/issues/2813 with your comments for visibility so anyone can pick this up

PR #2610, opened 2021-04-24
Automatic SSL certificate provisioning for localhost

### LINKED ISSUE #2257: Automatic SSL certificate provisioning for localhost?
https://github.com/socketry/localhost

ye-lin-aung:
### Description
Add automatic SSL certificate provisioning for localhost.

Closes #2257

I don't have much experience with FOSS contribution. This is one of my few contributions. Please let me know if the code doesn't work. I have added test cases as much as I can. I will create another pull request for docs once the commit is merged. I still can't run Rubocop. Its showing 

Error: The `Layout/Tab` cop has been renamed to `Layout/IndentationStyle`.


MSP-Greg:
@ye-lin-aung

Thanks for the contribution/PR.  A couple of suggestions from an old guy.

1. You can enable 'GitHub Actions' in your fork by clicking the 'Settings' tab, then 'Actions', then 'Allow all actions'.  We've got RuboCop locked to an older version for our CI because we're currently testing back to Ruby 2.2.

2. You've done your work in your fork's master branch.  You may find it easier to create new branches for PR work, as rebasing, keeping master clean, etc...

3. Puma works with TCP, UNIX, & SSL sockets.  The methods in Binder are separated for each of those.  You're adding ssl code to the `add_tcp_listener` method, the code should be added to the `add_ssl_listener` method.  Also, MRI Puma doesn't need Ruby's OpenSSL, but does need the OpenSSL library files (normally consider part of the OS).  I think.

In general, I'm probably not the best person to comment about the need for this.  I'm not a hardcore 'crypto' type, but I've worked with OpenSSL for quite a while, contributing here and elsewhere.  I also realize that there was a time when I knew nothing about it.

We've got the certs needed for our CI in the repo, and people can use them for additional testing, local use, etc.  I don't have any issue with adding a gem by @ioquatix.

ioquatix:
Looks great to me well done! Feedback from @MSP-Greg seems on point.

lib/puma/binder.rb
```diff
@@ -290,6 +290,19 @@ def add_tcp_listener(host, port, optimize_for_latency=true, backlog=1024)
 
       host = host[1..-2] if host and host[0..0] == '['
       tcp_server = TCPServer.new(host, port)
+
+      if defined? Localhost::Authority
+        raise "Puma compiled without SSL support" unless HAS_SSL
+        if ENV['RACK_ENV'] == "development" || ENV['RACK_ENV'] == "test"
+          authority = Localhost::Authority.fetch
```
ioquatix, line 297: Try to do this in the top level process (I assume this is) because there are some race conditions which I need to fix - if multiple processes try to get the same certificate at the same time, they may clobber each other - in theory not a big problem (especially for development) but you can simply avoid it by ensuring it's only called once per session and at some point I'll make sure it's more robust.

ioquatix: By the way in several years of using It, I've not even had a single bug report about it. So, it's kind of a non issue in practice.

ioquatix: Okay, I fixed it :) Still do it in the top level process to avoid generating a different certificate per worker if possible :)

MSP-Greg:
@ioquatix

A long time ago, Puma was setup to not use Ruby's OpenSSL.  An example of how we set up the context is:
```ruby
if Puma.jruby?
  ctx.keystore =  "#{CERT_PATH}/keystore.jks"
  ctx.keystore_pass = 'jruby_puma'
else
  ctx.key  = "#{CERT_PATH}/server.key"
  ctx.cert = "#{CERT_PATH}/server.crt"
  ctx.ca   = "#{CERT_PATH}/ca.crt"
end
```

Can that be done with localhost?

JFYI, the current context (`Puma::MiniSSL::Context`) is a ruby placeholder for info about the context, then it's created in minissl.c  to pass along to the server.

ioquatix:
I don't have any opinion about it. What problem are we trying to solve with multiple SSL providers?

ye-lin-aung:
@ioquatix  currently, `localhost` gem doesn't provide to get file path of cert or key via file path. `Puma::MiniSSL` accepts key and certs as file paths. Is it possible for `Localhost::Authority.fetch` to able to respond attributes like cert_path or key_path?  I can help with PR if you are ok.

ioquatix:
I am okay to receive a PR to expose the underlying paths of the certificates. The assumption here is that they will always be on disk, which I don't know if that's always true or not. The ability to create ephemeral certificates is useful too.

ye-lin-aung:
I have made the changes according to @MSP-Greg  suggestions. I have moved the usage of localhost authority inside `tcp_listener` to `ssl_listener`. And I have also make changes to use `Puma::MiniSSL::Context` with localhost authority and removed previous `OpenSSL Context` from commit. I have also moved the call of `Localhost::Authority.fetch` to `Binder` initialization according to @ioquatix  suggestions. Thank you everyone for the feedbacks.

pascalbetz:
Could this feature be explained/mentioned in the README?

README.md
```diff
@@ -187,6 +187,21 @@ Need a bit of security? Use SSL sockets:
 ```
 $ puma -b 'ssl://127.0.0.1:9292?key=path_to_key&cert=path_to_cert'
 ```
+#### Self-signed certificates:
```
pascalbetz, line 190: Thank you!

README.md
```diff
@@ -187,6 +187,21 @@ Need a bit of security? Use SSL sockets:
 ```
 $ puma -b 'ssl://127.0.0.1:9292?key=path_to_key&cert=path_to_cert'
 ```
+#### Self-signed certificates: 
+
+Puma supports [localhost](https://github.com/socketry/localhost) gem for self-signed certificates on non JRuby implementations. To use [localhost](https://github.com/socketry/localhost), you have to `require "localhost/authority"`:
```
nateberkopec, line 192: Is it enough to `gem "localhost", require: "localhost/authority"` in your Gemfile? Rails users are not used to modifying their `config.ru`.

lib/puma/binder.rb
```diff
@@ -273,6 +283,25 @@ def parse(binds, logger, log_msg = 'Listening')
       end
     end
 
+    def localhost_authority
+      @localhost_authority ||= Localhost::Authority.fetch if defined?(Localhost::Authority) && !defined?(JRUBY_VERSION)
+    end
+
+    def localhost_authority_context
+     if !localhost_authority.nil?
+        local_certificates_path = File.expand_path("~/.localhost")
```
nateberkopec, line 292: This should be after the `else`

lib/puma/binder.rb
```diff
@@ -273,6 +283,25 @@ def parse(binds, logger, log_msg = 'Listening')
...
+        if (localhost_authority.respond_to?(:key_path) && localhost_authority.respond_to?(:certificate_path))
+          key_path = localhost_authority.key_path
+          crt_path = localhost_authority.certificate_path
+        else
+          key_path = File.join(local_certificates_path, "localhost.key")
+          crt_path = File.join(local_certificates_path, "localhost.crt")
+        end
+        ctx = MiniSSL::ContextBuilder.new({ "key" => key_path, "cert" => crt_path}, @events).context
+        return ctx
```
nateberkopec, line 301: Unnecessary return (we should turn on that cop...)

lib/puma/binder.rb
```diff
@@ -338,6 +370,8 @@ def add_ssl_listener(host, port, ctx,
 
     def inherit_ssl_listener(fd, ctx)
       raise "Puma compiled without SSL support" unless HAS_SSL
+      # Puma will try to use local authority context if context is supplied nil
+      ctx = ctx || localhost_authority_context
```
nateberkopec, line 374: lets just use `||=`

lib/puma/binder.rb
```diff
@@ -311,6 +341,8 @@ def add_ssl_listener(host, port, ctx,
                          optimize_for_latency=true, backlog=1024)
 
       raise "Puma compiled without SSL support" unless HAS_SSL
+      # Puma will try to use local authority context if context is supplied nil
+      ctx = ctx || localhost_authority_context
```
nateberkopec, line 345: `||=`

lib/puma/request.rb
```diff
@@ -38,7 +38,7 @@ def handle_request(client, lines)
 
       env[PUMA_SOCKET] = io
 
-      if env[HTTPS_KEY] && io.peercert
+      if env[HTTPS_KEY] && io.methods.include?("peer_cert") && io.peercert
```
nateberkopec, line 41: Why is this necessary, and why are you checking for a different method than the one that you're actually calling?

test/test_puma_localhost_authority.rb
```diff
@@ -0,0 +1,168 @@
...
+  # net/http (loaded in helper) does not necessarily load OpenSSL
+  require "openssl" unless Object.const_defined? :OpenSSL
+
+  puts "", RUBY_DESCRIPTION, "RUBYOPT: #{ENV['RUBYOPT']}",
+    "                         OpenSSL",
+    "OPENSSL_LIBRARY_VERSION: #{OpenSSL::OPENSSL_LIBRARY_VERSION}",
+    "        OPENSSL_VERSION: #{OpenSSL::OPENSSL_VERSION}", ""
+end
+
+class TestPumaLocalhostAuthority < Minitest::Test
```
nateberkopec, line 31: Are these tests copied from another file?

I think for this feature we should be testing:

1. If the binder is SSL but there is no SSL cert provided, a self-signed one is generated (and used)
2. The self-signed cert is signed by the local authority
3. That's probably it. We know SSL works in Puma from other tests.

README.md
nateberkopec, line 192: I totally missed it: why doesn't this work on JRuby?

lib/puma/request.rb
ye-lin-aung: Its unnecessary, I will remove it.

lib/puma/binder.rb
ye-lin-aung: @nateberkopec  I am not sure what do you mean by this. Can i know where to put it?

test/test_puma_localhost_authority.rb
```diff
@@ -0,0 +1,117 @@
...
+      self.cert = ssl_socket.peercert
+    end
+  end
+
+
+  # net/http (loaded in helper) does not necessarily load OpenSSL
+  require "openssl" unless Object.const_defined? :OpenSSL
+
+
+  puts "", RUBY_DESCRIPTION, "RUBYOPT: #{ENV['RUBYOPT']}",
```
nateberkopec, line 28: This is not necessary because it's already logged in another SSL test file (can't remember which one)

See double output here: 
https://github.com/puma/puma/pull/2610/checks?check_run_id=2932850256#step:9:25

test/test_puma_localhost_authority.rb
```diff
@@ -0,0 +1,117 @@
...
+    @server = Puma::Server.new app, @events
+    @server.app = app
+
+    @server.add_ssl_listener @host, 0,nil
+
+
+    @http = Net::HTTP.new @host, @server.connected_ports[0]
+    @http.use_ssl = true
+
+    local_authority_key = OpenSSL::PKey::RSA.new File.read(File.join(Localhost::Authority.path,"localhost.key"))
```
nateberkopec, line 94: var is never used. I think you need to create this though, so maybe just don't assign it

test/test_puma_localhost_authority.rb
```diff
@@ -0,0 +1,117 @@
...
+    local_authority_crt = OpenSSL::X509::Certificate.new File.read(File.join(Localhost::Authority.path,"localhost.crt"))
+
+    @http.verify_mode = OpenSSL::SSL::VERIFY_NONE
+    @server.run
+    @cert = nil
+    client_error = false
```
nateberkopec, line 100: Variable never used

test/test_puma_localhost_authority.rb
```diff
@@ -0,0 +1,117 @@
...
+    begin
+      @http.start do
+        req = Net::HTTP::Get.new "/", {}
+        @http.request(req)
+        @cert = @http.peer_cert
+      end
+    rescue OpenSSL::SSL::SSLError, EOFError, Errno::ECONNRESET => e
```
nateberkopec, line 107: Variable never used

ioquatix:
When this is merged, can you please submit a PR to the localhost gem to add a link to puma under "See Also" in the `README.md`.

README.md
```diff
@@ -187,6 +187,21 @@ Need a bit of security? Use SSL sockets:
...
 $ puma -b 'ssl://127.0.0.1:9292?key=path_to_key&cert=path_to_cert'
 ```
+#### Self-signed SSL certificates (via _localhost_ gem, for development use): 
+
+Puma supports [localhost](https://github.com/socketry/localhost) gem for self-signed certificates. This is particularly useful if you want to use Puma with SSL locally, and self-signed certificates will work for your use-case. Currently, `localhost-authority` can be used only in MRI. To use [localhost](https://github.com/socketry/localhost), you have to `require "localhost/authority"`: 
+
+```ruby
+# config.ru
+require './app'
+require 'localhost/authority'
```
nateberkopec, line 197: can you `gem 'localhost', require: 'localhost/authority'`? Does that work?

nateberkopec:
@ye-lin-aung Since this is probably going to be a headline feature in the next release, why don't you take the honor of deciding the [codename](https://github.com/puma/puma/blob/af3675b2ee5485ec4fc62588869c41a5d5e278f7/lib/puma/const.rb#L104) for us?

ye-lin-aung:
@nateberkopec  sure. Thank you for giving me chance. I will let you know this weekend

README.md
bf4: presumably in my app I should *also* put localhost in the `development` group in my Gemfile since the release doesn't depend on the gem... :)

ioquatix:
I released an update to localhost gem so `require: true` should just work without any custom gem config.

Issue #1351, opened 2017-06-29
prune_bundler should optionally allow env variables to be refreshed


stereobooster:
There is `prune_bundler` which is good place for env variables reloading. See `imaginary code` part

```ruby
  def prune_bundler
      return unless defined?(Bundler)
      puma = Bundler.rubygems.loaded_specs("puma")
      dirs = puma.require_paths.map { |x| File.join(puma.full_gem_path, x) }
      puma_lib_dir = dirs.detect { |x| File.exist? File.join(x, '../bin/puma-wild') }

      unless puma_lib_dir
        log "! Unable to prune Bundler environment, continuing"
        return
      end

      deps = puma.runtime_dependencies.map do |d|
        spec = Bundler.rubygems.loaded_specs(d.name)
        "#{d.name}:#{spec.version.to_s}"
      end

      log '* Pruning Bundler environment'
      home = ENV['GEM_HOME']
      Bundler.with_clean_env do
        ######## imaginary code start
        ENV.update get_new_env_variables_callback
        ######## imaginary code end
        ENV['GEM_HOME'] = home
        ENV['PUMA_BUNDLER_PRUNED'] = '1'
        wild = File.expand_path(File.join(puma_lib_dir, "../bin/puma-wild"))
        args = [Gem.ruby, wild, '-I', dirs.join(':'), deps.join(',')] + @original_argv
        # Ruby 2.0+ defaults to true which breaks socket activation
        args += [{:close_others => false}] if RUBY_VERSION >= '2.0'
        Kernel.exec(*args)
      end
    end
```

salimane:
any update on this issue ? we are having the same problem

jeffling:
Hi, Any update here?

nateberkopec:
Hm, we already do this [here](https://github.com/puma/puma/blob/70f309b0260815ce5dec5faacf7e51cc742f0dd5/lib/puma/launcher.rb#L192). I don't understand, is there a problem with this?

nateberkopec:
No response, please open a new issue if you're having an issue with env replacement.

nateberkopec:
Nevermind, I misunderstood. The feature request is that we load NEW env variables when restarting, not replace with the old ones.

nateberkopec:
If anyone is watching this issue, commenting with your use case for this feature would help.

nateberkopec:
I've decided we will not support this.

[...]

PR #2611, opened 2021-04-25
Add option for refreshing env on restart

tlynam:
### Description
Closes https://github.com/puma/puma/issues/1351

Adds option for refreshing the ENV on restart

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ t/
 .rbx/
 Gemfile.lock
 .idea/
+.DS_Store
 vendor/
 /test/test_puma.state
 /test/test_server.sock
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -160,16 +160,7 @@ module Puma
 
     # Run the server. This blocks until the server is stopped
     def run
-      previous_env =
-        if defined?(Bundler)
-          env = Bundler::ORIGINAL_ENV.dup
-          # add -rbundler/setup so we load from Gemfile when restarting
-          bundle = "-rbundler/setup"
-          env["RUBYOPT"] = [env["RUBYOPT"], bundle].join(" ").lstrip unless env["RUBYOPT"].to_s.include?(bundle)
-          env
-        else
-          ENV.to_h
-        end
+      previous_env = original_bundler_env || ENV.to_h
 
       @config.clamp
 
@@ -188,7 +179,7 @@ module Puma
         graceful_stop
       when :restart
         log "* Restarting..."
-        ENV.replace(previous_env)
+        update_env(previous_env)
         @runner.stop_control
         restart!
       when :exit
@@ -538,6 +529,26 @@ module Puma
         .each { |config_key, value| log "- #{config_key}: #{value}" }
 
       log "\n"
+    end
+
+    def original_bundler_env
+      return unless defined?(Bundler)
+
+      env = Bundler::ORIGINAL_ENV.dup
+      # add -rbundler/setup so we load from Gemfile when restarting
+      bundle = "-rbundler/setup"
+      env["RUBYOPT"] = [env["RUBYOPT"], bundle].join(" ").lstrip unless env["RUBYOPT"].to_s.include?(bundle)
+      env
+    end
+
+    def update_env(previous_env)
+      replacement_env = if @options[:reload_env_on_restart]
+        ENV.to_h
+      else
+        previous_env
+      end
+
+      ENV.replace(replacement_env)
     end
   end
 end


lib/puma/launcher.rb
```diff
@@ -539,5 +530,25 @@ def log_config
...
+      env = Bundler::ORIGINAL_ENV.dup
+      # add -rbundler/setup so we load from Gemfile when restarting
+      bundle = "-rbundler/setup"
+      env["RUBYOPT"] = [env["RUBYOPT"], bundle].join(" ").lstrip unless env["RUBYOPT"].to_s.include?(bundle)
+      env
+    end
+
+    def update_env(previous_env)
+      replacement_env = if @options[:reload_env_on_restart]
+        ENV.to_h
```
tlynam, line 546: If I understand correctly, if this option is enabled, we'll fetch from the current `ENV` right?

lib/puma/launcher.rb
```diff
@@ -539,5 +530,25 @@ def log_config
...
+
+      env = Bundler::ORIGINAL_ENV.dup
+      # add -rbundler/setup so we load from Gemfile when restarting
+      bundle = "-rbundler/setup"
+      env["RUBYOPT"] = [env["RUBYOPT"], bundle].join(" ").lstrip unless env["RUBYOPT"].to_s.include?(bundle)
+      env
+    end
+
+    def update_env(previous_env)
+      replacement_env = if @options[:reload_env_on_restart]
```
tlynam, line 545: Is this how the option would look like?  Unsure about the name.  Maybe `refresh_env_on_restart`?

lib/puma/launcher.rb
```diff
@@ -160,16 +160,7 @@ def phased_restart
...
-        if defined?(Bundler)
-          env = Bundler::ORIGINAL_ENV.dup
-          # add -rbundler/setup so we load from Gemfile when restarting
-          bundle = "-rbundler/setup"
-          env["RUBYOPT"] = [env["RUBYOPT"], bundle].join(" ").lstrip unless env["RUBYOPT"].to_s.include?(bundle)
-          env
-        else
-          ENV.to_h
-        end
+      previous_env = original_bundler_env || ENV.to_h
```
tlynam, line 163: Thought we could shorten this method by moving this portion out to a private method

.gitignore
```diff
@@ -13,6 +13,7 @@ t/
 .rbx/
 Gemfile.lock
 .idea/
+.DS_Store
```
tlynam, line 16: I can take this out if you'd like 😄

nateberkopec: Surprised it wasn't already in there 👍

lib/puma/launcher.rb
nateberkopec: Either seems clear enough to me

lib/puma/launcher.rb
nateberkopec: I believe so, unless there's some weirdness around the `ENV` constant itself that I'm not aware of.

nateberkopec:
Seems fine to me, I've approved your CI.

MSP-Greg:
@tlynam

This was submitted when Actions was having Windows issues, and the Windows 2.2 thru 2.4 jobs failed.  I re-ran them, and it's passing now.

One issue with doing so is that the log of the failing run is replaced...

lib/puma/launcher.rb
tlynam: After trying to write an integration test for this, I'm confused on this would work.  It previously had `ENV.replace(previous_env)`.  I'm wondering if this feature would be accomplished by simply not calling `ENV.replace(previous_env)`?  It seems like `ENV` references the current environment variables.  Or would we need to somehow obtain the new environment variables?  Like loading them from another location?

tlynam: I was wondering how reloading ENV's during a hot restart would work in terms of how a user would update the `ENV`.  Could a user utilize `on_restart` to accomplish this and read new ENV's from a file?  Then maybe we don't need any code changes?
```
on_restart do
  ENV["TEST"] = "test2"
end
```

nateberkopec: Good point.

Maybe we would open a new process, have that process spit its `ENV` out so we can read it, then use `ENV.update` to read that all in?

nateberkopec:
Thought about this some more and decided, thanks to your investigation here, that this is not feasible to support. Users who want env variables to change should use rolling restarts.

Sorry for you to yak-shave this @tlynam!

tlynam:
Sure thing, sounds good @nateberkopec!

PR #2614, opened 2021-04-29
[changelog skip] Link to classes in the architecture docs.

dixpac:
Just started reading the architecture docs, for me as a first time
reader it would be useful to have direct link to stated class in the
markdown.

This commit passes through `architecture.md` only, if this is something
that makes sense to you I can pass through all the `docs/`


nateberkopec:
Nice!

PR #2620, opened 2021-05-05
test/helpers - Add fixtures sockets.rb, svr_base.rb, svr_in_proc.rb

MSP-Greg:
### Description

These are new fixtures files, test files will be rewritten to use them in subsequent PR's.

No tests use them, but running CI for RuboCop.

### DIFF
--- a/test/helpers/sockets.rb
+++ b/test/helpers/sockets.rb
@@ -1,0 +1,435 @@
+# frozen_string_literal: true
+
+require 'socket'
+require 'timeout'
+
+module TestPuma
+
+  READ_TIMEOUT = 10
+
+  module SktPrepend
+    def get_body(path = nil, dly: nil, len: nil, timeout: 10)
+      fast_write get_req(path, dly: dly, len: len)
+      read_body timeout
+    end
+
+    def get_response(path = nil, dly: nil, len: nil, timeout: 10)
+      fast_write get_req(path, dly: dly, len: len)
+      read_response timeout
+    end
+
+    def read_body(timeout = 10)
+      response = read_response timeout
+      response.split("\r\n\r\n").last
+    end
+
+    def read_response(timeout = 10)
+      content_length = nil
+      response = ''.dup
+      t_st = Time.now
+      loop do
+        begin
+          chunk = read_nonblock(16_384, exception: true)
+          if chunk.is_a? String
+            response << chunk
+            if (content_length ||= (t = response[/Content-Length: (\d+)/i , 1]) ? t.to_i : nil)
+              return response if response.split("\r\n\r\n", 2).last.bytesize == content_length
+            else
+              return response if response.split("\r\n\r\n", 2).length == 2
+            end
+          end
+        rescue EOFError # for debugging, intermittent
+          raise EOFError
+        rescue ::IO::WaitReadable
+        ensure
+          sleep 0.0002
+          if timeout < Time.now - t_st
+            raise ::Timeout::Error, 'Client Read Timeout'
+          end
+        end
+      end
+    end
+
+    def get(path = nil, dly: nil, len: nil)
+      fast_write get_req(path, dly: dly, len: len)
+    end
+
+    def write(str)
+      fast_write str
+    end
+
+    def <<(str)
+      fast_write str
+    end
+
+    def fast_write(str)
+      n = 0
+
+      while true
+        begin
+          n = syswrite str
+        rescue Errno::EAGAIN, Errno::EWOULDBLOCK => e
+          raise e unless IO.select(nil, [io], nil, 5)
+          retry
+        rescue Errno::EPIPE, SystemCallError, IOError => e
+          raise e
+        end
+
+        return if n == str.bytesize
+        str = str.byteslice(n..-1)
+      end
+    end
+
+    private
+
+    def get_req(path = nil, dly: nil, len: nil)
+      req = "GET /#{path} HTTP/1.1\r\n".dup
+      req << "Dly: #{dly}\r\n" if dly
+      req << "Len: #{len}\r\n" if len
+      req << "\r\n"
+    end
+  end
+
+  class SktTCP < ::TCPSocket
+    prepend SktPrepend
+  end
+
+  if Object.const_defined? :UNIXSocket
+    class SktUnix < ::UNIXSocket
+      prepend SktPrepend
+    end
+  end
+
+  if !Object.const_defined?(:Puma) || ::Puma.ssl?
+    require 'openssl'
+    class SktSSL < ::OpenSSL::SSL::SSLSocket
+      prepend SktPrepend
+    end
+  end
+
+  module Sockets
+
+    IS_JRUBY = Object.const_defined? :JRUBY_VERSION
+
+    IS_OSX = RUBY_PLATFORM.include? 'darwin'
+
+    IS_WINDOWS = !!(RUBY_PLATFORM =~ /mswin|ming|cygwin/ ||
+      IS_JRUBY && RUBY_DESCRIPTION.include?('mswin'))
+
+    IS_MRI = (RUBY_ENGINE == 'ruby' || RUBY_ENGINE.nil?)
+
+    HOST = TestPuma.const_defined?(:SvrBase) ? SvrBase::HOST : '127.0.0.1'
+
+    # OpenSSL::SSL::SSLError is intermittently raised on SSL connect?
+    # IOError is intermittent on all platforms (except maybe Windows), seems
+    # to be only be raised on SSL connect?
+    # macOS sometimes raises Errno::EBADF for socket init error
+    #
+    OPEN_WRITE_ERRORS = begin
+      ary = [Errno::ECONNREFUSED, Errno::ECONNRESET, Errno::ENOENT,
+        Errno::EPIPE, Errno::EBADF, IOError]
+      ary << Errno::ENOTCONN if IS_OSX
+      ary << OpenSSL::SSL::SSLError if Object.const_defined?(:OpenSSL)
+
+      ary.freeze
+    end
+
+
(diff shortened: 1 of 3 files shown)


test/helpers/sockets.rb
```diff
@@ -0,0 +1,435 @@
+# frozen_string_literal: true
+
+require 'socket'
+require 'timeout'
+
+module TestPuma
```
nateberkopec, line 6: I'd prefer `Puma::Test` unless you have a namespace-lookup related reason to do this

test/helpers/sockets.rb
```diff
@@ -0,0 +1,435 @@
...
+
+  READ_TIMEOUT = 10
+
+  module SktPrepend
+    def get_body(path = nil, dly: nil, len: nil, timeout: 10)
```
nateberkopec, line 11: No need for abbreviated variable names, lets spell out `delay` and `length`

test/helpers/sockets.rb
```diff
@@ -0,0 +1,435 @@
...
+
+    def read_body(timeout = 10)
+      response = read_response timeout
+      response.split("\r\n\r\n").last
+    end
+
+    def read_response(timeout = 10)
+      content_length = nil
+      response = ''.dup
+      t_st = Time.now
```
nateberkopec, line 29: ditto naming here. maybe `time_start`

test/helpers/sockets.rb
```diff
@@ -0,0 +1,435 @@
...
+      loop do
+        begin
+          chunk = read_nonblock(16_384, exception: true)
```
nateberkopec, line 32: let's pull the magic number out here into a constant (maybe `Puma::Test::SOME_CONST`) and name it appropriately

test/helpers/sockets.rb
```diff
@@ -0,0 +1,435 @@
...
+          if chunk.is_a? String
+            response << chunk
+            if (content_length ||= (t = response[/Content-Length: (\d+)/i , 1]) ? t.to_i : nil)
+              return response if response.split("\r\n\r\n", 2).last.bytesize == content_length
```
nateberkopec, line 36: Maybe we should pull out the double CRLF into a constant too? `DOUBLE_CRLF`

test/helpers/sockets.rb
```diff
@@ -0,0 +1,435 @@
...
+          n = syswrite str
+        rescue Errno::EAGAIN, Errno::EWOULDBLOCK => e
+          raise e unless IO.select(nil, [io], nil, 5)
+          retry
+        rescue Errno::EPIPE, SystemCallError, IOError => e
+          raise e
+        end
+
+        return if n == str.bytesize
+        str = str.byteslice(n..-1)
```
nateberkopec, line 79: str -> string in all cases

test/helpers/sockets.rb
```diff
@@ -0,0 +1,435 @@
...
+
+    def get_req(path = nil, dly: nil, len: nil)
+      req = "GET /#{path} HTTP/1.1\r\n".dup
+      req << "Dly: #{dly}\r\n" if dly
+      req << "Len: #{len}\r\n" if len
+      req << "\r\n"
+    end
+  end
+
+  class SktTCP < ::TCPSocket
```
nateberkopec, line 93: Hmm... it would definitely be confusing if we just called it `TCPSocket`...

Maybe `ExtendedTCPSocket`, `ExtendedUnixSocket` et cetera

test/helpers/sockets.rb
```diff
@@ -0,0 +1,435 @@
...
+  if !Object.const_defined?(:Puma) || ::Puma.ssl?
+    require 'openssl'
+    class SktSSL < ::OpenSSL::SSL::SSLSocket
+      prepend SktPrepend
+    end
+  end
+
+  module Sockets
+
+    IS_JRUBY = Object.const_defined? :JRUBY_VERSION
```
nateberkopec, line 112: Why does this have to be duped from `detect`?

test/helpers/sockets.rb
```diff
@@ -0,0 +1,435 @@
...
+    def fast_connect_raw(str = nil, type: nil, p: nil)
+      s = fast_connect type: type, p: p
+      s.fast_write str
+      s
+    end
+
+    # Used to open a socket.  Normally, the connection info is supplied by calling `bind_type`,
+    # but that can be set manually if needed for control sockets, etc.
+    # @param type: [Symbol] the type of connection, eg, :tcp, :ssl, :aunix, :unix
+    # @param p: [String, Integer] the port or path of the connection
```
nateberkopec, line 167: `p` -> `file_descriptor` perhaps?

nateberkopec: maybe even just `fd`, we do have precedent for using `fd` elsewhere

nateberkopec (changes requested):
Only got a third of the way through today but figured leaving my comments here would be good.

Overall, short variable names 🚫 and abbreviations in constant names 🚫 . The stuff you're doing to iron out platform differences here obviously is ✅ ✅

test/helpers/sockets.rb
MSP-Greg: Actually, it's both a namespace issue and a doc issue.  If we name it Puma::Test, all the test items will appear under `Puma` (class/module names will be shown as constants, etc).

`TestPuma` puts them outside of Puma, and an alpha list of classes (see https://msp-greg.github.io/puma/) puts all the test classes/modules below everything in the current list...

test/helpers/sockets.rb
MSP-Greg: Actually that is either the TCP/SSL port or the UNIXSocket path.  Not sure what to call it, but the doc/comments could be better.

test/helpers/sockets.rb
nateberkopec: Does Detect require that puma is loaded? Maybe we should change that...

test/helpers/sockets.rb
nateberkopec: maybe `socket`, because that's what both of those things are then

test/helpers/sockets.rb
```diff
@@ -0,0 +1,435 @@
...
+    #
+    # The response body must contain 'Hello World' on a line.
+    #
+    # ### Write Errors
+    # * Ubuntu & macOS
+    #   * tcp: Errno::ECONNREFUSED
+    #   * unix: Errno::EPIPE, Errno::ENOENT
+    #
+    # rubocop:disable Metrics/ParameterLists
+    def create_clients(replies, threads, clients_per_thread,
```
nateberkopec, line 209: I think all the methods in this class need clear docs on what they return

test/helpers/svr_in_proc.rb
```diff
@@ -0,0 +1,140 @@
...
+
+module TestPuma
+  class SvrInProc < SvrBase
+    include TmpPath
+    include TestPuma::Sockets
+
+    def teardown
+      return if skipped?
+      if @server
+        @server.stop true
```
nateberkopec, line 20: would prefer a named option here

test/helpers/svr_in_proc.rb
```diff
@@ -0,0 +1,140 @@
...
+
+      if @ios_to_close
+        @ios_to_close.each do |io|
+          io.close if io.is_a?(IO) && !io.closed?
+          io = nil
+        end
+      end
+
+      if defined?(@bind_port) && @bind_port
+        # ssl sockets need extra time to close?
```
nateberkopec, line 32: Oof, that's a long sleep...

test/helpers/svr_in_proc.rb
```diff
@@ -0,0 +1,140 @@
...
+      @server.inherit_binder Puma::Binder.new(@events, @conf)
+      @server.binder.parse @conf.options[:binds], @events
+      @server.run
+      sleep 0.01
+    end
+
+    def rack_env_to_body
+      lambda { |env|
+        body = ''.dup
+        env.sort.each { |a| body << "#{a.first}: #{a[1]}\n" }
```
nateberkopec, line 104: `env.sort.reduce("")`

test/helpers/sockets.rb
MSP-Greg: Maybe `port_path`.  Some might think `socket` is a socket instance?

test/helpers/sockets.rb
MSP-Greg: No.  It defines constants and predicate methods in the `Puma` namespace, but it doesn't require `Puma` (puma.rb) to be loaded.

test/helpers/svr_in_proc.rb
MSP-Greg: Let me check that....

MSP-Greg: I checked this, and can't find a reason for it.  But, there aren't any tests using `svr_in_proc.rb` with ssl connections.

test/helpers/sockets.rb
```diff
@@ -0,0 +1,435 @@
+# frozen_string_literal: true
+
+require 'socket'
+require 'timeout'
+
+module TestPuma
+
+  READ_TIMEOUT = 10
+
+  module SktPrepend
```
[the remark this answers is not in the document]
dentarg: By now I've learned that "skt" means "socket" but I think we should spell out "socket" 

IMHO all but very common abbreviations can be confusing and takes some extra thinking when reading source code.

dentarg:
@MSP-Greg should we close this PR in favour of https://github.com/puma/puma/pull/2694?

MSP-Greg:
@dentarg Thanks for the reminder.  Yes, I'll look at the other PR's and either close them or change to drafts.  I'll start work on them again, reviewing/changing naming, comments, etc...

PR #2640, opened 2021-06-08
Rename the `Puma::JSON` module to `Puma::JSONSerialization`

### LINKED ISSUE #2639: Introduction of `Puma::JSON` in 5.1.0 is "breaking" in some cases
**Describe the bug**

In #2479, a new class `Puma::JSON` was introduced (later on shipped with 5.1.0). It has an innocent breaking change when it comes to the `JSON` constant lookup.

**Puma config:**
```ruby
require "json"

workers 2

before_fork do
  JSON.dump({}) # or any method not implementation on `Puma::JSON`
end
```

**To Reproduce**

Output from `bundle exec puma -C puma.rb hello.ru` using Puma >= 5.1.0, note the `WARNING`:

```
[37721] Puma starting in cluster mode...
[37721] * Puma version: 5.1.0 (ruby 3.0.1-p64) ("At Your Service")
[37721] *  Min threads: 0
[37721] *  Max threads: 5
[37721] *  Environment: development
[37721] *   Master PID: 37721
[37721] *      Workers: 2
[37721] *     Restarts: (✔) hot (✔) phased
[37721] * Listening on http://0.0.0.0:9292
[37721] Use Ctrl-C to stop
[37721] WARNING hook before_fork failed with exception (NoMethodError) undefined method `dump' for Puma::JSON:Module
Did you mean?  dup
[37721] - Worker 0 (PID: 37722) booted, phase: 0
[37721] - Worker 1 (PID: 37723) booted, phase: 0
```

The problem is the same for any "`Puma::DSL` hook", in our case we actually ran into this for the `lowlevel_error_handler` hook.

**Expected behavior**

While it's easy to fix the problem by simply changing to top-level lookup in the client-code (`::JSON`) it was a bit confusing at first. Ideally, it would be cool if clients do not need to change their implementations, but I understand that probably means the class has to be renamed or moved. Maybe a note in the release notes could be good though?

 - OS: macOS 10.15.7
 - Puma Version: >= 5.1.0

walro:
### Description

For the hooks defined in `Puma::DSL`, the previous naming had the unfortunate side-effect of causing `JSON` to resolve to `Puma::JSON` rather than, the likely more expected, `::JSON` module.

Closes #2639

### DIFF
--- a/lib/puma.rb
+++ b/lib/puma.rb
@@ -12,7 +12,7 @@ require 'thread'
 
 require 'puma/puma_http11'
 require 'puma/detect'
-require 'puma/json'
+require 'puma/json_serialization'
 
 module Puma
   autoload :Const, 'puma/const'
@@ -60,7 +60,7 @@ module Puma
 
   # @!attribute [rw] stats_object
   def self.stats
-    Puma::JSON.generate @get_stats.stats
+    Puma::JSONSerialization.generate @get_stats.stats
   end
 
   # @!attribute [r] stats_hash
--- a/lib/puma/app/status.rb
+++ b/lib/puma/app/status.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
-require 'puma/json'
+require 'puma/json_serialization'
 
 module Puma
   module App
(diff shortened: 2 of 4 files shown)


walro:
Regarding the fails on GHA:

* `non_MRI / ubuntu-20.04 truffleruby-head` – seems to fail on master too
* `MRI / macos-10.15 2.3 (pull_request)` – not sure why it failed here, it [ran successfully](https://github.com/walro/puma/runs/2775660852?check_suite_focus=true) on my fork. Flaky tests? Hopefully the rename didn't trigger flakiness though :)

test/test_json.rb
```diff
@@ -1,6 +1,6 @@
 require_relative "helper"
 require "json"
-require "puma/json"
+require "puma/json_serialization"
 
 class TestJSON < Minitest::Test
```
dentarg, line 5:
```suggestion
class TestJSONSerialization < Minitest::Test
```

Maybe rename the file too?

walro: Ah yes, thanks!

walro: New version up -> cf697af

nateberkopec:
Thanks for the fix @walro!

PR #2653, opened 2021-06-30
improve remote_addr behavior

### LINKED ISSUE #2652: When remote_addr_header is used, but not passed, remote_addr defaults to localhost
**Describe the bug**
When the `header` option is passed to `set_remote_address` but the requisite header is not passed on a request, the request is attributed to localhost instead of to the actual peeraddr.

**Puma config:**

```
set_remote_address header: "X-Client-IP"
```

**To Reproduce**
Make an HTTP request from non-localhost without setting the `X-Client-IP`. Note that the Rack `REMOTE_ADDR` variable is set to `127.0.0.1` instead of your actual IP.

**Expected behavior**
When the header is not set, requests should have `REMOTE_ADDR` set to the actual remote addr. If it's not available, it is arguably more correct to use 0.0.0.0 rather than 127.0.0.1 but I'm less picky about that case.

 - OS: Linux
 - Puma Version: 5.3.2

Roguelazer:
Closes #2652

### Description
See #2652 for details; this makes two changes:

 - When using the `set_remote_address header: "header_name"` functionality, if the header is not passed, `REMOTE_ADDR` is now set to the physical peeraddr instead of always being set to 127.0.0.1
 - When an error occurs preventing the physical peeraddr from being fetched, `REMOTE_ADDR` is now set to the unspecified source address ('0.0.0.0') instead of to '127.0.0.1'

### DIFF
--- a/lib/puma/const.rb
+++ b/lib/puma/const.rb
@@ -176,6 +176,7 @@ module Puma
     PORT_443 = "443".freeze
     LOCALHOST = "localhost".freeze
     LOCALHOST_IP = "127.0.0.1".freeze
+    UNSPECIFIED_IP = "0.0.0.0".freeze
 
     SERVER_PROTOCOL = "SERVER_PROTOCOL".freeze
     HTTP_11 = "HTTP/1.1".freeze
--- a/lib/puma/const.rb
+++ b/lib/puma/const.rb
@@ -175,8 +175,10 @@ module Puma
     PORT_80 = "80".freeze
     PORT_443 = "443".freeze
     LOCALHOST = "localhost".freeze
-    LOCALHOST_IP = "127.0.0.1".freeze
-    UNSPECIFIED_IP = "0.0.0.0".freeze
+    LOCALHOST_IPV4 = "127.0.0.1".freeze
+    LOCALHOST_IPV6 = "::1".freeze
+    UNSPECIFIED_IPV4 = "0.0.0.0".freeze
+    UNSPECIFIED_IPV6 = "::".freeze
 
     SERVER_PROTOCOL = "SERVER_PROTOCOL".freeze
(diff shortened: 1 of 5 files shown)


lib/puma/const.rb
```diff
@@ -176,6 +176,7 @@ module Const
     PORT_443 = "443".freeze
     LOCALHOST = "localhost".freeze
     LOCALHOST_IP = "127.0.0.1".freeze
+    UNSPECIFIED_IP = "0.0.0.0".freeze
```
nateberkopec, line 179: I wonder if we might have trouble here with ipv6 setups.

nateberkopec:
Had a think about it and I think both of these changes are an improvement. However, they both have the potential to break people's setups, so I think we need to park these until 6.0.

Your other option is to put one or both changes behind a temporary config option that we will remove in 6.0.

I don't have a lot of insight right now on whether or not the next release will be a minor or major, so no promises on how long this could stay parked.

Had one small concern about ipv6, see comment.

lib/puma/const.rb
Roguelazer: I guess we could ask the socket for its type and then return `0.0.0.0` or `::` depending? I can make that change if you'd like.

nateberkopec: @Roguelazer Please do - I think it makes sense.

nateberkopec:
Removed waiting for review. This is only blocked on next major release.

nateberkopec:
I'm thinking next release might be a major.

nateberkopec:
Merged!

dentarg:
Why not squash merging @nateberkopec ?

nateberkopec:
Did it on the CLI and forgot 😆

PR #2654, opened 2021-06-30
add support for the PROXY protocol (v1 only)

### LINKED ISSUE #2651: PROXY Protocol
Many HTTP servers speak the [PROXY Protocol](https://haproxy.1wt.eu/download/1.5/doc/proxy-protocol.txt) to easily forward remote connection information below layer 7. It does not appear that Puma has such support.

It would be great if puma supported the PROXY protocol as another option of the `set_remote_address` DSL. When this is set, Puma could attempt to read the PROXY protocol (versions 1 and/or 2) from the socket prior to running the HTTP parser, and, if it successfully read a PROXY header, would use that to set the remote_addr before calling into the application.

Current main alternatives:

   - Use `X-Forwarded-For` and have Rack handle it in the request middleware
   - Use `X-Remote-IP` or somesuch and the `remote_address_header` option
   - Use a different HTTP server (Passenger + Apache, uWSGI)

Roguelazer:
### Description
This adds support for parsing the [PROXY Protocol](http://www.haproxy.org/download/1.8/doc/proxy-protocol.txt) (version 1 only). It's behind the `set_remote_address` feature when invoked like `set_remote_address proxy_protocol: :v1`. I put the `v1` in there in case some bold soul wants to add v2 support at some point.

This gracefully handles the case where the client doesn't send the PROXY protocol and falls back to normal logic.

Closes #2651 

### DIFF
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -828,7 +828,10 @@ module Puma
     #    `set_remote_address header: "X-Real-IP"`.
     #    Only the first word (as separated by spaces or comma) is used, allowing
     #    headers such as X-Forwarded-For to be used as well.
-    # 4. **\<Any string\>** - this allows you to hardcode remote address to any value
+    # 4. **proxy_protocol: v1**- set the remote address to the value read from the
+    #    HAproxy PROXY protocol, version 1. If the request does not have the PROXY
+    #    protocol attached to it, will fall back to :socket
+    # 5. **\<Any string\>** - this allows you to hardcode remote address to any value
     #    you wish. Because Puma never uses this field anyway, it's format is
     #    entirely in your hands.
     #
@@ -846,6 +849,13 @@ module Puma
         if hdr = val[:header]
           @options[:remote_address] = :header
           @options[:remote_address_header] = "HTTP_" + hdr.upcase.tr("-", "_")
+        elsif protocol_version = val[:proxy_protocol]
+          @options[:remote_address] = :proxy_protocol
+          protocol_version = protocol_version.downcase.to_sym
+          if ! [:v1].include?(protocol_version)
+            raise "Invalid value for proxy_protocol - #{protocol_version.inspect}"
+          end
+          @options[:remote_address_proxy_protocol] = protocol_version
         else
           raise "Invalid value for set_remote_address - #{val.inspect}"
         end
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -818,7 +818,7 @@ module Puma
     # a kernel syscall is required which for very fast rack handlers
     # slows down the handling significantly.
     #
-    # There are 4 possible values:
+    # There are 5 possible values:
     #
     # 1. **:socket** (the default) - read the peername from the socket using the
     #    syscall. This is the normal behavior.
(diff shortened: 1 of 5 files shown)


lib/puma/dsl.rb
```diff
@@ -846,6 +849,13 @@ def set_remote_address(val=:socket)
         if hdr = val[:header]
           @options[:remote_address] = :header
           @options[:remote_address_header] = "HTTP_" + hdr.upcase.tr("-", "_")
+        elsif protocol_version = val[:proxy_protocol]
+          @options[:remote_address] = :proxy_protocol
+          protocol_version = protocol_version.downcase.to_sym
+          if ! [:v1].include?(protocol_version)
```
dentarg, line 855:
```suggestion
          unless [:v1].include?(protocol_version)
```

lib/puma/dsl.rb
```diff
@@ -828,7 +828,10 @@ def wait_for_less_busy_worker(val=0.005)
     #    `set_remote_address header: "X-Real-IP"`.
     #    Only the first word (as separated by spaces or comma) is used, allowing
     #    headers such as X-Forwarded-For to be used as well.
-    # 4. **\<Any string\>** - this allows you to hardcode remote address to any value
+    # 4. **proxy_protocol: v1**- set the remote address to the value read from the
```
dentarg, line 831: Should this be

```suggestion
    # 4. **proxy_protocol: :v1**- set the remote address to the value read from the
```

?

dentarg (commented):
This PR makes it 5

https://github.com/puma/puma/blob/8264d20caeb73d65564c5def9267fecfc9473c16/lib/puma/dsl.rb#L821

lib/puma/dsl.rb
Roguelazer: It gets `.to_sym` called on it, so :v1 and "v1" both work, but I can recommend the sym form if you'd prefer!

Roguelazer:
I don't think any of the CI failures are related to my commit...

dentarg:
> I don't think any of the CI failures are related to my commit...

I started a re-run of CI

Roguelazer:
I found a bug in this in my own testing: on about 0.0025% of requests, parsing fails with `Puma::HttpParserError: Requested start is after data buffer end.`

It turns out that this happens when the initial read on the socket returned the PROXY protocol header and _only_ the PROXY protocol header (which depends on TCP_CORK and a bunch of other stuff on the client side, but is easily reproducible by hand). This is because if you call `@parser.execute` with a 0-byte buffer, it fails.

I'm going to push an update to this branch that ensures we don't do that.

Roguelazer:
This is good to be reviewed, for what it's worth

nateberkopec:
Just commenting to say I haven't forgotten about this. This is up for review by anyone other than the original author (we allow code review from anyone as we discuss in [CONTRIBUTING.md](https://github.com/puma/puma/blob/master/CONTRIBUTING.md))

nateberkopec:
Putting out to the puma-contrib channel that I want one more review on this but I think this would be a great addition for the next release

lib/puma/server.rb
```diff
@@ -321,6 +321,8 @@ def handle_servers
           remote_addr_value = @options[:remote_address_value]
         when :header
           remote_addr_header = @options[:remote_address_header]
+        when :proxy_protocol
+          remote_addr_proxy_protocol = @options[:remote_address_proxy_protocol]
```
dentarg, line 325: Do we need to define `remote_addr_proxy_protocol` outside of the case block? 

That's done for `remote_addr_value`, `remote_addr_header` (lines 316, 317)

dentarg: Does not seem like we need to:

```
$ ruby -e 'case 1;when 2;foo="foo";end;p foo'
nil
```

(I think lines 316, 317 can be removed, but that could/should be done in another PR)

dentarg (approved):
Looks good to me! I tried it out with this HAProxy config:

```haproxy
defaults
    timeout connect 10s
    timeout client 30s
    timeout server 30s
    # send everything to stdout
    log stdout format raw daemon

frontend www
    bind *:4444
    use_backend proxy

frontend www2
    bind *:5555
    use_backend noproxy

backend proxy
    mode http
    server server1 127.0.0.1:9292 send-proxy

backend noproxy
    mode tcp
    server server1 127.0.0.1:9292
```

nateberkopec:
Thanks for the review @dentarg!

PR #2220, opened 2020-04-10
ThreadPool concurrency refactoring

wjordan:
### Description
This PR refactors some tricky concurrency parts of `TestThreadPool`, in order to make the unit tests more stable (all tests pass on jruby), faster (no more 1-second sleeps), and a bit more readable. Many of the tests now use a subclass of `ThreadPool` that overrides `#<<` to wait on the existing `@not_full` condition variable, to ensure work begins executing on a worker thread before continuing the test.

There are also some small changes to the `ThreadPool` class itself:
- Add `#with_mutex` to make it easier to chain multiple `@mutex.synchronize`-wrapped methods together, and to extend existing mutex-wrapped methods in unit tests.
- Wait for initially-spawned threads to enter the wait loop before returning from `#initialize`. This fixes a subtle edge-case concurrency bug mostly affecting unit-test reliability, demonstrated by the following simple test that fails on `master`: https://github.com/puma/puma/blob/8d610275f6ca0c6094e75e93a3529ea61732fd57/test/test_thread_pool.rb#L239-L242
- Fix a concurrency bug in `#trim` where a trim could still be requested even when enough work to utilize all threads has been queued but not yet picked up by workers (e.g., `waiting > 0` but `waiting - todo.size == 0`). (I believe this was the cause of some jruby test flakiness in some `trim` unit tests.)
- Refactor the main wait-loop in `spawn_thread` to make it a bit simpler and easier to read and reason about.

### Your checklist for this pull request

- [x] I have reviewed the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.
- [x] I have added an entry to [History.md](../blob/master/History.md) if this PR fixes a bug or adds a feature. If it doesn't need an entry to HISTORY.md, I have added `[changelog skip]` the pull request title.
- [x] I have added appropriate tests if this PR fixes a bug or adds a feature.
- [x] My pull request is 100 lines added/removed or less so that it can be easily reviewed.
- [x] If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.
- [ ] If this closes any issues, I have added "Closes `#issue`" to the PR description or my commit messages.
- [ ] I have updated the documentation accordingly.
- [ ] All new and existing tests passed, including Rubocop.

PR #2656, opened 2021-07-02
Fix deadlock issue in thread pool

### LINKED ISSUE #2655: Thread pool deadlock
**Describe the bug**
The Puma worker thread pool can enter a deadlock state and become unable to process new requests.

**Puma config:**

In our case, the issue only happens on the control server worker and not the main server worker, so I assume most of the config is irrelevant since the configuration for the control server is mostly hardcoded. We have tried using both Unix sockets and TCP sockets for the control server and have observed the issue in both cases.

**To Reproduce**
Unfortunately I have not yet been able to produce a minimal example to reproduce the issue. The best I can do for now is describe our setup. (I will update this post if I can come up with a reliable way to reproduce the issue.)

(EDIT: I have been able to reproduce the issue with a minimal Puma setup, cf. https://github.com/puma/puma/issues/2655#issuecomment-872524754)

The deadlock issue only happens in the worker for the Puma control server, not the main server. We have a monitoring script that sends a `/stats` request to the Puma control server every 2 seconds. We use the response to aggregate statistics to monitor the overall availability of our services across multiple hosts.

The client side timeout is fairly short (0.5 seconds). Occasionally a request will time out (presumably because the Puma Ruby process is busy doing GC or something else and does not yield to the control server thread in time). That's not an issue, since the next request will likely succeed. However, sometimes the control server will become completely stuck and unable to process any further requests, which causes us to report the worker as unavailable (despite the fact that the main server thread is still running fine).

Using gdb to attach to a Puma process in this deadlock state, we were able to determine that the control server thread is stuck on this line: https://github.com/puma/puma/blob/v5.3.1/lib/puma/thread_pool.rb#L236.

As far as we can tell, the thread pool for the Puma control server is empty at that time. So it looks like what happens is that, once the bug is triggered, the next incoming request will call [`pool.wait_until_not_full`](https://github.com/puma/puma/blob/v5.3.1/lib/puma/server.rb#L334) and wait forever for a thread to become available, but the only way for a thread to become available would be to call `spawn_thread` which happens _after_ the call to `pool.wait_until_not_full`.

**Expected behavior**
The thread pool should never deadlock.

 - OS: Linux (uname: `Linux <hostname> 5.8.0-1038-aws #40~20.04.1-Ubuntu SMP Thu Jun 17 13:25:28 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`)
 - Puma Version: 5.3.1

**Additional notes:**
I realize that without clear steps to reproduce, there might not be much you can do here. However I would appreciate any insights you can offer. I can also instrument the Puma code as needed.

olivierbellone:
### Description
Fix deadlock issue in thread pool.

This deadlock can happen when a thread is being trimmed, if the main server thread has already accepted a new incoming request and is waiting for the thread pool to not be full.

I was unable to write a test for this (race conditions are hard to reproduce!), but I was able to confirm this is a regression introduced in 5.0 by #2220. Previously, Puma did call `not_full.signal` when a thread was being trimmed: https://github.com/puma/puma/blob/v4.3.8/lib/puma/thread_pool.rb#L109.

I'm happy to add a test if anyone has suggestions for how to write one :)

Fixes #2655.

### DIFF
--- a/lib/puma/thread_pool.rb
+++ b/lib/puma/thread_pool.rb
@@ -119,6 +119,7 @@ module Puma
                 @trim_requested -= 1
                 @spawned -= 1
                 @workers.delete th
+                not_full.signal
                 Thread.exit
               end
 


cjlarose:
@wjordan Any thoughts on if it'd be possible to write a test for this?

History.md
```diff
@@ -1,3 +1,8 @@
+## 5.3.3 / 2021-07-
```
nateberkopec, line 1: Do we still have in our notes somewhere that people should edit changelogs? This isn't necessary anymore, can be removed.

olivierbellone: 👍  Removed.

olivierbellone:
@cjlarose Do you think there's any chance this could be merged and released this week? This bug is affecting our production servers and I'd much rather do a clean Puma upgrade rather than monkey patch my fix in. Totally understand if that doesn't work for you though.

cjlarose (approved):
I've reviewed the changes and this looks great. I'm comfortable merging without a regression test given the nature of the problem.

cjlarose:
@olivierbellone I'm comfortable merging, but I can't guarantee a release this week. I'll bring it up with the team, but recall that you can always pull dependencies directly from `git` in your Gemfile.

olivierbellone:
@cjlarose Thanks! Appreciate the quick merge. Using an untagged version is unfortunately not an option for us, but I'll figure something out.

cjlarose:
@olivierbellone This change is available in Puma 5.4.0 now.

olivierbellone:
Awesome! Thank you very much.

PR #2685, opened 2021-09-01
Prefer to use "parallel" rather than "concurrent". [ci skip]

ioquatix:
Puma uses threads, which execute in parallel. Threads which can't execute in parallel may be considered concurrent, but it's a degenerate form of parallelism i.e. multiple threads which are sharing a single processor would be considered to running concurrently. It would be good to see a consistent use of this terminology, although I suspect there will be people who have a different interpretation.

Parallelism should be used to describe simultaneous execution. Concurrency should be used to describe interleaved execution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
   <img src="https://puma.io/images/logos/puma-logo-large.png">
 </p>
 
-# Puma: A Ruby Web Server Built For Concurrency
+# Puma: A Ruby Web Server Built For Parallelism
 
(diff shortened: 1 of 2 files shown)


cjlarose:
Thanks so much for your contribution @ioquatix ! It's always nice to have contributions from a Ruby core team member!

Changes look good to me, but I'll leave the PR open for a bit in case other folks want to chime in.

One thing that might be good to update, though, would be the `gemspec`, which I think also uses `concurrent` where really we should be trying to sell Puma on its ability to leverage _parallelism_.

ioquatix:
Thanks!

PR #2688, opened 2021-09-02
Preserve BUNDLE_APP_CONFIG on worker fork

### LINKED ISSUE #2687: with_bundled_env causes trouble if BUNDLE_APP_CONFIG was set
**Describe the bug**

In an environment where BUNDLE_APP_CONFIG is set, one set of bundler settings are used to load puma, and then a different set of bundler settings are used when forking off workers. This causes, eg:

```console
> bundle exec puma -C config.rb hello.ru
[3026] * Pruning Bundler environment
[3026] Puma starting in cluster mode...
[3026] * Puma version: 5.4.0 (ruby 2.7.3-p183) ("Super Flight")
[3026] *  Min threads: 0
[3026] *  Max threads: 5
[3026] *  Environment: development
[3026] *   Master PID: 3026
[3026] *      Workers: 2
[3026] *     Restarts: (✔) hot (✔) phased
[3026] * Listening on http://0.0.0.0:9292
[3026] Use Ctrl-C to stop
[3027] + Gemfile in context: /Users/jon/Developer/puma-env/Gemfile
[3028] + Gemfile in context: /Users/jon/Developer/puma-env/Gemfile
Could not find rake-13.0.6 in any of the sources
Could not find rake-13.0.6 in any of the sources
Run `bundle install` to install missing gems.
Run `bundle install` to install missing gems.
[3028] ! Unable to start worker
[3027] ! Unable to start worker
[3028] /Users/jon/.gem/ruby/2.7.0/gems/bundler-2.2.16/lib/bundler/setup.rb:17:in `exit'
[3027] /Users/jon/.gem/ruby/2.7.0/gems/bundler-2.2.16/lib/bundler/setup.rb:17:in `exit'
[3028] Early termination of worker
[3027] Early termination of worker
```

I ran into this in a docker project where the base ruby image sets BUNDLE_APP_CONFIG=/usr/local/bundle, but I also happened to have a local .bundle directory which I was using outside of docker.

**Puma config:**

```rb
workers 2
prune_bundler
```

```
bundle exec puma -C config.rb hello.ru
```

**To Reproduce**

There's a reproduction repo here - https://github.com/jdelStrother/puma-vs-bundler

```
export BUNDLE_APP_CONFIG=/dev/null
bundle install
bundle exec puma -C config.rb hello.ru
```

Perhaps Puma::Launcher#prune_bundler could be tweaked to something like:

```diff
    def prune_bundler
      return if ENV['PUMA_BUNDLER_PRUNED']
      return unless defined?(Bundler)
      require_rubygems_min_version!(Gem::Version.new("2.2"), "prune_bundler")
      unless puma_wild_location
        log "! Unable to prune Bundler environment, continuing"
        return
      end

      dirs = files_to_require_after_prune

      log '* Pruning Bundler environment'
      home = ENV['GEM_HOME']
      bundle_gemfile = Bundler.original_env['BUNDLE_GEMFILE']
+     bundle_app_config = Bundler.original_env['BUNDLE_APP_CONFIG']
      with_unbundled_env do
        ENV['GEM_HOME'] = home
        ENV['BUNDLE_GEMFILE'] = bundle_gemfile
+       ENV['BUNDLE_APP_CONFIG'] = bundle_app_config
        ENV['PUMA_BUNDLER_PRUNED'] = '1'
        args = [Gem.ruby, puma_wild_location, '-I', dirs.join(':')] + @original_argv
        # Ruby 2.0+ defaults to true which breaks socket activation
        args += [{:close_others => false}]
        Kernel.exec(*args)
      end
    end
```

?

**Desktop**
 - OS: MacOS, & LinuxKit in Docker
 - Puma Version: 5.4.0

jdelStrother:
Without this, if puma is launched with BUNDLE_APP_CONFIG set, that will
be lost inside of unbundled_env, which causes the worker processes to use a
potentially different bundler configuration.

Closes #2687

### DIFF
--- a/test/test_preserve_bundler_env.rb
+++ b/test/test_preserve_bundler_env.rb
@@ -39,6 +39,27 @@ class TestPreserveBundlerEnv < TestIntegration
     assert_match("Gemfile.bundle_env_preservation_test", new_reply)
   end
 
+  def test_worker_forking_preserves_bundler_config_path
+    skip_unless_signal_exist? :TERM
+
+    @tcp_port = UniquePort.call
+    env = {
+      # Disable the .bundle/config file in the bundle_app_config_test directory
+      "BUNDLE_APP_CONFIG" => "/dev/null",
+      # Don't allow our (rake test's) original env to interfere with the child process
+      "BUNDLE_GEMFILE" => nil,
+      "BUNDLER_ORIG_BUNDLE_GEMFILE" => nil
+    }
+    cmd = "bundle exec puma -q -w 1 --prune-bundler -b tcp://#{HOST}:#{@tcp_port}"
+    Dir.chdir File.expand_path("bundle_app_config_test", __dir__) do
+      @server = IO.popen(env, cmd.split, "r")
+    end
+    wait_for_server_to_boot
+    @pid = @server.pid
+    reply = read_body(connect)
+    assert_equal("Hello World", reply)
+  end
+
   def test_phased_restart_preserves_unspecified_bundle_gemfile
     skip_unless_signal_exist? :USR1
 
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -319,10 +319,12 @@ module Puma
       log '* Pruning Bundler environment'
       home = ENV['GEM_HOME']
       bundle_gemfile = Bundler.original_env['BUNDLE_GEMFILE']
+      bundle_app_config = Bundler.original_env['BUNDLE_APP_CONFIG']
       with_unbundled_env do
         ENV['GEM_HOME'] = home
         ENV['BUNDLE_GEMFILE'] = bundle_gemfile
         ENV['PUMA_BUNDLER_PRUNED'] = '1'
+        ENV["BUNDLE_APP_CONFIG"] = bundle_app_config
         args = [Gem.ruby, puma_wild_location, '-I', dirs.join(':')] + @original_argv
         # Ruby 2.0+ defaults to true which breaks socket activation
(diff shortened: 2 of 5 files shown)


test/test_preserve_bundler_env.rb
```diff
@@ -39,6 +39,25 @@ def test_usr2_restart_preserves_bundler_environment
     assert_match("Gemfile.bundle_env_preservation_test", new_reply)
   end
 
+  def test_worker_forking_preserves_bundler_config_path
```
cjlarose, line 42: Locally, if I undo the change introduced to `prune_bundler`,

```sh
git restore --source HEAD^ -- lib
```

This test is still green. In other words, the test does not appear to guard against a regression for the new functionality.

Do you think it is possible to write this test in a way that it fails on `master` in the way you observed?

cjlarose (changes requested):

test/test_preserve_bundler_env.rb
jdelStrother: Bah, I was testing with `ruby -Ilib:test test/test_preserve_bundler_env.rb` which does fail on master, but looks like the same thing inside of `bundle exec` doesn't.

I've force-pushed an update that sets BUNDLE_GEMFILE in the test, which should persuade it to fail in either scenario (and then subsequently be fixed by the BUNDLE_APP_CONFIG change).

cjlarose: Awesome, thanks!

cjlarose (approved):

PR #2694, opened 2021-09-08
Add new test framework files, test for PR #2675

MSP-Greg:
### Description

This PR includes some of the test files for a new test framework that (hopefully) will make it easier to write simpler tests.  Also, more complete and reliable tests, more data to see the affects of PR's, etc.

The also adds a test for PR #2675, `TestPOpenSSL#test_persistent_close`.  The test sends twenty requests to a Puma server, after each request, it pauses 1.5 seconds, with config `persistent_timeout` set to 1 second.

Request time results for master, using Ubuntu 20.04 & Ruby head:
```
  20 successful requests (10 loops of 2 clients * 1 request per client)
          10%    20%    40%    50%    60%    80%    90%    95%    97%    99%   100% 
    mS      0      0      0    130    330   5740   9540  13020  13020  13020  13520
```

Request time results for PR #2675,  using Ubuntu 20.04 & Ruby head:
```
  20 successful requests (10 loops of 2 clients * 1 request per client)
          10%    20%    40%    50%    60%    80%    90%    95%    97%    99%   100% 
    mS  2.927  3.049  3.252  3.288  3.320  3.390  3.838  4.876  4.876  4.876  5.707
```

This shows the benefit of the fixes in PR #2675.  Also, there is a new file in the docs folder (test_puma_intro.md), which is the start  of a description of the framework, along with comments in the files.  Going forward, I may make changes, add benchmark files (some soon), and start to convert the current tests to the new framework.  Work in my fork shows they can be much more consistent...

### Your checklist for this pull request
- [x] I have reviewed the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.
- [x] I have added (or updated) appropriate tests if this PR fixes a bug or adds a feature.
- [ ] My pull request is 100 lines added/removed or less so that it can be easily reviewed.
- [ ] If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.
- [ ] If this closes any issues, I have added "Closes `#issue`" to the PR description or my commit messages.
- [x] I have updated the documentation accordingly.
- [x] All new and existing tests passed, including Rubocop.

PR #2695, opened 2021-09-11
benchmarks - wrk overload testing files [ci skip]

MSP-Greg:
### Description

After working with wrk for a while, I thought having a script that ran several sets would be helpful.  Also, after each run, collecting  Puma stats and output from smem would show additional info.  The summary output is as below:
```
────────wrk────────  ─Request─time─distribution─(ms)─  Worker─requests  ─wrk─requests─
 -t    -c   req/sec   50%    75%    90%    99%   100%  spread   total     total   bad
 10    50    10504    1.3   24.6   38.8   48.2   58.9   0.29   158708    158708     0
 13    65    10536    1.4   34.1   52.8   64.7   74.4   0.28   159153    159153     0
 17    85    10580    1.5   46.3   71.2   86.4   96.5   0.19   159859    159859     0
 23   115    10627    1.7   65.5  100.1  121.0  131.5   0.27   160631    160631     0
 30   150    10604    2.4   86.9  132.4  159.6  170.9   0.22   160249    160249     0
             10570                                    Totals   798600    798600
══════════════════════════════════════════════════════════════════════════════════════

══════════════════════════════════ Memory Change per Request
WRK Threads            10     13     17     23     30
    Connections        50     65     85    115    150    All
    Av Resp Size    52974  52960  52995  53008  53000  52987
════════════════════════════════════════════════════════════
worker-0-00    USS     42     25     17    -17      9     15
worker-0-01    USS     23     21     19    -12      6     11
All            USS     32     23     18    -15      7     13
════════════════════════════════════════════════════════════
```

The top section shows data from the five wrk runs, which start at the total number of threads the Puma server has, and stop at three times that number.  The 'spread' column shows how consistent the requests are spread between workers.  It's interesting that as the load increases, throughput (req/sec) stays about the same, but the response time increases, especially in the slowest responses.

The bottom section shows memory used after each wrk run, and should remain low.  Oddly, using this with Ruby head lead to the discover of a memory leak in Ruby head, as the 'All' column was over 500 bytes per request.

Some notes:

1. Since CI is not consistent enough to check performance, a good wrk script is helpful for determining whether changes improve performance.
2. The code is commented in `overload/benchmarks/local/bench_overload_wrk.rb`.  It uses smem for memory info, and also uses @ioquatix's [fork of wrk](https://github.com/ioquatix/wrk).  It only works with Puma in cluster mode (with workers).
3. wrk doesn't support UNIXSockets, so I may try to get code running using some of the files in #2694, which shares rackup files with this PR.

### DIFF
--- a/benchmarks/local/bench_base.rb
+++ b/benchmarks/local/bench_base.rb
@@ -1,0 +1,229 @@
+# frozen_string_literal: true
+
+require 'optparse'
+
+module TestPuma
+  class BenchBase
+
+    # Signal used to trigger worker GC
+    GC_SIGNAL = 'SIGALRM'
+
+    # We're running under GitHub Actions
+    IS_GHA = ENV['GITHUB_ACTIONS'] == 'true'
+
+    WRK_PERCENTILE = [0.50, 0.75, 0.9, 0.99, 1.0].freeze
+
+    def initialize
+      sleep 5 # wait for server to boot
+
+      @thread_loops       = nil
+      @clients_per_thread = nil
+      @req_per_client     = nil
+      @body_kb            = 10
+      @dly_app            = nil
+      @bind_type          = :tcp
+
+      @ios_to_close = []
+
+      setup_options
+
+      unless File.exist? @state_file
+        puts "can't fined state file '#{@state_file}'"
+        exit 1
+      end
+
+      mstr_pid = File.binread(@state_file)[/^pid: +(\d+)/, 1].to_i
+      begin
+        Process.kill 0, mstr_pid
+      rescue Errno::ESRCH
+        puts 'Puma server stopped?'
+        exit 1
+      rescue Errno::EPERM
+      end
+
+      case @bind_type
+      when :ssl, :tcp
+        @bind_port = ENV.fetch('PORT', 40010).to_i
+      when :unix
+        @bind_path = "#{Dir.home}/skt.unix"
+      when :aunix
+        @bind_path = "@skt.aunix"
+      else
+        exit 1
+      end
+    end
+
+    def setup_options
+      OptionParser.new do |o|
+        o.on "-l", "--loops LOOPS", OptionParser::DecimalInteger, "create_clients: loops/threads" do |arg|
+          @thread_loops = arg.to_i
+        end
+
+        o.on "-c", "--connections CONNECTIONS", OptionParser::DecimalInteger, "create_clients: clients_per_thread" do |arg|
+          @clients_per_thread = arg.to_i
+        end
+
+        o.on "-r", "--requests REQUESTS", OptionParser::DecimalInteger, "create_clients: requests per client" do |arg|
+          @req_per_client = arg.to_i
+        end
+
+        o.on "-b", "--body_kb BODYKB", OptionParser::DecimalInteger, "CI RackUp: size of response body in kB" do |arg|
+          @body_kb = arg.to_i
+        end
+
+        o.on "-d", "--dly_app DELAYAPP", Float, "CI RackUp: app response delay" do |arg|
+          @dly_app = arg.to_f
+        end
+
+        o.on "-s", "--socket SOCKETTYPE", String, "Bind type: tcp, ssl, unix, aunix" do |arg|
+          @bind_type = arg.to_sym
+        end
+
+        o.on "-S", "--state STATEFILE", String, "Puma Server: state file" do |arg|
+          @state_file = arg
+        end
+
+        o.on "-t", "--threads THREADS", String, "Puma Server: threads" do |arg|
+          @threads = arg[/\d+\z/].to_i
+        end
+
+        o.on "-w", "--workers WORKERS", OptionParser::DecimalInteger, "Puma Server: workers" do |arg|
+          @workers = arg.to_i
+        end
+
+        o.on "-T", "--time TIME", OptionParser::DecimalInteger, "wrk: duration" do |arg|
+          @wrk_time = arg.to_i
+        end
+
+        o.on "-W", "--wrk_bind WRK_STR", String, "wrk: bind string" do |arg|
+          @wrk_bind_str = arg
+        end
+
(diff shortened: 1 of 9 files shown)


benchmarks/local/bench_base.rb
```diff
@@ -0,0 +1,229 @@
...
+      @body_kb            = 10
+      @dly_app            = nil
+      @bind_type          = :tcp
+
+      @ios_to_close = []
+
+      setup_options
+
+      unless File.exist? @state_file
+        puts "can't fined state file '#{@state_file}'"
```
jacobherrington, line 31: I noticed a small typo here! 🙈 

```suggestion
        puts "can't find state file '#{@state_file}'"
```

benchmarks/local/bench_base.rb
```diff
@@ -0,0 +1,229 @@
...
+    def run_wrk_parse(cmd)
+      print cmd.ljust 55
+
+      wrk_output = %x[#{cmd}]
+
+      wrk_data = "#{wrk_output[/\A.+ connections/m]}\n#{wrk_output[/  Thread Stats.+\z/m]}"
+
+      print " |#{wrk_data[/^ +\d+ +requests.+/].rstrip}\n"
+
+      # puts '', wrk_data, ''     # for debugging or output format changes
```
jacobherrington, line 136: Should this be removed or uncommented?

```suggestion
```

benchmarks/local/bench_base.rb
```diff
@@ -0,0 +1,229 @@
...
+        else
+          0
+        end
+      end).to_h
+      hsh
+    end
+
+    def parse_stats
+      obj = @puma_info.run 'stats'
+      # puts ''; pp obj; puts ''
```
jacobherrington, line 187: Same as above!

```suggestion
```

benchmarks/local/bench_base.rb
```diff
@@ -0,0 +1,229 @@
...
+
+    def parse_smem
+      @puma_info.run 'gc'
+      sleep 1
+
+      hsh_smem = Hash.new []
+      pids = @pids.keys
+
+      smem_info = %x[smem -c 'pid rss pss uss command']
+      # puts '', smem_info, ''
```
jacobherrington, line 214: Same as above!

```suggestion
```

benchmarks/local/bench_overload_wrk.rb
```diff
@@ -0,0 +1,379 @@
...
+        info[:smem] = parse_smem
+        run_summary info
+      end
+
+      puts '::[endgroup]' if IS_GHA
+
+      overall_summary
+      leak_calc
+      puts RUBY_DESCRIPTION, ''
+      # puts ''; pp @info; puts ''
```
jacobherrington, line 152: Same as above!

```suggestion
```

jacobherrington (commented):
This is a cool idea!

I noticed some nit-picky things (that might have been left in on purpose) and left notes on them, but you're welcome to disregard them. 😄

ioquatix:
To confirm, by `USS` you mean unique set size right? i.e. private non-sharable data.

MSP-Greg:
@jacobherrington

Thanks for reviewing and trying it.

I've updated things quite a bit, adding another generic rackup file, which, based on a request header, switches the body type between array, chunked, string, or io, and also the body size.  I'll be adding some more benchmark scripts.  I'll remove all the debug code as your comments above.

The code currently outputs to the console, but I've set it up so all the measurement data is a Ruby object, so it be can used for 'asserts' or pass/fail code.  CI machines vary quite a bit, and I haven't (yet) figured out a way to determine how 'fast' they're running.  But, if that can be done, we can add some performance testing to CI...

test/rackup/hello.ru
```diff
@@ -1 +1,3 @@
-run lambda { |env| [200, {"Content-Type" => "text/plain"}, ["Hello World"]] }
+hdrs = {'Content-Type'.freeze => 'text/plain'.freeze}.freeze
```
nateberkopec, line 1: Freezing these is a good call - we're here to benchmark Puma's overhead, not the Rack app 👍

nateberkopec (approved):
I'm good with this merging whenever you're ready Greg

MSP-Greg:
@nateberkopec

Thanks.  I'm still working on it, what I've expanded it to is a wrk script and a 'ruby connections' script that both are able to test the four types of request bodies (array, chunked, string, and IO), and also allow specifying the size of the body.  Without a parameter for the body type/size, they run a matrix of all the combinations, but you can select a vector of the matrix, or a single 'cell/element'.  The output shows requests per second, response time, response size, etc.

The 'ruby connections' script was using code that I originally wrote to send a stream of connections for use in CI tests, for things like restarts, shutdown, etc.  Without a reasonably fast connection stream, testing for those state changes in Puma might not show real metrics about request handling, etc.

A couple of things:

One might ask 'why the Ruby script?'   First, wrk doesn't support UNIXSockets.  Secondly, 'requests per second' (RPS) is somewhat separate from response time (RT).  With 'wrk', one can fill up the listener backlog.  When that happens, RPS stays almost the same, but RT increases, first in the longest 20% or so, then starts to affect all the requests.  One can query for the backlog size with a TCP listener, and I added it to stats so I could see it's value.  I haven't looked at whether it can be added to stats for production.

Anyway, I'm still working on it, and also working on comments/documentation.  Even, today, I'm still renaming methods that aren't clear...

MSP-Greg:
I'll be posting a large update to this, please **do not** merge!  Sorry for the delay...

MSP-Greg:
This code is similar to the code already committed in PR #2895.  That code used a constant wrk setup, and allowed a matrix of body types and sizes.  The code here increases the load on Puma, and shows behavior in terms of RPS (requests per sec) and metrics on response time.  Or, how Puma behaves as the incoming requests increase to the point of overloading it.

Given that there are several other issues that I'd like to work on, and also the merge conflicts shown above, closing with the expectation that I'll hoepfully reopen with updated code.

PR #2698, opened 2021-09-15
Fix missing `Concurrency` -> `Parallelism`.

ioquatix:
Also, the GitHub project description mentions "Concurrency" but I believe it's better to use "Parallelism" here for consistency.


cjlarose:
Screen Shot 2021-09-19 at 23 20 51

Updated the Github project description

PR #2700, opened 2021-09-15
More elaborate exception handling.

### LINKED ISSUE #2699: Puma listen thread can fail with infinite loop.
The following middleware can cause Puma to enter an infinite loop:

``` ruby
require 'objspace'

run lambda{|env|
	ios = ObjectSpace.each_object(IO).to_a.select{|io| io.is_a?(TCPServer)}
	
	ios.each(&:close)
	
	[200, [], [ios.inspect]]
}
```

This probably seems odd, but we are actually seeing some variant of this in a production system. We don't know who is calling close (suspect threading bug) but it causes an EBADF error in the list loop, followed by infinite `IOError`s:

```
> puma -w4 -t4
[95960] Puma starting in cluster mode...
[95960] * Puma version: 5.4.0 (ruby 3.0.2-p107) ("Super Flight")
[95960] *  Min threads: 4
[95960] *  Max threads: 4
[95960] *  Environment: development
[95960] *   Master PID: 95960
[95960] *      Workers: 4
[95960] *     Restarts: (✔) hot (✔) phased
[95960] * Listening on http://0.0.0.0:9292
[95960] Use Ctrl-C to stop
[95960] - Worker 1 (PID: 95962) booted in 0.02s, phase: 0
[95960] - Worker 0 (PID: 95961) booted in 0.02s, phase: 0
[95960] - Worker 2 (PID: 95963) booted in 0.02s, phase: 0
[95960] - Worker 3 (PID: 95964) booted in 0.02s, phase: 0
2021-09-15 20:56:09 +1200 Listen loop: #<Errno::EBADF: Bad file descriptor>
2021-09-15 20:56:09 +1200 Listen loop: #<IOError: closed stream>
2021-09-15 20:56:09 +1200 Listen loop: #<IOError: closed stream>
2021-09-15 20:56:09 +1200 Listen loop: #<IOError: closed stream>
2021-09-15 20:56:09 +1200 Listen loop: #<IOError: closed stream>
2021-09-15 20:56:09 +1200 Listen loop: #<IOError: closed stream>
2021-09-15 20:56:09 +1200 Listen loop: #<IOError: closed stream>
2021-09-15 20:56:09 +1200 Listen loop: #<IOError: closed stream>
2021-09-15 20:56:09 +1200 Listen loop: #<IOError: closed stream>
2021-09-15 20:56:09 +1200 Listen loop: #<IOError: closed stream>
2021-09-15 20:56:09 +1200 Listen loop: #<IOError: closed stream>
2021-09-15 20:56:09 +1200 Listen loop: #<IOError: closed stream>
2021-09-15 20:56:09 +1200 Listen loop: #<IOError: closed stream>
2021-09-15 20:56:09 +1200 Listen loop: #<IOError: closed stream>
```

At the very least, we should probably handle this more gracefully (i.e. exit the loop as soon as the listening socket becomes invalid for any reason, rather than trying to select/accept again).

ioquatix:
Fixes #2699.

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -356,7 +356,10 @@ module Puma
                 pool << client
               end
             end
-          rescue Object => e
+          rescue IOError, Errno::EBADF
+            # In the case that any of the sockets are unexpectedly close.
+            raise
+          rescue StandardError => e
             @events.unknown_error e, nil, "Listen loop"
           end
         end
--- a/test/helpers/integration.rb
+++ b/test/helpers/integration.rb
@@ -50,7 +50,7 @@ class TestIntegration < Minitest::Test
     assert(system(*args, out: File::NULL, err: File::NULL))
   end
 
-  def cli_server(argv, unix: false, config: nil)
+  def cli_server(argv, unix: false, config: nil, merge_err: false)
     if config
       config_file = Tempfile.new(%w(config .rb))
       config_file.write config
@@ -64,7 +64,11 @@ class TestIntegration < Minitest::Test
       @tcp_port = UniquePort.call
       cmd = "#{BASE} #{puma_path} #{config} -b tcp://#{HOST}:#{@tcp_port} #{argv}"
     end
-    @server = IO.popen(cmd, "r")
+    if merge_err
+      @server = IO.popen(cmd, "r", :err=>[:child, :out])
+    else
+      @server = IO.popen(cmd, "r")
+    end
     wait_for_server_to_boot
     @pid = @server.pid
     @server
--- a/test/rackup/close_listeners.ru
+++ b/test/rackup/close_listeners.ru
@@ -1,0 +1,6 @@
+require 'objspace'
+
+run lambda { |env|
+  ios = ObjectSpace.each_object(::TCPServer).to_a.tap { |a| a.each(&:close) }
+  [200, [], ["#{ios.inspect}\n"]]
+}
--- a/test/test_integration_single.rb
+++ b/test/test_integration_single.rb
@@ -178,4 +178,29 @@ class TestIntegrationSingle < TestIntegration
     @server.close unless @server.closed?
     @server = nil
   end
+
+  # listener is closed 'externally' while Puma is in the IO.select statement
+  def test_closed_listener
+    skip_unless_signal_exist? :TERM
+
+    cli_server "test/rackup/close_listeners.ru", merge_err: true
+    read_body connect
+
+    begin
+      Timeout.timeout(5) do
+        begin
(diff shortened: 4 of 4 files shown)


nateberkopec:
Can you adapt the code you put in your issue report into a test? Otherwise LGTM

ioquatix:
@nateberkopec please tell me any nits or other issues with this PR. I am not fully across the Puma test suite so I just cobbled something together from some of the other integration tests.

MSP-Greg:
@ioquatix

I was waiting for CI.  For a test, see:

https://github.com/MSP-Greg/puma/commit/c358b3b14d939ba04e800662b237d79fbeb9a24e.patch

Test & Master https://github.com/MSP-Greg/puma/actions/runs/1239898360
All appropriate runs failed 'TestIntegrationSingle#test_closed_listener'

Test & PR2700 https://github.com/MSP-Greg/puma/actions/runs/1239909432
All appropriate runs passed 'TestIntegrationSingle#test_closed_listener'

Windows is skipped due to the test using SIGTERM.  As mentioned, I tested closing the server, which is where the infinite loop happened.  I needed to add a timeout so the infinite loop was caught if it happened, like with master...

ioquatix:
@MSP-Greg do you want to just add your commit to my commit? I am happy for you to delete my test changes.

MSP-Greg:
@ioquatix

Thanks.  Sorry for the delay.  I squashed to remove your test, I saved the code in a patch file if...

MSP-Greg (approved):

ioquatix:
My test was a total hack. I am happy you did something better :) Thanks for your commitment to Puma.

ioquatix:
Also it should be said, that there is a material change to the listen loop - `rescue Object` changed to `rescue StandardError`. This means non-`StandardError` exceptions will propagate out and I think that's desirable since these often represent critical failures like memory allocation errors, etc. Basically, at this point, we should let the instance fail. In our production bug... the load balancer continues to try and send traffic for a short while and we end up with a whole batch of failed requests. I'd rather have it fail fast.

nateberkopec:
> Also it should be said, that there is a material change to the listen loop

Yeah, I think I'm willing to call this a bugfix rather than an API change.

dentarg:
> Maybe I'm misunderstanding what's going on, but we are now re-raising the exception on `EBADF`.
> 
> I would expect this to log from this line: https://github.com/puma/puma/blame/master/lib/puma/server.rb#L376 ([permalink](https://github.com/puma/puma/blame/d048ea0308de196972e38e18a0fab0c70ec848fe/lib/puma/server.rb#L376))

@ioquatix That showed up in CI, see https://github.com/puma/puma/pull/2748

PR #2702, opened 2021-09-17
Add support for APP_ENV environment variable

jacobherrington:
### Description
Using APP_ENV to set the environment is supported by Sinatra and
Sidekiq, so it makes sense to support the same behavior in Puma.

Like Sidekiq, APP_ENV will take precedence over RACK_ENV and RAILS_ENV.
APP_ENV defers to any argument passed via the --environment flag.

### DIFF
--- a/test/test_cli.rb
+++ b/test/test_cli.rb
@@ -426,6 +426,19 @@ class TestCLI < Minitest::Test
     assert_equal %w[a b], extra_dependencies
   end
 
+  def test_environment_app_env
+    ENV['RACK_ENV'] = @environment
+    ENV['RAILS_ENV'] = @environment
+    ENV['APP_ENV'] = 'test'
+
+    cli = Puma::CLI.new []
+
+    assert_equal 'test', cli.environment
+
+    ENV.delete 'APP_ENV'
+    ENV.delete 'RAILS_ENV'
+  end
+
   def test_environment_rack_env
     ENV.delete 'RACK_ENV'
 
--- a/README.md
+++ b/README.md
@@ -270,7 +270,7 @@ You can also provide a configuration file with the `-C` (or `--config`) flag:
 $ puma -C /path/to/config
 ```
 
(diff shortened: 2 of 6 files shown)


test/test_cli.rb
```diff
@@ -426,6 +426,19 @@ def test_extra_runtime_dependencies
...
   end
 
+  def test_environment_app_env
+    ENV['RACK_ENV'] = @environment
+    ENV['RAILS_ENV'] = @environment
+    ENV['APP_ENV'] = 'test'
+
+    Puma::CLI.new []
+
+    assert_equal 'test', ENV['RACK_ENV']
```
nateberkopec, line 436: `assert_equal 'test', cli.environment`

test/test_cli.rb
```diff
@@ -426,6 +426,19 @@ def test_extra_runtime_dependencies
     assert_equal %w[a b], extra_dependencies
   end
 
+  def test_environment_app_env
+    ENV['RACK_ENV'] = @environment
+    ENV['RAILS_ENV'] = @environment
+    ENV['APP_ENV'] = 'test'
+
+    Puma::CLI.new []
```
nateberkopec, line 434: cli = Puma::CLI.new []

nateberkopec:
One change, otherwise LGTM and thanks for implementing!

jacobherrington:
@nateberkopec Updated and force pushed the branch!

jacobherrington:
@nateberkopec It looks like `Puma::CLI.new []` doesn't have an `environment` method, is that unexpected?

nateberkopec:
Fixed it up: https://github.com/puma/puma/commit/f21ae53ab87e4625aaf37a3737c1f2a442014550

I couldn't remember the API.

What I was trying to get across was that the original test didn't actually test anything, it was making assertions about the setup and not about the action you took during the test. Unfortunately all the other tests around this behavior are exactly the same (I just deleted them).

jacobherrington:
@nateberkopec I see. I was curious about how this worked, but I assume the old tests were trustworthy.

I'll look into writing some new tests to cover the old cases, thanks for your help!

nateberkopec:
That would be great! If you want to refactor how the CLI object works a bit too so we don't have to `send` or `instance_variable_get` that would be nice. 

Puma's an old project... some of the tests are good, others are not 😆 I don't blame you for following what was already there.

Issue #2635, opened 2021-05-26
Allow configuration of on_booted event hooks


nateberkopec:
Originally requested in #1230 

Similar to `after_worker_boot`, but for the `on_booted` event.

will-in-wi:
Hey! I wanted to add my use case to this ticket. I'm trying to get some prometheus instrumentation set up (using the discourse plugin), but it uses the stats object. I'm struggling to find a reasonable place to initialize this where I know that `Puma.stats` is available. I've tried a Rails initializer and a Puma plugin, but the only thing that seems to work is:

```ruby
require "prometheus_exporter/instrumentation"

Puma::Plugin.create do
  def start(_launcher)
    in_background do
      loop do
        # So ugly... We really just need to loop until Puma.stats is initialized. This is happening in another thread, and I can't seem to find a better way.
        sleep(1)
        Puma.stats
        break
      rescue NoMethodError
        # Let the loop retry
      end

      # Defaults to sampling every 30 seconds.
      PrometheusExporter::Instrumentation::Puma.start
    end
  end
end
```

Having some kind of hook where I could initialize them would be excellent! I'm threaded, so the after_worker_boot hook doesn't work.

I'm happy to take a whack at this if someone can point me to where things are set up.

x-yuri:
My use case. Right now the app connects to db on boot. And in clustered mode it disconnects in master, then connects in workers. Which seems pointless (to connect only to disconnect soon). I wanted to make it connect to db in the only process in single mode, and in workers in clustered mode. But... right now it seems impossible.

Issue #1230, opened 2017-03-01
Callbacks for single mode?


swrobel:
### Steps to reproduce

config/puma.rb:
```ruby
after_worker_boot do
  puts 'hello from a worker'
end
```

example:
```
$ bundle exec puma
Puma starting in single mode...
* Version 3.7.0 (ruby 2.3.2-p217), codename: Snowy Sagebrush
* Min threads: 0, max threads: 16
* Environment: development
* Listening on tcp://0.0.0.0:9292
Use Ctrl-C to stop

$ bundle exec puma -w 2
[59172] Puma starting in cluster mode...
[59172] * Version 3.7.0 (ruby 2.3.2-p217), codename: Snowy Sagebrush
[59172] * Min threads: 0, max threads: 16
[59172] * Environment: development
[59172] * Process workers: 2
[59172] * Phased restart available
[59172] * Listening on tcp://0.0.0.0:9292
[59172] Use Ctrl-C to stop
hello from a worker
```

### Expected behavior

`after_worker_boot` or some equivalent callback is available in single mode after the server has booted successfully.

### Actual behavior

`after_worker_boot` callback doesn't run in single mode.

### System configuration

**Ruby version**: 2.3.2
**Rails version**: 5.0.1
**Puma version**: 3.7.0

nateberkopec:
`after_worker_boot` only runs in cluster mode (more than 1 worker).

swrobel:
So you don't see the value in having callbacks in single mode?

nateberkopec:
*shrug* Just pointing out that it's clearly documented that this is a cluster_mode only callback: https://github.com/puma/puma/blob/f5f23aaac7aaccff1b6b138d93dd4b1755ebf1c2/lib/puma/dsl.rb#L337

What's your use case for doing something after boot in single mode?

swrobel:
I'll respectfully disagree with your definition of "clearly documented" but that's sort of moot.

My use case is that I'm using the `tmp_restart` plugin and I want to display a notification using `terminal-notifier` when the restart completes.

swrobel:
I discovered that I'm able to get access to the callbacks by setting `workers 1` which basically forces clustered mode. Seems like an arbitrary distinction to me, but I guess *shrug*, it works.

swrobel:
[...]

PR #2709, opened 2021-09-22
Add on_booted event

rodzyn:
### Description
Closes https://github.com/puma/puma/issues/2635

I'm not sure if exactly that is what was intended. But I'm happy to get a feedback :)

### DIFF
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -664,6 +664,16 @@ module Puma
 
     alias_method :after_worker_boot, :after_worker_fork
 
+    # Code to run after puma is booted (works for both: single and clustered)
+    #
+    # @example
+    #   on_booted do
+    #     puts 'After booting...'
+    #   end
+    def on_booted(&block)
+      @config.options[:events].on_booted(&block)
+    end
+
     # When `fork_worker` is enabled, code to run in Worker 0
     # before all other workers are re-forked from this process,
     # after the server has temporarily stopped serving requests
--- a/README.md
+++ b/README.md
@@ -157,6 +157,15 @@ before_fork do
 end
 ```
 
+You can also specify a block to be run after puma is booted using `on_booted`:
+
+```ruby
+# config/puma.rb
+on_booted do
+  # configuration here
+end
+```
+
 ### Error handling
 
 If puma encounters an error outside of the context of your application, it will respond with a 500 and a simple
--- a/lib/puma/cli.rb
+++ b/lib/puma/cli.rb
@@ -93,7 +93,7 @@ module Puma
     #
 
     def setup_options
-      @conf = Configuration.new do |user_config, file_config|
+      @conf = Configuration.new({}, {events: @events}) do |user_config, file_config|
         @parser = OptionParser.new do |o|
           o.on "-b", "--bind URI", "URI to bind to (tcp://, unix://, ssl://)" do |arg|
             user_config.bind arg
--- a/lib/rack/handler/puma.rb
+++ b/lib/rack/handler/puma.rb
@@ -27,7 +27,9 @@ module Puma
         end
       end
 
-      conf = ::Puma::Configuration.new(options, default_options) do |user_config, file_config, default_config|
+      @events = options[:events] || ::Puma::Events.new
+
+      conf = ::Puma::Configuration.new(options, default_options.merge({events: @events})) do |user_config, file_config, default_config|
         if options.delete(:Verbose)
           require 'rack/common_logger'
           app = Rack::CommonLogger.new(app, STDOUT)
@@ -59,11 +61,11 @@ module Puma
     end
 
     def run(app, **options)
-      conf   = self.config(app, options)
+      conf = self.config(app, options)
 
       log_writer = options.delete(:Silent) ? ::Puma::LogWriter.strings : ::Puma::LogWriter.stdio
 
-      launcher = ::Puma::Launcher.new(conf, :log_writer => log_writer)
+      launcher = ::Puma::Launcher.new(conf, :log_writer => log_writer, events: @events)
 
       yield launcher if block_given?
       begin
--- a/test/config/event_on_booted.rb
+++ b/test/config/event_on_booted.rb
@@ -1,0 +1,3 @@
+on_booted do
+  puts "on_booted called"
+end
--- a/test/config/event_on_booted_exit.rb
+++ b/test/config/event_on_booted_exit.rb
@@ -1,0 +1,5 @@
+on_booted do
+  pid = Process.pid
+  Process.kill :TERM, pid
+  Process.wait pid
+end
--- a/test/test_integration_cluster.rb
+++ b/test/test_integration_cluster.rb
@@ -130,6 +130,16 @@ class TestIntegrationCluster < TestIntegration
     assert_equal 0, status
   end
 
+  def test_on_booted
+    cli_server "-w #{workers} -C test/config/event_on_booted.rb -C test/config/event_on_booted_exit.rb test/rackup/hello.ru", no_wait: true
+
+    output = []
+
+    output << $_ while @server.gets
+
+    assert output.any? { |msg| msg == "on_booted called\n" } != nil
+  end
+
   def test_term_worker_clean_exit
     cli_server "-w #{workers} test/rackup/hello.ru"
 
--- a/test/test_integration_single.rb
+++ b/test/test_integration_single.rb
@@ -49,6 +49,16 @@ class TestIntegrationSingle < TestIntegration
     assert_equal 15, status
   end
 
+  def test_on_booted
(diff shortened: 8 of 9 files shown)


lib/puma/dsl.rb
```diff
@@ -588,6 +588,18 @@ def after_worker_fork(&block)
       @options[:after_worker_fork] << block
     end
 
+    # Code to run after puma is booted in a single mode
+    #
+    # @note Single mode mode only.
```
nateberkopec, line 593: Typo

lib/puma/single.rb
```diff
@@ -52,6 +52,8 @@ def run
       @server = server = start_server
       server_thread = server.run
 
+      @launcher.config.run_hooks :on_booted, nil, @launcher.events
```
nateberkopec, line 55: Rather than add this line, should we somehow make  use the already-existing `on_booted` event, I wonder? It's maybe not great that we have this `run_hooks` system which has the exact same name of a trigger as an `Event`... feels like they could be combined.

nateberkopec:
Thank you so much for your first contrib to Puma! Welcome to the team!

rodzyn:
> Thank you so much for your first contrib to Puma! Welcome to the team!

Thanks for a warm welcome!

lib/puma/single.rb
rodzyn: Ok, that makes sense. I am not familiar yet how the events are working but I'll dig into that tomorrow and try to provide alternative solution that is addressing your suggestions. Thanks for this extremely quick feedback

nateberkopec: It's also possible that we decide it's possible, but we won't do it in this PR and leave the refactor for someone else later. But if you can do it on your own/it's easy enough, let's do it!

rodzyn:
@nateberkopec One simple way that it could be achieved is to give a possibility for DSLs to have access to the events object. I did that in the latest commit

lib/puma/dsl.rb
```diff
@@ -588,6 +592,17 @@ def after_worker_fork(&block)
...
+    # Code to run after puma is booted in a single mode
+    #
+    # @note Single mode only.
+    # @example
+    #   on_booted do
+    #     puts 'After booting...'
+    #   end
+    def on_booted(&block)
+      @options[:events].on_booted(&block)
+    end
```
jacobherrington, line 604: I'm not sure *exactly* why the tests are failing, but (for what it's worth) changing this method makes the tests pass:
```suggestion
    def on_booted(&block)
      @options[:on_booted] ||= []
      @options[:on_booted] << block
    end
```

test/test_config.rb
```diff
@@ -317,6 +317,10 @@ def test_run_hooks_before_fork
     assert_run_hooks :before_fork
   end
 
+  def test_run_hooks_on_booted
+    assert_run_hooks :on_booted
```
jacobherrington, line 321: This seems to be failing inside of `assert_run_hooks`. I think it is specifically this line:
https://github.com/puma/puma/blob/a6cdca85a15b2c277f64a5fa7572effb786fa9a1/test/test_config.rb#L379

jacobherrington (commented):
I don't have much useful guidance on the implementation, but I was able to make the test you added pass with a change that might help you find the right path for this feature.

lib/puma/dsl.rb
rodzyn: Thanks for the feedback! When you scroll through the history of this PR you'll actually see that was my initial implementation. I've the a suggestion that it might be better to use Puma event system instead

test/test_config.rb
rodzyn: This is leftover from the initial implementation

nateberkopec:
Needs a test and I think it's good to go?

lib/puma/dsl.rb
```diff
@@ -588,6 +592,17 @@ def after_worker_fork(&block)
       @options[:after_worker_fork] << block
     end
 
+    # Code to run after puma is booted in a single mode
+    #
+    # @note Single mode only.
```
johnnyshields, line 597: Why this would be Single mode only? Cluster mode also calls `fire(:on_booted)`

lib/puma/dsl.rb
```diff
@@ -251,6 +251,10 @@ def bind_to_activated_sockets(bind=true)
       @options[:bind_to_activated_sockets] = bind
     end
 
+    def events(events)
+      @options[:events] = events
+    end
```
dentarg, line 256: Do we need to add `events` to the DSL? Should we document it? Looks like some kind of hack to me? :) Should it be a public thing in the DSL?

lib/puma/dsl.rb
dentarg: Johnny is correct:

https://github.com/puma/puma/blob/5f255fcd132a3d67608a3d82b31c540bdbf69bcf/lib/puma/cluster.rb#L468 https://github.com/puma/puma/blob/5f255fcd132a3d67608a3d82b31c540bdbf69bcf/lib/puma/cluster.rb#L477

I guess the docs was written like this because the request (#2635, https://github.com/puma/puma/issues/1230#issuecomment-839953916) came from a user using single mode where `after_worker_boot` (aliased to `after_worker_fork`) doesn't exist. https://github.com/puma/puma/blob/5f255fcd132a3d67608a3d82b31c540bdbf69bcf/lib/puma/cluster.rb#L104

lib/puma/dsl.rb
dentarg: Should the other DSL objects also have access to `@events`?

https://github.com/puma/puma/blob/5f255fcd132a3d67608a3d82b31c540bdbf69bcf/lib/puma/configuration.rb#L146-L148

johnnyshields:
We just merged https://github.com/puma/puma/pull/2798 which affects events, so please pull latest master to this branch.

lib/puma/dsl.rb
rodzyn: It was long time ago, but I think it was result of this discussion: https://github.com/puma/puma/pull/2709#discussion_r713992970

It was made as a temporary measure, to be refactored later I guess. I'm not sure if that is still right approach

rodzyn:
Sorry for abandoning this work for a long time. I had a very busy personal life in the past few months. I'm back with more time so I did a rebase. I will also try to look once again into a solution.

NickLarsenNZ:
Any update on this?

indirect:
If you don't have multiple processes, can't you just... run your code directly in `puma.rb`? For example, it seems like I have embedded Sidekiq working in a single non-worker Puma process like this:

```ruby
# [the default rails puma.rb file goes here]

require "sidekiq"
require_relative "initializers/sidekiq"

sidekiq = Sidekiq.configure_embed do |config|
  # config.logger.level = Logger::DEBUG
  config.queues = %w[critical default low]
  config.concurrency = 2
end
sidekiq.run

at_exit do
  sidekiq&.stop
end
```

axos88:
Not necessarily. My use case involves running both single process and multi process based on environment. And I need the on booted event to fire correctly in both cases.

lib/puma/dsl.rb
rodzyn: Thanks for catching this, correct it runs for both, clustered and single mode, I've changed the description. The question is, what would be the expected behavior? At the end the name of the event is `on_booted` so I think it makes sense to be run on both: single and clutered. Any thoughts @nateberkopec ?

lib/puma/dsl.rb
rodzyn: I have changed the behavior, so now there is no need for this method and the `Puma::Events` instance is passed early to `Puma::Configuration` so it can be used by any type of DSL if needed

nateberkopec:
Looks like you've done a git-whoopsie and deleted your test.

nateberkopec (changes requested):
Fix tests

rodzyn:
@nateberkopec @dentarg - feel free to take a look if you have time whether these changes make sense

nateberkopec (approved):

nateberkopec:
Merged in #7c8cb02

nateberkopec:
Crud, I screwed up the merge. @MSP-Greg can you take a look and try? the changes to rack/handler meant that my merge didn't work.

MSP-Greg:
Looking at it now.

MSP-Greg:
Test are running.  I think the Windows mswin job will fail, it may be a recent change in Ruby master, re an error message.  All the Actions master builds are done at different times, so they're not all the same commit in ruby/ruby...
```
I think the older msg is:

undefined method `to_i' for an instance of Array

newer:

undefined method `to_i' for [0, 1]:Array
```

MSP-Greg:
Done...

nateberkopec:
🙇 Thanks!

Issue #2708, opened 2021-09-22
ssl_bind not compatible with localhost integration


erikaxel:
Hi,

I'm really happy that it is now (from v 5.5.0) possible to run Puma with self signed certificates generated by the localhost gem, However, I met a small bump trying to set it up.

I can run self signed SSL from the command line like this:
```puma -b 'ssl://localhost:9292' config.ru```
I would therefore expect one of the following to work in `puma.rb`:
```
ssl_bind 'localhost', '9292'
ssl_bind 'localhost', '9292', {}
```
However this does not work, I have to set it up like this:
```
  ssl_bind 'localhost', '9292', {
    key: "#{ENV["HOME"]}/.localhost/localhost.key",
    cert: "#{ENV["HOME"]}/.localhost/localhost.crt"
  }
```
I found this approach by reading #2706.

I think it would make sense to either support the first version (best in my opinion) or document the second one.

nateberkopec:
As mentioned in #2096, you must `require localhost` inside your application (in your case, inside `config.ru`).

nateberkopec:
Oh! I'm sorry, I didn't read your report closely enough. You're right, that doesn't work and it totally should. Marking as bug.

rodzyn:
@erikaxel Is it still a bug? I failed to reproduce that in the latest master

dentarg:
@rodzyn How did you try?

Using current master branch at 909f51e711ceada8e34820fdb9f07148281984a5, this still fails for me:

<Details>
<Summary><code>ssl_bind "localhost", "9292"</code></Summary>

```shell
$ cat test/config/ssl_self_signed_config.rb
require "localhost"
ssl_bind "localhost", "9292"
app do |env|
  [200, {}, ["embedded app"]]
end

$ b e bin/puma -C test/config/ssl_self_signed_config.rb
bundler: failed to load command: bin/puma (bin/puma)
/Users/dentarg/src/puma/lib/puma/dsl.rb:465:in `ssl_bind': wrong number of arguments (given 2, expected 3) (ArgumentError)
  from test/config/ssl_self_signed_config.rb:2:in `_load_from'
  from /Users/dentarg/src/puma/lib/puma/dsl.rb:81:in `instance_eval'
  from /Users/dentarg/src/puma/lib/puma/dsl.rb:81:in `_load_from'
  from /Users/dentarg/src/puma/lib/puma/configuration.rb:216:in `block in load'
[...]

PR #2711, opened 2021-09-23
Support `localhost` integration in `ssl_bind`

rodzyn:
### Description
Hello, here is my attempt to fix https://github.com/puma/puma/issues/2708

Closes https://github.com/puma/puma/issues/2708

### DIFF
--- a/lib/puma/binder.rb
+++ b/lib/puma/binder.rb
@@ -57,7 +57,6 @@ module Puma
 
       @envs = {}
       @ios = []
-      localhost_authority
     end
 
     attr_reader :ios
@@ -229,11 +228,6 @@ module Puma
 
           params = Util.parse_query uri.query
 
-          # If key and certs are not defined and localhost gem is required.
-          # localhost gem will be used for self signed
-          # Load localhost authority if not loaded.
-          ctx = localhost_authority && localhost_authority_context if params.empty?
-
           ctx ||= MiniSSL::ContextBuilder.new(params, @events).context
 
           if fd = @inherited_fds.delete(str)
@@ -246,9 +240,11 @@ module Puma
             ios_len = @ios.length
             io = add_ssl_listener uri.host, uri.port, ctx
 
+            uri_query = params.map { |k, v| "#{k}=#{v}" }.join('&')
+
             @ios[ios_len..-1].each do |i|
               addr = loc_addr_str i
-              logger.log "* #{log_msg} on ssl://#{addr}?#{uri.query}"
+              logger.log "* #{log_msg} on ssl://#{addr}?#{uri_query}"
             end
           end
 
@@ -292,22 +288,6 @@ module Puma
       end
     end
 
-    def localhost_authority
-      @localhost_authority ||= Localhost::Authority.fetch if defined?(Localhost::Authority) && !Puma::IS_JRUBY
-    end
-
-    def localhost_authority_context
-      return unless localhost_authority
-
-      key_path, crt_path = if [:key_path, :certificate_path].all? { |m| localhost_authority.respond_to?(m) }
-        [localhost_authority.key_path, localhost_authority.certificate_path]
-      else
-        local_certificates_path = File.expand_path("~/.localhost")
-        [File.join(local_certificates_path, "localhost.key"), File.join(local_certificates_path, "localhost.crt")]
-      end
-      MiniSSL::ContextBuilder.new({ "key" => key_path, "cert" => crt_path }, @events).context
-    end
-
     # Tell the server to listen on host +host+, port +port+.
     # If +optimize_for_latency+ is true (the default) then clients connecting
     # will be optimized for latency over throughput.
@@ -348,7 +328,7 @@ module Puma
 
       raise "Puma compiled without SSL support" unless HAS_SSL
       # Puma will try to use local authority context if context is supplied nil
-      ctx ||= localhost_authority_context
+      ctx ||= MiniSSL::ContextBuilder.new({}, @events).context
 
       if host == "localhost"
         loopback_addresses.each do |addr|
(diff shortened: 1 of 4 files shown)


lib/puma/dsl.rb
```diff
@@ -46,6 +46,14 @@ def self.ssl_bind_str(host, port, opts)
         else ''
         end
 
+      if ['127.0.0.1', 'localhost'].include?(host)
+        key = "#{ENV["HOME"]}/.localhost/localhost.key"
+        cert = "#{ENV["HOME"]}/.localhost/localhost.crt"
```
dentarg, line 51: Just in case the paths ever change, I think we should do what we do in the Binder class does and ask the localhost gem for the paths (and only fallback to static paths if the gem can't answer that):

https://github.com/puma/puma/blob/af3675b2ee5485ec4fc62588869c41a5d5e278f7/lib/puma/binder.rb#L299-L309

(it does have fallback

dentarg (commented):
Is it possible to add a test for this?

lib/puma/dsl.rb
dentarg: Would be good to not repeat the same code twice, but I'm not sure where would be a good place to share this code between Binder and DSL. @ye-lin-aung might have ideas?

lib/puma/dsl.rb
```diff
@@ -46,6 +46,14 @@ def self.ssl_bind_str(host, port, opts)
         else ''
         end
 
+      if ['127.0.0.1', 'localhost'].include?(host)
```
nateberkopec, line 49: Needs ipv6 detection too. I think we do this somewhere else (`const.rb`? dont rememebr)

lib/puma/dsl.rb
nateberkopec: Agreed on both points

lib/puma/dsl.rb
```diff
@@ -46,6 +46,14 @@ def self.ssl_bind_str(host, port, opts)
...
+        key = "#{ENV["HOME"]}/.localhost/localhost.key"
+        cert = "#{ENV["HOME"]}/.localhost/localhost.crt"
+      end
+
+      key = opts[:key] if opts[:key]
```
nateberkopec, line 54: rather than assign a new variable, just re-use opts I think

lib/puma/dsl.rb
nateberkopec: Also the PR as-written I think does not respect the same conditions as the code in `Binder` - dev environment only, and localhost must be defined.

nateberkopec (changes requested):
Thank you for your contribution! Agreed with all @dentarg's points and added a few of my own.

lib/puma/dsl.rb
rodzyn: Yeah, I was struggling with that a bit, because DSL is evaluated before rackup file is loaded. So, even if we have `require "localhost/authority"` in `.ru` file we can't check it on DSL because the file is not loaded yet. Any suggestions on how to tackle this problem?

rodzyn:
@nateberkopec @dentarg 
I made some more bold changes in this approach. It's more of a draft for now so the code is messy but I just wanted to double-check if that might be a good direction to follow.

It still needs some additional tests to be added (I only tested it on my local example-apps)

lib/puma/binder.rb
```diff
@@ -246,9 +240,11 @@ def parse(binds, logger, log_msg = 'Listening')
             ios_len = @ios.length
             io = add_ssl_listener uri.host, uri.port, ctx
 
+            uri_query = params.map { |k, v| "#{k}=#{v}" }.join('&')
+
             @ios[ios_len..-1].each do |i|
               addr = loc_addr_str i
-              logger.log "* #{log_msg} on ssl://#{addr}?#{uri.query}"
+              logger.log "* #{log_msg} on ssl://#{addr}?#{uri_query}"
```
nateberkopec, line 247: Why change this? Looks to me like `uri.query` is still available?

nateberkopec:
LGTM, tests is the only thing it really needs 👍

lib/puma/binder.rb
rodzyn: So `uri.query` as `uri` comes from https://github.com/puma/puma/blob/master/lib/puma/dsl.rb#L66-L67 (where there is no possibility to detect localhost gem, that was in my try in the first attempt) so it gets the URI with empty `key` and `cert` values (`&key=&cert=`).
These empty values are transformed to Hash and stored in `params` variable here https://github.com/puma/puma/blob/master/lib/puma/binder.rb#L230, and then the current implementation is transforming them (default localhost gem paths). In the end only `params` is "enriched" by new values, thus I'm transforming it back from Hash to String

test/test_integration_ssl.rb
```diff
@@ -54,14 +46,35 @@ def generate_config(opts = nil)
 end
 RUBY
 
+    @localhost_config = <<RUBY
+require 'localhost/authority'
```
dentarg, line 50:
```suggestion
require 'localhost'
```

test/test_integration_ssl.rb
```diff
@@ -81,12 +94,23 @@ def stop_server
...
+  def test_ssl_run_with_localhost_authority
+    body = nil
+    start_server("#{BASE} bin/puma -C #{generate_config(@localhost_config)}")
+    @http.start do
+      req = Net::HTTP::Get.new '/', {}
+      @http.request(req) { |resp| body = resp.body }
+    end
+    assert_equal 'https', body
+    stop_server
+  end unless ::Puma::IS_JRUBY
```
dentarg, line 115: I would prefer `skip ...` or `skip_if` over so we can see from the test output that we are lacking support in JRuby. I also think we need to skip Windows? ([reference](https://github.com/socketry/localhost/blob/664d28071e458375f6f378c48897820a84658d40/.github/workflows/development.yml#L12-L20))

dentarg: My bad, looks like Windows works: https://github.com/puma/puma/runs/3796685009?check_suite_focus=true#step:9:373 

I was thinking `local_certificates_path = File.expand_path("~/.localhost")` wouldn't work in Windows, but that code is probably never exercised because the localhost gem can respond to the `_path` methods

nateberkopec (approved):

dentarg:
@rodzyn After https://github.com/puma/puma/pull/2728 this only needed a small change so I made it in https://github.com/puma/puma/pull/2764 and added you as co-author

PR #2715, opened 2021-09-25
Clean up some duplicated code

jacobherrington:
### Description
Looking through this file, I noticed some identical code that could be cleaned up.

This is only better subjectively, but I didn't see anything re: refactors in the contributing doc. If Puma tends to avoid style-oriented refactors like this one without a strong rationale, feel free to close this PR.

### DIFF
--- a/lib/puma/runner.rb
+++ b/lib/puma/runner.rb
@@ -15,6 +15,16 @@ module Puma
       @app = nil
       @control = nil
       @started_at = Time.now
+      @wakeup = nil
+    end
+
+    def wakeup!
+      return unless @wakeup
+
+      @wakeup.write "!" unless @wakeup.closed?
+
+    rescue SystemCallError, IOError
+      Thread.current.purge_interrupt_queue if Thread.current.respond_to? :purge_interrupt_queue
     end
 
     def development?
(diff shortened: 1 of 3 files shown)


lib/puma/runner.rb
```diff
@@ -15,6 +15,17 @@ def initialize(cli, events)
       @app = nil
       @control = nil
       @started_at = Time.now
+      @wakeup = nil
+    end
+
+    def wakeup!
+      return unless @wakeup
+
+      begin
```
MSP-Greg, line 24: Maybe remove `begin` and move the rescue to 'method level'?

lib/puma/runner.rb
```diff
@@ -159,5 +166,13 @@ def start_server
       server.inherit_binder @launcher.binder
       server
     end
+
+    private
+
+    def ensure_output_directory_exists(path)
```
MSP-Greg, line 172: Maybe something like:
```ruby
def ensure_output_directory_exists(path, io_name)
```
and pass `'STDOUT'` or `'STDERR'` to it, along with changing:
```ruby
raise "Cannot redirect #{io_name} to #{path}"

MSP-Greg:
> If Puma tends to avoid style-oriented refactors

Fine by me, not sure about others.  Most older repos always have code that can be refactored.  And, ruby/ruby is always refactoring...

jacobherrington:
@MSP-Greg Thanks for the review, I pushed a new commit including your suggestions!

nateberkopec:
We're 100% ok with refactors here. I wouldn't even call this style-oriented, you're removing code, not just changing quotation marks and whitespace.

PR #2717, opened 2021-09-29
Update instructions for GitHub actions [ci skip]

jacobherrington:
### Description
I think the location of this setting moved!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -168,7 +168,7 @@ help.
 
 There's no need to update the changelog ([`History.md`](History.md)); that is done [when a new release is made](Release.md).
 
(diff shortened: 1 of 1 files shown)


MSP-Greg:
In my fork, I have to go to the 'Settings' tab (uri is the repo uri with `/settings` added), then click the 'Actions' tab on the left.

Do you have it in a different place?

jacobherrington:
By default, my fork had all of those options "enabled," but I had to navigate to `https://github.com/jacobherrington/puma/actions` and affirm a prompt to *actually* enable.

It is possible that this is only true for newer forks?

dentarg:
It is true for the first time after forking, just tried it with another GitHub account. 

If you disable actions (via Settings) you need to enable it via Settings, but then you know how to find it because you managed to disable, so I don't think we need any guidance about that. :)

PR #2719, opened 2021-10-05
Support for cert and key objects with ssl_bind DSL

dalibor:
### Description

We need a way to specify cert and key objects in Puma configuration without relaying on file paths. The use-case is when deploying to cloud providers and fetching certificates from Secrets Managers on application boot-up to avoid persist the certificates on disk for security reasons.

### DIFF
--- a/ext/puma_http11/mini_ssl.c
+++ b/ext/puma_http11/mini_ssl.c
@@ -199,6 +199,9 @@ sslctx_alloc(VALUE klass) {
   return TypedData_Wrap_Struct(klass, &sslctx_type, ctx);
 }
 
+X509 *GetX509CertPtr(VALUE);
+EVP_PKEY *GetPrivPKeyPtr(VALUE);
+
 VALUE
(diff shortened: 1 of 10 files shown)


MSP-Greg:
@dalibor

Thanks for the update.  This looks good, and it also seems to restart fine.  A long time I thought it would be helpful, but was busy with other things...

As to the API, it's a problem with Puma, as a lot of things have never been marked as private (either by actually doing so or comments).  There are also some things that have been done just to make CI easier, but not noted as such.

A while ago I cleaned up a pattern that I don't like:
```ruby
t = SomeClass.new opts
t.x = opts.x
t.y = opts.y
t.run
```

Well, that broke code in [capybara](https://github.com/teamcapybara/capybara/blob/5e51607fde5e47c0e91cacf5b7c9a45d3efcfb1a/lib/capybara/registrations/servers.rb#L13-L45).  This code will not break that, but it does change the `Binder#listeners` signature.

Anyone have any thoughts re the API change?

PR #2730, opened 2021-10-27
[jruby] a couple refactorings - avoid copy-ing bytes

kares:
### Description
Was reviewing the JRuby native bits and noticed a few low hanging fruits around unnecessary byte[] copying.
The `String` instances retrieved from `read` / `extract` will now be ASCII-8BIT just like the C bits (w `rb_str_new`).

Also, the exception handling seems to be incomplete thus it would be better to allow to raise the original exception instead of dummy wrapping into a Java `RuntimeException`, and where wrapping is present let's have the original Java cause around.


nateberkopec:
Thanks!

PR #2733, opened 2021-10-27
Keep thread names under 15 characters

ob-stripe:
### Description
Keep thread names under 15 characters.

Linux thread names are limited to 15 characters. Trying to set a longer name will just silently do nothing 😡 

This PR ensures that all threads created by Puma have names that are 15 characters or shorter. Who needs vowels anyway.

### DIFF
--- a/lib/puma/runner.rb
+++ b/lib/puma/runner.rb
@@ -69,7 +69,7 @@ module Puma
 
       control.binder.parse [str], self, 'Starting control server'
 
-      control.run thread_name: 'control'
+      control.run thread_name: 'ctl'
       @control = control
     end
 
--- a/lib/puma/cluster/worker.rb
+++ b/lib/puma/cluster/worker.rb
@@ -34,7 +34,7 @@ module Puma
         Signal.trap "SIGCHLD", "DEFAULT"
 
        Thread.new do
-          Puma.set_thread_name "worker check pipe"
+          Puma.set_thread_name "wrkr check"
           @check_pipe.wait_readable
           log "! Detected parent died, dying"
           exit! 1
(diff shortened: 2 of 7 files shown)


lib/puma/runner.rb
```diff
@@ -69,7 +69,7 @@ def start_control
 
       control.binder.parse [str], self, 'Starting control server'
 
-      control.run thread_name: 'control'
+      control.run thread_name: 'ctl'
```
nateberkopec, line 72: Suggestion: lets not change any thread names that are always less than 15 characters.

ob-stripe: The problem is that this thread name is also used as a prefix for its accompanying threadpool threads. So if we don't shorten it, we end up with names like `puma control tp 001` which is longer than 15 characters.

ob-stripe:
I added a Linux-only test that uses `ps` to assert that the thread name is correctly set at the OS level.

dentarg:
Looks like the added test doesn't work on jruby/truffleruby (but do in their -head versions? 🤔) Even fails when running only that test `bundle e m test/test_thread_pool.rb:65` (testing in `docker run --rm -it -v $PWD:/app -w /app ghcr.io/flavorjones/truffleruby:21.3.0 bash`), any ideas @ob-stripe?

dentarg:
Ah I guess we can't see the thread names in `/proc` when we run on the JVM

dentarg:
Okay, always one step behind @MSP-Greg (https://github.com/puma/puma/commit/fa178f3c829610d82b9c5013db748ff0207e5773) 😄

ob-stripe:
Thanks! I was struggling to get the test suite to pass with all different environments.

PR #2737, opened 2021-10-30
Do not run rubocop on non-CRuby

eregon:
It does not seem useful (redundant with running it on CRuby), makes builds slower, and also sometimes makes builds fail: https://github.com/puma/puma/runs/4047984193?check_suite_focus=true

### DIFF
--- a/.github/workflows/non_mri.yml
+++ b/.github/workflows/non_mri.yml
@@ -62,10 +62,6 @@ jobs:
       - name: compile
         run:  bundle exec rake compile
 
-      - name: rubocop
-        if: endsWith(matrix.ruby, '-head') == false
-        run: bundle exec rake rubocop
-
       - name: test
         id: test
         timeout-minutes: 12


dentarg (approved):
I agree

dentarg:
I'll note that this has been discussed before at https://github.com/puma/puma/pull/2198#discussion_r396870521 but I think we prefer green (and faster) tests over keeping it the same.

dentarg:
> sometimes makes builds fail: https://github.com/puma/puma/runs/4047984193?check_suite_focus=true

Do you have any idea as to why @eregon?

eregon:
> I'll note that this has been discussed before at [#2198 (comment)](https://github.com/puma/puma/pull/2198#discussion_r396870521) but I think we prefer green (and faster) tests over keeping it the same.

Looks like the conclusion was the same: only run it on CRuby: https://github.com/puma/puma/pull/2198/files#diff-ce0dee93a528f6c7648f4cd671a3ec7be6a80f976c88f3aa1c322b7a80828fbaR46

> Do you have any idea as to why @eregon?

Not sure, maybe it's a out of memory error. It seems GitHub Actions is really terse when a process gets killed.

nateberkopec:
Thanks for all fixes @eregon

PR #2746, opened 2021-11-02
[CI] Fix Timeout Errors when reading responses

MSP-Greg:
### Description

1. Currently, integration tests stop the CI if a response read timeout happens.  Catch the error and flunk the test.

2. There are two methods in `test_integration_cluster.rb` that are used when generating several requests.  One had a read timeout set to 20 seconds, the other (which has failed recently) had the timeout set to the default of 10.  Increase to 15.

These timeout errors seem to have recently increased, may be due to a change in Actions runners...

### DIFF
--- a/test/helpers/integration.rb
+++ b/test/helpers/integration.rb
@@ -165,6 +165,8 @@ class TestIntegration < Minitest::Test
         sleep 0.01
       end
     end
+  rescue Timeout::Error
+    flunk "response read_body timeout error"
   end
 
   # gets worker pids from @server output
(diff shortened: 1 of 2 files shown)


dentarg (approved):

dentarg:
Updating the PR title doesn't seem to change the proposed commit message for the squash commit in the GitHub UI, might be good to know.

MSP-Greg:
Noticed that with another recent PR.  Maybe the PR title is only used if there's more than one commit?

dentarg:
I've changed title with multiple commits but the proposed commit message hasn't changed then, so my guess it isn't propagating everywhere it could. Maybe some developer at GitHub will see our discussion a fix it. :)

PR #2312, opened 2020-07-20
Improve worker shutdown reliability, CI fixes

MSP-Greg:
### Description

CI is unreliable, with random failures across all Ruby types.  Small changes to lib files for shutdown, minor test changes.

Note that incomplete JRuby SSL implementation (`Puma::MiniSSL::Engine` `#init?` & `#shutdown`) will still   intermittently cause CI failure.

### Your checklist for this pull request
- [x] I have reviewed the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.
- [x] I have added an entry to [History.md](../blob/master/History.md) if this PR fixes a bug or adds a feature. If it doesn't need an entry to HISTORY.md, I have added `[changelog skip]` the pull request title.
- [ ] I have added appropriate tests if this PR fixes a bug or adds a feature.
- [x] My pull request is 100 lines added/removed or less so that it can be easily reviewed.
- [ ] If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.
- [ ] If this closes any issues, I have added "Closes `#issue`" to the PR description or my commit messages.
- [x] I have updated the documentation accordingly.
- [x]  All new and existing tests passed, including Rubocop.

PR #2748, opened 2021-11-03
Rescue `Errno::EBADF` in `Client#close`

dentarg:
Spotted happening in CI: https://github.com/puma/puma/runs/4095295281?check_suite_focus=true

    2021-11-03 16:34:54 +0000 Exception handling servers: #<Errno::EBADF: Bad file descriptor>/Users/runner/work/puma/puma/lib/puma/client.rb:163:in `close': Bad file descriptor (Errno::EBADF)
    TestRackServer#test_large_post_body =
            from /Users/runner/work/puma/puma/lib/puma/client.rb:163:in `close'

Related to #2700? (that changed `rescue Object` to `IOError, Errno::EBADF` and re-raises these two)

See also #2312 that added `Errno::EBADF` (and #2745 that improves on that)

### DIFF
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -161,7 +161,7 @@ module Puma
     def close
       begin
         @io.close
-      rescue IOError
+      rescue IOError, Errno::EBADF
         Puma::Util.purge_interrupt_queue
       end
     end


MSP-Greg:
@dentarg

These macOS errors are getting a little frustrating.  I had this and some other changes to test code, running more CI tonite.

MSP-Greg (approved):

dentarg:
@MSP-Greg Feel free to add/change anything in this PR if you want. I created this quickly to be able to discuss it.

Is some of the various exceptions we see in CI the fall out from #2700?

MSP-Greg:
> Is some of the various exceptions we see in CI the fall out from #2700?

I don't think so, but I'll look at it again.  I'm not sure what the cause is, but most of the problems have been in integration tests, which involve sub-processes.  I.think.

PR #2759, opened 2021-11-23
Add worker_check_interval configuration option

ob-stripe:
### Description
This PR adds support for changing the worker check interval. Previously, the interval was hardcoded to 5 seconds.

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -135,7 +135,7 @@ module Puma
     def check_workers
       return if @next_check >= Time.now
 
-      @next_check = Time.now + Const::WORKER_CHECK_INTERVAL
+      @next_check = Time.now + @options[:worker_check_interval]
 
       timeout_workers
       wait_workers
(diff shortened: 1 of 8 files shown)


nateberkopec:
Why did the default interval not work for your use case?

nateberkopec:
Also, you marked this as `WIP` - what remains to be done?

ob-stripe:
> Why did the default interval not work for your use case?

Currently, Stripe runs multiple Puma servers in single mode on each host, using [Einhorn](https://github.com/stripe-archive/einhorn) as the socket manager / controller process. Our monitoring system queries Einhorn every 2 seconds to get the list of all live workers, then queries each worker's control server to compute overall availability.

Einhorn is deprecated and we're in the process of replacing it with Puma cluster mode. While updating the monitoring system to support querying the Puma cluster main process rather than Einhorn, I noticed that worker statuses are only updated every 5 seconds. I raised this internally and there were concerns around changing the granularity from 2 seconds to 5 seconds. For the time being we're monkey-patching Puma's `WORKER_CHECK_INTERVAL` constant but having a dedicated configuration option would be much cleaner.

> Also, you marked this as WIP - what remains to be done?

Nothing, actually. This is ready for review. Happy to answer any questions or address feedback!

PR #2760, opened 2021-11-24
Fix error message in test_verify_fail_if_client_unknown_ca

lucaskanashiro:
### Description

In Ubuntu 22.04 (development release), we have got the following error:
```
  1) Failure:
TestPumaServerSSLClient#test_verify_fail_if_client_unknown_ca [/tmp/autopkgtest.nnOHcD/build.YFu/src/test/test_puma_server_ssl.rb:306]:
Expected /self\ signed\ certificate\ in\ certificate\ chain/ to match # encoding: ASCII-8BIT
\#    valid: true
"OpenSSL certificate verification error: self-signed certificate in certificate chain - 19".
```
Changing the expected error message from "self signed certificate" to "self-signed certificate" fixes the test failure.

### DIFF
--- a/test/test_puma_server_ssl.rb
+++ b/test/test_puma_server_ssl.rb
@@ -303,7 +303,7 @@ class TestPumaServerSSLClient < Minitest::Test
   end
 
   def test_verify_fail_if_client_unknown_ca
-    assert_ssl_client_error_match('self signed certificate in certificate chain', '/DC=net/DC=puma/CN=CAU') do |http|
+    assert_ssl_client_error_match(/self[- ]signed certificate in certificate chain/, '/DC=net/DC=puma/CN=CAU') do |http|
       key = "#{CERT_PATH}/client_unknown.key"
       crt = "#{CERT_PATH}/client_unknown.crt"
       http.key = OpenSSL::PKey::RSA.new File.read(key)


MSP-Greg:
@lucaskanashiro

Thanks for the PR.

1. I've tested Puma using OpenSSL 3.0 locally, and that was the only failure, excluding the expected deprecation warning during compile.  Are you testing against 3.0?  If so, do you by chance know how soon various distros will be upgrading to it?

2. Given the above, I think the assert needs to pass both 'self signed' and 'self-signed'?  Or, rather than a text match, it needs to be a regex?

lucaskanashiro:
@MSP-Greg thanks for the prompt reply.

>     1. I've tested Puma using OpenSSL 3.0 locally, and that was the only failure, excluding the expected deprecation warning during compile.  Are you testing against 3.0?  If so, do you by chance know how soon various distros will be upgrading to it?

Yes, I am testing against OpenSSL 3. In Ubuntu, we are moving to OpenSSL 3 right now. In Debian, there is an ongoing discussion about it, not sure about the outcome yet. IDK about the other distros.

>     2. Given the above, I think the assert needs to pass both 'self signed' and 'self-signed'?  Or, rather than a text match, it needs to be a regex?

You are right, not everyone will be using OpenSSL 3. In `assert_ssl_client_error_match` method, should we use `assert_match` instead of `assert_equal`?

MSP-Greg:
> we are moving to OpenSSL 3 right now

I hate it when that happens.  Wonder when it will hit WSL2/Ubuntu (what I'm using locally)...  I suppose I could build it, but...

Sorry, I didn't look at `assert_ssl_client_error_match`, but assumed with 'match' in the name.  Yes, please use `assert_match` instead of `assert_equal`.

MSP-Greg (changes requested):
Change to the following should work.  Sorry, intermittently AFK and now with coffee...

```ruby
assert_ssl_client_error_match(/self[- ]signed certificate in certificate chain/, '/DC=net/DC=puma/CN=CAU') do |http|
```

lucaskanashiro:
Thanks @MSP-Greg , I applied the suggested change.

MSP-Greg:
@lucaskanashiro

Thanks.  Hate to ask, but it needs to be a regex, not a string...

lucaskanashiro:
@MSP-Greg sorry :) it's fixed now.

MSP-Greg (approved):

PR #2770, opened 2021-12-14
Custom logger

### LINKED ISSUE #2511: Supporting Custom Logger Class

It is possible to have custom loggers that allow for json log aggregation as well as formatting and custom attributes.

Without json log aggregation results in poorly formatted logs and duplication of log formatting rules.

It would be great to pass a custom logging class to puma that would then be logged to.

I would expect this to result in a new method in the DSL.

vzajkov:
### Description
Closes #2511


MSP-Greg:
@vzajkov

1. Could you please rebase?
2. `test/test_config.rb` runs parallel.  The added test uses `capture_subprocess_io`, which remaps io.  This may intermittently cause other tests to fail.  So, it needs to run in a separate class.  I push a commit that changes it, as below, which is already in master:

https://github.com/puma/puma/blob/65a821aa7eabaddd3039067cd2b51cdd85346dba/test/test_config.rb#L415-L425

MSP-Greg:
@vzajkov I reverted the test changes above so master will pass.  See 65f0ced587

nateberkopec:
In order to merge, all places where Puma currently logs will have to be switched to this new logger object.

johnnyshields:
We just merged https://github.com/puma/puma/pull/2798 which affects logging, so you'll want to pull latest master to this branch.

Overload119:
Hey @vzajkov 

Do you need help with this? Let me know if I can help.

vzajkov:
Hey thanks - I will rebase and let you know if I have any questions. Appreciate it.

lib/puma/log_writer.rb
```diff
@@ -59,6 +59,8 @@ def log(str)
       @stdout.puts(format(str)) if @stdout.respond_to? :puts
 
       @stdout.flush unless @stdout.sync
+
+      @custom_logger.write(format(str)) if @custom_logger
```
nateberkopec, line 63: I think we would want to write to the custom logger _instead_ of stdout, right?

vzajkov: Sorry for the late reply - I think that makes sense, added the change.

phyzical:
any update on this? 

i am looking for a way to filter request logs matching a specific domain was hoping to handle this via a custom logger (unless someone knows of an existing way to achieve this for an app that doesn't use rails)

lib/puma/log_writer.rb
```diff
@@ -57,7 +57,9 @@ def self.null
 
     # Write +str+ to +@stdout+
     def log(str)
-      if @stdout.respond_to? :puts
+      if @custom_logger
+        @custom_logger.write(format(str))
```
dentarg, line 61: Should there be a guard here checking that the custom logger responds to write?

seuros (approved):

svoop:
@vzajkov Thanks a lot for your work on this! Would be really great to get custom loggers and DRY at least my code quite a bit.

nateberkopec:
For watchers of this thread: we cannot merge this because it breaks the tests. If you would like to pick this up and get the tests passing, you can do so.

lib/puma/dsl.rb
```diff
@@ -419,6 +419,11 @@ def log_requests(which=true)
       @options[:log_requests] = which
     end
 
+    # Pass in a custom logging class instance
+    def logger(custom_logger)
+      @options[:logger] = custom_logger
+    end
```
phyzical, line 425:
```suggestion
    def custom_logger(custom_logger)
      @options[:custom_logger] = custom_logger
    end
```

lib/puma/launcher.rb
```diff
@@ -76,6 +76,8 @@ def initialize(conf, launcher_args={})
       @log_writer.formatter = LogWriter::PidFormatter.new if clustered?
       @log_writer.formatter = options[:log_formatter] if @options[:log_formatter]
 
+      @log_writer.custom_logger = options[:logger] if @options[:logger]
```
phyzical, line 79:
```suggestion
      @log_writer.custom_logger = options[:custom_logger] if @options[:custom_logger]
```

test/config/custom_logger.rb
```diff
@@ -0,0 +1,13 @@
...
+  end
+
+  def write(msg)
+    @output.puts 'Custom logging: ' + msg
+    @output.flush
+  end
+end
+
+log_requests
+logger CustomLogger.new(STDOUT)
```
phyzical, line 13:
```suggestion
custom_logger CustomLogger.new(STDOUT)
```

test/test_config.rb
```diff
@@ -555,6 +555,18 @@ def assert_run_hooks(hook_name, options = {})
...
 
+# contains tests that cannot run parallel
+class TestConfigFileSingle < TestConfigFileBase
+  def test_custom_logger_from_DSL
+    conf = Puma::Configuration.new { |c| c.load 'test/config/custom_logger.rb' }
+
+    conf.load
+    out, _ = capture_subprocess_io { conf.options[:logger].write 'test' }
+
+    assert_equal 'Custom logging: test', out
```
phyzical, line 566:
```suggestion
    assert_equal "Custom logging: test\n", out
```

test/test_config.rb
```diff
@@ -555,6 +555,18 @@ def assert_run_hooks(hook_name, options = {})
   end
 end
 
+# contains tests that cannot run parallel
+class TestConfigFileSingle < TestConfigFileBase
+  def test_custom_logger_from_DSL
+    conf = Puma::Configuration.new { |c| c.load 'test/config/custom_logger.rb' }
+
+    conf.load
+    out, _ = capture_subprocess_io { conf.options[:logger].write 'test' }
```
phyzical, line 564:
```suggestion
    out, _ = capture_subprocess_io { conf.options[:custom_logger].write 'test' }
```

phyzical (changes requested):
I was able to get it passing with these changes, i noticed `options[:logger]` is already being used in the dsl.rb though im not sure if that was by design but it seemed that the change in the dsl is what caused all the other tests to output puma startup strings and cause failures

vzajkov:
Thanks @phyzical! I've added your changes but it seems to be failing 1 test after merging in master. 

@nateberkopec Is it possible that this is an intermittent failure?

PR #2773, opened 2021-12-17
Alternate worker culling strategy

ob-stripe:
### Description
This PR adds support for an alternate worker culling strategy: instead of culling the workers at the end of the internal worker list (which in all likelihood are the most recently started workers), setting the new `worker_culling_strategy` configuration option to `oldest` will cull the workers which were started the longest time ago.

This strategy makes it easier to implement a rolling restart of workers. This would be useful for us because our workers tend to leak memory over time. While of course the right thing to do would be write good code that does not leak memory :D in practice it's very hard and replacing workers every so often is a cheap and easy strategy.

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -111,7 +111,14 @@ module Puma
 
       debug "Culling #{diff.inspect} workers"
 
-      workers_to_cull = @workers[-diff,diff]
+      workers_to_cull =
+        case @options[:worker_culling_strategy]
+        when :youngest
+          @workers.sort_by(&:started_at)[-diff,diff]
+        when :oldest
+          @workers.sort_by(&:started_at)[0,diff]
+        end
+
       debug "Workers to cull: #{workers_to_cull.inspect}"
 
       workers_to_cull.each do |worker|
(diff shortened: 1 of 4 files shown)


lib/puma/cluster.rb
```diff
@@ -122,10 +129,10 @@ def cull_workers
...
     # @!attribute [r] next_worker_index
     def next_worker_index
-      all_positions =  0...@options[:workers]
-      occupied_positions = @workers.map { |w| w.index }
-      available_positions = all_positions.to_a - occupied_positions
-      available_positions.first
+      occupied_positions = @workers.map(&:index)
+      idx = 0
+      idx += 1 until !occupied_positions.include?(idx)
+      idx
```
ob-stripe, line 135: The reason I changed this method's implementation is because I ran into an edge case while writing tests for this PR: if `@options[:workers]` is decreased (e.g. because the process received a TTOU signal), the previous implementation could potentially return `nil`.

MSP-Greg (approved):

lib/puma/dsl.rb
```diff
@@ -795,6 +795,32 @@ def worker_shutdown_timeout(timeout)
       @options[:worker_shutdown_timeout] = Integer(timeout)
     end
 
+    # Set the strategy for worker culling.
+    #
+    # There are two possible values:
+    #
+    # 1. **:default** (the default) - the excess workers at the end of the
```
nateberkopec, line 802: Let's call it `:youngest`. More descriptive.

ob-stripe:
Renamed `default` to `youngest` as requested.

PR #2780, opened 2021-12-24
Allow backlog parameter to be set with ssl_bind DSL

dalibor:
### Description

We're adding option for setting the socket backlog value with the `ssl_bind` DSL, and updating the docs with a hint to the `net.core.somaxconn` sysctl value. What's interesting here is that on older Linux kernels it defaults to 128 ([reference](https://github.com/torvalds/linux/blob/ca2ef2d9f2aad7a28d346522bb4c473a0aa05249/Documentation/networking/ip-sysctl.rst#tcp-variables)), so the actual backlog value is capped to it.

> somaxconn - INTEGER
> Limit of socket listen() backlog, known in userspace as SOMAXCONN. Defaults to 4096. (Was 128 before linux-5.4) See also tcp_max_syn_backlog for additional tuning for TCP sockets.

Even with the default backlog value of 1024 (when kernel allows it), we are seeing some requests get dropped (504 errors with AWS ALB). That happens with a super fast end-points that respond in < 5ms occasionally when receiving a blast of requests within few seconds and there are other slower end-points that saturate the queue causing some requests to get dropped. So, I'm tweaking slightly the original doc about `backlog` that were misleading:

> Generally, you'll never hit the backlog cap in production.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -31,10 +31,10 @@ _workers_, and we sometimes call the threads created by Puma's
 ![https://bit.ly/2zwzhEK](images/puma-connection-flow.png)
 
 * Upon startup, Puma listens on a TCP or UNIX socket.
-  * The backlog of this socket is configured (with a default of 1024). The
-    backlog determines the size of the queue for unaccepted connections.
-    Generally, you'll never hit the backlog cap in production. If the backlog is
-    full, the operating system refuses new connections.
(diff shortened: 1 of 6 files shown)


nateberkopec:
Makes total sense. Not having this set on ssl_bind was an oversight and I agree with your docs changes as well.

nateberkopec (approved):

PR #2779, opened 2021-12-24
Patch Rubocop to work with newer Psych versions

dnasevic-godaddy:
### Description

Most recent RubyGems versions go with Psych >= 4.0 that breaks the RuboCop rake task. We use an old RuboCop version to target Ruby 2.2. So, we cannot upgrade RuboCop to latest and we have to patch it to work with the newer Psych versions.

We're also fixing some CI warnings for Ruby versions 2.3 - 2.5 for RubyGems versions having a bug that prevents `required_ruby_version` from working for Bundler.

You can see the failing CI builds here: https://github.com/puma/puma/actions/runs/1619499856.

### Your checklist for this pull request

- [x] I have reviewed the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.
- [x] I have added (or updated) appropriate tests if this PR fixes a bug or adds a feature.
- [x] My pull request is 100 lines added/removed or less so that it can be easily reviewed.
- [x] If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.
- [x] If this closes any issues, I have added "Closes `#issue`" to the PR description or my commit messages.
- [x] I have updated the documentation accordingly.
- [x] All new and existing tests passed, including Rubocop.

PR #2783, opened 2021-12-31
Bump RuboCop to work with newer Psych versions

dentarg:
From https://github.com/puma/puma/pull/2779

>Most recent RubyGems versions go with Psych >= 4.0 that breaks the RuboCop rake task. We use an old RuboCop version to target Ruby 2.2. So, we cannot upgrade RuboCop to latest and we have to patch it to work with the newer Psych versions.
>
>We're also fixing some CI warnings for Ruby versions 2.3 - 2.5 for RubyGems versions having a bug that prevents `required_ruby_version` from working for Bundler.
>
>You can see the failing CI builds here: https://github.com/puma/puma/actions/runs/1619499856.

but instead of patching RuboCop we bump it to the last version that still supports Ruby 2.2.

Close https://github.com/puma/puma/pull/2779

### DIFF
--- a/.github/workflows/mri.yml
+++ b/.github/workflows/mri.yml
@@ -72,8 +72,6 @@ jobs:
         run:  bundle exec rake compile
 
       - name: rubocop
-        # 2021-05-20 - RuboCop won't run with Psych 4.0 - remove when fixed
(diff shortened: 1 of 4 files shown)


dalibor:
@dentarg Thanks, this is great!

PR #2784, opened 2022-01-01
Remove yaml (psych) requirement in StateFile

MSP-Greg:
### Description

Update code and tests for StateFile:

1. remove yaml requirement for lib code
2. improve tests
3. deprecate lib constants only used for testing

### DIFF
--- a/lib/puma/state_file.rb
+++ b/lib/puma/state_file.rb
@@ -1,15 +1,40 @@
 # frozen_string_literal: true
 
-require 'yaml'
-
 module Puma
+
+  # Puma::Launcher uses StateFile to write a yaml file for use with Puma::ControlCLI.
+  #
+  # In previous versions of Puma, YAML was used to read/write the state file.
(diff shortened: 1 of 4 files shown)


lib/puma/state_file.rb
```diff
@@ -1,15 +1,37 @@
 # frozen_string_literal: true
 
-require 'yaml'
-
 module Puma
+
+  # In previous versions of Puma, YAML was used to read/write the state file.
```
nateberkopec, line 5: We should be clearer if this is a refactor or if the state files are incompatible across versions I think

nateberkopec (approved):

lib/puma/state_file.rb
MSP-Greg: There is no change to the state file from past versions.  Since we're just using numeric and string values, writing it with code was QED.  The tests are still using yaml/psych to parse the generated file.

Should I add something stating no changes from previous?

MSP-Greg: I updated the class comments?

PR #2791, opened 2022-01-08
fork_worker! should not require worker.phase == 0

johnnyshields:
In `Puma::Cluster`, currently `fork_worker!` requires `worker.phase == 0` to be invoked. This looks incorrect based on a reading of the code. (I have not actually observed this failing in production.)

If worker0 (i.e. the worker at index 0) dies, `spawn_worker` will create a new worker at index 0 with the latest `@phase` value. If happens after any worker forking or phase restart, then `fork_worker!` will no longer be invoked.

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -281,7 +281,6 @@ module Puma
         if (fork_requests = @options[:fork_worker].to_i) > 0
           @launcher.events.register(:ping!) do |w|
             fork_worker! if w.index == 0 &&
-              w.phase == 0 &&
               w.last_status[:requests_count] >= fork_requests
           end
         end


lefermierperso:
what the status for this commit? 
I've just add `fork_worker 2500` config to my server but all my workers are in `phase 3` and having `requests_count: 8400`
I'm using puma 5.5.0 and ruby 2.7.4

johnnyshields:
@lefermierperso this was not merged, as it was based on an incorrect understanding. Actually the behavior you are seeing is correct. The auto re-fork of worker 0 intentionally happens only once, unless if you trigger a restart of your server.

lefermierperso:
@johnnyshields thx for your answer, so there no way to automaticaly launch phased-restart after the worker 0 run xxx requests ?

johnnyshields:
@lefermierperso again it happens after the "first xxx requests only". There's no currently no way to do after "each xxx requests".

johnnyshields:
(however, if you wanted to implement that "each xxx mode", it would be the change in this PR.)

nateberkopec:
@lefermierperso Sounds like you might want to check out `puma-worker-killer`

PR #2794, opened 2022-01-09
In fork_worker mode, worker_culling_strategy "oldest" shouldn't cull worker 0

johnnyshields:
PR #2773 introduced `worker_culling_strategy = :oldest` which was designed to improve memory efficiency by culling "leaky" older workers first.

In `:fork_worker` mode however, this option culls worker0 (which tends to be the oldest), which seems contrary to the intention of improving memory efficiency. This will spawn an entirely new worker0 process, and the other childs (n >= 1) will not have any CoW memory share with this process.

Instead, it's better to exclude worker0 from the culling in `:fork_worker` mode.

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -109,21 +109,24 @@ module Puma
       diff = @workers.size - @options[:workers]
       return if diff < 1
 
-      debug "Culling #{diff.inspect} workers"
+      debug "Culling #{diff} workers"
 
-      workers_to_cull =
-        case @options[:worker_culling_strategy]
-        when :youngest
-          @workers.sort_by(&:started_at)[-diff,diff]
-        when :oldest
-          @workers.sort_by(&:started_at)[0,diff]
-        end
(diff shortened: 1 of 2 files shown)


nateberkopec:
Interesting. I guess that makes sense - in fork_worker mode, worker0 is sort of like our master process. Makes sense to me and prevents an accidental "footgun".

lib/puma/cluster.rb
```diff
@@ -108,25 +108,32 @@ def spawn_worker(idx, master)
...
-      workers_to_cull.each do |worker|
+      workers.each do |worker|
         log "- Worker #{worker.index} (PID: #{worker.pid}) terminating"
         worker.term
       end
     end
 
+    def workers_to_cull(diff)
+      workers = @workers.sort_by(&:started_at)
+      workers.reject! { |w| w.index == 0 } if @options[:fork_worker]
```
nateberkopec, line 124: Let's leave a comment here saying the same thing you did - that in fork_worker mode, worker 0 is our "master" and culling it would impact CoW gains

johnnyshields: Done

PR #2799, opened 2022-01-12
Log environment using option value

ags:
### Description

It's possible for the configured environment to be something other than `RACK_ENV`, e.g. `APP_ENV` or `RAILS_ENV`, but the runner will always log the value of `RACK_ENV`.

This changes the log output to show the configured environment value.

See https://github.com/puma/puma/blob/v5.5.2/lib/puma/configuration.rb#L203


nateberkopec:
Makes sense. That's a bug.

PR #2801, opened 2022-01-16
[CI] Update actions, fix intermittent test error

MSP-Greg:
### Description

Currently CI is running on a 'special' branch of MSP-Greg/setup-ruby-pkgs.  It and ruby/setup-ruby have been updated to work with Ruby 3.1 and the new windows-2022 image.  Revert to the normal branches.

Fixes an intermittent error in `test/test_integration_pumactl.rb`

### DIFF
--- a/.github/workflows/mri.yml
+++ b/.github/workflows/mri.yml
@@ -44,14 +44,14 @@ jobs:
         uses: actions/checkout@v2
 
       - name: load ruby
-        uses: MSP-Greg/setup-ruby-pkgs@win-ucrt-2
+        uses: MSP-Greg/setup-ruby-pkgs@v1
         with:
           ruby-version: ${{ matrix.ruby }}
           apt-get: ragel
           brew: ragel
           mingw: openssl ragel
           bundler-cache: true
-          setup-ruby-ref: MSP-Greg/ruby-setup-ruby/win-ucrt-1
+          setup-ruby-ref: MSP-Greg/ruby-setup-ruby/gem-update
         timeout-minutes: 10
 
       # Windows error thrown, doesn't affect CI
--- a/test/test_integration_pumactl.rb
+++ b/test/test_integration_pumactl.rb
@@ -89,7 +89,7 @@ class TestIntegrationPumactl < TestIntegration
 
     _, status = Process.wait2(@pid)
     assert_equal 0, status
-    assert_operator Time.now - start, :<, (DARWIN ? 8 : 5)
+    assert_operator Time.now - start, :<, (DARWIN ? 8 : 6)
     @server = nil
   end
 


dentarg:
What's up with Ruby 2.5 and your repos/branches (`MSP-Greg/setup-ruby-pkgs@win-ucrt-2`, `MSP-Greg/ruby-setup-ruby/win-ucrt-1`)? 

I can't make sense of [the error](https://github.com/puma/puma/runs/4833455128?check_suite_focus=true#step:3:131) but it looks like it happens consistently in that setup.

It doesn't happen with plain ruby/setup-ruby: https://github.com/dentarg/puma/runs/4834198864?check_suite_focus=true#step:9:71 (tested only with Ubuntu)

dentarg:
Would it be possible to only use `ruby/setup-ruby` in Puma? To make it more clear what is needed to run/test Puma on different platforms. We could make use of [reusable workflows](https://docs.github.com/en/actions/learn-github-actions/reusing-workflows#creating-a-reusable-workflow) to re-use things.

MSP-Greg:
> What's up with Ruby 2.5

It appears to be a weird (and intermittent) error due to old RubyGems being incompatible with new Psych.  The reason it wasn't failing for a while was that the original code I had was doing `gem update --system` for Rubies 2.3 thru 2.5 instead of just installing Bundler, as is now done with code in ruby/setup-ruby.   The branch I used here was changed since it was also the branch used for the PR in setup-ruby.  I'm trying to get a repo case...

> Would it be possible to only use ruby/setup-ruby in Puma?

Depends.  To test Windows Rubies 2.2 thru 3.1 (or master) requires three OpenSSL versions and two ragel versions.  The other OS's are simpler.  The readme could state that ragel and openssl are needed.

dentarg:
>It appears to be a weird (and intermittent) error due to old RubyGems being incompatible with new Psych.

Yeah, but not intermittent? Looks like fails every time now

MSP-Greg:
I really don't know.  Kind of dealing with some normal things.

Take a look at my fork of setup-ruby:
https://github.com/MSP-Greg/ruby-setup-ruby/actions/runs/1705378024

So I pushed it as a PR to setup-ruby, see:
https://github.com/ruby/setup-ruby/actions/runs/1705416015

My fork had the same failure as here, but the PR passed.  I think it's affected by Actions caching.  Not.Sure.

That wasn't using either of my repos (the fork or MSP-Greg/setup-ruby-pkgs).

MSP-Greg:
Screw it, I just added the code to run `gem update --system` for Rubies 2.3 thru 2.5, as I had in the CI that was running before that worked...

dentarg:
> I think it's affected by Actions caching

You are on to something. See https://github.com/dentarg/puma/commit/4ceaec5a421e6537282b63a337d065bf56737e68 where I switched off cache for the `MSP-Greg/setup-ruby-pkgs` jobs (Ruby 2.5 passed) and switched it on for `ruby/setup-ruby` and Ruby 2.5 failed!

I'll guess we continue to discuss this in https://github.com/ruby/setup-ruby/pull/264

dentarg (approved):

PR #2803, opened 2022-01-18
Adds a --silent option to the CLI

gottfrois:
### Description

This PR adds a `-s`, `--silent` option which deactivate Puma messages displayed on the prompt.

Without the option:

```sh
$ bundle exec bin/puma test/rackup/hello.ru
* Puma version: 5.5.2 (ruby 2.6.6-p146) ("Zawgyi")
*  Min threads: 0
*  Max threads: 5
*  Environment: development
*          PID: 88245
Use Ctrl-C to stop
=== puma shutdown: 2022-01-18 16:38:07 +0100 ===
- Goodbye!
```

With the option:

```sh
$ bundle exec bin/puma --silent test/rackup/hello.ru
^C%
```

### DIFF
--- a/lib/puma/cli.rb
+++ b/lib/puma/cli.rb
@@ -185,7 +185,7 @@ module Puma
             user_config.restart_command cmd
           end
 
-          o.on "-s", "--silent", "Do not log prompt messages other than errors" do |arg|
+          o.on "-s", "--silent", "Do not log prompt messages other than errors" do
             @events = Events.new NullIO.new, $stderr
           end
 
--- a/lib/puma/cli.rb
+++ b/lib/puma/cli.rb
@@ -185,8 +185,8 @@ module Puma
             user_config.restart_command cmd
           end
 
-          o.on "-s", "--silent", "Do not log prompt messages" do |arg|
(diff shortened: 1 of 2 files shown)


MSP-Greg:
Do we really want to silence STDERR?

gottfrois:
Good catch, might want to leave the stderr as it is

MSP-Greg:
Using `@events = Events.new NullIO.new, STDERR` seems to work, as errors are written.

I think the test will always pass, as `Puma::Events.null.class` or any call to `Puma::Events.new` will have a class of `Puma::Events`.  

Not sure about this, let's see what others think...

nateberkopec:
Agree with Greg. Also, isn't this option usually called `quiet`?

dentarg:
Puma has `-q, --quiet`, but it is used to silent the request logging (and it is on by default), from `puma --help`,

    -q, --quiet                      Do not log requests internally (default true)

(we have `-v, --log-requests` to turn it on)

gottfrois:
I've reverted the stderr not to be silent as suggested.

PR #2731, opened 2021-10-27
Always send lowlevel_error response to client

baelter:
### Description
Always send what the lowlevel_error handler returns to the client.

Refactors some code out of `Request#handle_request` to a new `Request#write_response` that can be used from `Server` to send the response from `lowlevel_error`
Closes #2341

### Your checklist for this pull request
- [x] I have reviewed the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.
- [x] I have added (or updated) appropriate tests if this PR fixes a bug or adds a feature.
- [x] My pull request is 100 lines added/removed or less so that it can be easily reviewed.
- [x] If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.
- [x] If this closes any issues, I have added "Closes `#issue`" to the PR description or my commit messages.
- [x] I have updated the documentation accordingly.
- [x] All new and existing tests passed, including Rubocop.

PR #2809, opened 2022-01-26
Revert "Always send lowlevel_error response to client (#2731)"

dentarg:
This reverts commit f8acac1f0702fea1a4f88d68a40bb2f53650b14c.

In response to https://github.com/puma/puma/issues/2808

### DIFF
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -46,7 +46,11 @@ module Puma
       env[HIJACK_P] = true
       env[HIJACK] = client
 
-      env[RACK_INPUT] = client.body
+      body = client.body
+
+      head = env[REQUEST_METHOD] == HEAD
+
+      env[RACK_INPUT] = body
       env[RACK_URL_SCHEME] ||= default_server_port(env) == PORT_443 ? HTTPS : HTTP
 
       if @early_hints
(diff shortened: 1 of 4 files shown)


olleolleolle:
@dentarg The issue with truffleruby there, could that be sidestepped by a `gem install bundler` _just_ for that Ruby, in that job?

dentarg:
@olleolleolle Maybe, but I don't think we should need to do that? On the other hand, all MRI jobs have this the argument `setup-ruby-ref: MSP-Greg/ruby-setup-ruby/gem-update` to `MSP-Greg/setup-ruby-pkgs@v1` whereas non-MRI jobs don't have that

 Ping @MSP-Greg

dentarg:
Hmm, it failed with plain ruby/setup-ruby too: https://github.com/dentarg/puma/runs/4971018457?check_suite_focus=true

@eregon Is this an issue in ruby/setup-ruby or with TruffleRuby?

eregon:
Already reported as https://github.com/oracle/truffleruby/issues/2586.
Seems the combination of latest bundler + truffleruby 22.0 is problematic (truffleruby-head + bundler shipped with truffleruby works, 21.3 + latest bundler works too).

PR #2832, opened 2022-02-16
User supplied host overrides config file's port

waiting-for-dev:
### Description
I'm starting a WIP PR to discuss some apparently weird behavior. When a user supplies a host but no port, the port specified in puma's config file is ignored.

I added a failing test to show the issue. I wonder if that's the desired behavior, as the [responsible code](https://github.com/puma/puma/blob/2d5582a60bd9e896652dfeb1a0b8ad4a598aec72/lib/rack/handler/puma.rb#L100-L102) looks pretty intentional.

To reproduce it on Rails:
- Change the default port on `config/puma.rb` to `2000`.
- Run `bin/rails server --binding 127.0.0.1`
- The server will listen to `http://127.0.0.1:3000` instead of `http://127.0.0.1:2000`

### DIFF
--- a/test/test_rack_handler.rb
+++ b/test/test_rack_handler.rb
@@ -106,6 +106,31 @@ class TestUserSuppliedOptionsHostIsSet < Minitest::Test
   end
 end
 
+class TestUserSuppliedOptionsHostDontOverridePort < Minitest::Test
+  def setup
+    @options = {}
+    @options[:user_supplied_options] = [:Host]
+  end
+
+  def test_port_from_config_wins_over_default
+    file_port = 6001
+
+    Dir.mktmpdir do |d|
+      Dir.chdir(d) do
+        FileUtils.mkdir("config")
+        File.open("config/puma.rb", "w") { |f| f << "port #{file_port}" }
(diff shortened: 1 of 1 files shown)


waiting-for-dev:
Thanks for your answer, @nateberkopec.

>  When you pass something into the Rack handler, we assume that is a command line option to a utilty like rails server and it takes precedence.

Yes, I get that. However, the issue is that when **HOST** is supplied by a user, **PORT** is overridden. I guess we shouldn't mix them! On Hanami, we're trying not to have `hanami server` defaults, so we're just sending host if the user provides a host and port if the user provides a port. However, we're experiencing that when we send a host, the port defaults to Rack default port instead of `config/puma.rb`'s port.

nateberkopec:
🤔 I guess that is a little surprising... we've had issues with how host and port are intertwined in Puma's config DSL in the past. Would you want to take a stab at making your test pass? Next version is probably a major so we can break stuff if need be 🤷

waiting-for-dev:
Sure @nateberkopec, I will :+1:

PR #2833, opened 2022-02-22
Periodically send status to systemd

### LINKED ISSUE #2604: Utilize systemd-notify to add status details
I come from a [discussion](https://community.theforeman.org/t/puma-status-monitoring-not-working-at-the-moment/23154) in the Foreman community where we want to get the status for debugging and scaling the system. In the past the [puma-systemd](https://github.com/sj26/puma-plugin-systemd/blob/master/lib/puma/plugin/systemd.rb) plugin was used and now [puma-status](https://github.com/ylecuyer/puma-status), but having this directly in puma without an additional plugin and configuration would be a great feature for us and our users.

We primarily need the number and state of the workers like the systemd plugin had, but the resource usage of the status plugin could also be helpful. Just think about would a sysadmin perhaps wants to see and can influence and provide this information if gathering it is doable without cause performance issues.

We add the status plugin and describe how to use it at the moment or perhaps add some graphical interface for monitoring it into the Foreman webinterface in the future.

**Additional context**
While the considerations are coming from how we use or plan to use it in the Foreman project, I think also other projects and users can benefit from it.

ekohl:
### Description
With systemd it's possible to send a status line. This will show up when users run systemctl status puma.service.

Most of this code is based on puma-plugin-systemd.

At the moment this is completely untested, but I had written some code that I wanted to share. If anyone wants to finish this before I do, they should feel welcome to.

Closes #2604


lib/puma/systemd.rb
```diff
@@ -29,6 +29,29 @@ def start_watchdog
       end
     end
 
+    def start_status_loop
+      instance = self
+      sleep_time = 1.0
+      log "Sending status to systemd every #{sleep_time.round(1)} sec"
+
+      Thread.new do
```
nateberkopec, line 37: Rather than create threads ourselves, `puma` should provide an API that a puma plugin can use to manage this process.

nateberkopec:
Happy to make this happen, but we should take the approach of providing APIs necessary for a plugin to provide this functionality.

lib/puma/systemd.rb
ekohl: That sounds like a good idea because it should also be used to stop. Just like the thread for the watchdog. It may be more important in the test suite, but right now it may be leaking the thread.

And there is actually for plugins some code for plugins with threads:
https://github.com/puma/puma/blob/eb59ed00d71f66766d75d7c40f671d7016b9d5ae/lib/puma/plugin.rb#L60-L71

Perhaps that should be utilized instead?

nateberkopec: TIL 😆 Yup that's fine too.

ekohl:
This is still very much a draft since I didn't actually test this yet, but I took a stab at rewriting the systemd integration as a plugin that is automatically loaded. I wonder if this is a good direction to take. If so, I can clean up the patch and split it in two (first one that transforms it into a plugin, then one that adds status watching).

nateberkopec:
To me this looks great. I think if you rebase and add tests it's good to go.

ekohl:
I've rebased it to resolve conflicts, but I don't have time to write tests right now. If anyone does have the time, I'd highly welcome it.

nateberkopec:
Adding the `contrib-wanted` label.

dentarg:
Closing in favour of https://github.com/puma/puma/pull/3006

PR #2846, opened 2022-03-23
OpenSSL - allow session reuse via a 'reuse' ssl_bind method or bind string query parameter

### LINKED ISSUE #2845: SSL ServerHello should send "session id"
**Describe the bug**

We have set up a Rails server with SSL and peer certificates, and it is working well serving REST requests.  However one client requires us to set a SSL session id, and the log shows that the SSL session id sent in the "ServerHello" message from our application is empty.  I have done a done a ton of searching, but cannot find how to get Ruby/Rails/Puma/OpenSSL to send a "session id" in the "ServerHello" handshake.

Any help greatly appreciated.

**Puma config:**

```ruby
ssl_bind '0.0.0.0', 3000, {
  key: 'key.pem',
  cert: 'cert.pem',
  verify_mode: 'force_peer',
  ca: "cacerts.pem",
}
```

**To Reproduce**

Start the server and make a request.  The "ServerHello" should contain a non-empty "session id".

**Expected behavior**

A "ServerHello" handshake response with a session id.

 - OS: macOS, Linux, and Windows
 - Puma Version 5.6.2 (ruby 3.1.1-p18)

MSP-Greg:
### Description

This PR allows enabling OpenSSL TLSv1.2 session reuse.

This PR only affects code related to https connections and does not change current behavior.

OpenSSL and TLSv1.2 support session reuse via what is referred to as session id.  Opening a tls connection with 'reuse' uses less network resources but requires additional memory on the server.  It may also take cpu time when the session cache is flushed, which happens every 255 connections.

Note that session reuse on TLSv1.3 uses what is referred to as a session ticket, rather than a session id.

Switching session reuse on can be done in a config file, using the `:reuse` key in the options hash passed to [DSL#ssl_bind](https://msp-greg.github.io/puma/Puma/DSL.html#ssl_bind-instance_method).  Using a value of `true` turns on reuse with default OpenSSL setting.  Passing a hash with `:size` and/or `:timeout` keys with integer values uses the values for reuse setup.

There is a single bind query parameter used (which DSL#ssl_bind creates via DSL.ssl_bind_str).  The syntax for it is listed in [MiniSSL::Context](https://msp-greg.github.io/puma/Puma/MiniSSL/Context.html), along with the other query parameters.

Closes #2845.

### DIFF
--- a/ext/puma_http11/mini_ssl.c
+++ b/ext/puma_http11/mini_ssl.c
@@ -210,25 +210,28 @@ sslctx_alloc(VALUE klass) {
 VALUE
 sslctx_initialize(VALUE self, VALUE mini_ssl_ctx) {
   SSL_CTX* ctx;
-
-#ifdef HAVE_SSL_CTX_SET_MIN_PROTO_VERSION
-  int min;
-#endif
   int ssl_options;
   VALUE key, cert, ca, verify_mode, ssl_cipher_filter, no_tlsv1, no_tlsv1_1,
     verification_flags, session_id_bytes, cert_pem, key_pem;
-#ifndef HAVE_SSL_CTX_SET_DH_AUTO
-  DH *dh;
-#endif
   BIO *bio;
   X509 *x509;
   EVP_PKEY *pkey;
-
+#ifdef HAVE_SSL_CTX_SET_MIN_PROTO_VERSION
+  int min;
+#endif
+#ifndef HAVE_SSL_CTX_SET_DH_AUTO
+  DH *dh;
+#endif
 #if OPENSSL_VERSION_NUMBER < 0x10002000L
   EC_KEY *ecdh;
 #endif
+#ifdef HAVE_SSL_CTX_SET_SESSION_CACHE_MODE
+  VALUE reuse, reuse_cache_size, reuse_timeout;
 
-  TypedData_Get_Struct(self, SSL_CTX, &sslctx_type, ctx);
+  reuse = rb_funcall(mini_ssl_ctx, rb_intern_const("reuse"), 0);
+  reuse_cache_size = rb_funcall(mini_ssl_ctx, rb_intern_const("reuse_cache_size"), 0);
+  reuse_timeout = rb_funcall(mini_ssl_ctx, rb_intern_const("reuse_timeout"), 0);
+#endif
 
   key = rb_funcall(mini_ssl_ctx, rb_intern_const("key"), 0);
 
@@ -247,6 +250,8 @@ sslctx_initialize(VALUE self, VALUE mini_ssl_ctx) {
   no_tlsv1 = rb_funcall(mini_ssl_ctx, rb_intern_const("no_tlsv1"), 0);
 
   no_tlsv1_1 = rb_funcall(mini_ssl_ctx, rb_intern_const("no_tlsv1_1"), 0);
+
+  TypedData_Get_Struct(self, SSL_CTX, &sslctx_type, ctx);
 
   if (!NIL_P(cert)) {
     StringValue(cert);
@@ -314,8 +319,6 @@ sslctx_initialize(VALUE self, VALUE mini_ssl_ctx) {
 
   SSL_CTX_set_min_proto_version(ctx, min);
 
-  SSL_CTX_set_options(ctx, ssl_options);
-
 #else
   /* As of 1.0.2f, SSL_OP_SINGLE_DH_USE key use is always on */
   ssl_options |= SSL_OP_NO_SSLv2 | SSL_OP_NO_SSLv3 | SSL_OP_SINGLE_DH_USE;
@@ -326,10 +329,23 @@ sslctx_initialize(VALUE self, VALUE mini_ssl_ctx) {
   if(RTEST(no_tlsv1_1)) {
     ssl_options |= SSL_OP_NO_TLSv1 | SSL_OP_NO_TLSv1_1;
   }
-  SSL_CTX_set_options(ctx, ssl_options);
 #endif
 
-  SSL_CTX_set_session_cache_mode(ctx, SSL_SESS_CACHE_OFF);
+#ifdef HAVE_SSL_CTX_SET_SESSION_CACHE_MODE
+  if (!NIL_P(reuse)) {
+    SSL_CTX_set_session_cache_mode(ctx, SSL_SESS_CACHE_SERVER);
+    if (!NIL_P(reuse_cache_size)) {
+      SSL_CTX_sess_set_cache_size(ctx, NUM2INT(reuse_cache_size));
+    }
+    if (!NIL_P(reuse_timeout)) {
+      SSL_CTX_set_timeout(ctx, NUM2INT(reuse_timeout));
+    }
+  } else {
+    SSL_CTX_set_session_cache_mode(ctx, SSL_SESS_CACHE_OFF);
+  }
+#endif
+
+  SSL_CTX_set_options(ctx, ssl_options);
 
   if (!NIL_P(ssl_cipher_filter)) {
     StringValue(ssl_cipher_filter);
@@ -340,8 +356,7 @@ sslctx_initialize(VALUE self, VALUE mini_ssl_ctx) {
   }
 
 #if OPENSSL_VERSION_NUMBER < 0x10002000L
-  // Remove this case if OpenSSL 1.0.1 (now EOL) support is no
-  // longer needed.
+  // Remove this case if OpenSSL 1.0.1 (now EOL) support is no longer needed.
   ecdh = EC_KEY_new_by_curve_name(NID_X9_62_prime256v1);
   if (ecdh) {
     SSL_CTX_set_tmp_ecdh(ctx, ecdh);
--- a/lib/puma/minissl.rb
+++ b/lib/puma/minissl.rb
@@ -22,6 +22,7 @@ module Puma
         @socket = socket
         @engine = engine
         @peercert = nil
+        @reuse = nil
       end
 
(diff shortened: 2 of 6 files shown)


ext/puma_http11/mini_ssl.c
```diff
@@ -339,20 +344,6 @@ sslctx_initialize(VALUE self, VALUE mini_ssl_ctx) {
...
-#else
-    rb_const_get(rb_cRandom, rb_intern_const("DEFAULT")),
-#endif
-    rb_intern_const("bytes"),
-    1, ULL2NUM(SSL_MAX_SSL_SESSION_ID_LENGTH));
-
-  SSL_CTX_set_session_id_context(ctx,
-                                 (unsigned char *) RSTRING_PTR(session_id_bytes),
-                                 SSL_MAX_SSL_SESSION_ID_LENGTH);
-
```
dentarg, line 355: Something we no longer need?

dentarg:
`PUMA_REUSE_ID` — can we make it less ambiguous by including `TLS` or `SSL` somewhere in that string?

ext/puma_http11/mini_ssl.c
MSP-Greg: I don't think so, but I thought I'd sit on it for a bit...

MSP-Greg:
> `PUMA_REUSE_ID` — can we make it less ambiguous by including TLS or SSL somewhere in that string?

Yeah, thought about that, figured I'd change before (and if) this becomes a real PR.  Maybe start all ssl related compile flags with `PUMA_SSL`, so this would be `PUMA_SSL_REUSE_ID`?

A more general discussion, but the question of what should be 'compile time' options vs runtime options?  This could be runtime, so Puma could have an ssl listener using it, and another not.  But, that keeps adding to the bind parameters, and makes things rather messy...

ext/puma_http11/mini_ssl.c
MSP-Greg: I was wrong.  Removing it causes issues with TLSv1.3 connections.  Kind of odd.  Regardless, I added the code back in...

donv:
Hi!

I tried adding

```ruby
ENV['PUMA_SSL_REUSE_ID'] = 'YES'
gem "puma", "~> 5.0", git: 'https://github.com/MSP-Greg/puma.git'
```

to my Gemfile, but it seems that does not compile the C extension.  Is that right?

Do I need to pack a gemfile, or is there a way to use the granch directly from Git?

donv:
SUCCESS!

I followed the excellent contribution guide and compiled and used the 00_issue-2845 branch, and our tests passed!

It would be awesome if this could be merged to master.

I must ask why this option should not always be enabled?  Why the need for an environment variable?

MSP-Greg:
@donv

Thank you for testing the PR/patch.

> I must ask why this option should not always be enabled?

Good question.  It's a breaking change, but some quick testing showed that some large sites (google, etc) support it.

Caching has a few defaults, some of which could be exposed.  Two in particular are TTL for each session, and the max cache size.  I believe defaults are 300 secs for TTL, and size is 20k.  Some may find those default limits are higher than needed.

Depending on the clients, caching may result in fewer network 'trips', and may reduce CPU time.  But, it also takes memory, and when OpenSSL flushes the cache, other operations may be impacted.

Note that the above does not apply to TLSv1.3 reuse, which needs to be enabled (which this PR does) and defaults to 'stateless' session reuse.  Also, if most connections are using keep-alive, session reuse may not really be beneficial.

Lastly, I'm thinking about using a string like '120,8192' or 't120s8192` to be used for configuring caching, which would set the TTL and max size.  Using zero for the value would use the default.  Not sure whether it should be compile time or runtime.

If anyone has configured other servers for this (nginx, etc) and have any thoughts, looking for input...

dentarg:
I think it should be runtime, if possible, sounds more useful.

Re: nginx, `ssl_session_cache shared:SSL:10m;` ("about 40000 sessions") seems to be a common configuration: https://ssl-config.mozilla.org/#server=nginx&version=1.17.7&config=intermediate&openssl=1.1.1k&guideline=5.6

Some more links

- the nginx docs https://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_session_cache
- a guide https://haydenjames.io/nginx-tuning-tips-tls-ssl-https-ttfb-latency/

MSP-Greg:
@dentarg

Thanks for looking into it, links are helpful.

> I think it should be runtime, if possible, sounds more useful.

Ok.  I'll look into adding to the bind parameters, similar to what I mentioned above.  Probably this weekend.  I may see if ALPN can be added also, tired of seeing the messages every time I run curl...

seriousCookies:
This sounds really great, we need this as well; any chance that this can be merged in sometime soon? :)

MSP-Greg:
@seriousCookies

Sorry for the delay; a few things came up that I needed to work on.  I've started work on changing it to a runtime option.  So, should be a few days...

MSP-Greg:
@donv @seriousCookies

If you have time to check this, the update requires one to add the following line to the ssl_bind hash in one's config file:
```ruby
reuse: 'dflt'
```
or add `&reuse=dflt` to the bind string.

Doing so should turn on session caching/reuse.

lib/puma/minissl.rb
```diff
@@ -289,6 +295,36 @@ def check
...
+        # * 'off' - matches the behavior of Puma 5.6 and earlier.  This is included
+        #   in case reuse 'on' is made the default in future Puma versions.
+        # * 'dflt' - sets session reuse on, with OpenSSL default cache size of
+        #   20k and default timeout of 300 seconds.
+        # * 's,t' - where s and t are integer strings, for size and timeout.
+        # * 's' - where s is an integer strings for size.
+        # * ',t' - where t is an integer strings for timeout.
+        # reuse can also be used as a bind query parameter with the above values.
+        #
+        def reuse=(reuse_str)
```
nateberkopec, line 309: I suggest a different API:

```ruby
def reuse=(reuse, size: nil, timeout: nil)
end
```

We'll do the work of parsing the connection string on the user's behalf in `dsl.rb`. 

Should we also have a separate DSL method for setting this?

MSP-Greg: I'll look at it.

I thought using the same string for a query parameter and `ssl_bind` calls.  Part of that is based on the fact that SSL query strings keep getting longer and longer as we add more options.  And, we need them for restarts...

So, I thought combining size & timeout in one parameter was best?

seriousCookies:
> 

thanks for that, it works! :)

lib/puma/minissl.rb
MSP-Greg: Confused.

> We'll do the work of parsing the connection string on the user's behalf in dsl.rb.

When one has an ssl bind, `Binder#parse` converts the bind query string to a hash via `Util.parse_query`.  Then the 'query' hash is passed to `MiniSSL::ContextBuilder.new` which initializes a `MiniSSL::Context`.  Hence, rather than dsl.rb, do you mean `MiniSSL::ContextBuilder.new`?

> Should we also have a separate DSL method for setting this?

DSL really has only one SSL method, `#ssl_bind`, which calls a convenience method `.ssl_bind_str`. At present, ssl options are passed to `#ssl_bind` as a hash, with the values essentially matching the query parameter string values.  It assembles the bind query string.

MSP-Greg: @nateberkopec Still not quite sure what you're looking for here.  As we keep adding SSL options, the query string gets more 'not human readable'.  The code in this PR was an attempt to keep it so.  It's easy to use SSl's defaults, and people can change the setup if desired...

MSP-Greg: > We'll do the work of parsing the connection string on the user's behalf in dsl.rb.
> Should we also have a separate DSL method for setting this?

dsl.rb has `ssl_bind`, which translates an options hash into a bind string.  So that's an odd place for it.

1. `def reuse=(reuse, size: nil, timeout: nil)` - it's somewhat odd to have a attribute setter accept multiple parameters?
2. The method should probably go in either [MiniSSL::ContextBuilder](https://msp-greg.github.io/puma/Puma/MiniSSL/ContextBuilder.html) or [MiniSSL::Context](https://msp-greg.github.io/puma/Puma/MiniSSL/Context.html). Most of the methods in MiniSSL::Context operate on the query value, and most of the logic is also in that class.  Some is in MiniSSL::ContextBuilder, it would probably be a good idea to refactor a bit.

Thoughts?

nateberkopec: I think at this point, let's not add too many blockers to getting 6.0 out. Do what you think is best 👍

MSP-Greg: Sorry.  I know this is a bit of a PITA, but the ssl bind strings have seen a large increase in size with all the additional SSL options.  At present, the `ssl_bind` call in a config file is just taking simple objects (numbers, strings, boolean) as values in its hash, so currently it looks like:
```ruby
  ssl_bind '#{HOST}', '#{bind_port}', {
    cert: cert,
    key:  key,
    ca: ca,
    verify_mode: 'none',
    reuse: 'dflt'
  }
```
We could change it to the following, where the `size` and `timeout` keys are optional?
```ruby
  ssl_bind '#{HOST}', '#{bind_port}', {
    cert: cert,
    key:  key,
    ca: ca,
    verify_mode: 'none',
    reuse: {use: true, size: 2_000, timeout: 400}
  }
```

What do you think of that?  We still need to encode it to a string and then unpack it to create the SSL context.

MSP-Greg: @nateberkopec I didn't quite use the style shown above, but something similar to what you were looking for.  Changed opening comment.

Issue #1339, opened 2017-06-19
SSL3_GET_CLIENT_HELLO:no shared cipher after updating SSL certificate


jbeyer05:
After updating my SSL certificate, any requests to Puma produce the following error messages from Puma's side:

2017-06-19 12:30:19 -0400: SSL error, peer: 73.10.182.229, peer cert: , #<Puma::MiniSSL::SSLError: OpenSSL error: error:1408A0C1:SSL routines:SSL3_GET_CLIENT_HELLO:no shared cipher - 336109761>
2017-06-19 12:30:19 -0400: SSL error, peer: 73.10.182.229, peer cert: , #<Puma::MiniSSL::SSLError: OpenSSL error: error:1408A0C1:SSL routines:SSL3_GET_CLIENT_HELLO:no shared cipher - 336109761>
2017-06-19 12:30:19 -0400: SSL error, peer: 73.10.182.229, peer cert: , #<Puma::MiniSSL::SSLError: OpenSSL error: error:1408A0C1:SSL routines:SSL3_GET_CLIENT_HELLO:no shared cipher - 336109761>

Both certificates came from the same issuer.  Both seem identical in setup, with the one exception being that the new certificate is for a wildcard domain.

Thanks in advance!

rickpeyton:
I have the same issue.

```
web_1  | 2017-07-24 22:58:10 +0000: SSL error, peer: 69.58.181.15, peer cert: , #<Puma::MiniSSL::SSLError: OpenSSL error: error:140760FC:SSL routines:SSL23_GET_CLIENT_HELLO:unknown protocol - 336027900>
web_1  | 2017-07-24 22:58:18 +0000: SSL error, peer: 69.58.181.15, peer cert: , #<Puma::MiniSSL::SSLError: OpenSSL error: error:1408A10B:SSL routines:SSL3_GET_CLIENT_HELLO:wrong version number - 336109835>
web_1  | 2017-07-24 22:58:26 +0000: SSL error, peer: 69.58.181.15, peer cert: , #<Puma::MiniSSL::SSLError: OpenSSL error: error:1408A0C1:SSL routines:SSL3_GET_CLIENT_HELLO:no shared cipher - 336109761>
[...]

PR #2847, opened 2022-03-26
Fail hard if SSL certs or keys cannot be read by user

stanhu:
### Description

Previously if an SSL cert or key could not be read, Puma would bind to
the configured SSL port but not accept any connections. The only
indication that something went awry is an obscure log message:

```
 #<Puma::MiniSSL::SSLError: OpenSSL error: error:1417A0C1:SSL routines:tls_post_process_client_hello:no shared cipher - 193>
```

We now fail hard with an exception if this happens to make it clear
that the permissions need to be fixed.

Relates to https://github.com/puma/puma/issues/1339

### DIFF
--- a/lib/puma/minissl.rb
+++ b/lib/puma/minissl.rb
@@ -214,6 +214,11 @@ module Puma
         @cert_pem = nil
       end
 
+      def check_file(file, desc)
+        raise ArgumentError, "#{desc} file '#{file}' does not exist" unless File.exist? file
+        raise ArgumentError, "#{desc} file '#{file}' is not readable" unless File.readable? file
+      end
+
       if IS_JRUBY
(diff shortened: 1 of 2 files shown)


MSP-Greg:
@stanhu

Thanks for the PR, good catch/idea.  Maybe add a method for file checking, something like:
```ruby
      def check_file(file, desc)
        raise ArgumentError, "#{desc} file '#{file}' does not exist" unless File.exist? file
        raise ArgumentError, "#{desc} file '#{file}' is not readable" unless File.readable? file
      end
```

Also, perform the same update to the JRuby keystore file?

See https://github.com/MSP-Greg/puma/commit/0b6426635ff7e8885ecf444b1d19fd23ebf138b9, I can push the commit to your branch.

stanhu:
Works for me!

MSP-Greg (approved):

PR #2848, opened 2022-03-27
Fail hard if SSL certs or keys are invalid

stanhu:
### Description

Previously if an invalid SSL cert or key (e.g. blank file, mismatching
private and public key, etc.) were specified, Puma would quietly
accept them, bind to the port, and mysteriously fail with an error
only after a client initiates a connection:

```
SSL error, peer: ::1, peer cert: : #<Puma::MiniSSL::SSLError: OpenSSL error: error:1417A0C1:SSL routines:tls_post_process_client_hello:no shared cipher - 193>
```

We now check the OpenSSL return values and raise an exception if the
cert or key cannot be loaded. For example:

```
puma/lib/puma/minissl.rb:330:in `initialize': SSL_CTX_use_certificate_chain_file: error in file '/tmp/cert.txt': error:0909006C:PEM routines:get_name:no start line (Puma::MiniSSL::SSLError)
```

### DIFF
--- a/ext/puma_http11/mini_ssl.c
+++ b/ext/puma_http11/mini_ssl.c
@@ -30,6 +30,12 @@ typedef struct {
 
 VALUE eError;
 
+NORETURN(void raise_file_error(const char* caller, const char *filename));
+
+void raise_file_error(const char* caller, const char *filename) {
+  rb_raise(eError, "%s: error in file '%s': %s", caller, filename, ERR_error_string(ERR_get_error(), NULL));
+}
+
 void engine_free(void *ptr) {
   ms_conn *conn = ptr;
   ms_cert_buf* cert_buf = (ms_cert_buf*)SSL_get_app_data(conn->ssl);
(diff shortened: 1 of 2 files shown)


MSP-Greg (approved):

dentarg:
Does it make sense to add a test to the test suite for this?

MSP-Greg:
@dentarg @stanhu

Maybe something like the following added to `TestPumaServerSSL` in `test_puma_server_ssl.rb`?
```ruby
  unless Puma.jruby?
    def test_invalid_cert
      assert_raises(Puma::MiniSSL::SSLError) do
        start_server { |ctx| ctx.cert = __FILE__ }
      end
    end

    def test_invalid_key
      assert_raises(Puma::MiniSSL::SSLError) do
        start_server { |ctx| ctx.key = __FILE__ }
      end
    end

    def test_invalid_cert_pem
      assert_raises(Puma::MiniSSL::SSLError) do
        start_server { |ctx|
          ctx.instance_variable_set(:@cert, nil)
          ctx.cert_pem = 'Not a valid pem'
        }
      end
    end

    def test_invalid_key_pem
      assert_raises(Puma::MiniSSL::SSLError) do
        start_server { |ctx|
          ctx.instance_variable_set(:@key, nil)
          ctx.key_pem = 'Not a valid pem'
        }
      end
    end

    def test_invalid_ca
      assert_raises(Puma::MiniSSL::SSLError) do
        start_server { |ctx|
          ctx.ca = __FILE__
        }
      end
    end
  end
```

All five fail on master, pass with the PR.  I can push a commit...

@stanhu - thanks for the PR, input testing is always good.  After your previous PR, I was thinking the same thing (raise on invalid)...

stanhu:
Great idea on the test!

PR #2849, opened 2022-03-30
[jruby] support a truststore option

kares:
which might be a completely different file than the *keystore* ...

due backwards compatibility we assume `truststore = keystore`
(`truststore_pass = keystore_pass`)

### Description

Since the SSL implementations are different on MRI vs JRuby and the JRuby versions follows Java convention, in terms of using a keystore, this PR takes the convention used in Java one step further to have a separate trust-store (with trusted certificates) and key-store which is expected to hold the private key related material.


Rakefile
```diff
@@ -72,6 +72,8 @@ else
 
   Rake::JavaExtensionTask.new("puma_http11", gemspec) do |ext|
     ext.lib_dir = "lib/puma"
+    ext.source_version = '1.8'
```
nateberkopec, line 75: I'm going to mark as "breaking" b/c of this change (though minor!)

test/test_puma_server_ssl.rb
```diff
@@ -332,6 +332,26 @@ def test_verify_client_cert
       http.verify_mode = OpenSSL::SSL::VERIFY_PEER
     end
   end
+
+  def test_verify_client_cert_using_truststore
```
nateberkopec, line 336: Wouldn't this test be green on master branch (minus the lines setting the truststore, of course)?

nateberkopec (changes requested):
Concerned that test would be green on master - do we have a way of specifically testing a cert that would fail unless the trust store was set?

Rakefile
kares: happy to revert this part if it's to delay shipping the feature.
my reasoning was - theoretically this would break users using JRuby 9.1 (9.1.17.0 released on April 23 2018) which was the last to work with Java 1.7 that could still install latest Puma. 
But I do not think anyone would be upgrading smt on 9.1 and not use at least Java 8, so the "breaking" risk seemed almost non-existent given there's JRuby 9.2 already in maintenance mode and latest stable in JRuby 9.3 (all of which require Java 8+).

test/test_puma_server_ssl.rb
kares: yeah, right due the backwards compatibility when `truststore` isn't set ... 
what's confusing is that even if I set it to a `truststore` that does not contain the expected CA it still does not behave - seems like the SSL errors are not propagated properly from the Java MiniSSL impl, going to investigate.

MSP-Greg: @kares

> SSL errors are not propagated properly from the Java MiniSSL imp

Until PR's #2847 and #2848, the MiniSSL c implementation was assumed that all cert/key/ca data was valid.  I'm pretty sure the Java implementation needs the same fixes.

Thanks for your help with JRuby in Puma...

Rakefile
nateberkopec: I think next version release will be a major so this won't delay it. Thanks for the explanation.

nateberkopec:
Thanks @kares!

PR #2855, opened 2022-04-06
Escape SSL cert and filenames

stanhu:
### Description

Filenames with `+` would fail to load because it would be
misinterpreted as a space. Since Puma already unescapes the query
string, we should escape the value to fix this.

We also make `cert` and `key` conditional in the query string because
that gives a clearer error message:

```
ERROR: Please specify the SSL key via 'key=' or 'key_pem='
```

Rather than:

```
puma/lib/puma/minissl.rb:218:in `check_file': Key file '' does not exist (ArgumentError)
```

### DIFF
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'puma/const'
+require 'puma/util'
 
 module Puma
   # The methods that are available for use inside the configuration file.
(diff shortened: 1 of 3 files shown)


lib/puma/dsl.rb
```diff
@@ -65,7 +66,10 @@ def self.ssl_bind_str(host, port, opts)
         v_flags = (ary = opts[:verification_flags]) ?
           "&verification_flags=#{Array(ary).join ','}" : nil
 
-        "ssl://#{host}:#{port}?cert=#{opts[:cert]}&key=#{opts[:key]}" \
+        cert_flags = (cert = opts[:cert]) ? "cert=#{Puma::Util.escape(opts[:cert])}" : nil
+        key_flags = (cert = opts[:key]) ? "&key=#{Puma::Util.escape(opts[:key])}" : nil
```
dentarg, line 70: No need for the `cert` assignment here, right?

stanhu: Oops, thanks: https://github.com/puma/puma/pull/2857

PR #2861, opened 2022-04-16
Fix rack.after_reply exceptions breaking connections

### LINKED ISSUE #2856: raise in rack.after_reply has bad behavior
noted here https://github.com/rack/rack/pull/1802#discussion_r846544710

BlakeWilliams:
### Description

Currently, when a `rack.after_reply` callable raises an exception we
attempt to handle it like other client errors by responding with an HTTP
500 response. This however doesn't work because `rack.after_reply`
callbacks are called after the response body has already been written to
the client.

This can cause issues with re-used connections. This is because 2 HTTP
responses are being returned for a single request. If a second HTTP
request is made before the error handling logic completes the timing can
line up causing the second HTTP response to be served a 500 from the
first HTTP requests `rack.after_reply` callbacks raising.

That may look roughly like:

1. Request 1 starts, opening a reusable TCP connection
2. Request 1 is written to and "completed"
3. Request 1 `rack.after_reply` callables are called
4. Request 2 starts, reusing the same TCP connection as request 1
5. `rack.after_reply` raises, calls `client_error` and serves a 500
   response
6. Request 2 receives the 500 response.

This is somewhat difficult to reproduce using HTTP clients since it's a
race condition whether or not the 500 is written at the "correct" time
or not.

To prevent this issue the `rack.after_reply` callables are now wrapped
in a begin/rescue/end block that rescues from `StandardError` and logs
instead of attempting to serve a 500 response.

Closes #2856

### DIFF
--- a/test/test_rack_server.rb
+++ b/test/test_rack_server.rb
@@ -144,7 +144,7 @@ class TestRackServer < Minitest::Test
     # * If we would block trying to read from the socket, we can assume that
     #   the erroneous 500 response wasn't/won't be written.
     sleep 0.1
-    assert_raises IO::EAGAINWaitReadable do
+    assert_raises IO::WaitReadable do
       content = socket.read_nonblock(12)
       refute_includes content, "500"
     end
--- a/test/test_rack_server.rb
+++ b/test/test_rack_server.rb
@@ -36,8 +36,8 @@ class TestRackServer < Minitest::Test
   def setup
     @simple = lambda { |env| [200, { "X-Header" => "Works" }, ["Hello"]] }
     @server = Puma::Server.new @simple
-    port = (@server.add_tcp_listener "127.0.0.1", 0).addr[1]
-    @tcp = "http://127.0.0.1:#{port}"
+    @port = (@server.add_tcp_listener "127.0.0.1", 0).addr[1]
+    @tcp = "http://127.0.0.1:#{@port}"
     @stopped = false
   end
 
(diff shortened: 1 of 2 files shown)


test/test_rack_server.rb
```diff
@@ -108,6 +108,52 @@ def test_after_reply
...
+    # This sleeps to give the server time to write the invalid/extra HTTP
+    # response.
+    #
+    # * If we can read from the socket, we know that extra content has been
+    #   written to the connection and assert that it's our erroneous 500
+    #   response.
+    # * If we would block trying to read from the socket, we can assume that
+    #   the erroneous 500 response wasn't/won't be written.
+    sleep 0.1
+    assert_raises IO::EAGAINWaitReadable do
```
BlakeWilliams, line 147: I'm not the biggest fan of this test, but using `Net::HTTP` was flaky and the problem is pretty low-level so this is what I ended up with. It is technically testable using `Net::HTTP` but you have to `instance_variable_get(:@socket)` on the connection and I wasn't a huge fan of diving into stdlib internals.

Happy to get feedback/suggestions on alternatives here if anyone has any.

nateberkopec:
Now this is a high-quality first time contribution! 🥇

nateberkopec:
Looks good to me but I'll leave it open for a bit to see if anyone else has anything to say

MSP-Greg:
This passes for me locally.  The one thing not sure about is whether the begin/rescue block should be inside or outside the `after_reply.each` loop?

BlakeWilliams:
> This passes for me locally. The one thing not sure about is whether the begin/rescue block should be inside or outside the `after_reply.each` loop?

Same, I'm wondering if it's platform differences since most of the failures are on Windows. I'll push up what I think may fix it in a bit.

I chose to put the rescue outside of the `each` block to keep the existing behavior where a callback raising halts the rest of the chain. Happy to change it if that's the direction y'all prefer, but it may be somewhat of a breaking change. There's a broader discussion about this behavior in the rack spec [over in the rack repo](https://github.com/rack/rack/pull/1802#discussion_r846540916) if you're interested.

BlakeWilliams:
It looks like https://github.com/puma/puma/pull/2861/commits/d044ed2d0a9cd5c074cf2e3a64b9c64e5fdf1876 fixed the failures for the windows builds but all [Ruby HEAD builds are still failing](https://github.com/BlakeWilliams/puma/runs/6050963245?check_suite_focus=true). I ran the suite against master in my fork too and [it looks like Ruby HEAD is failing there too](https://github.com/BlakeWilliams/puma/runs/6050963427?check_suite_focus=true) (my fork is 1 commit behind though, if that matters).

MSP-Greg:
@BlakeWilliams

Ignore the `TestIntegrationPumactl#test_prune_bundler_with_multiple_workers` failures, as this PR doesn't affect those tests.  

Puma may need some changes for issues related to default gems...

MSP-Greg:
Also, OS socket errors may differ between Ubuntu, macOS, and Windows.  Kind of a pita when one is wondering what going wrong...

PR #2868, opened 2022-04-29
Puma::ControlCLI - allow refork command to be sent as a request

### LINKED ISSUE #2866: "Invalid request command: pumactl refork"
**Describe the bug**
I run puma in refork mode, but when I run `pumactl refork` I just get:
> Invalid request command: refork

`refork` does appear in the list of available commands when I run `pumactl --help`. Sending SIGURG to the main puma process works fine. Other `pumactl` commands, such as gc and stats, work fine.

 - OS: Ubuntu Linux
 - Puma Version 5.5.0

MSP-Greg:
### Description

Currently, the refork command will only be sent as a signal by `Puma::ControlCLI`.  Change to allow it be sent as a request if the control server is setup with a uri.

Closes #2866

### DIFF
--- a/lib/puma/control_cli.rb
+++ b/lib/puma/control_cli.rb
@@ -32,11 +32,13 @@ module Puma
     # @deprecated 6.0.0
     COMMANDS = CMD_PATH_SIG_MAP.keys.freeze
 
+    # Uncomment below statement & clause in run method statement if needed
+    #
     # commands that cannot be used in a request
-    NO_REQ_COMMANDS = %w{refork}.freeze
+    # NO_REQ_COMMANDS = %w[].freeze
 
     # @version 5.0.0
-    PRINTABLE_COMMANDS = %w{gc-stats stats thread-backtraces}.freeze
+    PRINTABLE_COMMANDS = %w[gc-stats stats thread-backtraces].freeze
 
     def initialize(argv, stdout=STDOUT, stderr=STDERR)
       @state = nil
@@ -185,8 +187,6 @@ module Puma
 
       if @command == 'status'
         message 'Puma is started'
-      elsif NO_REQ_COMMANDS.include? @command
-        raise "Invalid request command: #{@command}"
       else
         url = "/#{@command}"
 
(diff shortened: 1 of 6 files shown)


lib/puma/control_cli.rb
```diff
@@ -33,10 +33,10 @@ class ControlCLI
     COMMANDS = CMD_PATH_SIG_MAP.keys.freeze
 
     # commands that cannot be used in a request
-    NO_REQ_COMMANDS = %w{refork}.freeze
+    NO_REQ_COMMANDS = %w[].freeze
```
nateberkopec, line 36: Probably should just remove this constant then, yes?

nateberkopec: I feel like I'm missing something.

MSP-Greg: I can, but I thought I'd leave the code in case a 'signal only' command was added.  Or, previous to this PR, `refork` was a signal only command, and I never tested it using a control server that had a control uri.

Leaving as is means that CI will check that if one is added.  Ok, I know it's a little odd...

MSP-Greg: I commented out the constant and the clause that used it.  Same for the test (skipped if the constant is undefined).

I can remove if you want...  I just don't want myself or someone else making the same mistake again, so the commented code leave a **big** hint...

lib/puma/control_cli.rb
```diff
@@ -32,11 +32,13 @@ class ControlCLI
     # @deprecated 6.0.0
     COMMANDS = CMD_PATH_SIG_MAP.keys.freeze
 
+    # Uncomment below statement & clause in run method statement if needed
+    #
     # commands that cannot be used in a request
-    NO_REQ_COMMANDS = %w{refork}.freeze
+    # NO_REQ_COMMANDS = %w[].freeze
```
johnnyshields, line 38: Should just remove this altogether?

MSP-Greg:
I rebased and added 'signal only' commands that were previously unavailable using `ControlCLI`. See https://github.com/puma/puma/pull/2868/commits/eb87fe59dc4098c1add5f2f9b19027fe1ae8b91b

I think I found all of them.  Hence, `ControlCLI` can be used to send all signal commands that work with the main Puma process.

PR #2869, opened 2022-05-02
Github Actions. Run test only if linter finished successfully.

alexeevit:
### Description
Hello. Now if we run CI it installs all the dependencies, compiles the extension and only then runs rubocop for each MRI version on every OS. And if we have a linting error it'll ruin all the started jobs.

I'm not sure, but it looks like it doesn't matter what OS and ruby version we use for linting and we don't need to compile the extension for it. So I moved the linter before the tests and made MRI and NON-MRI tests to depend on its result.

If you think it's not what we need, I'm open for discussion.


MSP-Greg (approved):

nateberkopec:
@alexeevit Could you remove the non-mri badge in the README as well, please?

alexeevit:
> @alexeevit Could you remove the non-mri badge in the README as well, please?

Done 👌

MSP-Greg:
@nateberkopec - thanks, I was looking at this PR , and...

Should we do the same with the ragel testing?  Also, I'll add some path restrictions to ragel, so it only runs when it or a file in the ext folder changes...

alexeevit:
@MSP-Greg I could do this if you wish

MSP-Greg:
@alexeevit

> I could do

Fine by me.  For ragel's 'on' maybe something like the below?  We could filter further, but ext files and workflows don't change that often...

```yaml
on:
  push:
    paths:
      - 'ext/**'
      - '.github/workflows/**'
  pull_request:
    paths:
      - 'ext/**'
      - '.github/workflows/**'
  workflow_dispatch:
```

Also, the rubocop job could use `ruby/set-ruby@v1` instead of `MSP-Greg/setup-ruby-pkgs@v1`.  `MSP-Greg/setup-ruby-pkgs` uses the code from `ruby/set-ruby`, and I'd prefer it is only used when the additional inputs are needed.  Helps keep questions/issues directed to the correct repo...

MSP-Greg:
> do we need ruby linter for the ext CI in the first place?

Actually, never checked, don't know if rubocop runs any checks, but about all it can do is whitespace issues?

MSP-Greg:
> I used it in the beginning but switched to your MSP-Greg/setup-ruby-pkgs@v1 for consistency

That's valid.  I don't have a strong opinion.  I just don't want issues with `setup-ruby` being directed to `setup-ruby-pkgs`...

MSP-Greg:
I checked RuboCop on minissl.c with whitespace added, and no error.  But, we should probably run RuboCop on `ext/puma_http11/extconf.rb`...

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,170 @@
...
+    name: 'Rubocop linting'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: MSP-Greg/setup-ruby-pkgs@v1
+        with:
+          ruby-version: 2.6
+          bundler-cache: true
+      - name: bundle install
+        run: bundle install --jobs 4 --retry 3
```
olleolleolle, line 16: This is now a repeated install, the bundler-cache does install and cache gems.

MSP-Greg: @olleolleolle

Thanks, you're correct.  Notes to self - quick glances at the code are not good, and always look at the logs...

@alexeevit can you catch thhe above on the update discussed above?

olleolleolle: Since this is a little hidden, I usually tack on a line comment "bundle install and cache gems" to make this clearer and more discoverable.

alexeevit: @olleolleolle but doesn't `with: - bundler-cache: true` covers bundle caching?
https://github.com/puma/puma/pull/2869/files#diff-95557d53bf91069e59d82b9d8fcfaf52ec84a762858983edd5ef3c9b3e4c8191R14

alexeevit: In the last CI run it used the cache:

https://github.com/puma/puma/runs/6292533678?check_suite_focus=true

alexeevit: I thought maybe it makes sense to use the repository and `vendor/bundle` as artifacts and re-use them in the next jobs

olleolleolle: @alexeevit Hello!

What I meant was, I usually note the setting like this, to show that it actually does a `bundle install` on its own, no need to issue a second `bundle install` after it.

```yaml
bundler-cache: true # 'bundle install' and cache the gems
```

PR #2871, opened 2022-05-06
Fixup SERVER_PROTOCOL & HTTP_VERSION headers

### LINKED ISSUE #2870: Puma treats the HTTP protocol version as an header, loses track of real protocol version when there is a client-provided header named 'version'
**Describe the bug**
Puma treats the HTTP protocol version as an header and if there is a client-provided header named 'version' Puma adds the two values, later Puma uses this value when matching which protocol to respond, having now a bogus value it will always fallback to HTTP/1.0.

There is a related issue that was closed:

[Issue 1565 - Requests to Puma hanging due to issues with Keep-Alive, Content-Length, and HTTP_VERSION headers](https://github.com/puma/puma/issues/1565)

And this comment pretty much sums what is happening
https://github.com/puma/puma/issues/1565#issuecomment-382793873

> > Unfortunately I don't know how to dig deeper to figure out where the "doubled" version is coming from
> 
> It comes from [here](https://github.com/puma/puma/blob/master/ext/puma_http11/puma_http11.c#L264-L268), set via [this](https://github.com/puma/puma/blob/master/ext/puma_http11/puma_http11.c#L302) and lives in Ragel [here](https://github.com/puma/puma/blob/master/ext/puma_http11/http11_parser_common.rl#L41) and [here](https://github.com/puma/puma/blob/master/ext/puma_http11/http11_parser.rl#L63-L65)
> 
> As part of the HTTP spec every client-provided header gets transformed to uppercase, has `-` replaces with `_` and gets the prefix of `HTTP_` added. In the case of a `Version` header that'll transform it to `HTTP_VERSION`.
> 
> The problem is we're already setting `HTTP_VERSION` to `HTTP/1.1` in Puma (via the C/Java extensions) we fall into the RFC's behaviour for when you set the same header more than once: We add the client header value to create `HTTP/1.1, HTTP/1.1`. You could send a request with `Version:Blabbedy` and get `HTTP_VERSION: HTTP/1.1, Blabbedy` grimacing
> 
> Though it's not in the Rack spec, this is also a problem in Rack itself ([rack/rack#970](https://github.com/rack/rack/issues/970)). We can fix Puma and we can open PRs on Thin/Unicorn too but I'd love some thoughts from the maintainers of Puma on it before I do

The related Rack issue was fixed on the 3.0.0 version https://github.com/rack/rack/pull/1691

**Puma config:**
NA

**To Reproduce**

Start puma hello.ru server.
```
> echo 'run lambda { |env| [200, {"Content-Type" => "text/plain"}, ["Hello World"]] }' > hello.ru
> puma hello.ru
```

Test without version header - Puma responds with HTTP/1.1:
```
time curl -v -k http://localhost:9292 
*   Trying 127.0.0.1:9292...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 9292 (#0)
> GET / HTTP/1.1
> Host: localhost:9292
> User-Agent: curl/7.68.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Type: text/plain
< Content-Length: 11
< 
* Connection #0 to host localhost left intact
Hello World
real	0m0,010s
user	0m0,001s
sys	0m0,009s
```

Test with version header - Puma responds with HTTP/1.0:
```
time curl -v -k http://localhost:9292 --header 'version: abc'
*   Trying 127.0.0.1:9292...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 9292 (#0)
> GET / HTTP/1.1
> Host: localhost:9292
> User-Agent: curl/7.68.0
> Accept: */*
> version: abc
> 
* HTTP 1.0, assume close after body
< HTTP/1.0 200 OK
< Content-Type: text/plain
< Content-Length: 11
< 
* Closing connection 0
Hello World
real	0m0,011s
user	0m0,004s
sys	0m0,007s
```

 - OS: Ubuntu 20.4
 - Puma Version 5.6.4

MSP-Greg:
### Description

The code for the headers HTTP_VERSION & SERVER_PROTOCOL has existed and essentially remained unchanged since possibly the original Puma commit [2011-Sep-22  'Name change: Mongrel => Puma'](https://github.com/puma/puma/tree/190a81c55a1cbf215658b8fd6bb7ce2a0326b6c6).

Currently, SERVER_PROTOCOL is fixed as [`HTTP_11`](https://github.com/puma/puma/blob/20df56b7cabb4e4cada493d51894b08c7406bc4a/lib/puma/binder.rb#L54).  HTTP_VERSION is first set to the protocol from the request 'first line', but if a client defines the header, the value is appended.  Many would consider it to be a mistake to use a header name that could also be used by a client, but that decision was made a long time ago...

For a summary of app servers responses, see @jeremyevans PR comment in [rack/rack](https://github.com/rack/rack/pull/1883#issuecomment-1118673046).

He noted Rails code that checks for env[HTTP_VERSION] == HTTP_10, and rack has similar code for [HTTP_11](https://github.com/sinatra/sinatra/blob/b4f61f59cf8f938db261a6561a69ca61d01e8581/lib/sinatra/base.rb#L299).

Hence, by setting HTTP_VERSION to a combination of a derived value and values set by a client, Puma is creating code that can cause issues with both Rack and Sinatra.

Rack will be changing its handling of these headers in Rack 3, and may ignore the HTTP_VERSION header in Rack 4.

This PR changes code that SERVER_PROTOCOL is correctly set (simply replacing HTTP_VERSION with SERVER_PROTOCOL), then in Request#normalize_env sets `env[HTTP_VERSION] = env[SERVER_PROTOCOL]`.  Since env[HTTP_VERSION] is replaced, a value set by the client is overwritten, as it conflicts with the Rails and Sinatra code mentioned above.  Note that the line can be possibly be ommitted when Rack 4 is released, since it will no longer depend on HTTP_VERSION.

Although one might consider these to be breaking changes, they are really needed to fix long existing quirks due to the choice of HTTP_VERSION as a derived header.

Closes #2870

### DIFF
--- a/ext/puma_http11/http11_parser.c
+++ b/ext/puma_http11/http11_parser.c
@@ -297,7 +297,7 @@ case 13:
 tr18:
 #line 65 "ext/puma_http11/http11_parser.rl"
 	{
-    parser->http_version(parser, PTR_TO(mark), LEN(mark, p));
+    parser->server_protocol(parser, PTR_TO(mark), LEN(mark, p));
   }
 	goto st14;
 tr26:
--- a/ext/puma_http11/http11_parser.h
+++ b/ext/puma_http11/http11_parser.h
@@ -46,7 +46,7 @@ typedef struct puma_parser {
   element_cb fragment;
   element_cb request_path;
   element_cb query_string;
-  element_cb http_version;
+  element_cb server_protocol;
   element_cb header_done;
 
   char buf[BUFFER_LEN];
--- a/ext/puma_http11/http11_parser.java.rl
+++ b/ext/puma_http11/http11_parser.java.rl
@@ -39,8 +39,8 @@ public class Http11Parser {
     Http11.query_string(runtime, parser.data, parser.buffer, parser.query_start, fpc-parser.query_start);
   }
 
-  action http_version {
-    Http11.http_version(runtime, parser.data, parser.buffer, parser.mark, fpc-parser.mark);
+  action server_protocol {
+    Http11.server_protocol(runtime, parser.data, parser.buffer, parser.mark, fpc-parser.mark);
   }
 
   action request_path {
--- a/ext/puma_http11/http11_parser.rl
+++ b/ext/puma_http11/http11_parser.rl
@@ -62,8 +62,8 @@ static void snake_upcase_char(char *c)
     parser->query_string(parser, PTR_TO(query_start), LEN(query_start, fpc));
   }
 
-  action http_version {
-    parser->http_version(parser, PTR_TO(mark), LEN(mark, fpc));
+  action server_protocol {
+    parser->server_protocol(parser, PTR_TO(mark), LEN(mark, fpc));
   }
 
   action request_path {
--- a/ext/puma_http11/http11_parser_common.rl
+++ b/ext/puma_http11/http11_parser_common.rl
@@ -38,8 +38,8 @@
   Method = ( upper | digit | safe ){1,20} >mark %request_method;
 
   http_number = ( digit+ "." digit+ ) ;
-  HTTP_Version = ( "HTTP/" http_number ) >mark %http_version ;
-  Request_Line = ( Method " " Request_URI ("#" Fragment){0,1} " " HTTP_Version CRLF ) ;
+  Server_Protocol = ( "HTTP/" http_number ) >mark %server_protocol ;
+  Request_Line = ( Method " " Request_URI ("#" Fragment){0,1} " " Server_Protocol CRLF ) ;
 
   field_name = ( token -- ":" )+ >start_field $snake_upcase_field %write_field;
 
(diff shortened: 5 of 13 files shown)


dentarg:
Could do this for 6.0?

MSP-Greg:
In the above, I was concerned about getting all the detail regarding the issue mentioned.  Maybe a simple summary was left out.

1. Currently SERVER_PROTOCOL is fixed to HTTP/1.1.  This PR sets it to HTTP/1.0 if the client request is using HTTP/1.0.

2. Currently, since both Rails & Sinatra use an equality conditional with HTTP_VERSION in their code, if a client provides a VERSION header, the conditional will be incorrectly evaluated.  This PR removes the client provided value.  Once other software changes to using SERVER_PROTOCOL and Rack 4 is released, the client value can be passed thru.

MSP-Greg:
Been a while, rebased...

nateberkopec (approved):

PR #2874, opened 2022-05-11
Explain `WEB_CONCURRENCY` and relation to preloading [ci skip]

jonathanhefner:
### Description

This explains `WEB_CONCURRENCY` as an alternative to the `-w` flag, and how `WEB_CONCURRENCY` can automatically enable preloading.

This also includes some minor edits to "Clustered mode" section for improved flow and clarity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,15 +106,23 @@ Puma also offers "clustered mode". Clustered mode `fork`s workers from a master 
 $ puma -t 8:32 -w 3
 ```
 
+Or with the `WEB_CONCURRENCY` environment variable:
+
+```
+$ WEB_CONCURRENCY=3 puma -t 8:32
+```
+
 Note that threads are still used in clustered mode, and the `-t` thread flag setting is per worker, so `-w 2 -t 16:16` will spawn 32 threads in total, with 16 in each worker process.
 
-In clustered mode, Puma can "preload" your application. This loads all the application code *prior* to forking. Preloading reduces total memory usage of your application via an operating system feature called [copy-on-write](https://en.wikipedia.org/wiki/Copy-on-write) (Ruby 2.0+ only). Use the `--preload` flag from the command line:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -106,23 +106,33 @@ Puma also offers "clustered mode". Clustered mode `fork`s workers from a master
...
+```
+$ WEB_CONCURRENCY=3 puma -t 8:32
+```
+
 Note that threads are still used in clustered mode, and the `-t` thread flag setting is per worker, so `-w 2 -t 16:16` will spawn 32 threads in total, with 16 in each worker process.
 
-In clustered mode, Puma can "preload" your application. This loads all the application code *prior* to forking. Preloading reduces total memory usage of your application via an operating system feature called [copy-on-write](https://en.wikipedia.org/wiki/Copy-on-write) (Ruby 2.0+ only). Use the `--preload` flag from the command line:
+In clustered mode, Puma can "preload" your application. This loads all the application code *prior* to forking. Preloading reduces total memory usage of your application via an operating system feature called [copy-on-write](https://en.wikipedia.org/wiki/Copy-on-write).
+
+If the `WEB_CONCURRENCY` environment variable is set to a value > 1, preloading will be enabled by default. Otherwise, you can use the `--preload` flag from the command line:
```
dentarg, line 119: ...if `prune_bundler` isn't used: https://github.com/puma/puma/blob/8a4ef0c16c235e42b3839971e88075cd01962ff1/lib/puma/configuration.rb#L150

jonathanhefner: My thinking was that: (1) `--prune-bundler` is not mentioned anywhere else in the README, so mentioning it here might be abrupt, and (2) it is [documented elsewhere](https://github.com/puma/puma/blob/20df56b7cabb4e4cada493d51894b08c7406bc4a/lib/puma/dsl.rb#L702) as being incompatible with preloading, so specifically calling it out here might not be necessary.

I've now added it as a parenthetical, but I can make it more prominent if you prefer.

dentarg:
Great, thanks!

PR #2875, opened 2022-05-11
Minor: Export puma/ruby versions in /stats

ylecuyer:
### Description
I added puma/ruby versions to the /stats endpoint this way when monitoring puma we have an easy way to know which versions are running, here is an example:

Screenshot_2022-05-11_23-54-55

### DIFF
--- a/lib/puma/runner.rb
+++ b/lib/puma/runner.rb
@@ -178,5 +178,18 @@ module Puma
         raise "Cannot redirect #{io_name} to #{path}"
       end
     end
+
+    def stats
+      {
+        versions: {
+          puma: Puma::Const::PUMA_VERSION,
+          ruby: {
+            engine: RUBY_ENGINE,
+            version: RUBY_VERSION,
+            patchlevel: RUBY_PATCHLEVEL
+          }
+        }
+      }
+    end
   end
 end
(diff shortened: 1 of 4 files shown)


lib/puma/runner.rb
```diff
@@ -178,5 +178,18 @@ def ensure_output_directory_exists(path, io_name)
         raise "Cannot redirect #{io_name} to #{path}"
       end
     end
+
+    def stats
+      {
+        versions: {
+          puma: Puma::Const::PUMA_VERSION,
+          ruby: {
```
misdoro, line 186: It may be interesting to expose the whole `RbConfig::CONFIG` here if someone wants a complete environment information.

nateberkopec: I took a look, I don't think there's anything else in there that we really want to expose. Exposing all of that info would probably 10x the size of the stats hash, we do need to be conscious of that.

nateberkopec:
Thanks @ylecuyer !

ylecuyer:
Thanks @nateberkopec, I forgot to add a line in the changelog do you want me to open another PR for that?

nateberkopec:
No don't sweat it, I mostly create those myself now.

PR #2881, opened 2022-05-19
Add ENV['PUMA_NO_RUBOCOP'] to disable RuboCop install, add mswin ci

MSP-Greg:
### Description

1. Setting `ENV['PUMA_NO_RUBOCOP']` disables installation and use of RuboCop.  Since we're now doing RuboCop as a separate workflow, the other workflows do not need to install RuboCop and it's dependencies.  This lightens the resolution Bundler needs to do, and makes the bundle caches smaller.  Note that the extension gem [jaro_winkler](https://rubygems.org/gems/jaro_winkler) is a dependency of RuboCop, and it will not compile on mswin.

2. Small changes to allow Puma to build using Ruby Windows mswin build. Add to Actions CI.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -2,16 +2,19 @@ require "bundler/setup"
 require "rake/testtask"
 require "rake/extensiontask"
 require "rake/javaextensiontask"
-require "rubocop/rake_task"
 require_relative 'lib/puma/detect'
 require 'rubygems/package_task'
 require 'bundler/gem_tasks'
 
+begin
+  # Add rubocop task
+  require "rubocop/rake_task"
+  RuboCop::RakeTask.new
+rescue LoadError
+end
+
 gemspec = Gem::Specification.load("puma.gemspec")
 Gem::PackageTask.new(gemspec).define
-
-# Add rubocop task
-RuboCop::RakeTask.new
 
 # generate extension code using Ragel (C and Java)
 desc "Generate extension code (C and Java) using Ragel"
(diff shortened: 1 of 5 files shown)


Rakefile
```diff
@@ -11,7 +13,7 @@ gemspec = Gem::Specification.load("puma.gemspec")
 Gem::PackageTask.new(gemspec).define
 
 # Add rubocop task
-RuboCop::RakeTask.new
+RuboCop::RakeTask.new unless NO_RUBOCOP
```
dentarg, line 16: WDYT of this instead?

```ruby
begin
  require "rubocop/rake_task"
  RuboCop::RakeTask.new
rescue LoadError
end

**Review by dentarg — APPROVED** [review]

File: Rakefile (hunk 1 above)
Comment (dentarg) [reply Rakefile:16]: (for all of the changes to `Rakefile`, not only the above line)

File: ext/puma_http11/extconf.rb [hunk 2]
```diff
@@ -14,6 +14,8 @@
   found_ssl = if (!$mingw || RUBY_VERSION >= '2.4') && (t = pkg_config 'openssl')
     puts 'using OpenSSL pkgconfig (openssl.pc)'
     true
+  elsif have_library('libcrypto', 'BIO_read') && have_library('libssl', 'SSL_CTX_new')
+    true
```
dentarg, line 18: Is this change related to this PR?

MSP-Greg: This started as a PR about the mswin build, and it might be a bug in mkmf.  The RuboCop ENV variable involved the same code, so I added it.  An issue or PR mentioned the idea of having a way to skip RuboCop, but I couldn't find it.

But, almost all current OpenSSL implementations name the files 'libcrypto' and 'libssl'.  Older installs did not.  Older Windows OpenSSL used name 'libeay32' and 'ssleay32'.  Anyway, it is needed for mswin, and most other builds use pkgconfig.

dentarg (approved):

PR #2882, opened 2022-05-19
Fix "log_writer.rb:59:in `write': deadlock; recursive locking (ThreadError)"

MSP-Greg:
### Description

CI intermittently fails with "log_writer.rb:59:in `write': deadlock; recursive locking (ThreadError)".  This is happening in log writes/puts, and this code maybe called in signal handlers.  Mutex and other locking mechanisms cannot be used in signal handlers, so this PR is an attempt to fix the problem.

Example log:
```
/Users/runner/work/puma/puma/lib/puma/log_writer.rb:59:in `write': deadlock; recursive locking (ThreadError)
	from /Users/runner/work/puma/puma/lib/puma/log_writer.rb:59:in `puts'
	from /Users/runner/work/puma/puma/lib/puma/log_writer.rb:59:in `log'
	from /Users/runner/work/puma/puma/lib/puma/launcher.rb:329:in `log'
	from /Users/runner/work/puma/puma/lib/puma/launcher.rb:205:in `close_binder_listeners'
	from /Users/runner/work/puma/puma/lib/puma/cluster.rb:329:in `block in setup_signals'
	from /Users/runner/work/puma/puma/lib/puma/log_writer.rb:59:in `write'
	from /Users/runner/work/puma/puma/lib/puma/log_writer.rb:59:in `puts'
	from /Users/runner/work/puma/puma/lib/puma/log_writer.rb:59:in `log'
	from /Users/runner/work/puma/puma/lib/puma/runner.rb:44:in `log'
	from /Users/runner/work/puma/puma/lib/puma/cluster.rb:460:in `run'
```

While doing recent work in my fork I have also seen the error.  Can't repo locally, and tried to get code independent of Puma to fail, again, no luck.

Wondering if anyone has a better idea?

Second commit contains minor test fixes for intermittent issues.

### DIFF
--- a/lib/puma/log_writer.rb
+++ b/lib/puma/log_writer.rb
@@ -3,6 +3,7 @@
 require 'puma/null_io'
 require 'puma/error_logger'
 require 'stringio'
+require 'io/wait'
 
 module Puma
 
@@ -56,14 +57,17 @@ module Puma
 
     # Write +str+ to +@stdout+
     def log(str)
+      @stdout.is_a?(IO) and @stdout.wait_writable(1)
       @stdout.puts(format(str)) if @stdout.respond_to? :puts
-
       @stdout.flush unless @stdout.sync
-    rescue Errno::EPIPE
+    rescue Errno::EPIPE, Errno::EBADF
     end
 
     def write(str)
+      @stdout.is_a?(IO) and @stdout.wait_writable(1)
       @stdout.write(format(str))
+      @stdout.flush unless @stdout.sync
+    rescue Errno::EPIPE, Errno::EBADF
     end
 
     def debug(str)
--- a/test/test_integration_pumactl.rb
+++ b/test/test_integration_pumactl.rb
@@ -62,7 +62,7 @@ class TestIntegrationPumactl < TestIntegration
     skip_unless :fork
     cli_server "-q -w #{workers} test/rackup/sleep.ru --control-url unix://#{@control_path} --control-token #{TOKEN} -S #{@state_path}", unix: true
 
-    start = Time.now
+    start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
 
     s = UNIXSocket.new @bind_path
     @ios_to_close << s
@@ -89,7 +89,7 @@ class TestIntegrationPumactl < TestIntegration
 
     _, status = Process.wait2(@pid)
     assert_equal 0, status
-    assert_operator Time.now - start, :<, (DARWIN ? 8 : 6)
+    assert_operator Process.clock_gettime(Process::CLOCK_MONOTONIC) - start, :<, (DARWIN ? 8 : 7)
     @server = nil
   end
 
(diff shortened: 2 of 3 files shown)


lib/puma/log_writer.rb
```diff
@@ -56,14 +57,17 @@ def self.null
 
     # Write +str+ to +@stdout+
     def log(str)
+      @stdout.is_a?(IO) and @stdout.wait_writable(1)
```
dentarg, line 60: https://ruby-doc.org/stdlib-2.7.1/libdoc/io/wait/rdoc/IO.html#wait_writable-method

>Waits until [IO](https://ruby-doc.org/stdlib-2.7.1/libdoc/io/wait/rdoc/IO.html) is writable without blocking and returns `self` or `nil` when times out.

Sounds like this shouldn't hurt to have

test/test_integration_pumactl.rb
```diff
@@ -89,7 +89,7 @@ def test_phased_restart_cluster
 
     _, status = Process.wait2(@pid)
     assert_equal 0, status
-    assert_operator Time.now - start, :<, (DARWIN ? 8 : 6)
+    assert_operator Time.now - start, :<, (DARWIN ? 8 : 7)
```
dentarg, line 92: Maybe this test should be using `Process.clock_gettime(Process::CLOCK_MONOTONIC)`? https://blog.dnsimple.com/2018/03/elapsed-time-with-ruby-the-right-way/

(but giving the test more time seems fine to me)

MSP-Greg: Done

lib/puma/log_writer.rb
MSP-Greg: > Sounds like this shouldn't hurt to have

My thought also.  Also, catching 'bad io' errors so a restart or shutdown isn't interrupted.

I intend to run it a few times in my fork to see if the intermittent error occurs...

nateberkopec (approved):

MSP-Greg:
Since this was merged, I've seen one "log_writer.rb:59:in `write': deadlock; recursive locking (ThreadError)" error running in my fork.

The following doesn't afford really perfect locking, but it's better than what we have now.  Any thoughts?  It runs fine locally, have tried it on Actions...

```ruby
WRITE_QUEUE = Thread::Queue.new
def lock
  locked = false
  @stdout.is_a?(IO) and @stdout.wait_writable(1)
  sleep 0.005 until WRITE_QUEUE.empty?
  WRITE_QUEUE.push Process.pid
  locked = true
  yield
  @stdout.flush unless @stdout.sync
rescue Errno::EPIPE, Errno::EBADF
ensure
  WRITE_QUEUE.pop if locked
end

# Write +str+ to +@stdout+
def log(str)
  lock { @stdout.puts(format str) } if @stdout.respond_to? :puts
end

def write(str)
  lock { @stdout.write format(str) }
end
```

PR #2883, opened 2022-05-20
NullIO#closed? should return false

MSP-Greg:
### Description

Currently returns `true`.

### DIFF
--- a/lib/puma/null_io.rb
+++ b/lib/puma/null_io.rb
@@ -53,8 +53,9 @@ module Puma
       self
     end
 
+    # This is used as singleton class, so can't have state.
     def closed?
-      true
+      false
     end
   end
 end
--- a/test/test_null_io.rb
+++ b/test/test_null_io.rb
@@ -65,7 +65,7 @@ class TestNullIO < Minitest::Test
     assert_equal nio, nio.flush
   end
 
-  def test_closed_returns_true
-    assert_equal true, nio.closed?
+  def test_closed_returns_false
+    assert_equal false, nio.closed?
   end
 end


MSP-Greg:
@eregon Sorry for the ping.  The test mentioned above fails frequently on truffleruby.  Every time I've seen the error, I check truffleruby-head, and it passes there.  The test is [here](https://github.com/puma/puma/blob/3d24dc5936d0e570376fa843a682cf2c5481041e/test/test_puma_server.rb#L887-L925).  If you have any ideas...

JFYI, Puma is testing on Windows mswin...

eregon:
I have no specific idea, I think adding the skip is fine here if needed, although a check against `RUBY_ENGINE_VERSION` would be better since it seems already fixed on truffleruby-head (and so should be for the next release as well).

The `sleep 0.1` in that test are probably very timing-sensitive though (and make the test slow), that should probably be avoided if possible.

MSP-Greg:
> a check against RUBY_ENGINE_VERSION

I'm a bit TruffleRuby challenged.  Would `RUBY_ENGINE_VERSION < '22.2'` be okay?  Or, something a bit more complicated?

> The sleep 0.1 in that test are probably very timing-sensitive though (and make the test slow

Given how long a few other tests take, not concerned about time.  Also, some of the timing is related to Actions (one VM on a four VM system).  Or, many of the test delays aren't needed when I run locally...

eregon:
`omit unless Gem::Version.new(RUBY_ENGINE_VERSION) > Gem::Version.new('22.1')` seems more robust and is true for both `22.2.0-dev-8839057f` and upcoming 22.2.

MSP-Greg:
@eregon - Thanks, updated.

Issue #2188, opened 2020-03-16
[Question] What's the purpose of Puma having its own MiniSSL?


eregon:
What's the purpose of Puma having its own MiniSSL, instead of using the OpenSSL standard library?

When seeing things like https://github.com/puma/puma/pull/2181 it seems to me this choice has a significant maintenance cost.
cc @nateberkopec @fzakaria

In other words, I'd prefer if everyone optimized the standard library SSL vs doing per-gem specific optimizations for SSL.

HoneyryderChuck:
+1, I'd like to know it as well.

However, the linked PR is introducing the Netty OpenSSL variant instead of the JRuby Bouncy Castle-based `openssl`, which is massively outdated, so it's technically not the same thing.

dentarg:
https://github.com/puma/puma/pull/1836, https://github.com/puma/puma/pull/1832, https://github.com/puma/puma/pull/1562 might be better (recent) examples of the maintenance cost

eregon:
> However, the linked PR is introducing the Netty OpenSSL variant instead of the JRuby Bouncy Castle-based `openssl`, which is massively outdated, so it's technically not the same thing.

Sounds like something worth fixing in JRuby itself then. I would think almost the entire Ruby ecosystem is using the OpenSSL stdlib for SSL.

HoneyryderChuck:
> I would think almost the entire Ruby ecosystem is using the OpenSSL stdlib for SSL.

It should. JRuby lags massively behind, however. The java ecosystem is mostly to blame, as they chose to ignore the TLS elephant in the room. AFAIK there's only one HTTP/2-enabled server in Java, Netty, which maintains their own TLS stack, which FFI's openssl (the one being integrated in the linked PR), and although I don't know all the reasons, I'd guess it's because all other alternatives don't support TLS 1.2 features such as ALPN negotiation, which you need to have an HTTP/2 server.

There are also API discrepancies between CRuby's openssl and JRuby's, an example being the ASN API. All in all, JRuby would be better off FFI'ing to openssl, but that would be a massive change.

None of the reasons above seem to apply to puma, however. So, looking forward to hear about it from the maintainers, as that's been hanging on my head for a while.

MSP-Greg:
First of all, two somewhat connected issues are being discussed here.  First, why is JRuby's SSL support lagging, and second, why is Puma essentially bypassing Ruby OpenSSL.

[...]

PR #2884, opened 2022-05-23
[jruby] improve the truststore option

kares:
### Description

This PR is a follow up on https://github.com/puma/puma/pull/2849
Namely it adds test/asserts that were missing on GH-2849 that confirm the `truststore` option work as expected + a minor refactoring to only keep the leaf cert around.

Also, introduces an option to set `truststore = :default` which will allow on JRuby for Puma to only use the JVM's trust defaults (Java has it's own *cacert* file). The reason why this option is necessary is due backwards compatibility, before GH-2849 only `keystore = ...` was supported and the assumption was that the keystore file also contains the trust chain.

The commits in this PR tell the story in detail.


MSP-Greg:
@kares Thanks for all your work with JRuby SSL.

Both [Puma:DSL](https://github.com/puma/puma/blob/3d24dc5936/lib/puma/dsl.rb)and [Puma::MiniSSL:ContextBuilder](https://github.com/puma/puma/blob/3d24dc5936d0e570376fa843a682cf2c5481041e/lib/puma/minissl/context_builder.rb) need to be 'synced' with the properties of the [Puma::MiniSSL::Context](https://github.com/puma/puma/blob/3d24dc5936/lib/puma/minissl.rb#L200-L308) class.  This PR doesn't make those changes,  and there may also be issues with other properties.

Puma users set SSL listener connection properties using `Pums::DSL#ssl_bind` or via the query parameters of a bind listener URI.  Some of the JRuby properties, along with the changes in #2846, lead one to consider whether the values used Pums::DSL#ssl_bind should match the string values in the URI query parameters, or whether they should have different API's.

Lastly, more and more parameters have been added, so query strings are increasing in complexity.  Code may be adding query parameters that should be handled by default values in Puma::MiniSSL:ContextBuilder.

Also, how one sets TCP properties of an SSL listener connection, not.sure.  Some of these questions are separate from this PR, or could be addressed in another PR.

Personally, I need to spend more time reviewing current code, and also wondering about others' opinions.

kares:
Thanks for taking a look at the PR :green_heart: 

> Both [Puma:DSL](https://github.com/puma/puma/blob/3d24dc5936/lib/puma/dsl.rb)and [Puma::MiniSSL:ContextBuilder](https://github.com/puma/puma/blob/3d24dc5936d0e570376fa843a682cf2c5481041e/lib/puma/minissl/context_builder.rb) need to be 'synced' with the properties of the [Puma::MiniSSL::Context](https://github.com/puma/puma/blob/3d24dc5936/lib/puma/minissl.rb#L200-L308) class. This PR doesn't make those changes, and there may also be issues with other properties.

right, wasn't aware of these I went ahead and sync-ed up.
for the use-case I am after we're using the `Context` API directly thus I never really looked at exposing the properties elsewhere.

> Puma users set SSL listener connection properties using `Pums::DSL#ssl_bind` or via the query parameters of a bind listener URI. Some of the JRuby properties, along with the changes in https://github.com/puma/puma/pull/2846, lead one to consider whether the values used Pums::DSL#ssl_bind should match the string values in the URI query parameters, or whether they should have different API's.

I just went ahead and exposed the (optional) `truststore`, further with the truststore the password should be optional (historically setting a `keystore` assumed a `keystore-pass` as well). as for naming, there was an already established pattern of `keystore-pass` thus the other related option follow the same: `keystore-type`, `truststore-pass` and `truststore-type`.

there's special care for setting a `truststore=default` param to map it into a string accepted by the `Context`. if there are better naming suggestion I am happy to accept those ...

> Lastly, more and more parameters have been added, so query strings are increasing in complexity. 
Code may be adding query parameters that should be handled by default values in `Puma::MiniSSL:ContextBuilder`.

not sure what to do here - maybe the parameters should just be passed using reflection -> read what writers `Context` have and accept those ...

MSP-Greg:
> not sure what to do here - maybe the parameters should just be passed using reflection -> read what writers Context have and accept those ...

First of all, I should have divided my post into two, or opened a separate issue for the more general points.  So, sorry for mixing them, and thanks for making the changes.

I think one thing that should be done is add a markdown doc that describes configuration of the bind sockets.  If I get one started, I may ping you to check the JRuby sections.

Specific to JRuby, does JRuby OpenSSL support session reuse and/or ALPN?  They should be added to the MRI side soon.

kares:
> I think one thing that should be done is add a markdown doc that describes configuration of the bind sockets. If I get one started, I may ping you to check the JRuby sections.

:ok: 

> Specific to JRuby, does JRuby OpenSSL support session reuse and/or ALPN? They should be added to the MRI side soon.

kind of yes and yes:

- session re-use is automatic but there isn't much control (no API exposed on the native Java side) 
  - `session_reused?` kind of works
- ALPN support was, finally, added ~ 2 months ago (in **0.12.2**)

---

And just a note on the original issue (of `truststore` vs `keystore` JRuby specifics) - these could be avoided and JRuby could have the same `Context` API as MRI if jruby-openssl was brought in but the whole idea of `MiniSSL` in Puma would turn just to using `OpenSSL` API which I guess is smt the `MiniSSL` abstraction wanted to avoid (on MRI) ...

MSP-Greg:
> would turn just to using OpenSSL API which I guess is smt the MiniSSL abstraction wanted to avoid (on MRI) ...

That decision was made a long time ago, and I often think 'this could be a lot easier if...'

But, one consideration is that a fundamental feature of Puma is allowing users to reload an app.  Let's say Ruby OpenSSL has a new version that is due to some squirrelly obtuse crypto issue (which may not affect Puma use).  Since the main Puma process has the previous extension gem loaded, loading the new one can be problematic.  I.think.that.made.sense.

Thanks again for your JRuby help.

dentarg:
We do have this issue about replacing MiniSSL and integrating directly with OpenSSL https://github.com/puma/puma/issues/2188

PR #2897, opened 2022-06-18
Load the configuration before passing it to the binder:

Edouard-chin:
Hi 👋 ! Thanks for your work on maintaining Puma!

- ### Problem

  The feature to manually set the "rack.url_scheme" header with
  Puma's `rack_url_scheme` option (introduced in commit https://github.com/puma/puma/commit/32852f715af9e6eb809284d910e647b8fd6e2ea6) has no effect.

  ### Context

  The Configuration object is not yet loaded at the time it's being
  passed to the Binder so any options set inside the config file (i.e.
  `config/puma.rb`) remains nil.
  https://github.com/puma/puma/blob/6a5bda5b4eeb8c86c64468d6916f7d3917e5640a/lib/puma/binder.rb#L45

  Worth noting that this is only true for options set inside the
  config file. Options passed as args to the CLI
  (i.e. `puma -t min_thread:max_thread`) are correctly set
  (they get loaded differently).
  https://github.com/puma/puma/blob/6a5bda5b4eeb8c86c64468d6916f7d3917e5640a/lib/puma/binder.rb#L42-L43

  ### Solution

  I originally tought I'd fix the issue by adding a new CLI flag for
  `rack_url_scheme` but ended up loading the config a bit earlier as
  it felt error prone for the future to have only some options
  available in the Binder.

  I did some git archeology to try to figure out why the config was
  loaded after passing it to the binder but couldn't find a legit
  reason, sorry if I missed it

  <details>
    <summary>Git archeology</summary>

    The Binder class originally didn't need the config so the ordering didn't
    matter.
    https://github.com/puma/puma/commit/ffae12c40f46f8372cc5ff68b84ef6091670c184

    It was later added in https://github.com/puma/puma/commit/47f933d076c57d5c2500360b09e55630c5c4a3e2
  </details>

### Description

### DIFF
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -52,16 +52,16 @@ module Puma
       @original_argv = @argv.dup
       @config        = conf
 
+      # Advertise the Configuration
+      Puma.cli_config = @config if defined?(Puma.cli_config)
+
+      @config.load
+
       @binder        = Binder.new(@log_writer, conf)
       @binder.create_inherited_fds(ENV).each { |k| ENV.delete k }
       @binder.create_activated_fds(ENV).each { |k| ENV.delete k }
 
       @environment = conf.environment
-
-      # Advertise the Configuration
-      Puma.cli_config = @config if defined?(Puma.cli_config)
-
-      @config.load
 
       if @config.options[:bind_to_activated_sockets]
         @config.options[:binds] = @binder.synthesize_binds_from_activated_fs(
--- a/test/config/rack_url_scheme.rb
+++ b/test/config/rack_url_scheme.rb
@@ -1,0 +1,1 @@
+rack_url_scheme "https"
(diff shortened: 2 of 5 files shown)


dentarg:
Thanks for the PR! 👍🏻 on the "Git archeology" part, I like that 

Should we also remove the faulty test added in https://github.com/puma/puma/commit/32852f715af9e6eb809284d910e647b8fd6e2ea6#diff-267386f55151c0f4e7f1f02c72881be7560baa1e80d6e2e63d30b074dda37cd1 for this feature? Or does it still serve some purpose?

Edouard-chin:
Thanks for the quick review. 

> Should we also remove the faulty test added 

👍. I replace them with integration tests instead.

MSP-Greg:
Ran the PR tests on top of master, and `TestIntegrationSingle#test_conf_is_loaded_before_passing_it_to_binder` failed. So, LGTM.

MSP-Greg (approved):

Issue #2843, opened 2022-03-17
Bundler error prevents phased-restarting puma after upgrading to rails 7


joker-777:
**Describe the bug**

We saw two different errors

```
158586] ! Unable to start worker
[4158586] /usr/lib/ruby/gems/3.0.0/gems/bundler-2.3.9/lib/bundler/definition.rb:481:in `materialize' [4158594] + Gemfile in context: /var/local/kenhub/kenhub.d/878/Gemfile
```
which should have raised this kind of error `raise GemNotFound, "Could not find #{missing_specs.map(&:full_name).join(", ")}" in any of the sources"`

and
```
[16442] ! Unable to start worker
[16442] /usr/lib/ruby/gems/3.0.0/gems/bundler-2.3.9/lib/bundler/runtime.rb:309:in `check_for_activated_spec!'
```
which should have raised this kind of error `Gem::LoadError.new "You have already activated #{activated_spec.name} #{activated_spec.version}, but your Gemfile requires #{spec.name} #{spec.version}. #{suggestion}"`

Why didn't we see these error messages in the puma logs and how can we fix them?

**Puma config:**

```
#!/usr/bin/env puma

# The directory to operate out of.
#
# The default is the current directory.
#
directory "/var/local/kenhub/kenhub"

# Set the environment in which the rack's app will run. The value must be a string.
#
# The default is "development".
#
environment ENV["RAILS_ENV"] || "development"

# Store the pid of the server in the file at "path".
#
pidfile "/var/run/puma/puma.pid"

# Use "path" as the file to store the server info state. This is
# used by "pumactl" to query and control the server.
#
state_path "/var/run/puma/puma.state"

# Redirect STDOUT and STDERR to files specified. The 3rd parameter
# ("append") specifies whether the output is appended, the default is
# "false".
#
stdout_redirect "/var/log/kenhub/puma.log", "/var/log/kenhub/puma.log", true

# Configure "min" to be the minimum number of threads to use to answer
# requests and "max" the maximum.
#
# The default is "0, 16".
#
# threads 0, 16
threads 8, 8


# Bind the server to "url". "tcp://", "unix://" and "ssl://" are the only
# accepted protocols.
#
# The default is "tcp://0.0.0.0:9292".
#
bind 'unix:///var/run/puma/puma.sock'


# === Cluster mode ===

# How many worker processes to run.  Typically this is set to
# to the number of available cores.
#
# The default is "0".
#
workers ENV["PUMA_WORKERS"] || 16

[...]

PR #2903, opened 2022-07-02
Ruby 3.2 will have native IO#wait_* methods, don't require io/wait

MSP-Greg:
### Description

In Ruby 3.1 and earlier versions, IO#wait_* methods were contained in the standard library gem `io/wait`.  Ruby 3.2 has moved the methods to the main IO code.  See 'Move wait* methods to io.c' https://github.com/ruby/ruby/pull/6036.

So, `io/wait` does not need to be required for Ruby 3.2 and later.  Adjust code to only require it when needed.

### DIFF
--- a/lib/puma/detect.rb
+++ b/lib/puma/detect.rb
@@ -8,6 +8,8 @@ module Puma
   # @version 5.2.1
   HAS_FORK = ::Process.respond_to? :fork
 
+  HAS_NATIVE_IO_WAIT = ::IO.public_instance_methods(false).include? :wait_readable
+
   IS_JRUBY = Object.const_defined? :JRUBY_VERSION
 
   IS_OSX = RUBY_PLATFORM.include? 'darwin'
--- a/lib/puma/log_writer.rb
+++ b/lib/puma/log_writer.rb
@@ -3,7 +3,7 @@
 require 'puma/null_io'
 require 'puma/error_logger'
 require 'stringio'
-require 'io/wait'
+require 'io/wait' unless Puma::HAS_NATIVE_IO_WAIT
 
 module Puma
 
--- a/lib/puma/minissl.rb
+++ b/lib/puma/minissl.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 begin
-  require 'io/wait'
+  require 'io/wait' unless Puma::HAS_NATIVE_IO_WAIT
 rescue LoadError
 end
 
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -15,7 +15,7 @@ require 'puma/io_buffer'
 require 'puma/request'
 
 require 'socket'
-require 'io/wait'
+require 'io/wait' unless Puma::HAS_NATIVE_IO_WAIT
 require 'forwardable'
 
 module Puma


silva96:
I'm running ruby 3.1 and `::IO.public_instance_methods(false).include? :wait_readable` returns true. This causes the io/wait library not being loaded I suppose. Then my puma wont start:

```
 Early termination of worker
[1979180] + Gemfile in context: /home/ubuntu/www/releases/20221108181543/Gemfile
[1979176] ! Unable to start worker
[1979176] /home/ubuntu/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/bundler-2.3.12/lib/bundler/runtime.rb:309:in `check_for_activated_spec!'
```

MSP-Greg:
> I'm running ruby 3.1 and `::IO.public_instance_methods(false).include? :wait_readable` returns true

Something is loading `io/wait`, if I run that via the below, it returns false.
```
ruby -e "puts ::IO.public_instance_methods(false).include? :wait_readable"
```
Any changes re Ruby default gems in your Gemfile.lock when you upgraded?

silva96:
This is the gist of the upgrade:

https://gist.github.com/silva96/f69a42f9ef3510e7761a8331a13edd3e

MSP-Greg:
Can you try a patch?  Might be something particular with your setup.  There are [three files](https://github.com/puma/puma/pull/2903/files) that use have the conditional:
```ruby
unless Puma::HAS_NATIVE_IO_WAIT
```

change that to:
```ruby
unless ::IO.public_instance_methods(false).include?(:wait_readable)
```

MSP-Greg:
I assume you're having the problem with restarts?  Regardless, can you patch Puma, or work from a GitHub branch, etc?

The Puma code that's logging what you listed is:
```ruby
rescue Exception => e
  log e
  exit 1
end
```
So, the line `log e` is showing pretty much nothing.  Not sure if we can get more info, but maybe `e.class` and/or `e.message` may have more info.  Not.Sure.

silva96:
Thanks I'll give it a try maybe on friday and let you know the output.

MSP-Greg:
Thanks.

FYI, the above isn't fixing anything, but it's an attempt to get more info about what went wrong.

As in #2843, it's difficult to repo locally, and since you've got the problem, the above patch should (hopefully) show enough info to identify the problem...

silva96:
Just if it helps, this is my puma.rb config in production:

```ruby
#!/usr/bin/env puma

directory '/home/ubuntu/www/current'
rackup "/home/ubuntu/www/current/config.ru"
environment 'production'

tag ''

pidfile "/home/ubuntu/www/shared/tmp/pids/puma.pid"
state_path "/home/ubuntu/www/shared/tmp/pids/puma.state"
stdout_redirect '/home/ubuntu/www/shared/log/puma_access.log', '/home/ubuntu/www/shared/log/puma_error.log', true

threads 0,16

bind 'unix:///home/ubuntu/www/shared/tmp/sockets/puma.sock'

workers 2

prune_bundler

on_restart do
  puts 'Refreshing Gemfile'
  ENV["BUNDLE_GEMFILE"] = ""
end
```

silva96:
Ok I found the problem,

Context: I'm deploying using capistrano, and after deploy I restart puma using systemd

```
# changed this
ExecStart=/home/ubuntu/.asdf/shims/puma -C /home/ubuntu/www/shared/puma.rb
# to this
ExecStart=/usr/bin/bash -lc '/home/ubuntu/.asdf/shims/bundle exec puma -C /home/ubuntu/www/shared/puma.rb'
```

The thing is after deploying, asdf shims were not updating puma from 5.x to 6 so the Gemfile version differed from the puma that was running.

Using bundle exec would run the puma version recently installed.

PR #2906, opened 2022-07-06
Main server loop minor optimisations

saluzafa:
### Description
Minor optimisations made to the main server loop:
- alias `@binder` to a local `binder` variable ;
- use `shd` variable inside loop to store `shutting_down?` result instead of calling `shutting_down?` multiple times ;
- remove call to `Puma::Client#tap` to avoid creating a new local variable scope (Proc) to set client attributes.


lib/puma/server.rb
```diff
@@ -330,9 +331,10 @@ def handle_servers
           [nil, nil]
         end
 
-        while @status == :run || (drain && shutting_down?)
+        shd = false
```
dentarg, line 334:
```suggestion
        is_shutting_down = false
```

I think we should avoid cryptic abbreviations

saluzafa: 100% agree with you 😄 
Fixed!

lib/puma/server.rb
```diff
@@ -346,11 +348,12 @@ def handle_servers
                 rescue IO::WaitReadable
                   next
                 end
-                drain += 1 if shutting_down?
-                pool << Client.new(io, @binder.env(sock)).tap { |c|
-                  c.listener = sock
-                  c.send(addr_send_name, addr_value) if addr_value
-                }
+                drain += 1 if shd
```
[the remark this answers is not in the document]
saluzafa: Actually you are right, it can possibly affect the total connections count written to Puma log (line 367 `@log_writer.debug "Drained #{drain} additional connections." if drain`).
Do you want me to revert that change?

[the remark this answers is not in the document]
dentarg: Yes, I would prefer correct before performance but maybe there are other opinions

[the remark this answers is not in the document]
saluzafa: > Yes, I would prefer correct before performance but maybe there are other opinions

I reverted my changes :)

MSP-Greg:
Not sure about Puma 6, but there are a few classes that have code similar to Client:
```ruby
client = Client.new(io, @binder.env(sock))
client.listener = sock
client.send(addr_send_name, addr_value) if addr_value
```
Maybe it's time to add more arguments to `Client.new`?

saluzafa:
> Not sure about Puma 6, but there are a few classes that have code similar to Client:
> 
> ```ruby
> client = Client.new(io, @binder.env(sock))
> client.listener = sock
> client.send(addr_send_name, addr_value) if addr_value
> ```
> 
> Maybe it's time to add more arguments to `Client.new`?

I added a new argument `listener` to `Puma::Client.new`, I was not sure about using/not using a keyword argument for it.
I think the keyword argument version will "degrade" the performance IMHO (example below):
```ruby
def initialize(io, env = nil, listener: nil)
  [...]
end
```

MSP-Greg:
@saluzafa

Thanks.  The Puma API has typically been rather open, but I doubt anyone is using a Client object.  From what I've seen/recall, the Server and Config related objects are the only objects used by others?

Although objects with a lot of arguments in their `new` methods are sometimes a PITA, I personally dislike the pattern above.

Hence, if Puma 6 is about internal API changes that are unlikely to affect the vast majority of users, maybe we should add sock, addr_send_name, addr_value as arguments?   Not.sure.

saluzafa:
> @saluzafa
> 
> Thanks. The Puma API has typically been rather open, but I doubt anyone is using a Client object. From what I've seen/recall, the Server and Config related objects are the only objects used by others?
> 
> Although objects with a lot of arguments in their `new` methods are sometimes a PITA, I personally dislike the pattern above.
> 
> Hence, if Puma 6 is about internal API changes that are unlikely to affect the vast majority of users, maybe we should add sock, addr_send_name, addr_value as arguments? Not.sure.

Hello there!

Sorry my late reply 😨 , I refactored my code to not adding new arguments to `Client#initialize` method.
Can you tell me if it's better now or not? 🙂 

Cheers!

lib/puma/server.rb
```diff
@@ -320,41 +320,65 @@ def handle_servers
...
+                # set client remote address
+                # Prefer using direct method/setter instead of calling `#send` method
+                case remote_address_option
+                when :value
+                  c.peerip = remote_address_value_option
+                when :header
+                  c.remote_addr_header = remote_address_header_option
+                when :proxy_protocol
+                  c.expect_proxy_proto = remote_address_proxy_protocol_option
+                end
```
MSP-Greg, line 378: This case was originally outside of the while loop and used the following inside of it:
```ruby
c.send(addr_send_name, addr_value) if addr_value
```

dentarg:
Maybe each of these optimisations should be its own PR? We do squash merge things so _if_ there's some problem, bisecting would be more useful.

lib/puma/server.rb
```diff
@@ -320,41 +320,65 @@ def handle_servers
...
               else
                 pool.wait_until_not_full
-                pool.wait_for_less_busy_worker(@options[:wait_for_less_busy_worker])
 
-                io = begin
-                  sock.accept_nonblock
-                rescue IO::WaitReadable
-                  next
+                # Wait for less busy worker?
+                if wait_for_less_busy_worker_enabled
```
MSP-Greg, line 348: Maybe remove `if wait_for_less_busy_worker_enabled` and just use:
```ruby
if wait_for_less_busy_worker
```
It should either be a positive number or nil?  Both are checked in the method...

dentarg:
> Maybe each of these optimisations should be its own PR? We do squash merge things so if there's some problem, bisecting would be more useful.

I will repeat this. We really should try to do one thing at a time.

lib/puma/server.rb
saluzafa: Hello @MSP-Greg 
I reverted my code, the case is now outside like it was before.

Cheers

dentarg:
(Needs a rebase)

And I will repeat myself: see https://github.com/puma/puma/pull/2906#discussion_r916257603 and https://github.com/puma/puma/pull/2906#issuecomment-1291216777

I'm skeptical of the changes presented here, are they worth it?

PR #2916, opened 2022-08-26
5.7.0

nateberkopec:


History.md
```diff
@@ -1,4 +1,28 @@
-## 5.6.5 / 2022-07-dd
+## 5.7.0 / 2022-08-XX
+
+* Features
+  * Puma::ControlCLI - allow refork command to be sent as a request ([#2868], [#2866])
```
dentarg, line 4: This is listed under 5.6.5 too, but that looks incorrect (GitHub doesn't show that https://github.com/puma/puma/commit/6543f0959bd495b8b0326777034d6f13651dd4b3 has the v5.6.5 tag)

Will you fix that in this PR too? (I see you adjusted the date for 5.6.5)

dentarg: >but that looks incorrect (GitHub doesn't show that https://github.com/puma/puma/commit/6543f0959bd495b8b0326777034d6f13651dd4b3 has the v5.6.5 tag)

Can't seem to find that tag on any commits so I'm not sure what's right :/

MSP-Greg: cherry-pick?  See https://github.com/puma/puma/commit/6454710f1c

dentarg: That commit also doesn't show the tag (here on GitHub) Yes it is in 5-6-stable branch though 

I'm confused as how we make releases :)

MSP-Greg:
> I'm confused as how we make releases :)

Part of the issue is that PR #2653 is shown as a breaking change, it appears closed, but was merged.  Also, PR #2849, I think that's included, haven't checked.

There was also a security patch, which would normally be a patch release.  Or, feature commits wouldn't be included in a security patch release.

So, we may need to start a 5-7-stable branch for the next release. Not.sure.

nateberkopec:
> That commit also doesn't show the tag (here on GitHub) Yes it is in 5-6-stable branch though

Shoot. Mistake. That was tagged as bugfix _and_ feature and so when I did the release notes, I just looked at the bugfix list and pulled everything out of that list, without considering whether it was _also_ a feature.

I'll update release notes to match.

For anyone reading in the future, we try our best to be sem-ver-y, but it doesn't always work out.

nateberkopec:
> Part of the issue is that PR https://github.com/puma/puma/pull/2653 is shown as a breaking change, it appears closed, but was merged. Also, PR https://github.com/puma/puma/pull/2849, I think that's included, haven't checked.

Ah indeed. Probably should just call this 6.0 then.

dentarg:
@nateberkopec can the `5-7-0` branch be removed?

PR #2917, opened 2022-08-27
Allow `run_hooks` to pass a hash to blocks for use later

### LINKED ISSUE #2915: Callbacks for app lifecycle?
Hi, I'm wondering if puma has an API for applications to start and stop services that run within a puma process on start and stop. For example, Sidekiq has its `on(event)` API so that services can run inside each Sidekiq process:

```ruby
Sidekiq.configure_server do |config|
  config.on(:startup) do
    t = Thread.new do
      loop do
        puts Time.now
        sleep 1
      end
    end
    config.on(:shutdown) do
      t.raise Interrupt
    end
  end
end
```

I'm planning on adding support for Sidekiq running natively within other processes but the other process needs to offer something like this API so Sidekiq can be started and stopped "cleanly". What would you recommend here? The only clean shutdown option I can find today is `at_exit` which is less than ideal.

MSP-Greg:
### Description

See #2915 'Callbacks for app lifecycle?'.

When one calls DSL methods to register callbacks **inside of the fork block** (`on_worker_boot`, `on_worker_shutdown`, & `on_refork`), adds a parameter that is used as a key.  It can be a String or a Symbol, and it defines which callbacks a Hash object is passed to.

This allows the hooks to access objects stored in a hook called earlier.

See comments added to the DSL class.

Closes #2915

### DIFF
--- a/lib/puma/cluster/worker.rb
+++ b/lib/puma/cluster/worker.rb
@@ -23,6 +23,7 @@ module Puma
         @fork_pipe = pipes[:fork_pipe]
         @wakeup = pipes[:wakeup]
         @server = server
+        @hook_data = {}
       end
 
       def run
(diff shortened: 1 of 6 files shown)


mperham:
LGTM

MSP-Greg:
@nateberkopec

Sorry for the delay on this one.  I rebased and created a method in dsl.rb (`process_hook`) that removed all the repetitive code.

Added to the 6.0.0 milestone, as running in future versions of Sidekiq is important...

Issue #2902, opened 2022-06-28
Time to change to require_relative?


MSP-Greg:
Have a branch prepared for a PR, so I thought I'd ask first.

Any reason to not change all `require` statements to `require_relative`?

See https://github.com/ruby/ruby/search?q=require_relative, and https://github.com/puma/puma/pull/2381

Ruby master has now copied all the methods Puma used from `io/wait` to `IO`. I originally started the PR for that, as `io/wait` won't need to be required for Ruby 3.2 and later.  I.think.  I haven't tested it yet, but I think the only methods used in Puma are `wait_readable` and `wait_writable`...

nateberkopec:
Can we avoid breaking CentOS?

MSP-Greg:
> Can we avoid breaking CentOS?

I interpreted #2381 to mean that all ruby files could be loaded with `require_relative`, but any *.so files needed to be loaded with `require`.  We only have one `puma_http11`.

We could use `require_relative`, and if an error is raised, load via `require`...

nateberkopec:
Cool! Full speed ahead then 🚆

PR #2896, opened 2022-06-14
Response refactor, increase requests-per-second, body type handling

MSP-Greg:
### Description
This PR refactors response generation, speeding up operations with various body types and sizes.  It also adds code to better handle the various body types returned by Rails and/or Rack.  Much of this has been worked on by @guilleiguaran, so thanks for his work on it (see #2892).  Essentially, body types fall into five categories:

1. An Array - Puma will determine the content-length if not provided.
2. An Enum - this would be chunked by Puma or sent as is if chunked encoding has already been applied.
3. A File object - dependent on its size, it will either be sent with the headers, or sent after the headers using `IO.copy_stream`.
4. An object that responds to `to_path` - Puma will attempt to open the file, and if successful, will send as above.
5. An object that responds to `call`.

This PR has been around for a while, and I'm sure I've run millions of requests thru it.  Two items:

1. Duplicating client errors is difficult.  I believe I've got the error handling correct, but we may see some issues with it.  Not.sure.
2. Previously, Puma didn't ever call `to_path` on a response body.  This PR has been updated to do so when the body responds to it, but isn't a File object.  It opens the file (based on the name in `to_path`), and treats it as a File/IO object.  This may cause issues with the [Rack::Sendfile](https://msp-greg.github.io/rack/Rack/Sendfile.html) middleware.

Below summarizes performance using the code in benchmarks/local, all using:
benchmarks/local/response_time_wrk.sh -w2 -t5:5 -s tcp6 -Y
wrk -t8 -c16 -d10s
Server cluster mode -w2 -t5:5, bind: tcp6
ruby 3.2.0dev (2022-09-12T13:13:32Z master 2aa8edaec7) +YJIT [x86_64-linux]
Both include 'test/rackup/ci_*.ru files - cache response bodies' (#2952)
Windows WSL2/Ubuntu 22.04

```

Body    ────────── req/sec ──────────   ─────── req 50% times ─────── Master
 KB     array   chunk  string      io   array   chunk  string      io
1       13780    9719   13581   10964    0.65    1.22    0.68    0.90
10       5920    2769   13203   10831    1.78    4.04    0.80    0.95
100      1063     404   11002    9153   10.54   26.85    0.98    1.21
256       448     159   11264    9039   24.39   65.65    0.92    1.23
512       234      85   10757    8848   45.18  121.76    0.96    1.22
1024      129      43    8822    8543   79.86  237.58    0.77    1.21
2048       68      22    6342    8405  152.13  456.80    1.54    1.21
[...]

Issue #2388, opened 2020-09-28
Removal of `Puma::Server#min_threads=` breaks Capybara


nertzy:
**Describe the bug**
With Puma 5.0.0, the latest Capybara properly boots my Rails test server.

With Puma 5.0.1, it gives an error.

I know this is technically a Capybara issue, but I wanted to let you know since it is a breaking change to Puma's API between 5.0.0 and 5.0.1 and wanted to see if that was expected.

See https://github.com/teamcapybara/capybara/issues/2397

**Puma config:**

See https://github.com/teamcapybara/capybara/blob/952c4c21e70ba8fcea6de469ae7d555c279cebf5/lib/capybara/registrations/servers.rb#L42

**To Reproduce**

- Make a rails app that uses Capybara to run system specs
- Upgrade puma to 5.0.1
- Run system specs

```
/Users/dev/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/puma-5.0.1/lib/puma/server.rb:87: warning: deprecated Object#=~ is called on Proc; it always returns
 nil
#<Thread:0x00007f97548d5778 /Users/dev/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/capybara-3.33.0/lib/capybara/server.rb:76 run> terminated with exception
 (report_on_exception is true):
Traceback (most recent call last):
        4: from /Users/dev/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/capybara-3.33.0/lib/capybara/server.rb:77:in `block in boot'
        3: from /Users/dev/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/capybara-3.33.0/lib/capybara/config.rb:64:in `block in server='
        2: from /Users/dev/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/capybara-3.33.0/lib/capybara/registrations/servers.rb:40:in `block in <main>'
        1: from /Users/dev/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/capybara-3.33.0/lib/capybara/registrations/servers.rb:40:in `tap'
/Users/dev/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/capybara-3.33.0/lib/capybara/registrations/servers.rb:42:in `block (2 levels) in <main>': undefined
method `min_threads=' for #<Puma::Server:0x00007f9755b5c658> (NoMethodError)
[...]

PR #2918, opened 2022-08-28
6.0.0

nateberkopec:


History.md
```diff
@@ -1,3 +1,34 @@
...
+  * [jruby] support setting TLS protocols + rename ssl_cipher_list ([#2899])
+
+* Bugfixes
+  * Load the configuration before passing it to the binder ([#2897])
+  * Do not remove pre-existing socket after USR2+TERM ([#2816]) ([#2817], [#2816])
+  * Fix `Puma::StateFile#load` returns empty string instead of nil ([#2810])
+
+* Refactor
+  * log_writer.rb - add internal_write method ([#2888])
+  * [WIP] Refactor: Split out LogWriter from Events (no logic change) ([#2798])
```
dentarg, line 27: Drop WIP tag and redundant word
```suggestion
  * Split out LogWriter from Events (no logic change) ([#2798])
```

History.md
```diff
@@ -1,3 +1,34 @@
...
+  * No longer supporting Java 1.7 or below (JRuby 9.1 was the last release to
+    support this) ([#2849])
+
+* Features
+  * We now log entire backtrace when worker start fails ([#2891])
+  * [jruby] enable TLSv1.3 support ([#2886])
+  * Export puma/ruby versions in /stats ([#2875])
+  * Allow configuring request uri max length & request path max length ([#2840])
+  * [jruby] support a truststore option ([#2849], [#2904], [#2884])
+  * Add a couple of accessors ([#2774])
```
dentarg, line 16: Let's spell out what actually was added
```suggestion
  * Add read accessors for `Puma::Cluster`'s `@workers` and Puma::Runner's `@options`. ([#2774]) 
```

dentarg (commented):
Some comments, haven't done any comparison with any branch. Are we doing a release of everything in master now?

MSP-Greg:
Can we hold off a day or two to see about merging either #2892 or #2896?

They both fix use with [`Rack::BodyProxy`](https://msp-greg.github.io/rack/Rack/BodyProxy.html).

Meanwhile, everyone can review HISTORY.md...

MSP-Greg:
Drop Support for older Rubies?  Maybe 2.2 & 2.3?

nateberkopec:
> Are we doing a release of everything in master now?

Yes

MSP-Greg:
There are three bugs in the 6.0.0 list, I believe the last two are listed in 5.6.5?

I can update the link section in History...

nateberkopec:
Addressed Greg's comment above, waiting on #2896 and #2892

nateberkopec:
OK! Everything marked 6.0 is now closed... last call? Anyone? Anything else kicking around your head Greg?

dentarg:
Will 6.0 address https://github.com/puma/puma/issues/2921?

MSP-Greg:
Maybe add #2942? After that's merged, maybe add a PR that removes the following (tests may need changes):
https://github.com/puma/puma/blob/5199ff293a675a7cbfde79ed81ab47278b246b2f/lib/puma/server.rb#L50-L53

I thought about adding ALPN for SSL, but wrk doesn't use it, so it can't be used to see what kind of performance difference it makes.  It should lessen the network traffic needed to initialize an SSL connection, but not sure whether it's significant in today's world.  Regardless, it wouldn't be a breaking change, so it could be added in 6.1.

MSP-Greg:
A quick look thru some code showed the below:

1. Should `MIN_THREADS` & `MAX_THREADS` be removed?
2. Should `WEB_CONCURRENCY` change to `PUMA_WEB_CONCURRENCY` (or even `PUMA_WORKERS`)?  Not sure if it's used by other apps.

https://github.com/puma/puma/blob/2b56d801cc03c248896f7f51b4c6f8c88113157d/lib/puma/configuration.rb#L220-L231****

MSP-Greg:
I totally forgot about changing to `require_relative`.  Noticed it in some code, then found #2902.

dentarg:
> 2\. Should `WEB_CONCURRENCY` change

I don't think it should change name, it could be breaking in the opposite way that [Puma 5.0 was](https://github.com/puma/puma/blob/15b079439222e489915380021b9dca7b9fa5ff65/5.0-Upgrade.md). 

Puma 5.0 introduced the use of the `WEB_CONCURRENCY` ENV in the defaults in #2143 (maybe inspired by the [Heroku documentation on Puma](https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server)?) The possible ramifications on Heroku wasn't understood until after release though: https://github.com/puma/puma/issues/2393#issuecomment-702352208

dentarg:
> Things that will break:
> 
> Rails: [actioncable/test/client_test.rb](https://github.com/rails/rails/blob/5fb3e721e70a13a7c0facf2d9175c15fb5132cc4/actioncable/test/client_test.rb#L76-L80)

`Puma::Server#min_threads`, `Puma::Server#max_threads` used here too: https://github.com/teamcapybara/capybara/blob/f7ab0b5cd5da86185816c2d5c30d58145fe654ed/lib/capybara/registrations/servers.rb#L43

History.md
```diff
@@ -1,3 +1,44 @@
...
+  * Do not raise error raised on HTTP methods we don't recognize or support, like CONNECT ([#2932], [#1441])
+  * Fixed a memory leak when creating a new SSL listener ([#2956])
+
+* Refactor
+  * log_writer.rb - add internal_write method ([#2888])
+  * [WIP] Refactor: Split out LogWriter from Events (no logic change) ([#2798])
+  * Extract prune_bundler code into it's own class. ([#2797])
+  * Refactor Launcher#run to increase readability (no logic change) ([#2795])
+  * Ruby 3.2 will have native IO#wait_* methods, don't require io/wait ([#2903])
+  * Various internal API refactorings ([#2942], [#2921], [#2922], [#2955])
```
dentarg, line 40: I think we need to spell this out or otherwise we get will issues like #2388 again

MSP-Greg:
Server's initialize method signature has changed between 6.0.0 and 5.6.5

6.0.0 - `def initialize(app, log_writer=LogWriter.stdio, events=Events.new, options = {})`

5.6.5 - `def initialize(app, events=Events.stdio, options={})`

Wondering if we should make the `log_writer` part of options, then revert the method signature change?  Thoughts?

dentarg:
Sounds like a good idea to me

MSP-Greg:
> (@MSP-Greg) Wondering if we should make the log_writer part of options, then revert the method signature change? Thoughts?

> (@dentarg) Sounds like a good idea to me


@dentarg & @nateberkopec - see PR #2965.

nateberkopec:
Good catch. It's of course a major version release so we can break it if we want but it does feel unnecessary. So I'm in favor of changing the signature back.

nateberkopec:
@MSP-Greg re: your comments on env variables, we'll keep those the same. As @dentarg mentioned, WEB_CONCURRENCY especially is a wider convention among Heroku apps.

I'm OK with our other new config-related breakages we're introducing, but as @dentarg suggested we should spell them out individually.

MSP-Greg:
> we should spell them out individually.

I was looking at that also, no time today to work on it.  I believe both 6.0.0 and 5.6.5 set min & max threads from the opts parameter in `Server#intialize`, so Capybara could either delete the line or guard it with `Server.respond_to? :min_threads=`.

Rails will need to change its code, but QED.  I'll check Capybara & Rails tomorrow with master.

MSP-Greg:
Speaking of breaking changes, below is what I know of that's affected:
* [Rails CI - with_puma_server](https://github.com/rails/rails/blob/main/actioncable/test/client_test.rb#L76)
* [Capybara](https://github.com/teamcapybara/capybara/blob/master/lib/capybara/registrations/servers.rb#L41-L44)
* [GoogleCloudPlatform/functions-framework-ruby](https://github.com/GoogleCloudPlatform/functions-framework-ruby/blob/main/lib/functions_framework/server.rb#L85-L92)

I've got a patch tested for Rails:
old:
```ruby
server = ::Puma::Server.new(rack_app, ::Puma::Events.strings)
server.add_tcp_listener "127.0.0.1", port
server.min_threads = 1
server.max_threads = 4
```
new:
```ruby
opts = { min_threads: 1, max_threads: 4 }
if Puma::Const::PUMA_VERSION >= '6'
  opts[:log_writer] = ::Puma::LogWriter.strings
  server = ::Puma::Server.new(rack_app, nil, opts)
else
  # Puma 5
  server = ::Puma::Server.new(rack_app, ::Puma::Events.strings, opts)
  server.min_threads = opts[:min_threads]
  server.max_threads = opts[:max_threads]
end
```

Seem ok?  We need to add another autoload to puma.rb for LogWriter, I can do tomorrow...

6.0-Upgrade.md
```diff
@@ -0,0 +1,50 @@
...
+3. **wait_for_less_busy_worker is now default, meaning lower latencies for high-utilization servers** `wait_for_less_busy_worker` was an experimental feature in Puma 5 and it's now the default in Puma 6. This feature makes each Puma child worker in cluster mode wait before listening on the socket, and that wait time is proportional to N * `number_of_threads_responding_to_requests`. This means that it's more likely that a request is picked up by the least-loaded Puma child worker listening on the socket. Many users reported back that this option was stable and decreased average latency, particularly in environments with high load and utilization.
+
+### Experimental Rack 3 Support
+
+[Rack 3 is now out](https://github.com/rack/rack/blob/main/UPGRADE-GUIDE.md) and we've started on Rack 3 support. Please open a bug if you find any incompatibilites.
+
+### Sidekiq 7 Capsules
+
+Sidekiq 7 (releasing soon) introduces Capsules, which allows you to run a Sidekiq server inside your Puma server (or any other Ruby process for that matter). We've added support by allowing you to pass data into `run_hooks`, see issue #2845.
```
bdewater, line 34: This seems to be the wrong issue number - I think it should be 2917?

MSP-Greg:
The flurry of PR's resulted from running a lot of CI to address the three repos (listed above) use of Puma in a nonstandard way.

MSP-Greg:
Just piled PR's #2966, #2967, #2968, #2969, #2971 into one branch, ran the Test workflow seven times.

**All MRI jobs passed**, with one failure in 'NON-MRI: macos-11 jruby', which was a Java SSL error (I think), and one failure in (guarded) 'NON-MRI: macos-11 truffleruby' which was 'truffleruby: Unexpected internal exception in at_exit,'

Hence, I think the intermittent failure rate will drop considerably...

This weekend I did see a few test timeouts, but the duration of the compile step was several times longer than normal jobs.  May see if that can be picked up by Actions...

nateberkopec:
todd-trapani-QldMpmrmWuc-unsplash

(image from unsplash, posting here for permanent URL to put in upgrade guide)

nateberkopec:
OK phew. I'm back, was busy moving into a new place the last two weeks.

nateberkopec:
OK, looks good to me. Would appreciate everyone's review.

MSP-Greg:
Will review.  Question - should we mark some classes as internal only?  Maybe Client, Worker classes, etc?

I did try to determine what repos were using Puma in a non-standard way, and all the ones I could find were creating a server to send test requests to.  IOW, by passing Launcher, Runner, etc.  So, most the Config related classes and Server.

I expect changes to Client (timers, IOBuffer, etc) and maybe others, and saving those changes until v7 would be a PITA...

Thoughts?

nateberkopec:
> Will review. Question - should we mark some classes as internal only? Maybe Client, Worker classes, etc?

Sure, we can start doing that. What mechanism do you propose? Rails used `:nodoc:` IIRC.

MSP-Greg:
See #2988 for a PR.  Looking for input as to what classes/modules should be considered 'private'.

PR #2920, opened 2022-08-30
Add on_thread_exit hook

### LINKED ISSUE #2100: How can I know when the thread is removed from the pool?
How can I know when the thread is removed from the pool?

I need to perform some actions on thread die but I can't.
The thread removed from the pool without any hooks.
https://github.com/puma/puma/blob/master/lib/puma/thread_pool.rb#L142

Thanks

biinari:
Closes #2100

### Description

For apps that need to tidy up thread-local resources when a thread exits, add a new `on_thread_exit` hook.
This will be called when a thread is being trimmed, just before the thread exits.

An example config entry involving rabbitmq channels being held in thread local variables (one channel per thread):

```ruby
on_thread_exit do
  unless Thread.current[MY_KEY].nil?
    Thread.current[MY_KEY].disconnect
    Thread.current[MY_KEY] = nil
  end
end
```

The hook should be configured with a quick non-blocking operation as it will block new requests from being started in other threads. I wondered about moving the call to the hook outside of the mutex section, but that becomes quite a bit less readable and I can't see what anyone would need to do that would be a blocking operation.


dentarg:
LGTM 👌 Needs a rebase though.

One thought, should we call it `before_thread_exit_hooks`? As you can have multiple hooks. I guess you followed `out_of_band_hook` which also could use an `s` on the end IMHO.

MSP-Greg:
Should there be a hook on the front end, immediately after the thread block starts?

Also, current code in this PR rescues outside of `before_thread_exit_hook.each(&:call)`.  With other hooks it's done inside the iteration, so if one hook errors, the others are still called.

biinari:
Thanks for the suggestions. I've rebased and followed the naming from the `out_of_band` which now has no `_hook` suffix in the instance variable.

@MSP-Greg, it sounds like you're asking for an additional feature with a hook at the start of running a thread. If you have a use case for needing this, perhaps create a separate issue? (Or correct me if I've misunderstood).

Changed the rescue to be per hook call as suggested, that's a good idea.

MSP-Greg:
@biinari

> to tidy up thread-local resources

Not an additional feature, but possibly a more complete implementation of the feature.  In this case, a logical place to create the 'thread-local resources' might be a companion `on_thread_start` (or `on_thread_enter`) hook.

biinari:
@MSP-Greg, actually that does sound sensible to add in with this as it would be a place to create thread-local resources or to monitor when threads are starting and stopping. I'll take a look at it soon assuming nobody disagrees.

nateberkopec:
@biinari Works for me too, please do investigate 👍

nateberkopec:
It looks like OP needed to move on to other things.

I don't think adding `on_thread_start` should actually block this though? We can merge as-is and make adding the complementary hook a separate issue.

biinari:
Thanks for merging this in. Sorry you're right, I've had my focus other projects for a while.

nateberkopec:
No worries homie, no one's on a timeline here 👍 Thanks for your contribution 🙇

PR #2927, opened 2022-09-06
Rack 3.0.0 release

MSP-Greg:
### Description

Rack 3.0.0 was released today, and breaks CI.

This PR makes the minimal changes needed to pass CI with Rack 3.0.0.

Currently, the Gemfile is `gem "rack", ">= 1.6.13"`.  Not sure if we should add an Actions workflow that runs a couple of jobs with Rack 2.2.x?

### DIFF
--- a/test/test_cli.rb
+++ b/test/test_cli.rb
@@ -49,7 +49,7 @@ class TestCLI < Minitest::Test
     cli = Puma::CLI.new ["-b", "tcp://127.0.0.1:0",
                          "--control-url", url,
                          "--control-token", "",
-                         "test/rackup/lobster.ru"], @log_writer, @events
+                         "test/rackup/hello.ru"], @log_writer, @events
 
     t = Thread.new do
       cli.run
@@ -84,7 +84,7 @@ class TestCLI < Minitest::Test
     cli = Puma::CLI.new ["-b", "tcp://127.0.0.1:0",
                          "--control-url", control_url,
                          "--control-token", token,
-                         "test/rackup/lobster.ru"], @log_writer, @events
+                         "test/rackup/hello.ru"], @log_writer, @events
 
     t = Thread.new do
       cli.run
@@ -120,7 +120,7 @@ class TestCLI < Minitest::Test
                          "-w", "2",
                          "--control-url", url,
                          "--control-token", "",
-                         "test/rackup/lobster.ru"], @log_writer, @events
+                         "test/rackup/hello.ru"], @log_writer, @events
 
     # without this, Minitest.after_run will trigger on this test ?
     $debugging_hold = true
@@ -168,7 +168,7 @@ class TestCLI < Minitest::Test
     cli = Puma::CLI.new ["-b", "unix://#{@tmp_path2}",
                          "--control-url", url,
                          "--control-token", "",
-                         "test/rackup/lobster.ru"], @log_writer, @events
+                         "test/rackup/hello.ru"], @log_writer, @events
 
     t = Thread.new { cli.run }
 
@@ -195,7 +195,7 @@ class TestCLI < Minitest::Test
     cli = Puma::CLI.new ["-b", "unix://#{@tmp_path2}",
                          "--control-url", url,
                          "--control-token", "",
-                         "test/rackup/lobster.ru"], @log_writer, @events
+                         "test/rackup/hello.ru"], @log_writer, @events
 
     t = Thread.new { cli.run }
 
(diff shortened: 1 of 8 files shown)


test/test_rack_server.rb
```diff
@@ -20,21 +20,22 @@ def call(env)
         @app.call(env)
       rescue Exception => e
         @exception = e
+STDOUT.syswrite "\n#{e.message}\n"
```
dentarg, line 23: Missing indentation?

dentarg:
>Not sure if we should add an Actions workflow that runs a couple of jobs with Rack 2.2.x?

We should I think

test/test_cli.rb
```diff
@@ -49,7 +49,7 @@ def test_control_for_tcp
     cli = Puma::CLI.new ["-b", "tcp://127.0.0.1:0",
                          "--control-url", url,
                          "--control-token", "",
-                         "test/rackup/lobster.ru"], @log_writer, @events
+                         "test/rackup/hello.ru"], @log_writer, @events
```
dentarg, line 52: Why this change?

test/test_config.rb
```diff
@@ -16,6 +17,7 @@ def test_default_max_threads
   end
 
   def test_app_from_rackup
+    skip_if :rack3
     conf = Puma::Configuration.new do |c|
       c.rackup "test/rackup/hello-bind.ru"
```
dentarg, line 22: Can we use the rackup gem instead of skipping the test? https://github.com/rack/rack/blob/main/UPGRADE-GUIDE.md (On phone atm so haven't checked myself)

test/test_rack_server.rb
```diff
@@ -20,21 +20,22 @@ def call(env)
...
         [ 500, {}, ["Error detected"] ]
       end
     end
   end
 
   class ServerLint < Rack::Lint
     def call(env)
-      check_env env
+      check_env env if Rack::RELEASE < '3'
```
dentarg, line 31: Didn't see removal of this method mentioned in the changelog for rack 3 https://github.com/rack/rack/blob/main/CHANGELOG.md

fyi @ioquatix

dentarg: Looks like it was renamed to `check_environment` in https://github.com/rack/rack/commit/eef251d3758741fec514356c117ff791a4c23347

test/test_rack_server.rb
MSP-Greg: Sorry, that was my debug code.  I'll remove.

test/test_cli.rb
MSP-Greg: I believe all the tests involve a control server, so the app isn't important?

test/test_rack_server.rb
MSP-Greg: Actually, `Wrapper.new(@app, env).check_environment`.  Added.

test/test_config.rb
MSP-Greg: Loading a gem that has webrick as a dependency seems odd.  1,249 downloads since 03-Aug-22.

MSP-Greg:
Added two jobs to a new Rack_v2 workflow, runs at the same time as the RuboCop workflow, so it shouldn't add much time to CI

test/test_config.rb
dentarg: Oh right, yeah never mind...

test/test_cli.rb
dentarg: Just trying to understand the motivation behind making the change (to learn something)... has `Rack::Lobster` been removed?

MSP-Greg:
Rebased with [Prefix all environment variables with PUMA_](https://github.com/puma/puma/commit/75853f2520a2ee9014cf1a3f627764aba21e1210)

test/test_cli.rb
MSP-Greg: It's now part of rackup.  https://github.com/rack/rackup/blob/main/lib/rackup/lobster.rb

Sorry, when Rack v3.0.0 was released, I was in the middle of some Puma code, and wasn't too happy.

Maybe we should add a CI job that allows pre-releases and all dependencies are `>=`.  Give us better 'early warning'?

dentarg: Sure

nateberkopec:
Thanks for the work Greg!

ioquatix:
I've added this puma-head-rack-v3 to my conformance tests:

https://github.com/socketry/rack-conform/runs/8246190238?check_suite_focus=true

It failed because you may not be supporting Array headers. Can you check?

dentarg:
Added that as an issue so we don't forget to follow up https://github.com/puma/puma/issues/2931

PR #2928, opened 2022-09-08
Refactor/move configuration defaults

### LINKED ISSUE #2922: Puma 6 - Configuration constants
Some constants are defined in [`Puma::ConfigDefault`](https://msp-greg.github.io/puma/Puma/ConfigDefault.html), and some are in defined in [`Puma::Const`](https://msp-greg.github.io/puma/Puma/Const.html) (`FIRST_DATA_TIMEOUT`, `PERSISTENT_TIMEOUT`, `MAX_FAST_INLINE`, etc).

One issue involving 'config' is when the default values are added/merged.  This doesn't involve 'normal' CLI use of Puma, but repo's using Puma to run up a server (often for testing) do so by creating a server via `server = Server.new; ...; server.run`.

Wondering about getting all the config constants in one place, and not having any in `Puma::Const`, which has a lot of constants related to HTTP (strings, status codes, etc).

Any thoughts?

Sorry about the maybe self-serving links, they're easy for me to grab...

nateberkopec:
~~(this is not signed because I didn't have my key with me, will sign and force push later)~~

Signed now

Closes #2922

### DIFF
--- a/lib/puma/configuration.rb
+++ b/lib/puma/configuration.rb
@@ -155,6 +155,7 @@ module Puma
       tag: File.basename(Dir.getwd),
       tcp_host: '0.0.0.0'.freeze,
       tcp_port: 9292,
+      wait_for_less_busy_worker: 0.005,
       worker_boot_timeout: 60,
       worker_check_interval: 5,
       worker_culling_strategy: :youngest,
--- a/lib/puma/configuration.rb
+++ b/lib/puma/configuration.rb
@@ -5,17 +5,6 @@ require 'puma/plugin'
 require 'puma/const'
 
 module Puma
-
-  module ConfigDefault
-    DefaultRackup = "config.ru"
-
-    DefaultTCPHost = "0.0.0.0"
-    DefaultTCPPort = 9292
-    DefaultWorkerCheckInterval = 5
-    DefaultWorkerTimeout = 60
-    DefaultWorkerShutdownTimeout = 30
-  end
-
   # A class used for storing "leveled" configuration options.
   #
   # In this class any "user" specified options take precedence over any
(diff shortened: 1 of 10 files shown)


lib/puma/configuration.rb
```diff
@@ -136,7 +125,48 @@ def final_options
   # is done because an environment variable may have been modified while loading
   # configuration files.
   class Configuration
-    include ConfigDefault
+    DEFAULTS = {
```
nateberkopec, line 128: TODO: alphabetize

test/test_config.rb
```diff
@@ -539,7 +539,6 @@ def test_config_loads_correct_min_threads
 
   def test_config_loads_correct_max_threads
     conf = Puma::Configuration.new
-    assert_equal conf.default_max_threads, conf.options.default_options[:max_threads]
```
nateberkopec, line 542: this is now covered by another test

nateberkopec:
I think this is good but would like a review

lib/rack/handler/puma.rb
```diff
@@ -93,16 +93,16 @@ def self.set_host_port_to_config(host, port, config)
           config.bind "unix://#{host}"
         elsif host && host =~ /^ssl:\/\//
           uri = URI.parse(host)
-          uri.port ||= port || ::Puma::Configuration::DefaultTCPPort
+          uri.port ||= port || ::Puma::Configuration::DEFAULTS[:tcp_port]
```
JuanitoFatas, line 96: Question 🙋🏻‍♀️  

Should this read from a method `Puma.options` (that returns `#puma_default_options`), so it reads values from user configurations, otherwise it always defaults to puma’s default (`9292`)?

Same questions for other `::Puma::Configuration::DEFAULTS` in this Pull Request 🙇🏻‍♂️

JuanitoFatas: I see now we are only keeping the same behavior as before 👍🏼

Issue #1441, opened 2017-10-17
Puma doesn't handle HTTP CONNECT method


nirvdrum:
### Steps to reproduce

1) Run puma with any app of your choosing. Let's assume the app is running on port 9292.

2) Make an HTTPS request to this app, as if it were a proxy server: `env http_proxy=http://localhost:9393 wget https://github.com/`. It doesn't matter that the app doesn't handle this route, since puma will error out before ever reaching the app.

### Expected behavior

I was looking to build a proxy server with Puma. It's probably too much to expect Puma to properly handle HTTP CONNECT methods, but if it's not going to, a nice error message would be helpful.

### Actual behavior

Instead of my request being handled or a friendly error message printed, I see an unrelated error message and stack trace:

```
#<RuntimeError: No REQUEST PATH>
/home/kjmenard/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/puma-3.10.0/lib/puma/server.rb:526:in `normalize_env'
/home/kjmenard/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/puma-3.10.0/lib/puma/server.rb:580:in `handle_request'
/home/kjmenard/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/puma-3.10.0/lib/puma/server.rb:437:in `process_client'
/home/kjmenard/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/puma-3.10.0/lib/puma/server.rb:301:in `block in run'
/home/kjmenard/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/puma-3.10.0/lib/puma/thread_pool.rb:120:in `block in spawn_thread'
```

There is indeed no request path passed with an HTTP CONNECT method. That part of the validation is correct. But, since the request path isn't required for HTTP CONNECT, I think the error message is  a bit misleading. If Puma isn't going to handle the method, it should probably reject the request at an earlier stage.

### System configuration

**Ruby version**: 2.3.3
**Rails version**: N/A
**Puma version**: 3.10.0

nateberkopec:
I took a look at this again, I'm not sure why we have to reject requests without a path. The Rack spec clearly indicates `PATH_INFO` is optional, and `REQUEST_PATH` seems to be a non-spec addition which is I guess equal to `SCRIPT_NAME` plus `PATH_INFO`.

PR #2932, opened 2022-09-08
Do not raise error on CONNECT

mrzasa:
### Description

Closes https://github.com/puma/puma/issues/1441

I'm trying to fix https://github.com/puma/puma/issues/1441 and to do this I've prepare a proof of concept. I've done two things:
1. Don't raise error on empty `REQUEST_PATH` - it's a completely valid request to send CONNECT without a request path.
2. Handle the fact that puma doesn't support `CONNECT` gracefully by responding with 501.

I'm not sure if I handle both things in the right places. Right now I have a working code with a simple test. I'd be grateful for feedback and hints how to improve the code to make it mergeable.

### DIFF
--- a/lib/puma/const.rb
+++ b/lib/puma/const.rb
@@ -163,6 +163,14 @@ module Puma
 
     REQUEST_METHOD = "REQUEST_METHOD".freeze
     HEAD = "HEAD".freeze
+    GET = "GET".freeze
+    POST = "POST".freeze
+    PUT = "PUT".freeze
+    DELETE = "DELETE".freeze
+    OPTIONS = "OPTIONS".freeze
+    TRACE = "TRACE".freeze
+    PATCH = "PATCH".freeze
+    SUPPORTED_HTTP_METHODS = [HEAD, GET, POST, PUT, DELETE, OPTIONS, TRACE, PATCH].freeze
     # ETag is based on the apache standard of hex mtime-size-inode (inode is 0 on win32)
     LINE_END = "\r\n".freeze
     REMOTE_ADDR = "REMOTE_ADDR".freeze
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -73,8 +73,13 @@ module Puma
 
       begin
         begin
-          status, headers, res_body = @thread_pool.with_force_shutdown do
-            @app.call(env)
+          if SUPPORTED_HTTP_METHODS.include?(env[REQUEST_METHOD])
+            status, headers, res_body = @thread_pool.with_force_shutdown do
+              @app.call(env)
+            end
+          else
+            @log_writer.log "Unsupported HTTP method used: #{env[REQUEST_METHOD]}"
+            status, headers, res_body = [501, {}, ["#{env[REQUEST_METHOD]} method is not supported"]]
           end
 
           return :async if client.hijacked
@@ -266,14 +271,12 @@ module Puma
         uri = URI.parse(env[REQUEST_URI])
         env[REQUEST_PATH] = uri.path
 
-        raise "No REQUEST PATH" unless env[REQUEST_PATH]
-
         # A nil env value will cause a LintError (and fatal errors elsewhere),
         # so only set the env value if there actually is a value.
         env[QUERY_STRING] = uri.query if uri.query
       end
 
-      env[PATH_INFO] = env[REQUEST_PATH]
+      env[PATH_INFO] = env[REQUEST_PATH].to_s # #to_s in case it's nil
 
       # From https://www.ietf.org/rfc/rfc3875 :
       # "Script authors should be aware that the REMOTE_ADDR and
--- a/test/test_web_server.rb
+++ b/test/test_web_server.rb
@@ -79,6 +79,20 @@ class WebServerTest < Minitest::Test
     socket.close
   end
 
+  def test_unsupported_method
+    socket = do_test("CONNECT www.zedshaw.com:443 HTTP/1.1\r\nConnection: close\r\n\r\n", 100)
+    response = socket.read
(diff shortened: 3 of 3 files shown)


lib/puma/request.rb
```diff
@@ -73,8 +73,13 @@ def handle_request(client, lines, requests)
 
       begin
         begin
-          status, headers, res_body = @thread_pool.with_force_shutdown do
-            @app.call(env)
+          if env[REQUEST_METHOD] != CONNECT
```
dentarg, line 76: Is this necessary? What happens without it?

lib/puma/request.rb
```diff
@@ -266,14 +271,12 @@ def normalize_env(env, client)
         uri = URI.parse(env[REQUEST_URI])
         env[REQUEST_PATH] = uri.path
 
-        raise "No REQUEST PATH" unless env[REQUEST_PATH]
```
nateberkopec, line 269: Actually, the Rack spec says:

```
PATH_INFO: This may be an empty string, if the request URL targets the application root and does not have a trailing slash.
```

So this is also a rack conformance issue

lib/puma/request.rb
```diff
@@ -266,14 +271,12 @@ def normalize_env(env, client)
...
-
         # A nil env value will cause a LintError (and fatal errors elsewhere),
         # so only set the env value if there actually is a value.
         env[QUERY_STRING] = uri.query if uri.query
       end
 
-      env[PATH_INFO] = env[REQUEST_PATH]
+      env[PATH_INFO] = env[REQUEST_PATH].to_s # #to_s in case it's nil
```
nateberkopec, line 279: See above comment. PATH_INFO should always be a string.

lib/puma/request.rb
```diff
@@ -73,8 +73,13 @@ def handle_request(client, lines, requests)
...
       begin
         begin
-          status, headers, res_body = @thread_pool.with_force_shutdown do
-            @app.call(env)
+          if env[REQUEST_METHOD] != CONNECT
+            status, headers, res_body = @thread_pool.with_force_shutdown do
+              @app.call(env)
+            end
+          else
+            @log_writer.log "CONNECT method used, but it's not supported by puma."
```
nateberkopec, line 81: Actually, the [text on MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501) suggests a little different logic is possible:

```ruby 
if ALLOWED_HTTP_METHODS.include?(env[REQUEST_METHOD])
  # call app 
else 
  # return 501
end
```

test/test_web_server.rb
```diff
@@ -79,6 +79,13 @@ def test_file_streamed_request
     socket.close
   end
 
+  def test_unsupported_method
```
nateberkopec, line 82: Re: my suggested change above, let's also add one additional test with a nonsense/non-existent HTTP method, like:

```do_test(ASDJLKQW www.zedshaw.com...)```

**Review by nateberkopec — CHANGES_REQUESTED** [review]
Left some comments

File: lib/puma/request.rb (hunk 3 above)
Comment (mrzasa) [reply lib/puma/request.rb:279]: PATH_INFO - sure. but in the tested scenatio REQUEST_PATH is not a string but nil. To make sure we always assign a string (even if it's empty) to PATH_INFO, I'm calling `to_s`

File: lib/puma/request.rb (hunk 2 above)
Comment (mrzasa) [reply lib/puma/request.rb:269]: Rack spec specifies the PATH_INFO and REQUEST_SPEC seems to be a non-spec addition (to quote you) https://github.com/puma/puma/issues/1441#issuecomment-774784265. Should puma fail when it's nil? Maybe we could just assign an empty PATH_INFO then (as I do below)?

**Review by mrzasa — COMMENTED** [review]
Thanks for your feedback, folks!

File: test/test_web_server.rb (hunk 5 above)
Comment (mrzasa) [reply test/test_web_server.rb:82]: Done :)

**Review by mrzasa — COMMENTED** [review]
I've refactored the check of unsupported methods.

File: lib/puma/const.rb [hunk 6]
```diff
@@ -163,6 +163,14 @@ module Const
...
     REQUEST_METHOD = "REQUEST_METHOD".freeze
     HEAD = "HEAD".freeze
+    GET = "GET".freeze
+    POST = "POST".freeze
+    PUT = "PUT".freeze
+    DELETE = "DELETE".freeze
+    OPTIONS = "OPTIONS".freeze
+    TRACE = "TRACE".freeze
+    PATCH = "PATCH".freeze
+    SUPPORTED_HTTP_METHODS = [HEAD, GET, POST, PUT, DELETE, OPTIONS, TRACE, PATCH]
```
nateberkopec, line 173: Should be frozen too

lib/puma/request.rb
nateberkopec: It looks like for HTTP URIs it will always return at least an empty string, but if someone gives us a weird protocol maybe it could be nil.

I think nil.to_s is a reasonable conversion.

nateberkopec (changes requested):
Let's freeze the array. Mark this as reviewable and let's 🚢

lib/puma/request.rb
mrzasa: I was testing it with `config.ru` 
```
require "localhost"
run proc { [200, {"Content-Type" => "text/plain"}, ["Hello, World!"]] }
```
and a request
```
 wget https://github.com/ -e https_proxy=http://localhost:9292
```

With this request REQUEST_PATH was `nil`.

lib/puma/const.rb
mrzasa: done

nateberkopec (approved):

PR #2936, opened 2022-09-09
Allow header values to be arrays (Rack 3)

### LINKED ISSUE #2931: Does Puma support Array headers?
I've added this puma-head-rack-v3 to my conformance tests:

https://github.com/socketry/rack-conform/runs/8246190238?check_suite_focus=true

It failed because you may not be supporting Array headers. Can you check?

_Originally posted by @ioquatix in https://github.com/puma/puma/issues/2927#issuecomment-1240473585_

MSP-Greg:
### Description

Closes #2931

### DIFF
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -464,8 +464,15 @@ module Puma
           next
         end
 
-        if vs.respond_to?(:to_s) && !vs.to_s.empty?
-          vs.to_s.split(NEWLINE).each do |v|
+        ary = if vs.is_a?(::Array) && !vs.empty?
+          vs
+        elsif vs.respond_to?(:to_s) && !vs.to_s.empty?
+          vs.to_s.split NEWLINE
+        else
+          nil
+        end
+        if ary
+          ary.each do |v|
             next if illegal_header_value?(v)
             lines.append k, colon, v, line_ending
           end
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -167,6 +167,7 @@ module TestSkips
       when :fork    then MSG_FORK                       unless HAS_FORK
       when :unix    then MSG_UNIX                       unless Puma::HAS_UNIX_SOCKET
       when :aunix   then MSG_AUNIX                      unless Puma.abstract_unix_socket?
+      when :rack3   then "Skipped unless Rack >= 3.x"   unless ::Rack::RELEASE >= '3'
       else false
     end
     skip skip_msg, bt if skip_msg
--- a/test/test_response_header.rb
+++ b/test/test_response_header.rb
@@ -1,9 +1,13 @@
 require_relative "helper"
 require "puma/events"
 require "net/http"
+require "nio"
 
 class TestResponseHeader < Minitest::Test
   parallelize_me!
+
+  # this file has limited response length, so 10kB works.
+  CLIENT_SYSREAD_LENGTH = 10_240
 
   def setup
     @host = "127.0.0.1"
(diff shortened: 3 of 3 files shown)


test/test_response_header.rb
```diff
@@ -29,7 +30,7 @@ def server_run(app: @app, early_hints: false)
   end
 
   def send_http_and_read(req)
-    send_http(req).read
+    send_http(req).sysread 10_240
```
nateberkopec, line 33: 🤔

test/helper.rb
```diff
@@ -167,6 +167,7 @@ def skip_unless(eng, bt: caller)
       when :fork    then MSG_FORK                       unless HAS_FORK
       when :unix    then MSG_UNIX                       unless Puma::HAS_UNIX_SOCKET
       when :aunix   then MSG_AUNIX                      unless Puma.abstract_unix_socket?
+      when :rack3   then "Skipped unless Rack >= 3.x"   unless ::Rack::RELEASE >= '3'
```
nateberkopec, line 170: Accidental?

lib/puma/request.rb
```diff
@@ -464,8 +464,15 @@ def str_headers(env, status, headers, res_info, lines, requests, client)
...
         end
 
-        if vs.respond_to?(:to_s) && !vs.to_s.empty?
-          vs.to_s.split(NEWLINE).each do |v|
+        ary = if vs.is_a?(::Array) && !vs.empty?
+          vs
+        elsif vs.respond_to?(:to_s) && !vs.to_s.empty?
+          vs.to_s.split NEWLINE
+        else
+          nil
```
nateberkopec, line 472: if we just set this to `[]` then we can skip the `if ary` check

test/test_response_header.rb
```diff
@@ -141,4 +142,12 @@ def test_illegal_character_in_value_when_newline
 
     refute_match("X-header: First\000 line\r\nX-header: Second Lin\037e\r\n", data)
   end
+
+  def test_header_value_array
```
nateberkopec, line 146: Are we introducing incompatibilies with this kind of thing in a Rack 2 app? I suppose not, since a Rack 2 app working like this is sort of by definition actually a Rack 3 app?

nateberkopec (commented):
Just left some thoughts.

test/test_response_header.rb
MSP-Greg: `read` is evil.  I was working on this, and with `read`, the new test took 20 sec.  With `sysread`, it was what it should be.

But, then the Windows tests (only Ruby 3.1 and later), weren't getting the full response.  But, using PR #2896, no problem.

Some of the issue is using the same Process for the server and client...

test/helper.rb
MSP-Greg: Should have been in the previous Rack 3 PR/commit.  I thought I might need it here, but there aren't any Rack objects...

test/test_response_header.rb
MSP-Greg: I doubt a Rack 2 app will ever pass an array as a header value, they'll hit the `vs.to_s.split NEWLINE` code?

lib/puma/request.rb
MSP-Greg: Thanks.  Done.

test/test_response_header.rb
nateberkopec: I see 👍 Lets refactor the magic number into a constant somwhere then.

lib/puma/request.rb
MSP-Greg: This broke CI, I'm going to revert.

test/test_response_header.rb
MSP-Greg: Done.  Added comment also, as the value is dependent on the tests.  I've used looping code for responses with unknown length...

PR #2948, opened 2022-09-10
Removes debug info from tests by default

nateberkopec:
New output:

```
install -c -p -m 755 puma_http11.so /host/nateberkopec/Documents/Code.nosync/upstream/puma/lib/puma
Run options: --seed 52639

# Running:

SSS................................................................................................................................................................................................S...................................S...............................................S.S...S...........................................S..........................................................................................................................................................................................................

Fabulous run in 133.758721s, 3.9773 runs/s, 11.5731 assertions/s.

532 runs, 1548 assertions, 0 failures, 0 errors, 9 skips

You have skipped tests. Run with --verbose for details.
```

CI output remains the same.

### DIFF
--- a/.github/workflows/rack2.yaml
+++ b/.github/workflows/rack2.yaml
@@ -16,7 +16,7 @@ jobs:
     env:
       CI: true
       TESTOPTS: -v
-      PUMA_DEBUG: true
+      PUMA_TEST_DEBUG: true
       PUMA_NO_RUBOCOP: true
       PUMA_CI_RACK_2: true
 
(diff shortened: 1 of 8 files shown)


nateberkopec:
Hmm actually PUMA_DEBUG is probably a bad choice, it causes some tests to fail due to things being output when they weren't before.

MSP-Greg:
I was going to switch it to `PUMA_TEST_DEBUG`, and I noticed the change to [test/test_preserve_bundler_env.rb](https://github.com/puma/puma/pull/2948/files#diff-5a6691f581cb661446bb58166c52cb80c772519d1997fb6496884ea32e8c1e1b).

Looks like debugging the local bundler issue, should it be removed?

nateberkopec:
Cheers, thanks Greg!

PR #2956, opened 2022-09-15
minissl.c - immediate ruby_memcheck fixes

MSP-Greg:
### Description

Spent most of this evening working with [ruby_memcheck](https://github.com/Shopify/ruby_memcheck).  Code in this PR fixes a few memory leaks, some of which occur when improper data is used to create an SSL connection/listener.

All of the other results were calling `rb_define_` functions, which may be false positives.  Or, these are Ruby functions that are called when one requires Puma, as they are the functions that allow Ruby to know that a `Puma` module, or a `Puma::Server` class exist.

Note that the changes fix leaks in an object that is created once for each SSL listener.  I believe there are no leaks on the client/ request/accepted socket level.

I need to spend some more time working with ruby_memcheck...


nateberkopec:
Amazing! "Leaked on each SSL listener" is probably pretty low real world impact because those probably live for the lifetime of a Puma server 99% of the time but still glad we've fixed it.

PR #2964, opened 2022-09-20
require to require_relative

### LINKED ISSUE #2902: Time to change to require_relative?

MSP-Greg:
### Description

Change use of `require` to `require_relative` where appropriate.

Closes #2902

Actions macOS runners seem to be having some issues.

### DIFF
--- a/lib/puma.rb
+++ b/lib/puma.rb
@@ -10,14 +10,16 @@ require 'stringio'
 
 require 'thread'
 
+# use require, see https://github.com/puma/puma/pull/2381
 require 'puma/puma_http11'
-require 'puma/detect'
-require 'puma/json_serialization'
+
+require_relative 'puma/detect'
+require_relative 'puma/json_serialization'
 
 module Puma
-  autoload :Const, 'puma/const'
-  autoload :Server, 'puma/server'
-  autoload :Launcher, 'puma/launcher'
+  autoload :Const,    "#{__dir__}/puma/const"
+  autoload :Server,   "#{__dir__}/puma/server"
+  autoload :Launcher, "#{__dir__}/puma/launcher"
 
   # at present, MiniSSL::Engine is only defined in extension code (puma_http11),
   # not in minissl.rb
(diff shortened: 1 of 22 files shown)


MSP-Greg:
Actions has issues earlier. I reran the test workflow's failed jobs.  JFYI. passed

lib/puma.rb
```diff
@@ -10,14 +10,16 @@
...
-require 'puma/json_serialization'
+
+require_relative 'puma/detect'
+require_relative 'puma/json_serialization'
 
 module Puma
-  autoload :Const, 'puma/const'
-  autoload :Server, 'puma/server'
-  autoload :Launcher, 'puma/launcher'
+  autoload :Const,    "#{__dir__}/puma/const"
```
nateberkopec, line 20: This feels weird. Is this weird? I've never seen this before.

MSP-Greg: > This feels weird. Is this weird? I've never seen this before.

If it's an absolute path, it 'bypasses' the RubyGems resolution.  So

`autoload :Const, "puma/const"` resolved by RubyGems, similar to `require`
`autoload :Const, "#{__dir__}/puma/const"` resolved as a path, no RubyGems, similar to `require_relative`

The Ruby docs for `Kernel.autoload` show an absolute path, `Module.autoload` show a 'RubyGems' like path.

I think most code paths that are followed with 'normal' Puma use will see `require_relative` for all the files autoloaded here, so these are kind of helpers for running single test files.  `autload` only does something if the constant isn't defined...

PR #2983, opened 2022-10-03
CONTRIBUTING: spell out how to change ulimit [ci skip]

dentarg:
We link to https://wilsonmar.github.io/maximum-limits/, which still looks like a good resource, but it is ~4 years old now and talks a lot about older macOS versions. I found that `sudo launchctl limit maxfiles ...` didn't do anything for me in macOS Monterey. Let's just include the simple command to change the limit for the current session in our guide.


JuanitoFatas (approved):
Nice 👍🏻

PR #2997, opened 2022-10-23
[CI] - Update client socket methods, fix intermittent failures

MSP-Greg:
### Description

1. Implement a new file `test/helpers/sockets.rb`, which collects methods for creating client connections in tests.  Seems to increase CI speed and reliability.  Begin conversion of existing files to use the new methods.

2. `hot_restart_does_not_drop_connections` tests are run in both single and clustered integration tests.  There are intermittent failures when checking for whether the server has been restarted.  Most involve errors with the IO provided by `IO.popen`.  The main purpose of the tests are to verify that client connections are handled correctly.  So, add more error handling to the server checks.  Note that the there are mutliple restarts, and the client connections behave as expected.

3. Fix a TruffleRuby issue.  If an added error class is within the test class, the tests fail.  Moving it outside the class (so it's defined at root) solves the problem.

Locally, I noticed that test_puma_server.rb ran faster with this patch.  Interested in whether others find the same.

### DIFF
--- a/test/helpers/sockets.rb
+++ b/test/helpers/sockets.rb
@@ -48,13 +48,13 @@ module TestPuma
       while n < byte_size
         begin
           n += syswrite(n.zero? ? str : str.byteslice(n..-1))
-        rescue Errno::EAGAIN, Errno::EWOULDBLOCK
+        rescue Errno::EAGAIN, Errno::EWOULDBLOCK => e
           unless wait_writable WRITE_TIMEOUT
-            raise ConnectionError, SKT_WRITE_ERR_MSG
+            raise e
           end
           retry
-        rescue  Errno::EPIPE, SystemCallError, IOError
-          raise ConnectionError, SKT_WRITE_ERR_MSG
+        rescue => e
+          raise e
         end
       end
     end
--- a/test/helpers/sockets.rb
+++ b/test/helpers/sockets.rb
@@ -1,0 +1,171 @@
+require 'socket'
+
+module TestPuma
+
+  # @!macro [new] req
+  #   @param [String] path request uri path
+  #   @param [Float] dly: delay in app when using 'ci' rackups
+  #   @param [Integer, String] body_conf: response body type and size in kB when using 'ci' rackups
+
+  # @!macro [new] tout
+  #   @param [Float] timeout: read timeout for socket
+
+  # @!macro [new] ret_skt
+  #   @return [SktSSL, SktTCP, SktUNIX] the opened socket
+
+
+  RESP_READ_TIMEOUT = 10
+
+  module SktPrepend
+
+    RESP_READ_LEN = 1_024 * 64
+    RESP_SPLIT = "\r\n\r\n"
+
+    WRITE_TIMEOUT = 5
+
+    # Writes a string to the socket using `syswrite`.
+    # @param [String] str
+    # @return [Integer] number of bytes written
+    def write(str)
+      fast_write str
+    end
+
+    # Writes a string to the socket using `syswrite`.
+    # @param [String] str
+    # @return [self]
+    def <<(str)
+      fast_write str
+      self
+    end
+
+    # Writes a string to the socket using `syswrite`.
+    # @param [String] str the string to write
+    # @return [Integer] the number of bytes written
+    def fast_write(str)
+      return unless str.is_a? String
+      n = 0
+      byte_size = str.bytesize
+      while n < byte_size
+        begin
+          n += syswrite(n.zero? ? str : str.byteslice(n..-1))
+        rescue Errno::EAGAIN, Errno::EWOULDBLOCK
+          unless wait_writable WRITE_TIMEOUT
+            raise ConnectionError, SKT_WRITE_ERR_MSG
+          end
+          retry
+        rescue  Errno::EPIPE, SystemCallError, IOError
+          raise ConnectionError, SKT_WRITE_ERR_MSG
+        end
+      end
+    end
+
+    # Returns the response body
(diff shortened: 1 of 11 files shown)


test/helpers/sockets.rb
```diff
@@ -0,0 +1,171 @@
...
+      while n < byte_size
+        begin
+          n += syswrite(n.zero? ? str : str.byteslice(n..-1))
+        rescue Errno::EAGAIN, Errno::EWOULDBLOCK
+          unless wait_writable WRITE_TIMEOUT
+            raise ConnectionError, SKT_WRITE_ERR_MSG
+          end
+          retry
+        rescue  Errno::EPIPE, SystemCallError, IOError
+          raise ConnectionError, SKT_WRITE_ERR_MSG
```
dentarg, line 57: The definition of `SKT_WRITE_ERR_MSG` seems to be missing

MSP-Greg: Fixed.  I think `ConnectionError` was also a mistake.  Some of this code came from a larger file, which had code for generating a stream of clients, similar to what's in `integration.rb`'s `hot_restart_does_not_drop_connections`.  Sorry for the error.

test/helpers/sockets.rb
```diff
@@ -0,0 +1,171 @@
...
+    # The SSLSocket class used by the TestPuma framework.  The `SktPrepend` module
+    # is prepended.  The socket is opened with parameters set by `bind_type`.
+    class SktSSL < ::OpenSSL::SSL::SSLSocket
+      prepend SktPrepend
+    end
+  end
+
+  # The TCPSocket class used by the TestPuma framework.  The `SktPrepend` module
+  # is prepended.  The socket is opened with parameters set by `bind_type`.
+  class SktTCP < ::TCPSocket
```
nateberkopec, line 160: I think we should use `Socket` over `Skt` everywhere

test/helpers/sockets.rb
```diff
@@ -0,0 +1,171 @@
...
+          part = read_nonblock(RESP_READ_LEN, exception: false)
+          case part
+          when String
+            unless content_length || chunked
+              chunked ||= part.include? "\r\nTransfer-Encoding: chunked\r\n"
+              content_length = (t = part[/^Content-Length: (\d+)/i , 1]) ? t.to_i : nil
+            end
+
+            response << part
+            hdrs, body = response.split RESP_SPLIT, 2
```
nateberkopec, line 92: Same with these new constant names, let's spell out `RESPONSE_SPLIT` and `RESPONSE_*`

test/helpers/sockets.rb
```diff
@@ -0,0 +1,171 @@
...
+    end
+  end
+
+  # The TCPSocket class used by the TestPuma framework.  The `SktPrepend` module
+  # is prepended.  The socket is opened with parameters set by `bind_type`.
+  class SktTCP < ::TCPSocket
+    prepend SktPrepend
+  end
+
+  if Object.const_defined? :UNIXSocket
```
nateberkopec, line 164: What does moving this logic into a module which is prepended to a custom socket class get us that we didn't have before? I guess I'm not thrilled by having to change the method calls everywhere, and for newbies to Puma contributing to have to learn that we have a special socket class that we use.

MSP-Greg: > What does moving this logic into a module which is prepended to a custom socket class get us that we didn't have before?

Correct handling of sockets.  Using read, write, gets, etc uses buffering in Ruby, which can be preempted by other code.  As I saw recently, a test took 20 seconds with read, but less than a second with sysread/read_nonblock.

Also, this is the start of the conversion.  Two additional helper files handle creation of in-proc servers and popen servers.  Each has a method (`startServer` and `startPuma`) with similar setup.  Bind can be `tcp`, tcp4`, `tcp6`, the same for ssl, along with `aunix` and `unix`.  To create a client socket, a `connect` method is used, and it knows what kind of socket to create.

The server helper files and the sockets file are aware of each other.  In this PR I removed all the methods from sockets.rb that are dependent on the server helper files.

Additionally, we have issues with sockets not being closed, unix socket files remaining, etc.  If one looks over the current test suite, client sockets are created in different ways, same for servers.  IOW, there is probably a lot duplicated 'helper' code.

I think most people write tests by looking at other similar tests.  We can also doc the new test DSL methods, I thought placing them in a `TestPuma` namespace would allow them to be included in the docs without much distraction.  If the additional methods clutter up the methods list, we can just doc them with md files.

Lastly, test times
```
Ubuntu test_puma_server.rb
        This PR      master
serial    26.0       86.0
parallel   3.1       40.9 
```

**Off-topic**
Regardless of what we do with this, maybe it's time to divide up some of the test files?  It varies from file to file, but TestPumaServer has 96 tests...

test/test_integration_pumactl.rb
```diff
@@ -64,7 +64,7 @@ def test_phased_restart_cluster
 
     start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
 
-    s = UNIXSocket.new @bind_path
+    s = TestPuma::SktUNIX.new @bind_path
```
nateberkopec, line 67: This kind of change is what I'm not thrilled about. Are you saying that these tests should be converted to use methods from `test/helpers/integration.rb` anyway?

nateberkopec:
Closing as stale, Greg I'm sure will either circle back or fix this in a different way eventually 😆

dentarg:
https://github.com/puma/puma/pull/3179 maybe be the replacement for parts of this PR (see https://github.com/puma/puma/pull/3179#issuecomment-1598027679)

PR #3001, opened 2022-10-26
[CI] Fix TruffleRuby - TestPumaServer#test_lowlevel_error_message_without_backtrace

MSP-Greg:
### Description

Fixes failure on TruffleRuby

### DIFF
--- a/test/test_puma_server.rb
+++ b/test/test_puma_server.rb
@@ -5,6 +5,11 @@ require "net/http"
 require "nio"
 require "ipaddr"
 
+class WithoutBacktraceError < StandardError
+  def backtrace; nil; end
+  def message; "no backtrace error"; end
+end
+
 class TestPumaServer < Minitest::Test
   parallelize_me!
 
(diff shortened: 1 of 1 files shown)


MSP-Greg:
@eregon - This was needed to fix a failing test on TruffleRuby.  Seems like a binding issue, or maybe redefining `class` as a String.  Not.sure.

eregon:
Thanks for the fix.
What was error, do you have a link to the log or the backtrace?
That would make it much easier to understand what happens

MSP-Greg:
@eregon Just for you, I loaded TruffleRuby while taking a break (just kidding, probably should have done it a long time ago).  As I suspected, it's not really a TruffleRuby issue, but a quirk of MRI Ruby (and maybe JRuby).

The error is:
`<NoMethodError: undefined method `name' for "WithoutBacktraceError":String>`

I assume that the passing builds are calling `to_s`, but TruffleRuby is calling `name`, which should exist, but doesn't in the code before this PR, which was:
```ruby
class WithoutBacktraceError < StandardError
  def backtrace; nil; end
  def message; "no backtrace error"; end
  def class; "WithoutBacktraceError"; end  # this is returning an incorrect object
end
```

Not that because of the code using the error in Puma, the only way to see this is to look at the stderr output from the server, which is available as `@log_writer.stderr`...

Issue #3000, opened 2022-10-25
Sinatra streaming broken in Puma 6.0.0


dentarg:
See https://github.com/sinatra/sinatra/issues/1832 for more details but this simple app https://github.com/sinatra/sinatra/blob/v3.0.2/examples/stream.ru works fine with Puma 5.6.5 but with 6.0.0 it waits until all `sleep`s have finished before the response is returned (this is more noticeable if you increase the sleep time in the example).

As noted in https://github.com/sinatra/sinatra/issues/1832#issuecomment-1291175026 the example runs as expected in falcon (v0.42.3).

dentarg:
If I change `IO_BUFFER_LEN_MAX` to a low value like `1` the streaming works

https://github.com/puma/puma/blob/8159aa4a7783cd1f4bdb75e918dad6ff2492bac7/lib/puma/request.rb#L24-L25

Used here 

https://github.com/puma/puma/blob/03ed6c8e78b9ba0c76055af2ddb0bdd532807d6c/lib/puma/request.rb#L303-L325

The buffering was introduced in https://github.com/puma/puma/pull/2896

MSP-Greg:
What about something like 64kB?

dentarg:
Sure, it would probably help many cases. What about making it configurable? Is that a good idea?

I suspect a more reliable solution for Sinatra is to switch to the new Streaming body  https://github.com/puma/puma/pull/2740, that wouldn’t be buffered, or? Need to try this myself

MSP-Greg:
> What about making it configurable? Is that a good idea?

Yes.  Adding a PR with `size_to_first_byte` as an added DSL option.  Not sure what the default should be.  32 kB for now.

nateberkopec:
I don't like the idea of closing a bug with a config setting. You shouldn't need to configure Puma to make chunked responses work correctly. Am I reading this wrong?

dentarg:
I don't think you are, and I agree.

Then the buffering should be opt-in? Or removed completely? Or is there another way?

dentarg:
> I suspect a more reliable solution for Sinatra is to switch to the new Streaming body #2740, that wouldn’t be buffered, or? Need to try this myself

I tried this https://github.com/dentarg/testssl.web/commit/5e6b14bb3dd83d0db5044e9d9b560533f500e664 and then streaming works in Puma 6 (this code does not work in Puma 5), not sure everyone should need to make changes like this though?

dentarg:
[...]

PR #3002, opened 2022-10-26
Correctly close app body for all code paths

### LINKED ISSUE #2999: Puma 6.0.0 does not close the response bodies of hijacked requests
**Describe the bug**

When upgrading my rails application to puma 6, I noticed that the request hang for a noticeable amount of time every time I modify one of my views and reload my app. After a lot of digging, I concluded that the issue is due to a request opening a websocket for ActionCable that opens a read lock (in a rack middleware looking for changes in the view files) but never releases it. This lock will eventually be released when ActiveSupport notices that the previous request did not properly close, and completes it.

The request opening the websocket is hijacked and therefore, the close method is never called on its body. I assume that this should not happen, based on this comment in the previous version:
https://github.com/puma/puma/blob/1b6b8adfaeb4a2bb50f43853a7736f166d323dba/lib/puma/request.rb#L190-L192

So I wonder if this issue reopens [CVE-2022-23634](https://github.com/advisories/GHSA-rmj8-8hhh-gv5h). Moreover, I wonder if there is not an issue with these lines, where the old value of `res_body` is lost without first being closed:
https://github.com/puma/puma/blob/03ed6c8e78b9ba0c76055af2ddb0bdd532807d6c/lib/puma/request.rb#L107-L116

**To Reproduce**

I defined this simple application:
```ruby
require 'rack'
available = true
run ->(env) {
  if available
    available = false
    env['rack.hijack'].call
    [200, { 'Content-Type' => 'text/plain' }, ::Rack::BodyProxy.new(['Hello World']){ available = true }]
  else
    [500, { 'Content-Type' => 'text/plain' }, ['Unavailable']]
  end
}
```
with this simple Dockerfile:
```Dockerfile
FROM ruby:3.1.2
RUN gem install puma rack
CMD ["puma", "hello.ru"]
```
That I run with `docker run --rm -it -v ``pwd``/hello.ru:/hello.ru -p 9292:9292 puma_test`.

**Expected behavior**

On the first request, a resource is consumed but should be released even if the request is hijacked. On the next request, we should get "Hello World".
```bash
curl http://127.0.0.1:9292 # hangs, this is expected
curl http://127.0.0.1:9292 # should return "Hello World" but returns "Unaivalable"
```

MSP-Greg:
### Description

1. At present, some code paths will not close the app body.  Add code to always hold a reference to the app body, so when code calling `lowlevel_error` is executed, the reference to the app body is maintained.

2. test_rack_server.rb - add test_hijack_body_close - test for above

3. Small refactor of test_puma_server.rb.  Takes too damn long.

4. test_puma_server.rb - add `test_lowlevel_error_body_close` to verify body close call when lowlevel_error

Closes #2999.

### DIFF
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -48,6 +48,7 @@ module Puma
     def handle_request(client, io_buffer, requests)
       env = client.env
       socket  = client.io   # io may be a MiniSSL::Socket
+      app_body = nil
 
       return false if closed_socket?(socket)
 
@@ -85,13 +86,17 @@ module Puma
 
       begin
         if SUPPORTED_HTTP_METHODS.include?(env[REQUEST_METHOD])
-          status, headers, res_body = @thread_pool.with_force_shutdown do
+          status, headers, app_body = @thread_pool.with_force_shutdown do
             @app.call(env)
           end
         else
           @log_writer.log "Unsupported HTTP method used: #{env[REQUEST_METHOD]}"
-          status, headers, res_body = [501, {}, ["#{env[REQUEST_METHOD]} method is not supported"]]
+          status, headers, app_body = [501, {}, ["#{env[REQUEST_METHOD]} method is not supported"]]
         end
+
+        # app_body needs to always be closed, hold value in case lowlevel_error
+        # is called
+        res_body = app_body
 
         return :async if client.hijacked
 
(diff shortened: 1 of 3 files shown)


lib/puma/request.rb
```diff
@@ -22,7 +22,7 @@ module Request # :nodoc:
     IO_BODY_MAX = 1_024 * 64
 
     # max size for io_buffer, force write when exceeded
-    IO_BUFFER_LEN_MAX = 1_024 * 1_024 * 4
```
nateberkopec, line 25: This is the change to close #3000 right? We should separate these.

nateberkopec (changes requested):

lib/puma/request.rb
MSP-Greg: Done.  I'll create another PR, setting this as a DSL option called `size_to_first_byte`, as it affects TTFB.

MSP-Greg:
Added commit '[CI] Fix two intermittent failures?', see https://github.com/puma/puma/actions/runs/3348499190

MSP-Greg:
One person stated that this PR fixed the issue re #2999, see [comment](https://github.com/puma/puma/issues/2999#issuecomment-1295436729).

Re #3004, as mentioned, testing with [hotwired/turbo-rails](https://github.com/hotwired/turbo-rails), I may write a few more tests.

This will cause merge conflicts with #3004, so I'd like to merge this.

Thoughts?

PR #3006, opened 2022-10-29
Periodically send status to systemd

### LINKED ISSUE #2604: Utilize systemd-notify to add status details

QWYNG:
### Description
Hi, Thank you for great gem.
Add test for #2833. Thank you @ekohl 
it was not able to retrieve information from `Puma.stats_hash` properly.
So I have changed the logic in `lib/puma/plugin/systemd.rb` a little.

### Description from #2833

> With systemd it's possible to send a status line. This will show up when users run systemctl status puma.service.
> 
> Most of this code is based on puma-plugin-systemd.
> 
> At the moment this is completely untested, but I had written some code that I wanted to share. If anyone wants to finish this before I do, they should feel welcome to.
> 
> Closes #2604


dentarg:
@QWYNG are you up for rebasing this PR? Sounds like potential users would be happy with it (latest replies in https://github.com/puma/puma/issues/2604)

lib/puma/plugin/systemd.rb
```diff
@@ -51,11 +51,14 @@ def start(launcher)
   end
 
   def status
-    common = "workers: #{running}/#{max_threads} threads, #{pool_capacity} available, #{backlog} backlog"
     if clustered?
-      "Puma #{Puma::Const::VERSION} cluster: #{booted_workers}/#{workers} #{common}"
+      messages = Array.new(workers) do |i|
+        common_message(stats[:worker_status][i][:last_status])
```
ekohl, line 56: I didn't look at the actual source, but would it be possible to iterate over `stats[:worker_status]` instead? Untested, but something like:
```suggestion
      messages = stats[:worker_status].values do |worker|
        common_message(worker[:last_status])
```

ekohl (approved):
Thanks @QWYNG for taking my work and taking it over the finish line.

It does look the message gets trimmed so that could be a future improvement. Still, I think this is ready to merge.

lib/puma/plugin/systemd.rb
```diff
@@ -10,7 +10,7 @@
 Puma::Plugin.create do
   def start(launcher)
     begin
-      require 'sd_notify'
+      require_relative '../sd_notify'
     rescue LoadError
       launcher.log_writer.log "Systemd integration failed. It looks like you're trying to use systemd notify but don't have sd_notify gem installed"
```
ekohl, line 15: Now that it's vendored (https://github.com/puma/puma/commit/6d8b728b42a61bcf3c1e4c698c9165a45e6071e8), I don't think there's a chance that it will fail to load. Certainly the message is wrong now since the `sd_notify` gem is no longer needed.

lib/puma/plugin/systemd.rb
QWYNG: Thank you!
stats[:worker_status] is Array instanse.
So I modified it as follows

```ruby
      messages = stats[:worker_status].map do |worker|
        common_message(worker[:last_status])
      end.join(',')
```

lib/puma/systemd.rb
```diff
@@ -29,6 +29,29 @@ def start_watchdog
       end
```
ekohl, line 29: Something went wrong in your rebase. This file is supposed to be deleted in the `Rewrite systemd integration as a plugin` commit.

ekohl (approved):
Thanks!

QWYNG:
Hi @dentarg, @nateberkopec 
 Could you please check this PR?

dentarg (approved):
Looks good to me, haven't used it myself, but I trust people from the community who seems to have tried this out!

PR #3011, opened 2022-11-04
Add built-in systemd notify support

joaomarcos96:
### Description
This PR aims to remove the need to install the [sd_notify gem](https://github.com/agis/ruby-sdnotify) by vendoring its code, thus, making the support for systemd notify built-in.

I have considered what @mperham have done with [Sidekiq](https://github.com/mperham/sidekiq/blob/ebeff59/lib/sidekiq/sd_notify.rb) and said about `sd_notify` https://github.com/puma/puma/pull/2438#issuecomment-735950804:
> With Sidekiq I decided to vendor the code for the sd_notify gem because it's only ~50 lines of stable code. I didn't want to require the user to add the gem to their gemfile manually, that shouldn't be something they need to care about IMO.

- [x] I have added (or updated) appropriate tests if this PR fixes a bug or adds a feature. **PS.:** I haven't added tests as the `sd_notify` gem code was copied exactly as it is.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,6 @@ gem "minitest", "~> 5.11"
 gem "minitest-retry"
 gem "minitest-proveit"
 gem "minitest-stub-const"
-gem "sd_notify"
 
 gem "rack", (ENV['PUMA_CI_RACK_2'] ? "~> 2.2" : ">= 2.2")
 
--- a/docs/systemd.md
+++ b/docs/systemd.md
@@ -24,8 +24,7 @@ After=network.target
 
 [Service]
 # Puma supports systemd's `Type=notify` and watchdog service
-# monitoring, if the [sd_notify](https://github.com/agis/ruby-sdnotify) gem is installed,
-# as of Puma 5.1 or later.
+# monitoring, as of Puma 5.1 or later.
 # On earlier versions of Puma or JRuby, change this to `Type=simple` and remove
 # the `WatchdogSec` line.
 Type=notify
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -318,12 +318,7 @@ module Puma
     def integrate_with_systemd
       return unless ENV["NOTIFY_SOCKET"]
 
-      begin
-        require_relative 'systemd'
-      rescue LoadError
-        log "Systemd integration failed. It looks like you're trying to use systemd notify but don't have sd_notify gem installed"
-        return
-      end
+      require_relative 'systemd'
 
       log "* Enabling systemd notification integration"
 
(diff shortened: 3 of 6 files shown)


MSP-Greg:
Re the macOS failures, it looks like line 115 in `test_integration_single.rb` needs to change to:
```ruby
assert_match(/Connection refused|Couldn't connect to server/, rejected_curl_stderr.read)
```
Or, the error message reported by curl has changed?

Let me check...

joaomarcos96:
@MSP-Greg macOS-11 2.6 test is still failing. Any idea why?

MSP-Greg:
> macOS-11 2.6 test is still failing

Nothing to do with this PR.  Intermittent test failure from hell.

I just noticed something re comments in [sd_notify.rb](https://github.com/puma/puma/pull/3011/files#diff-c19f1f12bca167462adb2b05ed0f27ab5ef76c6515673a5d9e4d1cbb8521559c).  Most doc systems will attach the comments below `require "socket"` to the `Puma` namespace.  So, they should probably be moved so they're below `module Puma` and indented.  Doing so will attach them to `Puma::SdNotify`.

Maybe CI will pass...

joaomarcos96:
@MSP-Greg thank you! I did what you suggested and now all tests are passing.
If this PR gets merged, I'd like the changes to be ported to 5.x as well. How could I do that? Should I open a PR for each 5.x-stable branch?

dentarg:
I'm not sure how clear routines we have for backports (almost never happens in my experience) but https://github.com/puma/puma/tree/5-6-stable looks to be the same as https://github.com/puma/puma/releases/tag/v5.6.5 (I draw this conclusion from the release saying `139 commits to master since this release` and the branch saying `This branch is 23 commits ahead, 139 commits behind master.` => both says 139 ✔️ 😄 – you may want to double check that)

Then there is this branch https://github.com/puma/puma/tree/5-6-5 `This branch is 1 commit ahead, 77 commits behind master.` not sure why it exist and if it serves any purpose

I think @nateberkopec can provide more guidance on backporting

MSP-Greg (approved):

nateberkopec:
>If this PR gets merged, I'd like the changes to be ported to 5.x as well. How could I do that? Should I open a PR for each 5.x-stable branch?

@joaomarcos96  Yes, that is correct. In this case though, because we're adding a feature, we should probably create a new minor version `5.7`, so if you're still interested LMK and I will create a branch.

joaomarcos96:
@nateberkopec Yes, I'd like that.

nateberkopec:
👍 We now have a `5-7-stable` branch, you may make PRs against it and we will release @joaomarcos96.

voxik:
As a Fedora maintainer, I am disappointed with this PR and with bundling in general. What was the reason to not specify the dependency in .gemspec? If at least the code was put into dedicated directory to make this more obvious. And if at least the license field was updated to list the `MIT` license alongside the `BSD-3-Clause`.

PR #3013, opened 2022-11-05
Add IOBuffer to Client, remove from ThreadPool thread instances

MSP-Greg:
### Description

Currently, an [IOBuffer](https://github.com/puma/puma/blob/f323d129bc16/lib/puma/io_buffer.rb) object is created in each thread in each ThreadPool (one per worker).  This object is used to assemble response headers, and also may contain parts or all of the response body.

It is shared among all clients/connections processed in the thread.  The data is cleared after every request.  But, given all the code paths, etc, if it was not cleared, it could leak headers between responses from different clients.

No one has reported this occurring.  Regardless, it is a relatively lightweight object subclassed from StringIO (a default gem in Ruby), so rather than share an instance with clients in a thread, remove it from the thread, and add it to the Client class.  Doing so makes it impossible for data to go to the wrong client.

Also, this allows changing/simplifying some method signatures.  Previously, there was a calculation in `Request#str_headers` that calculated whether to force a 'keep-alive' connection closed.  Moved the calculation to immediately before the call to `str_headers` in `Request#prepare_response`.  The calcuation is time sensitive, but there are no 'blocking' operations preceding the code using the calculation result.

Lastly, as improvements to Ruby and/or Puma allow more concurrency, this extra isolation may be needed in the future.

Moving IOBuffer has been discussed somewhere, don't recall when/where...

### DIFF
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -9,6 +9,7 @@ class IO
 end
 
 require_relative 'detect'
+require_relative 'io_buffer'
 require 'tempfile'
 require 'forwardable'
 
@@ -65,6 +66,7 @@ module Puma
     def initialize(io, env=nil)
       @io = io
       @to_io = io.to_io
+      @io_buffer = IOBuffer.new
       @proto_env = env
       @env = env ? env.dup : nil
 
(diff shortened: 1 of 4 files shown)


MSP-Greg:
Actions CI is having issues.  The only failed job didn't complete the `setup-ruby` step. See [failed step](https://github.com/puma/puma/actions/runs/3398336039/jobs/5651292962#step:3:20).  Today, I've seen the same in my repo forks...

dentarg (approved):
Makes sense to me

Issue #3014, opened 2022-11-06
Puma 6 hard-codes "allowed" HTTP methods


kyledrake:
Puma 6 has introduced a SUPPORTED_HTTP_METHODS constant and a check which fails other methods: https://github.com/puma/puma/blob/3d33475dd400c8d097889ddd93fab38d080c25e0/lib/puma/request.rb#L94

This will break, at a bare minimum, WebDAV implementations behind Puma, which use methods like PROPFIND and MKCOL.

I'm not sure what the "correct" approach here is, but we did have a working WebDAV implementation that is no longer working with Puma 6 and I've had to downgrade for now to figure out how to deal with this for the time being.

nateberkopec:
I think we could add a note about this in the upgrade guide/history under breaking changes. However, any support of non-standard HTTP verbs prior to Puma 6 was completely accidental.

When considering how Puma should react, [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html) lists the 8 standard methods. So we should of course at least support those (anyone want to create issues re: methods in that list we don't support, go ahead).
 
IMO HTTP methods are not a 100% Rack-level concern so it's Puma's job to make sure that we handle HTTP methods correctly. Supporting arbitrary methods is therefore not good.

dentarg:
What about allowing the user to specify methods that should also be considered supported? If the user knows their rack app can handle it.

I actually [had the thought](https://github.com/puma/puma/pull/2932#discussion_r966466501) that this was a breaking change but I failed to come up with a valid use case so I never said anything in the PR.

kyledrake:
The RFC as I interpret it doesn't say that the HTTP methods listed are required to be the only methods (only that GET and HEAD are required and the rest are optional). It says I should respond with 501 only if the origin server does not recognize or does not implement the method.

I think the ambiguity is who's in charge of making that determination, puma or the web app. Since puma is primarily used for highly customized programming, I feel like it should be the responsibility of the web application, not the HTTP server, to make this determination because in my use case, I do recognize the method and can handle it.

If this was causing a major security problem I might have a different opinion (but would probably still do the implementation upstream with something like nginx), but failing that check I think it's reasonable to move back to the original behavior and allow the web application to decide how to handle irregular methods.

MSP-Greg:
> The RFC as I interpret it doesn't say that the HTTP methods listed are required to be the only methods

[...]

PR #3015, opened 2022-11-07
Add note about HTTP method change in 6.0

nateberkopec:
Related to #3014


dentarg (approved):

dentarg:
Merged this so people can easier find the info. If it changes in some way, we can update the guide.

PR #3016, opened 2022-11-07
Refactor const.rb - freeze

MSP-Greg:
### Description

Ruby 2.3 implemented the magic comment `# frozen_string_literal: true`.  The below shows one interesting result, `join`:
```ruby
v1 = "Frozen?"
v2 = %w[HEAD GET POST PUT DELETE OPTIONS TRACE PATCH].freeze
v3 = v2.join ' '
v4 =  {501 => "HTTP/1.1 501 Not Implemented\r\n\r\n",
       503 => "HTTP/1.1 503 Service Unavailable\r\n\r\nBUSY"}

puts v1.frozen?      # -> true
puts v2[0].frozen?   # -> true
puts v3.frozen?      # -> false
puts v4[501].frozen? # -> true
```

So, remove `.freeze` when not needed in const.rb.  Also, remove extra constants (`GET POST PUT DELETE OPTIONS TRACE PATCH`) that were added in https://github.com/puma/puma/commit/1b6b8adfaeb4.

### DIFF
--- a/lib/puma/const.rb
+++ b/lib/puma/const.rb
@@ -5,7 +5,6 @@ module Puma
   class UnsupportedOption < RuntimeError
   end
 
-
   # Every standard HTTP code mapped to the appropriate message.  These are
   # used so frequently that they are placed directly in Puma for easy
   # access rather than Puma::Const itself.
@@ -100,8 +99,8 @@ module Puma
   # too taxing on performance.
   module Const
 
-    PUMA_VERSION = VERSION = "6.0.0".freeze
-    CODE_NAME = "Sunflower".freeze
+    PUMA_VERSION = VERSION = "6.0.0"
+    CODE_NAME = "Sunflower"
 
     PUMA_SERVER_STRING = ['puma', PUMA_VERSION, CODE_NAME].join(' ').freeze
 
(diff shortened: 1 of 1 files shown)


lib/puma/const.rb
```diff
@@ -112,28 +111,28 @@ module Const
     WRITE_TIMEOUT = 10
 
     # The original URI requested by the client.
-    REQUEST_URI= 'REQUEST_URI'.freeze
-    REQUEST_PATH = 'REQUEST_PATH'.freeze
-    QUERY_STRING = 'QUERY_STRING'.freeze
-    CONTENT_LENGTH = "CONTENT_LENGTH".freeze
+    REQUEST_URI= 'REQUEST_URI'
+    REQUEST_PATH = 'REQUEST_PATH'
+    QUERY_STRING = 'QUERY_STRING'
```
dentarg, line 116: Suggestion: now that we are touching these lines, let's take the opportunity to make things a tiny bit more consistent by using `"` over `'`

```suggestion
    REQUEST_URI= "REQUEST_URI"
    REQUEST_PATH = "REQUEST_PATH"
    QUERY_STRING = "QUERY_STRING"
```

lib/puma/const.rb
```diff
@@ -112,28 +111,28 @@ module Const
...
+    CONTENT_LENGTH = "CONTENT_LENGTH"
 
-    PATH_INFO = 'PATH_INFO'.freeze
+    PATH_INFO = 'PATH_INFO'
```
dentarg, line 119:
```suggestion
    PATH_INFO = "PATH_INFO"
```

dentarg (commented):
Usage of `'` comes from `HTTP_STATUS_CODES` and `PUMA_SERVER_STRING`, otherwise this file uses `"`, address that if you want or save it for later (in general I dislike doing changes like this because it makes `git blame` harder to use but if we are touching them anyways...)

MSP-Greg:
@dentarg

For you, I changed all `'` to `"` in Puma::Const.  Left `HTTP_STATUS_CODES` (in the Puma namespace) the same.

 Personally, I prefer `'`...

dentarg (approved):

MSP-Greg:
@dentarg Thaks for the review.

I've looked at it before, but (of course) now that 6.0.0 is released, I'm really looking at it.

There are a lot of constants that could be moved, as they're only used in one file.  For instance, the comment for `HTTP_STATUS_CODES` states "These are used so frequently that they are placed directly in Puma".  Today, the only file (including tests) I find them in is request.rb.  Same for some of the constant 'groups' in Puma::Const.

Any thoughts?

dentarg:
No real thoughts on that other than it sounds like a different PR :)

MSP-Greg:
> it sounds like a different PR

Of course...

MSP-Greg:
@nateberkopec

Thoughts?  Just checking, as this does remove some constants added to 6.0.0.  See updated first post.

francois-ferrandis (approved):
Seems like a good idea, I was wondering why all these freezes were still in place! :slightly_smiling_face:

Issue #3020, opened 2022-11-12
Log loaded extensions behind `PUMA_DEBUG`


dentarg:
Re the 'loaded extensions', maybe something like the below?

Single mode:
```
* Listening on http://127.0.0.1:40001
* Loaded Extensions:
*     /mnt/c/GitHub/puma/vendor/bundle/ubuntu22/ruby/3.2.0+1/gems/nio4r-2.5.8/lib/nio4r_ext.so
*     /usr/local/lib/ruby/3.2.0+1/x86_64-linux/date_core.so
*     /usr/local/lib/ruby/3.2.0+1/x86_64-linux/enc/encdb.so
*     /usr/local/lib/ruby/3.2.0+1/x86_64-linux/enc/trans/transdb.so
*     /usr/local/lib/ruby/3.2.0+1/x86_64-linux/etc.so
*     /usr/local/lib/ruby/3.2.0+1/x86_64-linux/io/wait.so
*     /usr/local/lib/ruby/3.2.0+1/x86_64-linux/monitor.so
*     /usr/local/lib/ruby/3.2.0+1/x86_64-linux/pathname.so
*     /usr/local/lib/ruby/3.2.0+1/x86_64-linux/socket.so
*     /usr/local/lib/ruby/3.2.0+1/x86_64-linux/stringio.so
Use Ctrl-C to stop
```

Cluster mode:
```
[16449] * Listening on http://127.0.0.1:40001
[16449] * Loaded Extensions:
[16449] *     /usr/local/lib/ruby/3.2.0+1/x86_64-linux/date_core.so
[16449] *     /usr/local/lib/ruby/3.2.0+1/x86_64-linux/enc/encdb.so
[16449] *     /usr/local/lib/ruby/3.2.0+1/x86_64-linux/enc/trans/transdb.so
[16449] *     /usr/local/lib/ruby/3.2.0+1/x86_64-linux/etc.so
[16449] *     /usr/local/lib/ruby/3.2.0+1/x86_64-linux/io/wait.so
[16449] *     /usr/local/lib/ruby/3.2.0+1/x86_64-linux/monitor.so
[16449] *     /usr/local/lib/ruby/3.2.0+1/x86_64-linux/pathname.so
[16449] *     /usr/local/lib/ruby/3.2.0+1/x86_64-linux/socket.so
[16449] *     /usr/local/lib/ruby/3.2.0+1/x86_64-linux/stringio.so
[16449] Use Ctrl-C to stop
```

[...]

PR #3030, opened 2022-12-10
Debug print loaded extensions if PUMA_DEBUG is set

jeremyevans:
### Description

This implements the feature requested in #3020, printing the loaded extensions if PUMA_DEBUG is set.

Not sure if this is implemented in the best place, but trying to move it to a separate method seemed challenging as the existing method appears to be called for multiple places, and I wanted to avoid redundancy.

  I didn't do this because there are no existing tests for printing the listening socket.
  I didn't do this as there is no existing documentation for PUMA_DEBUG, beyond a comment in the systemd doc.
  CI should be able to check this. I didn't run the tests nor Rubocop locally, though all CI jobs that ran in my fork passed.


MSP-Greg:
@jeremyevans

Thank you for the PR.

> Not sure if this is implemented in the best place

That's the messy part.  Placing the code in `Binder#parse` has some issues.

If one starts Puma with a control server using a socket connection, the control server calls `Binder#parse` for its bind address, so the loaded extensions will be logged twice.

If one is using a 'single' setup (with no workers), one can log from the main process, but if one is using workers (a clustered setup), it would probably be best to log from Worker index 0 (Worker0).  As an example, when running clustered, `nio` is loaded in the Worker0 process, it is not loaded in the main process.

I've got a patch for that, not sure about how often/when to log the Worker0 extensions, but it does log correctly when the server starts.

Also, the code below is using file extensions for system libraries for macOS & Windows, which are different from the file extensions used by Ruby extensions (`.bundle` & `.so`, respectively).
```ruby
`$LOADED_FEATURES.grep(/\.(so|dylib|dll)\z/i).each { |f| log_writer.debug("    #{f}") }`
```

So, in the patch I've got I used:
```ruby
re_ext = /\.#{RbConfig::CONFIG['DLEXT']}\z/i
$LOADED_FEATURES.grep(re_ext).each { |f| @log_writer.debug("  #{f}") }
```

Lastly, I have yet to try any of the code with non-MRI Rubies, not sure what they'll do.  Additionally, two items, first, when and how to log Worker0 on restarts, and second, I can't recall if $LOADED_FEATURES resolves symlinks or not.  People may prefer to see resolved (actual) paths.

I can push the patch if you'd like...  Again, thanks.

jeremyevans:
@MSP-Greg Please do push your patch, it's definitely better than what I submitted. `$LOADED_FEATURES` will generally store real paths (which resolves symlinks).

MSP-Greg:
@jeremyevans

Pushed the commit, the CI error is one of those intermittent issues that we have.  We normally squash PR commits, I always have 'merge' commits when rebasing PR branches.  Must be doing something wrong...

MSP-Greg:
@jeremyevans I made a mess of the branch.  Can't seem to force push to the branch.

I think the last commit [Merge branch 'debug-print-loaded-extensions-3020...](https://github.com/puma/puma/commit/b2762c47448b088b6aa2a2e335c76b8829267301) and the first set of commits ([Debug print loaded extensions if PUMA_DEBUG is set](https://github.com/puma/puma/commit/b475913a65c1fea9b25e7adc1b866977c4021923) and [Fix regexp used for determining loading extensions](https://github.com/puma/puma/commit/62c552ae3715b2cf5bfd9fd7c9120470d4884dae))  need to be deleted?  I hate it when this happens.  Sorry.

jeremyevans:
No big deal.  I think the best approach here is to delete this PR, and for you to push a PR with your changes.  Does that make sense?

PR #3035, opened 2022-12-14
Remove use of etc and time gems in Puma

MSP-Greg:
### Description

Several years ago the `etc` and `time` gems were added to Puma's requirements.  They are no longer needed, as a small code change replaces `time`, and I can't find any reference to `etc`.

Closes #3033.

### Your checklist for this pull request
- [x] I have reviewed the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.
- [x] I have added (or updated) appropriate tests if this PR fixes a bug or adds a feature.
- [x] My pull request is 100 lines added/removed or less so that it can be easily reviewed.
- [ ] If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.
- [x] If this closes any issues, I have added "Closes `#issue`" to the PR description or my commit messages.
- [ ] I have updated the documentation accordingly.
- [x] All new and existing tests passed, including Rubocop.

PR #3036, opened 2022-12-14
Log loaded extensions when `PUMA_DEBUG` is set

### LINKED ISSUE #3020: Log loaded extensions behind `PUMA_DEBUG`

MSP-Greg:
### Description

This implements the feature requested in https://github.com/puma/puma/issues/3020, printing the loaded extensions if PUMA_DEBUG is set.

When Puma is running 'single' (no workers), the log is done in the main Puma process, after `Server#run` has been called, as some extensions maybe loaded in that call. For Puma running 'clustered' (workers used), the log is generated in Worker index 0, after `Server#run` is called and also in the master process.

Log is also generated on phased-restarts.

Added a test for both, small refactor (moved functionality to `intergration.rb`) for using a control server in CI.

Closes #3020

### DIFF
--- a/lib/puma/cluster/worker.rb
+++ b/lib/puma/cluster/worker.rb
@@ -117,7 +117,7 @@ module Puma
           server_thread = server.run
 
           if @log_writer.debug? && index == 0
-            debug_loaded_extensions "────────────────────────────────── Loaded Extensions - worker 0:"
+            debug_loaded_extensions "Loaded Extensions - worker 0:"
           end
 
           stat_thread ||= Thread.new(@worker_write) do |io|
--- a/lib/puma/cluster/worker.rb
+++ b/lib/puma/cluster/worker.rb
@@ -115,6 +115,11 @@ module Puma
 
         while restart_server.pop
           server_thread = server.run
+
+          if @log_writer.debug? && index == 0
+            debug_loaded_extensions "────────────────────────────────── Loaded Extensions - worker 0:"
+          end
+
           stat_thread ||= Thread.new(@worker_write) do |io|
             Puma.set_thread_name "stat pld"
             base_payload = "p#{Process.pid}"
--- a/lib/puma/runner.rb
+++ b/lib/puma/runner.rb
@@ -201,7 +201,7 @@ module Puma
 
     # this method call should always be guarded by `@log_writer.debug?`
     def debug_loaded_extensions(str)
-      @log_writer.debug str
+      @log_writer.debug "────────────────────────────────── #{str}"
       re_ext = /\.#{RbConfig::CONFIG['DLEXT']}\z/i
       $LOADED_FEATURES.grep(re_ext).each { |f| @log_writer.debug("    #{f}") }
     end
--- a/lib/puma/runner.rb
+++ b/lib/puma/runner.rb
@@ -198,5 +198,12 @@ module Puma
         }
       }
     end
+
+    # this method call should always be guarded by `@log_writer.debug?`
+    def debug_loaded_extensions(str)
+      @log_writer.debug str
+      re_ext = /\.#{RbConfig::CONFIG['DLEXT']}\z/i
+      $LOADED_FEATURES.grep(re_ext).each { |f| @log_writer.debug("    #{f}") }
+    end
   end
 end
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -467,6 +467,7 @@ module Puma
                   @events.fire(:ping!, w)
                   if !booted && @workers.none? {|worker| worker.last_status.empty?}
                     @events.fire_on_booted!
+                    debug_loaded_extensions("────────────────────────────────── Loaded Extensions - master:") if @log_writer.debug?
                     booted = true
                   end
                 end
@@ -476,6 +477,7 @@ module Puma
             end
             if in_phased_restart && workers_not_booted.zero?
               @events.fire_on_booted!
+              debug_loaded_extensions("────────────────────────────────── Loaded Extensions - master:") if @log_writer.debug?
               in_phased_restart = false
             end
 
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -467,7 +467,7 @@ module Puma
                   @events.fire(:ping!, w)
                   if !booted && @workers.none? {|worker| worker.last_status.empty?}
                     @events.fire_on_booted!
-                    debug_loaded_extensions("────────────────────────────────── Loaded Extensions - master:") if @log_writer.debug?
+                    debug_loaded_extensions("Loaded Extensions - master:") if @log_writer.debug?
                     booted = true
                   end
                 end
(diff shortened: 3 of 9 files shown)


dentarg:
Re: tests, I see there's [`wait_for_server_to_include`](https://github.com/puma/puma/blob/7b6301009321d527891e4761da0d39a025b9015f/test/helpers/integration.rb#L129), maybe that could be used? Just to look for the `Loaded Extensions` string at least (when starting with `PUMA_DEBUG` set)

MSP-Greg:
Added tests.  After #3035, I may try to add a count of the extensions to the two tests, so we'll know if there are additions.

Question - while running locally, I wondered if Puma in 'cluster' mode should log **both** master process and worker0 loaded extensions?  Currently it only logs worker0.

dentarg:
> Added tests. After #3035, I may try to add a count of the extensions to the two tests, so we'll know if there are additions.

👍 
 
> Question - while running locally, I wondered if Puma in 'cluster' mode should log **both** master process and worker0 loaded extensions? Currently it only logs worker0.

I can only see it as a positive to log both

lib/puma/cluster/worker.rb
```diff
@@ -115,6 +115,11 @@ def run
 
         while restart_server.pop
           server_thread = server.run
+
+          if @log_writer.debug? && index == 0
+            debug_loaded_extensions "────────────────────────────────── Loaded Extensions - worker 0:"
+          end
```
dentarg, line 121:
```suggestion
          debug_loaded_extensions("────────────────────────────────── Loaded Extensions - worker 0:") if index == 0
```

dentarg (commented):
`debug_loaded_extensions` includes `if @log_writer.debug?` so no need to guard the method calls with it, right?

MSP-Greg:
@dentarg

> `debug_loaded_extensions` includes `if @log_writer.debug?` so no need to guard the method calls with it, right?

Not sure.  I felt it was unlikely that many people would regularly be using `PUMA_DEBUG`, so guard the method call so it's never performed.  The guard in the method could be removed, but once it's been called, `LogWriter#debug` has the conditional anyway, so adding one more seemed insignificant?

dentarg:
If we want to remove `if @log_writer.debug?` from the `debug_loaded_extensions` definition, we should at least be consistent and guard the call in `puma/lib/puma/single.rb` too, no?

https://github.com/puma/puma/blob/01b013a00da386fd8db66f6ad42460e7affc9cf1/lib/puma/single.rb#L60

MSP-Greg:
> we should at least be consistent and guard the call in puma/lib/puma/single.rb too, no?

Sorry and thanks for pointing that out, my mistake.  Fixed in single.rb, removed conditional in runner.rb, with a comment...

lib/puma/runner.rb
```diff
@@ -194,5 +194,12 @@ def stats
         }
       }
     end
+
+    # this method call should always be guarded by `@log_writer.debug?`
+    def debug_loaded_extensions(str)
```
dentarg, line 199: Why not have the guard in this method?

Also, I think "Puma running in debug mode" (`PUMA_DEBUG` set) is not something only `LogWriter` should know about. In the future, more things could be happening in "debug more" (e.g. collecting more details stats) in my mind.

What do you think of adding a constant `PUMA_DEBUG` to `lib/puma.rb`? And a method `Puma.debug?` (or `Puma.debug_mode?`) and use that here. That method can also be used in both `LogWriter` and `ErrorLogger` (both do `ENV.key?('PUMA_DEBUG')` today)

lib/puma/cluster/worker.rb
```diff
@@ -115,6 +115,11 @@ def run
 
         while restart_server.pop
           server_thread = server.run
+
+          if @log_writer.debug? && index == 0
+            debug_loaded_extensions "────────────────────────────────── Loaded Extensions - worker 0:"
```
nateberkopec, line 120: I'm not a huge fan of having this long formatted string right in the code

Maybe:

```ruby
debug_loaded_extensions("worker 0")
```

...and let `debug_loaded_extensions` add the `---` padding and such.

lib/puma/runner.rb
nateberkopec: I'm fine with either option, up to Greg.

nateberkopec (changes requested):

lib/puma/cluster/worker.rb
MSP-Greg: See [Fix !DRY code](https://github.com/puma/puma/pull/3036/commits/8d2da80a64587dbcf6f3d78a9ee309b974bd16e5)

Not my best code...  Sory for the double push, first was the commit, second was "maybe it's time to rebase" force push

Happy New Year!

nateberkopec (approved):

nateberkopec:
Wait on merge until I push 6.0.2 though 😄

nateberkopec:
@MSP-Greg Merge when ready 👍

dentarg (approved):

PR #3039, opened 2022-12-17
[CI] test_integration_cluster.rb - conditionally require  'time' gem

MSP-Greg:
### Description

Ruby 3.2's native `time` class/object can now parse a properly formatted text string, including ISO-8601.  With the change, tests no longer need to load the `time` gem for parsing 'stats'.

### DIFF
--- a/test/test_integration_cluster.rb
+++ b/test/test_integration_cluster.rb
@@ -1,10 +1,21 @@
 require_relative "helper"
 require_relative "helpers/integration"
 
-require "time"
-
 class TestIntegrationCluster < TestIntegration
   parallelize_me! if ::Puma.mri?
+
+  # Ruby 3.2 can parse iso8601 strings, others can't or throw error
+  begin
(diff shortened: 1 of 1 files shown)


dentarg:
Is this needed to catch potential bugs? E.g. related to https://github.com/puma/puma/pull/3035

Otherwise, I think this makes the tests more complex than they need to be

MSP-Greg:
> Is this needed to catch potential bugs? E.g. related to https://github.com/puma/puma/pull/3035

Probably not.

> I think this makes the tests more complex than they need to be

Well, 3.2 is adding an overload to `Time.new`.  I can't think of any way to simplify the code.  I'd prefer to always make use of new features in Ruby, and also have as few dependencies as possible in both Puma's 'lib' code and its test suite.

nateberkopec:
Yeah this one's a bit overboard for me, I'd rather just require `time` than write 15 lines of code we have to maintain.

PR #3040, opened 2022-12-20
Introduce the ability to return 413: payload too large for requests

shayonj:
### Description

When receiving large payload objects, the server can often slowdown or get fully exhausted if bunch of requests with large payload body size come in. When request with large payload comes in, lot of the time is spent reading the body into buffer, writing it to the IO for rack, before the request is passed to the rails app for further processing.

While there are some workarounds around limiting large request sizes, like at nginx layer by setting `client_max_body_size`, which would return a `413` back to the client, today that is not possible with puma.

This would be a very nice feature to have, especially when there is no reverse proxy in between client and server.

This approach - allows a user to set `http_content_length_limit` via  a config variable (defaults to `nil`). This value is then compared against `Content-Length` http header (when present) before reading the body into buffer. If the user value is higher than the header value, the request body is not loaded and an immediate `413` (`Payload too large`) http response is returned, from `Puma::Request.handle_request`.

Without having to buffer in the huge request and return the `413` immediately to the clients that send a `Content-Length` - could be a nice feature and helpful protection to have.

Additionally, when no `Content-Length` http header is present, it is compared against the size of the body of the request.
During this time, the body is already loaded, so there isn't much we can do, but it still respects the limit and
returns the `413` status from puma itself, before passing down the request to the app. 

### DIFF
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -156,6 +156,7 @@ module Puma
       @peerip = nil if @remote_addr_header
       @in_last_chunk = false
       @http_content_length_size = 0
+      @http_content_length_limit_exceeded = false
 
       if @buffer
         return false unless try_to_parse_proxy_protocol
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -214,8 +214,8 @@ module Puma
     end
 
     def try_to_finish
-      if env[CONTENT_LENGTH] && @http_content_length_limit
-        @http_content_length_limit_exceeded = env[CONTENT_LENGTH].to_i > @http_content_length_limit
+      if env[CONTENT_LENGTH] && above_http_content_limit(env[CONTENT_LENGTH].to_i)
+        @http_content_length_limit_exceeded = true
       end
 
       if @http_content_length_limit_exceeded
@@ -254,7 +254,7 @@ module Puma
 
       @parsed_bytes = @parser.execute(@env, @buffer, @parsed_bytes)
 
-      if @parser.finished? && @http_content_length_limit && (@parser.body.bytesize > @http_content_length_limit)
+      if @parser.finished? && above_http_content_limit(@parser.body.bytesize)
         @http_content_length_limit_exceeded = true
       end
 
@@ -612,6 +612,10 @@ module Puma
       end
       @requests_served += 1
       @ready = true
+    end
+
+    def above_http_content_limit(value)
+      @http_content_length_limit&.< value
     end
   end
 end
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -88,7 +88,6 @@ module Puma
 
       @http_content_length_limit = nil
       @http_content_length_limit_exceeded = false
-      @http_content_length_size = 0
 
       @peerip = nil
       @peer_family = nil
@@ -155,7 +154,6 @@ module Puma
       @body_remain = 0
       @peerip = nil if @remote_addr_header
       @in_last_chunk = false
-      @http_content_length_size = 0
       @http_content_length_limit_exceeded = false
 
       if @buffer
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -88,6 +88,7 @@ module Puma
 
       @http_content_length_limit = nil
       @http_content_length_limit_exceeded = false
+      @http_content_length_size = 0
 
       @peerip = nil
       @peer_family = nil
@@ -154,6 +155,7 @@ module Puma
       @body_remain = 0
       @peerip = nil if @remote_addr_header
       @in_last_chunk = false
+      @http_content_length_size = 0
 
       if @buffer
         return false unless try_to_parse_proxy_protocol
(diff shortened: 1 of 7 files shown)


shayonj:
Trying it out locally

```ruby
# config.rb

http_content_length_limit 100
```

```bash
 bundle exec bin/puma -C config.rb test/rackup/hello.ru
```

Sending a `curl` with standard body body above 100 bytes
```bash

curl -kso /dev/null -XPOST http://localhost:9292 -d "Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum" -w "==============\n\n        
| dnslookup: %{time_namelookup}\n    
| connect: %{time_connect}\n             
| appconnect: %{time_appconnect}\n 
| pretransfer: %{time_pretransfer}\n 
| starttransfer: %{time_starttransfer}\n 
| total: %{time_total}\n
| size: %{size_download}\n
| HTTPCode=%{http_code}\n\n"
==============

 
| dnslookup: 0.003017
 
| connect: 0.003273
 
| appconnect: 0.000000
 
| pretransfer: 0.003293
 
| starttransfer: 0.003995
 
| total: 0.004048
 
| size: 17
 
| HTTPCode=413
```

With a large json (40mb)
```bash
curl -kso /dev/null -XPOST http://localhost:9292/1 -d @40mb.json -w "==============\n\n      
| dnslookup: %{time_namelookup}\n
| connect: %{time_connect}\n
| appconnect: %{time_appconnect}\n
| pretransfer: %{time_pretransfer}\n
| starttransfer: %{time_starttransfer}\n
| total: %{time_total}\n
| size: %{size_download}\n
| HTTPCode=%{http_code}\n\n"
==============

 
| dnslookup: 0.002976
 
| connect: 0.003273
 
| appconnect: 0.000000
 
| pretransfer: 0.003293
 
| starttransfer: 0.003568
 
| total: 0.004242
 
| size: 17
 
| HTTPCode=413
```

lib/puma/dsl.rb
```diff
@@ -1022,6 +1022,19 @@ def mutate_stdout_and_stderr_to_sync_on_write(enabled=true)
       @options[:mutate_stdout_and_stderr_to_sync_on_write] = enabled
     end
 
+    # Specify how big the request payload should be.
```
nateberkopec, line 1025:
```suggestion
    # Specify how big the request payload should be, in bytes.
```

test/test_puma_server.rb
```diff
@@ -383,6 +383,20 @@ def test_early_hints_is_off_by_default
     assert_equal expected_data, data
   end
 
+  def test_request_payload_too_large
+    server_run(http_content_length_limit: 10)
+
+    sock = send_http "POST / HTTP/1.1\r\nHost: test.com\r\nContent-Type: text/plain\r\nContent-Length: 19\r\n\r\n"
+    sleep 0.5
```
nateberkopec, line 390: Are these sleeps necessary? 🤔

nateberkopec:
Great idea. I want to look at the code in a little more depth later, but I think this is a great feature for inclusion.

lib/puma/client.rb
```diff
@@ -210,6 +215,17 @@ def try_to_parse_proxy_protocol
     end
 
     def try_to_finish
```
nateberkopec, line 217: This is a very important path so we'll need to be careful with anything we add here.

test/test_puma_server.rb
shayonj: actually, not in this case, i saw some other tests were doing it for replicating a timeout, but its not necessary here. Will update.

shayonj:
Sounds good, appreciate the review. Yeah, definitely feels like something very useful. Especially when keeping puma workers snappy and fast by rejecting large requests. 

I have been using it locally with our @tines monolith and works well (but ofc its local only). I will try to see if I can roll this out on some test stacks to test internally as well.

lib/puma/client.rb
```diff
@@ -613,5 +613,11 @@ def set_ready
       @requests_served += 1
       @ready = true
     end
+
+    def above_http_content_limit(value)
+      return false if @http_content_length_limit.nil?
+
+      value > @http_content_length_limit
+    end
```
MSP-Greg, line 621: Maybe the following?   Looks a bit odd, but...
```ruby
  def above_http_content_limit(value)
    @http_content_length_limit&.< value
  end
```

shayonj: sure, updated! 🙌

MSP-Greg: Thanks.  I'm busy with some cluster-f*(& code due to XMas Ruby releases.  Thanks for adding code for chunked.  I can't review till later, maybe someone else can...

shayonj: haha! no rush, thanks for the suggestions and supporting xmas ruby releases. Have a fun xmas break! 🎄

nateberkopec:
Thank you for your change @shayonj 🙇

shayonj:
Thanks @nateberkopec! Any ideas, when we can expect a release with this change? 🙌

nateberkopec:
Probably in about 4 weeks? When I'm not busy, I try to release on a 4-week cadence. We just did a patch release so your commit was the first commit since that.

PR #3041, opened 2022-12-21
Add `supported_http_methods` option

### LINKED ISSUE #3014: Puma 6 hard-codes "allowed" HTTP methods

francois-ferrandis:
### Description

Closes #3014

This pr adds an option called `supported_http_methods`. This option allows puma users to specify which HTTP methods

### DIFF
--- a/6.0-Upgrade.md
+++ b/6.0-Upgrade.md
@@ -45,7 +45,16 @@ Check the following list to see if you're depending on any of these behaviors:
 1. We've removed the following constants: `Puma::StateFile::FIELDS`, `Puma::CLI::KEYS_NOT_TO_PERSIST_IN_STATE` and `Puma::Launcher::KEYS_NOT_TO_PERSIST_IN_STATE`, and `Puma::ControlCLI::COMMANDS`.
 1. We no longer support Ruby 2.2, 2.3, or JRuby on Java 1.7 or below.
 1. The behavior of `remote_addr` has changed. When using the set_remote_address header: "header_name" functionality, if the header is not passed, REMOTE_ADDR is now set to the physical peeraddr instead of always being set to 127.0.0.1. When an error occurs preventing the physical peeraddr from being fetched, REMOTE_ADDR is now set to the unspecified source address ('0.0.0.0') instead of to '127.0.0.1'
-1. Previously, Puma supported anything as an HTTP method and passed it to the app. We now only accept the 8 HTTP methods defined in RFC 9110.
+1. Previously, Puma supported anything as an HTTP method and passed it to the app. We now only accept the 8 HTTP methods defined in RFC 9110. If you want to override which HTTP methods are accepted, you may use the `supported_http_methods` option:
+    ```ruby
+      supported_http_methods %w[
+        ACL BASELINE-CONTROL BIND CHECKIN CHECKOUT CONNECT COPY DELETE
+        GET HEAD LABEL LINK LOCK MERGE MKACTIVITY MKCALENDAR MKCOL
+        MKREDIRECTREF MKWORKSPACE MOVE OPTIONS ORDERPATCH PATCH POST
+        PRI PROPFIND PROPPATCH PUT REBIND REPORT SEARCH TRACE UNBIND
+        UNCHECKOUT UNLINK UNLOCK UPDATE UPDATEREDIRECTREF VERSION-CONTROL
+      ].to_set.freeze
+    ```
 
 Then, update your Gemfile:
 
--- a/lib/puma/configuration.rb
+++ b/lib/puma/configuration.rb
@@ -133,6 +133,7 @@ module Puma
       debug: false,
       early_hints: nil,
       environment: 'development'.freeze,
+      supported_http_methods: Puma::Const::DEFAULT_SUPPORTED_HTTP_METHODS,
       # Number of seconds to wait until we get the first data for the request
       first_data_timeout: 30,
       io_selector_backend: :auto,
--- a/lib/puma/const.rb
+++ b/lib/puma/const.rb
@@ -1,6 +1,8 @@
 #encoding: utf-8
 # frozen_string_literal: true
 
+require "set"
+
 module Puma
   class UnsupportedOption < RuntimeError
   end
@@ -147,15 +149,7 @@ module Puma
     MAX_BODY = MAX_HEADER
 
     REQUEST_METHOD = "REQUEST_METHOD".freeze
-    HEAD = "HEAD".freeze
-    GET = "GET".freeze
-    POST = "POST".freeze
-    PUT = "PUT".freeze
-    DELETE = "DELETE".freeze
-    OPTIONS = "OPTIONS".freeze
-    TRACE = "TRACE".freeze
-    PATCH = "PATCH".freeze
-    SUPPORTED_HTTP_METHODS = [HEAD, GET, POST, PUT, DELETE, OPTIONS, TRACE, PATCH].freeze
+    DEFAULT_SUPPORTED_HTTP_METHODS = %w[HEAD GET POST PUT DELETE OPTIONS TRACE PATCH].to_set.freeze
     # ETag is based on the apache standard of hex mtime-size-inode (inode is 0 on win32)
     LINE_END = "\r\n".freeze
     REMOTE_ADDR = "REMOTE_ADDR".freeze
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -88,7 +88,7 @@ module Puma
       env[RACK_AFTER_REPLY] ||= []
 
       begin
-        if SUPPORTED_HTTP_METHODS.include?(env[REQUEST_METHOD])
+        if @supported_http_methods.include?(env[REQUEST_METHOD])
           status, headers, app_body = @thread_pool.with_force_shutdown do
             @app.call(env)
           end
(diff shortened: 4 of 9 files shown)


lib/puma/const.rb
```diff
@@ -147,15 +149,7 @@ module Const
...
-    HEAD = "HEAD".freeze
-    GET = "GET".freeze
-    POST = "POST".freeze
-    PUT = "PUT".freeze
-    DELETE = "DELETE".freeze
-    OPTIONS = "OPTIONS".freeze
-    TRACE = "TRACE".freeze
-    PATCH = "PATCH".freeze
-    SUPPORTED_HTTP_METHODS = [HEAD, GET, POST, PUT, DELETE, OPTIONS, TRACE, PATCH].freeze
+    DEFAULT_SUPPORTED_HTTP_METHODS = %w[HEAD GET POST PUT DELETE OPTIONS TRACE PATCH].to_set.freeze
```
francois-ferrandis, line 152: This files specifies

    # frozen_string_literal: true

so why freeze strings ?

dentarg: See https://github.com/puma/puma/commit/88e51fb08e0735a98a519db46649f01bcc88d03c, the magic comment was added when Puma still supported Ruby versions not supporting the magic comment

6.0-Upgrade.md
```diff
@@ -45,7 +45,16 @@ Check the following list to see if you're depending on any of these behaviors:
...
-1. Previously, Puma supported anything as an HTTP method and passed it to the app. We now only accept the 8 HTTP methods defined in RFC 9110.
+1. Previously, Puma supported anything as an HTTP method and passed it to the app. We now only accept the 8 HTTP methods defined in RFC 9110. If you want to override which HTTP methods are accepted, you may use the `supported_http_methods` option:
+    ```ruby
+      supported_http_methods %w[
+        ACL BASELINE-CONTROL BIND CHECKIN CHECKOUT CONNECT COPY DELETE
+        GET HEAD LABEL LINK LOCK MERGE MKACTIVITY MKCALENDAR MKCOL
+        MKREDIRECTREF MKWORKSPACE MOVE OPTIONS ORDERPATCH PATCH POST
+        PRI PROPFIND PROPPATCH PUT REBIND REPORT SEARCH TRACE UNBIND
+        UNCHECKOUT UNLINK UNLOCK UPDATE UPDATEREDIRECTREF VERSION-CONTROL
+      ].to_set.freeze
```
dentarg, line 56: Why do the user need to call `#to_set` and `#freeze`? I think it is implementation details that the user shouldn't need to care about

lib/puma/const.rb
```diff
@@ -1,6 +1,8 @@
 #encoding: utf-8
 # frozen_string_literal: true
 
+require "set"
```
dentarg, line 4: Do we need to use `Set`? We should be careful what dependencies we introduce.

[Ruby 3.0](https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/) made `set` a default gem (and there was also a recent change, can't remember/find exactly when, where you no longer need to require set). Anyway, I can't say for sure if this will be a problem, but similar cases [have been](https://github.com/puma/puma/discussions/3034).

lib/puma/request.rb
```diff
@@ -547,7 +547,7 @@ def str_headers(env, status, headers, res_body, io_buffer, force_keep_alive)
       colon = COLON
 
       resp_info = {}
-      resp_info[:no_body] = env[REQUEST_METHOD] == HEAD
+      resp_info[:no_body] = env[REQUEST_METHOD] == "HEAD"
```
dentarg, line 550: Probably still want to compare using a constant here?

test/test_supported_http_methods.rb
```diff
@@ -0,0 +1,70 @@
...
+  attr_reader :ran_test
+
+  def call(env)
+    @ran_test = true
+
+    [200, {"Content-Type" => "text/plain"}, ["hello!"]]
+  end
+end
+
+class WebServerTest < Minitest::Test
```
dentarg, line 17: I think you should look over the existing tests, looks like you have invented the wheel again for a few test helpers things in this file

dentarg (commented):
Some quick comments

lib/puma/const.rb
francois-ferrandis: As far as I can see, `Set` has been available since [at least Ruby 2.3](https://docs.ruby-lang.org/en/2.3.0/Set.html), so the Ruby versions supported by puma 7 will have it.

It seems to me that a set is the appropriate data structure here. We could use a `Hash`, but it will have an uglier interface, or we could use an `Array`, but it will have worse performance (about 4-5 times slower). However, that performance difference might be negligible compared to the rest of the code.

I'll let you decide, just tell me! :slightly_smiling_face:

lib/puma/request.rb
francois-ferrandis: Sure, I'll bring it back. :wink:

test/test_supported_http_methods.rb
francois-ferrandis: I might need your guidance here. 

I wanted to have a test case where I could pass the `supported_http_methods` option when instantiating the server, but in the other test files, the instantiation was done in the `setup` method, which is implicitly called and taks no parameter.

This is why I introduced the `run_server` method, which takes an `options:` parameter.

Is there a more elegant way to test the behavior for different values of the `supported_http_methods` option?

Thank you very much.

lib/puma/const.rb
francois-ferrandis: That's what I thought, thanks for pointing to the historical context. :+1:

lib/puma/const.rb
francois-ferrandis: What about making the option a proc? That way the implementation is not our responsibility, and the users are free to use any data structure they want. We can still recommend using a set in the example, since most users will have that available to them.

Thoughts? :upside_down_face:

MSP-Greg:
Re the issue of set, I think it uses hash internally.

Maybe assume an array for the DSL parameter, use `ary.zip([nil]).to_h.freeze`, then test with `hsh.key? <req method>`...

test/test_supported_http_methods.rb
francois-ferrandis: OK, I spent a bit more time and simplified the helpers quite a bit here!

Let me know if there is yet a more simple way to do things. :pray:

lib/puma/const.rb
```diff
@@ -147,7 +147,7 @@ module Const
 
     REQUEST_METHOD = "REQUEST_METHOD"
     HEAD = "HEAD"
-    SUPPORTED_HTTP_METHODS = %w[HEAD GET POST PUT DELETE OPTIONS TRACE PATCH].freeze
+    DEFAULT_SUPPORTED_HTTP_METHODS = %w[HEAD GET POST PUT DELETE OPTIONS TRACE PATCH].zip([nil]).to_h.freeze
```
dentarg, line 150: I think we should add a comment explaining why we do `.zip([nil]).to_h` here, it is not obvious 

I'll include what this is evaluated to for reference 

```ruby
irb(main):001:0> %w[HEAD GET POST PUT DELETE OPTIONS TRACE PATCH].zip([nil]).to_h
=> {"HEAD"=>nil, "GET"=>nil, "POST"=>nil, "PUT"=>nil, "DELETE"=>nil, "OPTIONS"=>nil, "TRACE"=>nil, "PATCH"=>nil}
```

6.0-Upgrade.md
```diff
@@ -45,7 +45,16 @@ Check the following list to see if you're depending on any of these behaviors:
...
-1. Previously, Puma supported anything as an HTTP method and passed it to the app. We now only accept the 8 HTTP methods defined in RFC 9110.
+1. Previously, Puma supported anything as an HTTP method and passed it to the app. We now only accept the 8 HTTP methods defined in RFC 9110. If you want to override which HTTP methods are accepted, you may use the `supported_http_methods` option:
+    ```ruby
+      supported_http_methods %w[
+        ACL BASELINE-CONTROL BIND CHECKIN CHECKOUT CONNECT COPY DELETE
+        GET HEAD LABEL LINK LOCK MERGE MKACTIVITY MKCALENDAR MKCOL
+        MKREDIRECTREF MKWORKSPACE MOVE OPTIONS ORDERPATCH PATCH POST
+        PRI PROPFIND PROPPATCH PUT REBIND REPORT SEARCH TRACE UNBIND
+        UNCHECKOUT UNLINK UNLOCK UPDATE UPDATEREDIRECTREF VERSION-CONTROL
+      ]
```
dentarg, line 56: So with this new option you can make Puma not to respond to the most common HTTP methods, may be a good feature for someone, but also a bit of a sharp knife if you forget to include a method. Docs could state something along the lines "make sure you include all the HTTP methods your app support here if you use this option"

test/test_supported_http_methods.rb
dentarg: Looks pretty clean to me now but I suspect it could be refactored to share even more with existing helpers, I haven't looked in detail though, so let' save that for some future contributor (if even possible) :)

dentarg (commented):
Looks simple enough to me! A few minor comments

lib/puma/const.rb
dentarg: Also, should we do this? The previous implementation was fine with a regular Array? Do we know if it matters?

dentarg:
Also needs a rebase

lib/puma/configuration.rb
```diff
@@ -133,6 +133,7 @@ class Configuration
       debug: false,
       early_hints: nil,
       environment: 'development'.freeze,
+      supported_http_methods: Puma::Const::DEFAULT_SUPPORTED_HTTP_METHODS,
```
dentarg, line 136: We're inside the `Puma` module so we can do this

```suggestion
      supported_http_methods: Const::DEFAULT_SUPPORTED_HTTP_METHODS,
```

PR #3042, opened 2022-12-21
Correct date for Puma 6.0.1 in History

dentarg:
[ci skip]


nateberkopec:
Thanks for the cleanups @dentarg 🙇

Issue #3046, opened 2022-12-25
CI: Rubocop lint job was incorrectly skipped


dentarg:
Skipped for https://github.com/puma/puma/commit/423d17c4af5ae55973b6a201c7f40e8d03d2d5b3 at https://github.com/puma/puma/actions/runs/3771511344/jobs/6411944506

> Run fkirc/skip-duplicate-actions@v5.2.0
>   with:
>     paths_ignore: ["**.md"]
>     paths: []
>     concurrent_skipping: same_content_newer
>     skip_after_successful_duplicate: true
>     github_token: ***
>     cancel_others: false
>     do_not_skip: ["workflow_dispatch", "schedule"]
> Skip execution because the exact same files are concurrently checked in older run https://github.com/puma/puma/actions/runs/3771511146

It can be seen at the next build https://github.com/puma/puma/actions/runs/3771551306/jobs/6412004298#step:4:13 that 423d17c4af5ae55973b6a201c7f40e8d03d2d5b3 introduced a offense

> Offenses:
> 
> lib/puma/const.rb:104:1: C: [Correctable] Layout/TrailingWhitespace: Trailing whitespace detected.
> 
> 172 files inspected, 1 offense detected, 1 offense auto-correctable

dentarg:
@MemunaHaruna any idea what happened here?

MemunaHaruna:
Hey @dentarg looking at https://github.com/puma/puma/actions/runs/3771511344/jobs/6411944506, it seems this is the reason:

```
Skip execution because the exact same files are concurrently checked in older run
https://github.com/puma/puma/actions/runs/3771511146
```

dentarg:
Yes but why? We wanted the job to run in this case, it missed an offense

MemunaHaruna:
It looks like the offence was introduced in this [commit](https://github.com/puma/puma/pull/3016/commits/d037b36d304a4f47b6968618990e5767f3957194) when the branch was updated with master. The [lint job ran and failed](https://github.com/puma/puma/actions/runs/3771511146) on that commit. 
When the PR was merged, the CI was skipped because the merge commit had the exact same content as the earlier run (i.e the run which failed due to the branch being updated with master).

dentarg:
Thanks for the digging. Looks like it works as expected. We just need to watch out for the failing builds before we merge things.

PR #3047, opened 2022-12-25
CI: always run RuboCop lint job

dentarg:
Related to https://github.com/puma/puma/issues/3046

Also fix the lint offense that slipped through.


dentarg:
Turns out we do skip jobs correctly but it's a bit quirky as later runs can be green when jobs are skipped (but the jobs did fail previously).

I'll go ahead and merge this as I think it makes sense. We can always change it back if it isn't desirable.

Issue #3007, opened 2022-10-29
Streaming bodies, PR #2740, WebSocket


MSP-Greg:
**Describe the bug**

PR #2740 added support for streaming bodies.  Still reviewing, but somewhat confused.  Code below after the PR:
```ruby
if res_body && !res_body.respond_to?(:each)
  response_hijack = res_body
else
  response_hijack = resp_info[:response_hijack]
end

cork_socket socket

# ** removed existing 'no_body code', doesn't affect this issue

if content_length
  io_buffer.append CONTENT_LENGTH_S, content_length.to_s, line_ending
  chunked = false
elsif !response_hijack and resp_info[:allow_chunked]
  io_buffer << TRANSFER_ENCODING_CHUNKED
  chunked = true
end

io_buffer << line_ending

if response_hijack
  fast_write_str socket, io_buffer.to_s
  response_hijack.call socket
  return :async
end
```

Note the line `elsif !response_hijack and resp_info[:allow_chunked]`.  This means that if `Content-Length` is specified, we will write the header, but if `resp_info[:allow_chunked]` is set, there will be no transfer-eoncoding header written.  Hence, the added test in test_puma_server.rb `test_streaming_body` is an invalid response.  So, should Puma be writing the header?  This could be changed by removing `!response_hijack` from the elsif conditional.

Also, I'm not sure if Puma should be looking for headers returned from the app that would indicate a WebSocket upgrade and changing the response headers accordingly...

@ioquatix ?  Hope.this.made.sense.

ioquatix:
Thanks for creating this issue.

I would suggest you open a corresponding PR in the rack conformance test suite <https://github.com/socketry/rack-conform>, it will test it against several different web servers. You could introduce a test for the failing behaviour and see how other servers handle it. It would be great to have more streaming tests, especially if there are some edge cases which need to be considered more carefully.

I am personally of the opinion that less special snowflake behaviour, the more robust, easier to understand the implementation will be, so hopefully somewhat understandably, I get a bit frustrated when I see a lot of complex logic to handle application <-> protocol mappings. My general thesis, is to try and avoid complexity, or when unavoidable, keep the complexity in well defined layers so that at least in insolation it's understandable.

[...]

PR #3058, opened 2023-01-16
WebSocket support?

dentarg:
### Description

I've rebased @MSP-Greg's branch https://github.com/MSP-Greg/puma/tree/00-size-to-first-byte-ws

With these changes, Puma passes the [Rack::Conform](https://github.com/socketry/rack-conform) ["Basic WebSocket connection test"](https://github.com/socketry/rack-conform/pull/12), see https://github.com/dentarg/rack-conform/actions/runs/3928289028/jobs/6715741130

Needs some tests 

Close https://github.com/puma/puma/issues/3007

### DIFF
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -173,7 +173,7 @@ module Puma
       # below converts app_body into body, dependent on app_body's characteristics, and
       # resp_info[:content_length] will be set if it can be determined
       if !resp_info[:content_length] && !resp_info[:transfer_encoding] && status != 204
-        if res_body.respond_to?(:to_ary) && (array_body = res_body.to_ary)
+        if res_body.respond_to?(:to_ary) && (array_body = res_body.to_ary) && array_body.is_a?(Array)
           body = array_body
           resp_info[:content_length] = body.sum(&:bytesize)
         elsif res_body.is_a?(File) && res_body.respond_to?(:size)
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -241,7 +241,6 @@ module Puma
 
       if response_hijack
         fast_write_str socket, io_buffer.read_and_reset
-        uncork_socket socket
         response_hijack.call socket
         return :async
       end
@@ -569,7 +568,7 @@ module Puma
         else
           io_buffer.append "#{HTTP_11} #{status} ", fetch_status_code(status), line_ending
 
-          resp_info[:no_body] ||= STATUS_WITH_NO_ENTITY_BODY[status]
+          resp_info[:no_body] ||= status < 200 || STATUS_WITH_NO_ENTITY_BODY[status]
         end
       else
         resp_info[:allow_chunked] = false
@@ -583,7 +582,7 @@ module Puma
           io_buffer.append "HTTP/1.0 #{status} ",
                        fetch_status_code(status), line_ending
 
-          resp_info[:no_body] ||= STATUS_WITH_NO_ENTITY_BODY[status]
+          resp_info[:no_body] ||= status < 200 || STATUS_WITH_NO_ENTITY_BODY[status]
         end
       end
 
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -241,6 +241,7 @@ module Puma
 
       if response_hijack
         fast_write_str socket, io_buffer.read_and_reset
+        uncork_socket socket
         response_hijack.call socket
         return :async
       end
@@ -568,7 +569,7 @@ module Puma
         else
           io_buffer.append "#{HTTP_11} #{status} ", fetch_status_code(status), line_ending
 
-          resp_info[:no_body] ||= status < 200 || STATUS_WITH_NO_ENTITY_BODY[status]
+          resp_info[:no_body] ||= STATUS_WITH_NO_ENTITY_BODY[status]
         end
       else
         resp_info[:allow_chunked] = false
@@ -582,7 +583,7 @@ module Puma
           io_buffer.append "HTTP/1.0 #{status} ",
                        fetch_status_code(status), line_ending
 
-          resp_info[:no_body] ||= status < 200 || STATUS_WITH_NO_ENTITY_BODY[status]
+          resp_info[:no_body] ||= STATUS_WITH_NO_ENTITY_BODY[status]
         end
       end
 
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -223,7 +223,7 @@ module Puma
         # 101 (Switching Protocols) doesn't return here or have content_length,
         # it should be using `response_hijack`
         unless status == 101
-          if content_length and status != 204
+          if content_length && status != 204
             io_buffer.append CONTENT_LENGTH_S, content_length.to_s, line_ending
           end
 
@@ -236,7 +236,7 @@ module Puma
         if content_length
           io_buffer.append CONTENT_LENGTH_S, content_length.to_s, line_ending
           chunked = false
-        elsif !response_hijack and resp_info[:allow_chunked]
+        elsif !response_hijack && resp_info[:allow_chunked]
           io_buffer << TRANSFER_ENCODING_CHUNKED
           chunked = true
         end
@@ -487,7 +487,7 @@ module Puma
       to_add = nil
 
       env.each do |k,v|
-        if k.start_with?("HTTP_") and k.include?(",") and k != "HTTP_TRANSFER,ENCODING"
+        if k.start_with?("HTTP_") && k.include?(",") && k != "HTTP_TRANSFER,ENCODING"
           if to_delete
             to_delete << k
           else
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -220,27 +220,33 @@ module Puma
       cork_socket socket
 
       if resp_info[:no_body]
-        if content_length and status != 204
-          io_buffer.append CONTENT_LENGTH_S, content_length.to_s, line_ending
-        end
+        # 101 (Switching Protocols) doesn't return here or have content_length,
+        # it should be using `response_hijack`
+        unless status == 101
+          if content_length and status != 204
+            io_buffer.append CONTENT_LENGTH_S, content_length.to_s, line_ending
+          end
 
-        io_buffer << LINE_END
-        fast_write_str socket, io_buffer.read_and_reset
-        socket.flush
-        return keep_alive
-      end
-      if content_length
-        io_buffer.append CONTENT_LENGTH_S, content_length.to_s, line_ending
-        chunked = false
-      elsif !response_hijack and resp_info[:allow_chunked]
-        io_buffer << TRANSFER_ENCODING_CHUNKED
-        chunked = true
+          io_buffer << LINE_END
+          fast_write_str socket, io_buffer.read_and_reset
+          socket.flush
+          return keep_alive
+        end
+      else
+        if content_length
+          io_buffer.append CONTENT_LENGTH_S, content_length.to_s, line_ending
+          chunked = false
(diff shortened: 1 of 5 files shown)


nateberkopec:
Amazing work @dentarg! Thank you for doing this.

ioquatix:
I would also be okay with adding local test gem files to rack-conform if that makes it easier.

MSP-Greg:
@ioquatix Please allow the Window's guy some slack with sed.  My original sed strings were terrible, updated to:
```
SRC="gem ['\"]puma['\"].*"
DST="gem 'puma', git: 'https://github.com/$GITHUB_REPOSITORY.git', ref: '$GITHUB_SHA'"
sed -i "s#$SRC#$DST#" gems/puma-head-rack-v3.rb
```

I think it will be a bit more stable.  I'm ok with this, as we're running the CI as if it's from the rack-conform repo, and changing the 'Gemfile' before `bundle install`.  Need to use `GITHUB_REPOSITORY`, as it allows running from a fork of Puma.

ioquatix:
This looks good to me!

MSP-Greg:
Thanks.  Please let us know when we can change to the main branch.

Don't know about everyone else, but after seeing all the strange-ass things various (external) response classes do, I think it's a good idea for Puma to test with repos like rake-conform and turbo-rails...

ioquatix:
Okay, I will work on merging the websockets PR on rack-conform, if we can merge this one soon it will help.

MSP-Greg:
One reason I paused on this code is the status code handling.  I believe any response with a status code < 200 is considered a 'no body/content' response.  [IANA's Hypertext Transfer Protocol (HTTP) Status Code Registry](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml) and [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status) only show four codes in that range.

But, here and there it's implied that servers/apps can/may define their own.  Thoughts?  This code does not handle undefined 100 'range' codes as 'no body' codes...

MSP-Greg:
> Okay, I will work on merging the websockets PR on rack-conform, if we can merge this one soon it will help.

As long as you're back in that repo, might it log the tests?  It's not like there's hundreds/thousands of them...

BTW, no hurry, and thanks for getting it up and running.  Should be in the rack account, but maybe that's planned for later.  And, any other servers can feel free to use the workflow in their CI...

ioquatix:
@jeremyevans what do you think about moving the rack-conform test suite to the rack org?

ioquatix:
> But, here and there it's implied that servers/apps can/may define their own. Thoughts? This code does not handle undefined 100 'range' codes as 'no body' codes...

I've argued this within the rack spec, since `Rack::Response` also assumes any response of status code 100 doesn't have an entity body and thus was stripping out the upgrade/streaming response body.

https://github.com/rack/rack/pull/1987

The current implementation assumes that a streaming body is a super-set of entity bodies, and thus you shouldn't prune the response body for 1xx responses with a streaming response body.

jeremyevans:
> @jeremyevans what do you think about moving the rack-conform test suite to the rack org?

I think we should be more conservative when adding new repositories to the rack org. In hindsight, I don't think rack-test or rack-attack should have been moved to it.

After looking at the rack-conform repository and dependencies, I'm not in favor of moving rack-conform to the rack org.  In general, I don't think we should move external repositories into the rack org unless we have multiple rack committers willing to maintain them.

MSP-Greg:
> unless we have multiple rack committers willing to maintain them.

Makes sense.  But, as I've seen with Puma, web/app servers may be affected by small behavior changes in Rack, and they may not be testing for them.  Some kind of conformance code might be helpful.  Haven't really thought about it much...

nateberkopec:
Marking as "waiting for changes" as it maybe needs a test or two in the Puma test suite.

@dentarg do you feel up to that or want me and Greg to do it? Don't want you to sign up for more than you want to! 😄

ioquatix:
Okay, so my read is, the bar for moving `rack-conform` is at least one other person willing to help maintain it, whom is willing to also join the rack core team for maintenance purposes.

As it stands, it's fine where it is, but long term, eventually I have to assume I'm no longer going to be around (hopefully not for a long time). But when that happens, it makes sense that there is a way to continue maintaining the code if servers are using it as part of validating their conformance to the rack spec.

MSP-Greg:
> Marking as "waiting for changes" as it maybe needs a test or two in the Puma test suite.

JFYI, the additional workflow (using rack-conform) is creating a websocket connection, and also testing that the server (Puma) is correctly returning a 101 ('Switching Protocols') response and passing the IO to the app.

dentarg:
@nateberkopec if you or @MSP-Greg can add the tests we need, that's fine with me! Currently I don't know when I have a chunk of time free for that

ioquatix:
I have merged the `websockets` branch and it is now failing for Puma as expected.

MSP-Greg:
@dentarg @nateberkopec

Re tests, see the last commit 'Create test_puma_server_partial_hijack.rb, pull 1 test from test_puma…
…_server.rb'.  It pulls one test from test_puma_server.rb.  The two new tests fails when the commit is cherry picked to master.

MSP-Greg:
I just added three commits, one which reverts the first commit, which is the 101 fixes.  The first commit changed the definition of `resp_info[:no_body]` such that new or proprietary 1xx status codes would not be included.

But, [RFC 9110 15.2](https://www.rfc-editor.org/rfc/rfc9110.html#name-informational-1xx) clearly states **"A 1xx response is terminated by the end of the header section; it cannot contain content or trailers."**

The last new commit "request.rb - properly handle 101 'Switching Protocols' responses" correctly handles the above cases.

The new tests seem to be failing on TruffleRuby, possibly due to using syswrite & sysread.  Not.sure.  I hope to look at it sometime today.

ioquatix:
I'd like to introduce `rack.protocol` which when combined with a streaming body indicates an upgrade response.

Essentially:

```
if protocol = headers['rack.protocol']
  headers['connection'] = 'upgrade'
  headers['upgrade'] = protocol
  send_headers(headers)
  body.call(io)
else
  # other response handling
```

This also helps us in HTTP/2 which has a different "upgrade" mechanism (which doesn't use 101 status code).

MSP-Greg:
Added one more commit, fixes the TruffleRuby issue.  Adding an `io.wait_readable` statement fixed the isssue.

All passed in the test workflow.  I consider this finished, any thoughts?

ioquatix:
Okay, let me give it a review. w.r.t. my previous comment, detecting 101 status is good enough for now.

ioquatix (approved):
LGTM.

MSP-Greg:
> I'd like to introduce rack.protocol

Discuss in Rack?  JFYI, I'm kind of meh about it, as, assuming most WebSocket is `wss` (ssl/tls), the app will be generating other headers anyway.  So why would the app/web server need to add some more?

> HTTP/2 which has a different "upgrade" mechanism

I have yet to really take a 'deep dive' into HTTP/2 (or HTTP/3), so I'm speaking mostly about HTTP/1...

ioquatix:
Yep, we have a discussion for it: https://github.com/rack/rack/issues/1946

Right now, you are expecting the response to include the `connection: upgrade` and `upgrade: websocket` headers, as well as status code 101. But those headers are invalid for HTTP/2 so applications need to version sniff to send the correct response, which isn't obvious if you don't know that it's required. It's another area where HTTP/2 compatibility can be tricky.

ioquatix:
I've just merged and released support for webrick (v1.8.0) bi-directional streaming, added support in rackup (v2.1.0), and tested this using the conformance test suite (fully passing now for webrick on Rack 3).

I've added support for both `rack.protocol` and `upgrade` response headers, you can check how it works here:

https://github.com/rack/rackup/blob/eaea24a3d64a1b117df943a9d06779e659bb61af/lib/rackup/handler/webrick.rb#L137-L144

I don't actually have a strong opinion about this, but it's useful not to have to special case HTTP v1 and v2 too much, given that the upgrade header should not be specified on an HTTP/2 response.

test/test_puma_server_partial_hijack.rb
```diff
@@ -0,0 +1,136 @@
...
+    end
+  end
+
+  def server_run(**options, &block)
+    options[:log_writer]  ||= @log_writer
+    options[:min_threads] ||= 1
+    @server = Puma::Server.new block || @app, @events, options
+    @port = (@server.add_tcp_listener @host, 0).addr[1]
+    @server.run
+    sleep 0.15 if Puma.jruby?
```
nateberkopec, line 45: What are we sleeping for here?

test/test_puma_server_partial_hijack.rb
```diff
@@ -0,0 +1,136 @@
...
+  end
+
+  def test_101_header
+    headers = {
+      'Upgrade' => 'websocket',
+      'Connection' => 'Upgrade',
+      'Sec-WebSocket-Accept' => 's3pPLMBiTxaQ9kYGzzhZRbK+xOo=',
+      'Sec-WebSocket-Protocol' => 'chat',
+      'rack.hijack' => -> (io) {
+        io.wait_readable 0.1 # TRUFFLE
```
nateberkopec, line 100: Could you expand this comment just a little bit more? Why does Truffle need this?

lib/puma/request.rb
```diff
@@ -220,27 +220,33 @@ def prepare_response(status, headers, res_body, requests, client)
       cork_socket socket
 
       if resp_info[:no_body]
-        if content_length and status != 204
+        # 101 (Switching Protocols) doesn't return here or have content_length,
+        # it should be using `response_hijack`
+        unless status == 101
+          if content_length and status != 204
```
nateberkopec, line 226: let's just clean this up to `&&` while we're here (also on line 239)

nateberkopec (changes requested):
A few minor clarifications and then I'm ✅

test/test_puma_server_partial_hijack.rb
MSP-Greg: Updated comment.

Threads with JRuby/Truffle are different than MRI Rubies.  Thread creation/switching may take a bit more time when using JRuby/Truffle, especially with parallel testing in a multi VM CI environment.

Sometimes code additions like these may not be necessary when testing locally, other additions are required to pass both locally and in cloud CI.  A few tests use `parallelize_me! if ::Puma.mri?` because of these timing issues.

Re this particular issue, the following line
(`io.syswrite io.sysread(256)`) threw
`Read: #<Errno::EAGAIN: Resource temporarily unavailable>`

test/test_puma_server_partial_hijack.rb
MSP-Greg: See below comment.  Note that the above code came from test_puma_server.rb:
https://github.com/puma/puma/blob/8763c177a0bc8d7ec02b2faf1eda8efc687aa24b/test/test_puma_server.rb#L42-L49.

Since test_puma_server.rb currently has 1,565 lines, I thought creating another test file was a good idea.  Maybe it should be renamed and include 'full hijack' tests, thought we'd leave that for another commit.

lib/puma/request.rb
MSP-Greg: Done.

test/test_puma_server_partial_hijack.rb
MSP-Greg: I just loaded JRuby head locally, and the statement `sleep 0.15 if Puma.jruby?` isn't needed.  It may be needed when CI is very busy.  I'll remove and we'll see what happens.  JFYI, it was added in Oct '22, so things may be different now.

ioquatix:
Looking good!

nateberkopec (approved):

ioquatix:
Nice!

PR #3061, opened 2023-01-20
Fixup code for use with rackup gem (may be used with rack 3)

### LINKED ISSUE #3057: Using rack 3. Receive deprecation warning.
**Describe the bug**
deprecation warning: "Rack::Handler is deprecated and replaced by Rackup::Handler"

Spent a little time to determine the issue. The issue seems to be when rackup.rb requires puma.rb. puma.rb does a require on rack/handler.rb on about line 3. rack/handler.rb throws deprecation warning.

**Expected behavior**
App should start without deprecation warning.

 - OS: Mac M1 arm
 - Puma Version 6.0.2
 - rack version 3.0.3

~Marvin Foster 
email: mmarvb7@gmail.com or marvinfoster@finsync.com

MSP-Greg:
### Description

Rack v2 included the files needed to use `rackup` from the command line.  With Rack v3, it is now a separate gem, and the namespace has changed.

Re tests, when Rack v3 was released, we (ok, I) blocked  [test/test_rack_handler](https://github.com/puma/puma/blame/3f4e3c8ca689e39218459836b939c39fa81e2692/test/test_rack_handler.rb#L5) from running with Rack v3, so tests do exist.  This PR unblocks the tests.

The changes to `lib/rack/handler/puma.rb` look excessive.  This PR is taking the previous code, and turning it into a HEREDOC, so it can be used in `module_eval`.  This is the work-around to allow it to be placed in either the `Rack::Handler` or the `Rackup::Handler` namespace.

Adjust code to work with both.

**Note:** labeled as a 'bug', not sure...

Closes #3057 

### DIFF
--- a/lib/puma/rack_default.rb
+++ b/lib/puma/rack_default.rb
@@ -2,8 +2,20 @@
 
 require_relative '../rack/handler/puma'
 
-module Rack::Handler
-  def self.default(options = {})
-    Rack::Handler::Puma
+if Object.const_defined? :Rackup
+  module Rackup
+    module Handler
+      def self.default(options = {})
+        ::Rackup::Handler::Puma
+      end
+    end
+  end
+else
+  module Rack
+    module Handler
+      def self.default(options = {})
+        ::Rack::Handler::Puma
+      end
+    end
   end
 end
--- a/lib/puma/rack_default.rb
+++ b/lib/puma/rack_default.rb
@@ -2,6 +2,7 @@
 
 require_relative '../rack/handler/puma'
 
+# rackup was removed in Rack 3, it is now a separate gem
 if Object.const_defined? :Rackup
   module Rackup
     module Handler
@@ -10,7 +11,7 @@ if Object.const_defined? :Rackup
       end
     end
   end
-else
+elsif Object.const_defined?(:Rack) && Rack::RELEASE < '3'
   module Rack
     module Handler
       def self.default(options = {})
@@ -18,4 +19,6 @@ else
       end
     end
   end
+else
+  raise "Rack 3 must be used with the Rackup gem"
 end
--- a/lib/rack/handler/puma.rb
+++ b/lib/rack/handler/puma.rb
@@ -1,26 +1,18 @@
 # frozen_string_literal: true
 
-# Code below is 'odd', but don't want to define a top level constant
-
-if Object.const_defined? :Rackup
-  require 'rackup/handler'
-  ::Rackup::Handler
-else
-  require 'rack/handler'
-  ::Rack::Handler
-end.module_eval <<'HEREDOC'
-  module Puma
+# This module is used as an 'include' file in code at bottom of file
+module Puma
+  module RackHandler
     DEFAULT_OPTIONS = {
       :Verbose => false,
       :Silent  => false
     }
 
-    def self.config(app, options = {})
-      # `require_relative` won't work with `module_eval`
-      require 'puma'
-      require 'puma/configuration'
-      require 'puma/log_writer'
-      require 'puma/launcher'
+    def config(app, options = {})
+      require_relative '../../puma'
+      require_relative '../../puma/configuration'
+      require_relative '../../puma/log_writer'
+      require_relative '../../puma/launcher'
 
       default_options = DEFAULT_OPTIONS.dup
 
@@ -66,7 +58,7 @@ end.module_eval <<'HEREDOC'
       conf
     end
 
-    def self.run(app, **options)
+    def run(app, **options)
       conf   = self.config(app, options)
 
       log_writer = options.delete(:Silent) ? ::Puma::LogWriter.strings : ::Puma::LogWriter.stdio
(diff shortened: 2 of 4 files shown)


lib/puma/rack_default.rb
```diff
@@ -2,8 +2,20 @@
 
 require_relative '../rack/handler/puma'
 
-module Rack::Handler
-  def self.default(options = {})
-    Rack::Handler::Puma
+if Object.const_defined? :Rackup
```
choallin, line 5: Although this works, I think testing for Rack version 3 is more meaningful, because this tells the reader that this was implemented in response to a change in the new Rack version.

lib/rack/handler/puma.rb
```diff
@@ -1,114 +1,118 @@
...
-          end
+# Code below is 'odd', but don't want to define a top level constant
+
+if Object.const_defined? :Rackup
+  require 'rackup/handler'
+  ::Rackup::Handler
+else
+  require 'rack/handler'
+  ::Rack::Handler
+end.module_eval <<'HEREDOC'
```
choallin, line 11: To get rid of this `module_eval` we could create a new lib/rackup/handler.rb file that is required here. The file could either duplicate the code, or we are putting the code inside a module and are including this new module inside the new rackup/handler and rack/handler :thinking:

MSP-Greg: The cleanest way is with `modulfe_eval`, but that also 'hides' the code from RuboCop, etc.  I changed the code to use include.  See the last commit.

MSP-Greg:
A note for those who haven't looked at rack/rackup code, as of the current version of rackup (2.0.0), the code by default tries `require 'rack/handler/puma'` first, so the file path is fixed.  If Puma has been activated, that require statement succeeds...

lib/puma/rack_default.rb
```diff
@@ -2,8 +2,23 @@
...
+elsif Object.const_defined?(:Rack) && Rack::RELEASE < '3'
+  module Rack
+    module Handler
+      def self.default(options = {})
+        ::Rack::Handler::Puma
+      end
+    end
+  end
+else
+  raise "Rack 3 must be used with the Rackup gem"
```
nateberkopec, line 23: Error message could be something like `you must install the rackup gem.`, i.e. tell the user what to do exactly

lib/rack/handler/puma.rb
```diff
@@ -1,114 +1,136 @@
...
+          config.port port, host
         end
       end
     end
+  end
+end
 
-    register :puma, Puma
+# rackup was removed in Rack 3, it is now a separate gem
+if Object.const_defined? :Rackup
```
nateberkopec, line 112: Can we just combine this with what's in `rack_default`? 🤔

test/test_rack_handler.rb
```diff
@@ -2,16 +2,22 @@
 
 require "rack"
 
-if Rack::RELEASE < '3'
-
-  require "rack/handler/puma"
+module TestRackUp
+  if Rack::RELEASE < '3'
+    require "rack/handler/puma"
```
nateberkopec, line 7: this can be moved out of the if statement because you use it in both branches

test/test_rack_handler.rb
```diff
@@ -2,16 +2,22 @@
...
+    RACK_HANDLER_MOD = ::Rack::Handler
+  else
+    require "rackup"
+    require "rack/handler/puma"
+    RACK_HANDLER_MOD = ::Rackup::Handler
```
nateberkopec, line 12: Maybe we could just test `Puma::RackHandler` directly for most of the tests instead?

test/test_rack_handler.rb
MSP-Greg: Updated, see commit 'Changes per comments'

test/test_rack_handler.rb
MSP-Greg: Two methods are called in `RACK_HANDLER_MOD`, `register` and `get`.  If we want to verify that things work with Rack/Rackup, it would be good to continue doing so...

lib/rack/handler/puma.rb
MSP-Greg: No.  As mentioned above, Rackup currently will look for several servers, and Puma is the first one.  It determines if Puma is available by requiring `rack/handler/puma`.

For info on `rack_default`, see https://github.com/puma/puma/pull/194.

lib/puma/rack_default.rb
MSP-Greg: Updated, see commit 'Changes per comments'

lib/puma/rack_default.rb
MSP-Greg: Sorry, I thought I responded.  The rackup namespace needs to be defined.  The Rack v3 gem could be loaded, but unless the rackup gem is also loaded, this will fail.  I added a comment.

MSP-Greg:
Rebased

ioquatix:
Just FYI, I expect the rackup gem to evolve more rapidly than it has before. While Rack v2 and Rackup gems are currently aligned, I imagine the handler implementation will diverge in the future (or be extended I compatible ways which are not part of the Rack 2 implementation).

PR #3062, opened 2023-01-22
Reduce memory usage for large file uploads

willkoehler:
The motivation for this change is to reduce memory used by a Rails application running on Heroku that handles large file uploads (~100Mb). In it's current state the application (running Puma 6.0.2 + Rails 5.2.8 + Ruby 2.7.6) starts using swap and triggering "memory quota exceeded errors" on the Heroku dyno after only a few file uploads.

This PR is a proof of concept that appears to significantly reduce the memory used by Puma during the data phase of the HTTP request. The change I made was to re-use a single read buffer in `Puma::Client#read_body` and `Puma::Client#read_chunked_body` vs allocating a new buffer on each call to `read_nonblock()`.

I proved this change worked locally by throttling the network speed in Chrome dev tools and watching application memory usage as Puma received the file upload request. I then reproduced this experiment on a Heroku dyno, uploading a 100 Mb file to my application while watching memory on the dyno with `top`. I grabbed screenshots just before Puma handed the request off to Rack to isolate any memory usage to Puma alone.

### Puma v6.0.2 - Initial memory
memory13_at_idle_puma_6_0_2

### Puma v6.0.2 - Memory at 95% uploaded
memory14_95percent_uploaded_puma_6_0_2

### This PR - Initial memory
memory11_at_idle_with_fix

### This PR - Memory at 95% uploaded
memory12_95percent_uploaded_with_fix

To demonstrate this change had a positive impact on the application memory footprint overall (not just the Puma data phase), I watched Heroku's memory usage graph while uploading 5 large (100Mb) files. This PR reduced memory usage by ~100Mb (25% reduction) compared to Puma 6.0.2. See screenshot below.

heroku_memory

My biggest concern is that this solution may not be thread-safe. To test this, I ran Puma locally with `WEB_CONCURRENCY=0 bin/rails s` and performed 3 simultaneous 100Mb file uploads. Each of the files was received successfully.

One final note: I did not make a change to the `read_nonblock` call in `Puma::Client#try_to_finish` Doing so creates a shared memory conflict between `data` and `@buffer` and `@read_buffer` (they are all just references to the same thing). This would need to be fixed by duping `data` before assigning to `@buffer`.

```
if @buffer
  @buffer << data
else
  @buffer = data.dup
end
```

I felt that this additional complexity may be confusing and easily broken in the future.

### DIFF
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -98,6 +98,8 @@ module Puma
       @body_remain = 0
 
       @in_last_chunk = false
+
+      @read_buffer = +""
     end
 
     attr_reader :env, :to_io, :body, :io, :timeout_at, :ready, :hijacked,
@@ -433,7 +435,7 @@ module Puma
       end
 
       begin
-        chunk = @io.read_nonblock(want)
+        chunk = @io.read_nonblock(want, @read_buffer)
       rescue IO::WaitReadable
         return false
       rescue SystemCallError, IOError
@@ -465,7 +467,7 @@ module Puma
     def read_chunked_body
       while true
         begin
-          chunk = @io.read_nonblock(4096)
+          chunk = @io.read_nonblock(4096, @read_buffer)
         rescue IO::WaitReadable
           return false
         rescue SystemCallError, IOError
--- a/benchmarks/local/sinatra/Gemfile
+++ b/benchmarks/local/sinatra/Gemfile
@@ -1,0 +1,13 @@
+source "http://rubygems.org"
+git_source(:github) { |repo| "https://github.com/#{repo}.git" }
+
+ruby "3.2.0"
+
+gem "sinatra"
+gem "puma_worker_killer"
+
+# current puma release
+gem "puma"
+
+# PR to reduce memory of large file uploads
+# gem "puma", github: "willkoehler/puma", branch: "reduce_read_body_memory"
--- a/benchmarks/local/sinatra/README.md
+++ b/benchmarks/local/sinatra/README.md
@@ -1,0 +1,92 @@
+# Large file upload demo
+
+This is a simple app to demonstrate memory used by Puma for large file uploads and
+compare it to proposed changes in PR https://github.com/puma/puma/pull/3062
+
+### Steps to test memory improvements in https://github.com/puma/puma/pull/3062
+
+- Run the app with puma_worker_killer: `bundle exec puma -p 9090 --config puma.rb`
+- Make a POST request with curl: `curl --form "data=@some_large_file.mp4" --limit-rate 10M http://localhost:9090/`
+- Puma will log memory usage in the console
+
+Below is example of the results uploading a 115MB video.
+
+### Puma 6.0.2
+
+```
+[11820] Puma starting in cluster mode...
+[11820] * Puma version: 6.0.2 (ruby 3.2.0-p0) ("Sunflower")
+[11820] *  Min threads: 0
+[11820] *  Max threads: 5
+[11820] *  Environment: development
+[11820] *   Master PID: 11820
+[11820] *      Workers: 1
+[11820] *     Restarts: (✔) hot (✔) phased
+[11820] * Listening on http://0.0.0.0:3000
+[11820] Use Ctrl-C to stop
+[11820] - Worker 0 (PID: 11949) booted in 0.06s, phase: 0
+[11820] PumaWorkerKiller: Consuming 70.984375 mb with master and 1 workers.
+[11820] PumaWorkerKiller: Consuming 70.984375 mb with master and 1 workers.
+
+...curl request made - memory increases as file is received
+
+[11820] PumaWorkerKiller: Consuming 72.796875 mb with master and 1 workers.
+[11820] PumaWorkerKiller: Consuming 75.921875 mb with master and 1 workers.
+[11820] PumaWorkerKiller: Consuming 78.953125 mb with master and 1 workers.
+[11820] PumaWorkerKiller: Consuming 82.15625 mb with master and 1 workers.
+[11820] PumaWorkerKiller: Consuming 85.265625 mb with master and 1 workers.
+[11820] PumaWorkerKiller: Consuming 88.046875 mb with master and 1 workers.
+
+...(clipped out lines) memory keeps increasing while request is received
+
+[11820] PumaWorkerKiller: Consuming 121.53125 mb with master and 1 workers.
+[11820] PumaWorkerKiller: Consuming 122.75 mb with master and 1 workers.
+[11820] PumaWorkerKiller: Consuming 125.40625 mb with master and 1 workers.
+
+...request handed off from Puma to Rack/Sinatra
+
+[11820] PumaWorkerKiller: Consuming 220.6875 mb with master and 1 workers.
+127.0.0.1 - - [26/Jan/2023:20:09:56 -0500] "POST /upload HTTP/1.1" 200 162 0.0553
+[11820] PumaWorkerKiller: Consuming 228.96875 mb with master and 1 workers.
+[11820] PumaWorkerKiller: Consuming 228.96875 mb with master and 1 workers.
+```
+
+### With PR https://github.com/puma/puma/pull/3062
+
+```
+[20815] Puma starting in cluster mode...
+[20815] * Puma version: 6.0.2 (ruby 3.2.0-p0) ("Sunflower")
+[20815] *  Min threads: 0
+[20815] *  Max threads: 5
+[20815] *  Environment: development
+[20815] *   Master PID: 20815
+[20815] *      Workers: 1
+[20815] *     Restarts: (✔) hot (✔) phased
+[20815] * Listening on http://0.0.0.0:3000
+[20815] Use Ctrl-C to stop
+[20815] - Worker 0 (PID: 20944) booted in 0.1s, phase: 0
+[20815] PumaWorkerKiller: Consuming 73.25 mb with master and 1 workers.
+[20815] PumaWorkerKiller: Consuming 73.25 mb with master and 1 workers.
+
+...curl request made - memory stays level as file is received
+
+[20815] PumaWorkerKiller: Consuming 73.28125 mb with master and 1 workers.
+[20815] PumaWorkerKiller: Consuming 73.296875 mb with master and 1 workers.
+[20815] PumaWorkerKiller: Consuming 73.34375 mb with master and 1 workers.
+[20815] PumaWorkerKiller: Consuming 73.359375 mb with master and 1 workers.
+[20815] PumaWorkerKiller: Consuming 73.359375 mb with master and 1 workers.
+[20815] PumaWorkerKiller: Consuming 73.359375 mb with master and 1 workers.
+
+...(clipped out lines) memory continues to stay level
+
+[20815] PumaWorkerKiller: Consuming 73.703125 mb with master and 1 workers.
+[20815] PumaWorkerKiller: Consuming 73.703125 mb with master and 1 workers.
+[20815] PumaWorkerKiller: Consuming 73.703125 mb with master and 1 workers.
+
+...request handed off from Puma to Rack/Sinatra
+
+[20815] PumaWorkerKiller: Consuming 181.96875 mb with master and 1 workers.
+127.0.0.1 - - [26/Jan/2023:20:27:16 -0500] "POST /upload HTTP/1.1" 200 162 0.0585
+[20815] PumaWorkerKiller: Consuming 183.78125 mb with master and 1 workers.
+[20815] PumaWorkerKiller: Consuming 183.78125 mb with master and 1 workers.
+```
--- a/benchmarks/local/sinatra/config.ru
+++ b/benchmarks/local/sinatra/config.ru
@@ -1,0 +1,7 @@
+require "sinatra"
+
+post "/" do
+  204
+end
+
+run Sinatra::Application
--- a/benchmarks/local/sinatra/puma.rb
+++ b/benchmarks/local/sinatra/puma.rb
@@ -1,0 +1,15 @@
+silence_single_worker_warning
+
+workers 1
+
+before_fork do
+  require "puma_worker_killer"
+
+  PumaWorkerKiller.config do |config|
+    config.ram = 1024 # mb
+    config.frequency = 0.3 # seconds
+    config.reaper_status_logs = true # Log memory: PumaWorkerKiller: Consuming 54.34765625 mb with master and 1 workers.
+  end
+
+  PumaWorkerKiller.start
+end


nateberkopec:
> My biggest concern is that this solution may not be thread-safe. 

Someone check me here but I believe Client objects are only accessed by the Reactor thread, so I don't think there can be a thread safety issue here because this state is not shared.

lib/puma/client.rb
```diff
@@ -98,6 +98,8 @@ def initialize(io, env=nil)
       @body_remain = 0
 
       @in_last_chunk = false
+
+      @read_buffer = +""
```
nateberkopec, line 102: What is this? I haven't seen this syntax before

dentarg: It is the short syntax for a non frozen String (if strings was to be frozen by default). Same as `String.new` which would read better IMHO.

johnnyshields: It also has better performance.

```
Benchmark.bm do |bm|
  bm.report { 100000000.times { String.new } }
  bm.report { 100000000.times { +'' } }
end
   4.359000   0.000000   4.359000 ( 10.255661)
   2.954000   0.000000   2.954000 (  6.286254)
```

dentarg:
Can you contribute a repro app where you see the same gain? I think it would be interesting for others to reproduce/play with this, to verify gains.

lib/puma/client.rb
willkoehler: I originally had `String.new` but this triggered a Rubocop `Performance/UnfreezeString` error.

I like `String.new` because it's intention is clear and we only allocate once. But I also see the benefit of using `+""` as a rule app-wide because performance is important in Puma. There are some places where this gain would be meaningful.

willkoehler:
> > My biggest concern is that this solution may not be thread-safe.
> 
> Someone check me here but I believe Client objects are only accessed by the Reactor thread, so I don't think there can be a thread safety issue here because this state is not shared.

This was my take as well. It appeared that `@read_buffer` would only be used by a single request at any given time.

willkoehler:
To keep things as simple as possible, I'm building a small Sinatra app to demonstrate the same behavior. I have a proof of concept running. I'll circle back here when I have the final version pushed up with reproducible steps to demonstrate the memory consumption.

lib/puma/client.rb
nateberkopec: Thanks, it's obviously not very easily googleable 😆

nateberkopec:
If it's just built in Sinatra, you could drop into the `benchmarks` folder of the repository as well.

willkoehler:
The demo app is up https://github.com/willkoehler/puma_demo

I see the same essential memory behavior in this demo app as the Rails app. However Ruby garbage collection appears to be a bit more efficient in the demo app. Memory doesn't grow in the same unbounded way over multiple uploads as I'm seeing in the Rails app. However the demo app still shows a clear benefit with the PR vs Puma 6.0.2 during the data phase of the HTTP request.

dentarg:
Thanks for the demo app, I simplified it a tiny bit (just the `config.ru`, upload using curl) and added `puma_worker_killer` (https://github.com/zombocom/puma_worker_killer) just to log the memory usage (every 0.3 second): https://github.com/dentarg/gists/tree/master/gists/puma_pr-3062-upload-file-mem



Puma 6.0.2

```
[36093] Puma starting in cluster mode...
[36093] * Puma version: 6.0.2 (ruby 3.1.3-p185) ("Sunflower")
[36093] *  Min threads: 0
[36093] *  Max threads: 5
[36093] *  Environment: development
[36093] *   Master PID: 36093
[36093] *      Workers: 1
[36093] *     Restarts: (✔) hot (✔) phased
[36093] * Listening on http://0.0.0.0:9090
[36093] Use Ctrl-C to stop
[36093] - Worker 0 (PID: 36094) booted in 0.07s, phase: 0
[36093] PumaWorkerKiller: Consuming 70.703125 mb with master and 1 workers.
[36093] PumaWorkerKiller: Consuming 70.703125 mb with master and 1 workers.
...
[36093] PumaWorkerKiller: Consuming 110.40625 mb with master and 1 workers.
[36093] PumaWorkerKiller: Consuming 161.328125 mb with master and 1 workers.
[36093] PumaWorkerKiller: Consuming 276.515625 mb with master and 1 workers.
127.0.0.1 - - [26/Jan/2023:22:51:24 +0100] "POST /upload HTTP/1.1" 200 7 0.1335
[36093] PumaWorkerKiller: Consuming 250.859375 mb with master and 1 workers.
[36093] PumaWorkerKiller: Consuming 250.875 mb with master and 1 workers.
[36093] PumaWorkerKiller: Consuming 250.9375 mb with master and 1 workers.
```

This PR

```
[36240] Puma starting in cluster mode...
[36240] * Puma version: 6.0.2 (ruby 3.1.3-p185) ("Sunflower")
[36240] *  Min threads: 0
[36240] *  Max threads: 5
[36240] *  Environment: development
[36240] *   Master PID: 36240
[36240] *      Workers: 1
[36240] *     Restarts: (✔) hot (✔) phased
[36240] * Listening on http://0.0.0.0:9090
[36240] Use Ctrl-C to stop
[36240] - Worker 0 (PID: 36242) booted in 0.07s, phase: 0
[36240] PumaWorkerKiller: Consuming 71.125 mb with master and 1 workers.
[36240] PumaWorkerKiller: Consuming 71.15625 mb with master and 1 workers.
...
[36240] PumaWorkerKiller: Consuming 71.578125 mb with master and 1 workers.
[36240] PumaWorkerKiller: Consuming 140.25 mb with master and 1 workers.
[36240] PumaWorkerKiller: Consuming 213.28125 mb with master and 1 workers.
127.0.0.1 - - [26/Jan/2023:22:51:51 +0100] "POST /upload HTTP/1.1" 200 7 0.1279
[36240] PumaWorkerKiller: Consuming 213.28125 mb with master and 1 workers.
[36240] PumaWorkerKiller: Consuming 213.3125 mb with master and 1 workers.
[36240] PumaWorkerKiller: Consuming 213.3125 mb with master and 1 workers.
```

dentarg:
Forgot to mention, the file I uploaded above was 322 MB (337322761 bytes)

willkoehler:
Nice 🙌 I like the trick of using`puma_worker_killer` to monitor memory. Using curl to make the request will allow us to use the `--limit-rate` param to throttle the bandwidth, which will more closely simulate behavior in production so we can see how memory grows and is garbaged collected during the data phase.

There's a lot going on here to unpack memory-wise and I want to make sure we're separating Puma's memory use from the rest of the app. In particular, once Puma hands the app off to Rack and Sinatra, I see additional memory allocation that we won't be able to control.

If we throttle the request bandwidth with `--limit-rate` I think what we will see is that, with Puma 6.0.2, memory grows during the data phase of the request (inside the Puma code) and it will look like this

```
[36093] PumaWorkerKiller: Consuming 70.703125 mb with master and 1 workers.
[36093] PumaWorkerKiller: Consuming 70.703125 mb with master and 1 workers.

**** Memory keeps growing - mostly but maybe sometimes dropping when garbage collection runs

[36093] PumaWorkerKiller: Consuming 110.40625 mb with master and 1 workers.
```

With my PR I expect it will look like this

```
[36240] PumaWorkerKiller: Consuming 71.125 mb with master and 1 workers.
[36240] PumaWorkerKiller: Consuming 71.15625 mb with master and 1 workers.

**** Memory stays about the same

[36240] PumaWorkerKiller: Consuming 71.578125 mb with master and 1 workers.
```

I believe the last few memory logs in each of the samples above (where the memory usage jumps) are after Puma has handed the request off to Rack

```
[36240] PumaWorkerKiller: Consuming 71.578125 mb with master and 1 workers.

**** Request handed from Puma to Rack/Sinatra. Memory use jumps here but is out of our control.

[36240] PumaWorkerKiller: Consuming 140.25 mb with master and 1 workers.
[36240] PumaWorkerKiller: Consuming 213.28125 mb with master and 1 workers.
127.0.0.1 - - [26/Jan/2023:22:51:51 +0100] "POST /upload HTTP/1.1" 200 7 0.1279
[36240] PumaWorkerKiller: Consuming 213.28125 mb with master and 1 workers.
[36240] PumaWorkerKiller: Consuming 213.3125 mb with master and 1 workers.
[36240] PumaWorkerKiller: Consuming 213.3125 mb with master and 1 workers.
```

I'm going to update my app locally with your changes, reproduce your test and see if I can add some clarity about where memory is being allocated and how this PR may help.

willkoehler:
I updated my app and tested with `puma_worker_killer` and `curl`. Thanks again for the suggestion 🙌  This is much nicer than the way I was testing.

Here are the results uploading a 115MB video.

Run the app: `bundle exec puma -p 9090 --config puma.rb -w 1`
Make a rate-limited request: `curl --form "video=@IMG_0009.MOV" --limit-rate 10M http://localhost:9090/upload`

### Puma 6.0.2

```
[11820] Puma starting in cluster mode...
[11820] * Puma version: 6.0.2 (ruby 3.2.0-p0) ("Sunflower")
[11820] *  Min threads: 0
[11820] *  Max threads: 5
[11820] *  Environment: development
[11820] *   Master PID: 11820
[11820] *      Workers: 1
[11820] *     Restarts: (✔) hot (✔) phased
[11820] * Listening on http://0.0.0.0:3000
[11820] Use Ctrl-C to stop
[11820] - Worker 0 (PID: 11949) booted in 0.06s, phase: 0
[11820] PumaWorkerKiller: Consuming 70.984375 mb with master and 1 workers.
[11820] PumaWorkerKiller: Consuming 70.984375 mb with master and 1 workers.

...curl request made - memory increases as file is received

[11820] PumaWorkerKiller: Consuming 72.796875 mb with master and 1 workers.
[11820] PumaWorkerKiller: Consuming 75.921875 mb with master and 1 workers.
[11820] PumaWorkerKiller: Consuming 78.953125 mb with master and 1 workers.
[11820] PumaWorkerKiller: Consuming 82.15625 mb with master and 1 workers.
[11820] PumaWorkerKiller: Consuming 85.265625 mb with master and 1 workers.
[11820] PumaWorkerKiller: Consuming 88.046875 mb with master and 1 workers.

...(clipped out lines) memory keeps increasing while request is received

[11820] PumaWorkerKiller: Consuming 121.53125 mb with master and 1 workers.
[11820] PumaWorkerKiller: Consuming 122.75 mb with master and 1 workers.
[11820] PumaWorkerKiller: Consuming 125.40625 mb with master and 1 workers.

...request handed off from Puma to Rack/Sinatra

[11820] PumaWorkerKiller: Consuming 220.6875 mb with master and 1 workers.
127.0.0.1 - - [26/Jan/2023:20:09:56 -0500] "POST /upload HTTP/1.1" 200 162 0.0553
[11820] PumaWorkerKiller: Consuming 228.96875 mb with master and 1 workers.
[11820] PumaWorkerKiller: Consuming 228.96875 mb with master and 1 workers.
```

### This PR

```
[20815] Puma starting in cluster mode...
[20815] * Puma version: 6.0.2 (ruby 3.2.0-p0) ("Sunflower")
[20815] *  Min threads: 0
[20815] *  Max threads: 5
[20815] *  Environment: development
[20815] *   Master PID: 20815
[20815] *      Workers: 1
[20815] *     Restarts: (✔) hot (✔) phased
[20815] * Listening on http://0.0.0.0:3000
[20815] Use Ctrl-C to stop
[20815] - Worker 0 (PID: 20944) booted in 0.1s, phase: 0
[20815] PumaWorkerKiller: Consuming 73.25 mb with master and 1 workers.
[20815] PumaWorkerKiller: Consuming 73.25 mb with master and 1 workers.

...curl request made - memory stays level as file is received

[20815] PumaWorkerKiller: Consuming 73.28125 mb with master and 1 workers.
[20815] PumaWorkerKiller: Consuming 73.296875 mb with master and 1 workers.
[20815] PumaWorkerKiller: Consuming 73.34375 mb with master and 1 workers.
[20815] PumaWorkerKiller: Consuming 73.359375 mb with master and 1 workers.
[20815] PumaWorkerKiller: Consuming 73.359375 mb with master and 1 workers.
[20815] PumaWorkerKiller: Consuming 73.359375 mb with master and 1 workers.

...(clipped out lines) memory continues to stay level

[20815] PumaWorkerKiller: Consuming 73.703125 mb with master and 1 workers.
[20815] PumaWorkerKiller: Consuming 73.703125 mb with master and 1 workers.
[20815] PumaWorkerKiller: Consuming 73.703125 mb with master and 1 workers.

...request handed off from Puma to Rack/Sinatra

[20815] PumaWorkerKiller: Consuming 181.96875 mb with master and 1 workers.
127.0.0.1 - - [26/Jan/2023:20:27:16 -0500] "POST /upload HTTP/1.1" 200 162 0.0585
[20815] PumaWorkerKiller: Consuming 183.78125 mb with master and 1 workers.
[20815] PumaWorkerKiller: Consuming 183.78125 mb with master and 1 workers.
```

With larger files garbage collection sometimes kicks in and improve the memory story. But I still think this change has benefits:

- I've found that garbage collection isn't perfect. Memory still tends to grow in real-world apps.
- If there's a straightforward way to avoid allocating the memory in the first place and making the garbage collector work then we reduce the overall load on the system.
- There's more improvement that can be made after Puma hands over the request to Rack. I have ideas there as well. But that's a topic for a separate PR

nateberkopec:
Any chance we could commit this demo app to `/benchmarks`? Say in a new folder?

willkoehler:
> Any chance we could commit this demo app to `/benchmarks`? Say in a new folder?

For sure 👍 I simplified the demo app and added it to `/benchmarks/local/sinatra` Feel free to suggest a different name or location.

nateberkopec:
Excellent. Thank you very much for your contribution.

willkoehler:
Thanks everyone for the review and vetting 🙌 I'm running https://github.com/puma/puma/commit/59737815895508cca2caf010b8a8ef79cd1f2730 in production. I'll report back if I see anything strange.

sanzstez:
@willkoehler Can you please check this strange behaviour https://github.com/puma/puma/issues/3145 produced by your PR?

willkoehler:
Hi @sanzstez I'd be happy to help look into that. Let's continue the conversation in #3145

boazsegev:
@willkoehler , thank you for the demo-app I used it to test `iodine` 👍🏻

I think this is a wonderful approach, except that now that the `@read_buffer` isn't re-allocated for every `read`, I would consider increasing the buffer size. The system's buffer wouldn't be less than 64kb and using smaller chunks increases system calls (which hurts performance)... I would push for a 32Kb or 64Kb buffer instead (IMHO).

Good luck!

willkoehler:
Hi @boazsegev, I'm glad the demo-app was helpful 🙌

This was my first Puma contribution, so the decision about the buffer size is above my "pay grade". However, it sounds plausible that a larger buffer size would be an improvement and it's probably worth opening a separate issue to start the conversation.

Feel free to tag me in the issue. I would be happy to explore the benefits with you and make a case for increasing the buffer size if it ends up improving performance or memory efficiency.

MSP-Greg:
Wondering whether this could be a config option?  Maybe `req_body_buffer_size` or something similar?

Allows for a typical config 'out-of-the-box`, but users dealing with large files can increase it...

Issue #2818, opened 2022-02-09
Self-signed SSL certificates (via the `localhost`) with Cluster mode is raised an error 


y-yagi:
**Describe the bug** 

I tried to use Self-signed SSL certificates. But it seems that doesn't work in Cluster mode. 

**To Reproduce**

```ruby 
# config.ru 
require "localhost"
run proc { [200, {"Content-Type" => "text/plain"}, ["Hello, World!"]] }
```

Run it with:

```bash
$ bundle exec puma -b "ssl://0.0.0.0:9292" -t 5:5 -w 2 config.ru 
[26770] Puma starting in cluster mode...
[26770] * Puma version: 5.6.1 (ruby 3.0.3-p157) ("Birdie's Version")
[26770] *  Min threads: 5
[26770] *  Max threads: 5
[26770] *  Environment: development
[26770] *   Master PID: 26770
[26770] *      Workers: 2
[26770] *     Restarts: (✔) hot (✔) phased
2022-02-09 10:45:00 +0900 [26770] ERROR: Please specify the SSL key via 'key=' or 'key_pem='
```

It works well in Single mode. 

```bash
$ bundle exec puma -b "ssl://0.0.0.0:9292" -t 5:5 config.ru
Puma starting in single mode...
* Puma version: 5.6.1 (ruby 3.0.3-p157) ("Birdie's Version")
*  Min threads: 5
*  Max threads: 5
*  Environment: development
*          PID: 27111
* Listening on ssl://0.0.0.0:9292?
Use Ctrl-C to stop 
```

**Desktop (please complete the following information):** 

 - OS: [Linux(Ubuntu 18.04.6 LTS)]
 - Puma Version [5.6.1]

nateberkopec:
Confirmed. I am very sad we forgot to add a test for this! 😞

nateberkopec:
I think we could add a test in `test_integration_ssl.rb` using the example config from #2764

maleksiuk:
I investigated this issue a little bit and wanted to report my findings in case they're helpful.

The localhost gem support works fine in clustered mode if `require "localhost"` appears in the Puma config file or if preloading the app. In other words, these two variations work:

```
bundle exec bin/puma -b "ssl://0.0.0.0:9292" -t 5:5 -w 2 --preload config.ru
```

(where config.ru is the same one specified in this issue's description)

```
bundle exec bin/puma -t 5:5 -w 2 -C test/config/ssl_self_signed_config.rb
```

[...]

PR #3068, opened 2023-01-28
Self-signed SSL certificates (via the localhost) with Cluster mode 

choallin:
### Description
This PR should fix issue #2818.
Right now I only created a failing test, so that we can start a discussion about it

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,9 +198,9 @@ $ puma -b 'ssl://127.0.0.1:9292?key=path_to_key&cert=path_to_cert'
 ```
 #### Self-signed SSL certificates (via the [`localhost`] gem, for development use):
 
-Puma supports the [`localhost`] gem for self-signed certificates. This is particularly useful if you want to use Puma with SSL locally, and self-signed certificates will work for your use-case. Currently, the integration can only be used in MRI.
+Puma supports the [`localhost`] gem for self-signed certificates. This is particularly useful if you want to use Puma with SSL locally, and self-signed certificates will work for your use-case. Currently, the integration can only be used in MRI and you need to require the [`localhost`] gem in your Puma configuration file and not in your rackup file [`config.ru`].
 
-Puma automatically configures SSL when the [`localhost`] gem is loaded in a `development` environment:
+Puma automatically configures SSL when the [`localhost`] gem is loaded:
 
 ```ruby
 # Add the gem to your Gemfile
@@ -226,10 +226,6 @@ $ puma -b 'ssl://localhost:9292' config.ru
 # The following options allow you to reach Puma over HTTP as well:
 $ puma -b ssl://localhost:9292 -b tcp://localhost:9393 config.ru
 ```
-
-There is a known limitation to the [`localhost`] gem in Puma. Puma can't start in cluster 
-mode when the user didn't preload the app or requires the [`localhost`] gem in a 
-configuration file.
 
 [`localhost`]: https://github.com/socketry/localhost
 
--- a/README.md
+++ b/README.md
@@ -198,7 +198,7 @@ $ puma -b 'ssl://127.0.0.1:9292?key=path_to_key&cert=path_to_cert'
 ```
 #### Self-signed SSL certificates (via the [`localhost`] gem, for development use):
 
-Puma supports the [`localhost`] gem for self-signed certificates. This is particularly useful if you want to use Puma with SSL locally, and self-signed certificates will work for your use-case. Currently, the integration can only be used in MRI and you need to require the [`localhost`] gem in your Puma configuration file and not in your rackup file [`config.ru`].
+Puma supports the [`localhost`] gem for self-signed certificates. This is particularly useful if you want to use Puma with SSL locally, and self-signed certificates will work for your use-case. Currently, the integration can only be used in MRI and you need to require the [`localhost`] gem in your Puma configuration file.
 
 Puma automatically configures SSL when the [`localhost`] gem is loaded:
 
@@ -211,20 +211,15 @@ end
 # And require it implicitly using bundler
 require "bundler"
 Bundler.require(:default, ENV["RACK_ENV"].to_sym)
-
-# Alternatively, you can require the gem in config.ru:
-require './app'
-require 'localhost'
-run Sinatra::Application
 ```
 
 Additionally, Puma must be listening to an SSL socket:
 
 ```shell
-$ puma -b 'ssl://localhost:9292' config.ru
+$ puma -b 'ssl://localhost:9292' -C path/to/puma-config.rb config.ru
 
 # The following options allow you to reach Puma over HTTP as well:
-$ puma -b ssl://localhost:9292 -b tcp://localhost:9393 config.ru
+$ puma -b ssl://localhost:9292 -b tcp://localhost:9393 -C path/to/puma-config.rb config.ru
 ```
 
 [`localhost`]: https://github.com/socketry/localhost
--- a/README.md
+++ b/README.md
@@ -227,6 +227,10 @@ $ puma -b 'ssl://localhost:9292' config.ru
 $ puma -b ssl://localhost:9292 -b tcp://localhost:9393 config.ru
 ```
 
+There is a known limitation to the [`localhost`] gem in Puma. Puma can't start in cluster 
+mode when the user didn't preload the app or requires the [`localhost`] gem in a 
+configuration file.
+
 [`localhost`]: https://github.com/socketry/localhost
 
 #### Controlling SSL Cipher Suites
--- a/lib/puma/minissl/context_builder.rb
+++ b/lib/puma/minissl/context_builder.rb
@@ -33,7 +33,7 @@ module Puma
           ctx.protocols = params['protocols'] if params['protocols']
         else
           if params['key'].nil? && params['key_pem'].nil?
-            log_writer.error "Please specify the SSL key via 'key=' or 'key_pem='"
+            log_writer.error "Please specify the SSL key via 'key=' or 'key_pem='\nIf you are using the 'localhost' gem and run Puma in cluster mode please use a config file and load it there so it is accessable for Puma."
           end
 
           ctx.key = params['key'] if params['key']
--- a/lib/puma/minissl/context_builder.rb
+++ b/lib/puma/minissl/context_builder.rb
@@ -33,7 +33,7 @@ module Puma
           ctx.protocols = params['protocols'] if params['protocols']
         else
           if params['key'].nil? && params['key_pem'].nil?
-            log_writer.error "Please specify the SSL key via 'key=' or 'key_pem='\nIf you are using the 'localhost' gem and run Puma in cluster mode please use a config file and load it there so it is accessable for Puma."
+            log_writer.error "Please specify the SSL key via 'key=' or 'key_pem=', or require the 'localhost' gem in your Puma config for automatic self-signed certificates."
           end
 
           ctx.key = params['key'] if params['key']
--- a/test/test_integration_ssl.rb
+++ b/test/test_integration_ssl.rb
@@ -190,10 +190,11 @@ RUBY
 
   def test_ssl_run_without_any_key_and_authority; require 'logger'
     skip_if :windows; require 'stringio'
+    skip_if :jruby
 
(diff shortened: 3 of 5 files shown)


test/test_integration_ssl.rb
```diff
@@ -189,6 +188,33 @@ def test_ssl_run_with_localhost_authority
     end
   end
 
+  def test_ssl_run_with_localhost_authority_in_cluster_mode
+    skip_if :jruby
+
+    cmd = "#{BASE} bin/puma -b \"ssl://0.0.0.0:9292\" -t 5:5 -w 2 test/rackup/self_signed.ru"
```
choallin, line 194: The issue occurs *only* with this setting, when using the env var WEB_CONCURRENCY or when using the puma DSL `workers`  puma starts correctly.
Now since I didn't want to change the `with_server` method I duplicated some code here

nateberkopec: Looks like you're on the right track 👍

choallin: I now have three possible ways to handle this scenario:
1. Require the user to use a config file that loads the `localhost` gem so that the methods are available in the `Binder` class.
2. Require the user to use --preload so that the `localhost` gem can be loaded before binding to a socket/port
3. We `require "localhost"` in the `Binder` class. For this we have to check if the gem is available in the current bundle

For options 1 and 2 an additional error message + documentation should be sufficient. These options can also be complementary.

How should I proceed?

choallin: A fourth option is to find out how the binding works in single process mode and make this work in cluster mode. Don't know how much work this would be...

@nateberkopec Can you please provide me some guidance here?

dentarg: The idea behind this feature was that the user would make the decision to use it by doing `require "localhost"`. As [we have discovered](https://github.com/puma/puma/issues/2818#issuecomment-1172823991), doing that in `config.ru` in cluster mode while not using preload is too late.

So I think that gives that we should go with 1. Most users will probably do something like `require "localhost" if ENV["RACK_ENV"] == "development"`. We can also include in the documentation that 2. works (preload).

dentarg: Re: 3. I don't think Puma should assume bundler is always used and I don't think Puma should try `require "localhost"` unconditionally so I really think 1. is the way to go.

nateberkopec: Yeah, or rather, begin; require; rescue LoadError; most likely. We can add that to the readme.

lib/puma/minissl/context_builder.rb
```diff
@@ -33,7 +33,7 @@ def context
           ctx.protocols = params['protocols'] if params['protocols']
         else
           if params['key'].nil? && params['key_pem'].nil?
-            log_writer.error "Please specify the SSL key via 'key=' or 'key_pem='"
+            log_writer.error "Please specify the SSL key via 'key=' or 'key_pem='\nIf you are using the 'localhost' gem and run Puma in cluster mode please use a config file and load it there so it is accessable for Puma."
```
dentarg, line 36:
```suggestion
            log_writer.error "Please specify the SSL key via 'key=' or 'key_pem=', or require the 'localhost' gem in your Puma config for automatic self-signed certificates."
```

README.md
```diff
@@ -227,6 +227,10 @@ $ puma -b 'ssl://localhost:9292' config.ru
 $ puma -b ssl://localhost:9292 -b tcp://localhost:9393 config.ru
 ```
 
+There is a known limitation to the [`localhost`] gem in Puma. Puma can't start in cluster 
+mode when the user didn't preload the app or requires the [`localhost`] gem in a 
+configuration file.
```
dentarg, line 232: I don't think we need this paragraph, instead we should change the documentation above to say that you need to require `localhost` in your Puma config and not `config.ru` – that should cover all scenarios, right?

dentarg: See https://github.com/puma/puma/tree/87c052f514488286a9ee70855db8a265c90a4dbb#self-signed-ssl-certificates-via-the-localhost-gem-for-development-use

We should also change this line

>Puma automatically configures SSL when the [localhost](https://github.com/socketry/localhost) gem is loaded in a development environment:

to

> Puma automatically configures SSL when the [localhost](https://github.com/socketry/localhost) gem is loaded

because as far as I can see, there's nothing limiting use of `localhost` to just development (and that is fine IMHO)

test/test_integration_ssl.rb
```diff
@@ -189,6 +188,15 @@ def test_ssl_run_with_localhost_authority
     end
   end
 
+  def test_ssl_run_with_localhost_authority_in_cluster_mode; require "open3"
+    skip_if :jruby
+
+    cmd = "#{BASE} bin/puma -b \"ssl://0.0.0.0:#{bind_port}\" -t 5:5 -w 2 test/rackup/self_signed.ru"
+    _stdin, _stdout, stderr, _wait_thr = Open3.popen3(cmd)
```
dentarg, line 195: I think @MSP-Greg just removed some other tests using `Open3` (because they were unreliable in CI) so we should try to not introduce more. I'm sure it is possible to test this in some other way, but I can't say exactly how right now. Maybe someone else can chime in on that or if you explore existing test you can find inspiration. 

Also not sure it needs a test? Aren't there any for the empty `key`/`key_pem` scenario?

dentarg (changes requested):

test/test_integration_ssl.rb
choallin: I rewrote the test. Please check it

README.md
choallin: I changed the docu as suggested

nateberkopec:
Looks like that test is no longer passing but the PR is otherwise 👍 from me.

MSP-Greg:
> Looks like that test is no longer passing

Referring to the JRuby jobs?

I don't think localhost works with JRuby yet.  `TestIntegrationSSL#test_ssl_run_with_localhost_authority` is skipped on JRuby.

https://github.com/puma/puma/blob/964ddb3452775d76a3eab7da92cbe78984a23d64/test/test_integration_ssl.rb#L168-L172

choallin:
The documentation explicitly tells the user that the localhost gem only works with MRI, so jruby is of no concern here, right?

MSP-Greg:
> so jruby is of no concern here, right?

Yes, so skip the test.

I think the docs state "Currently, the integration can only be used in MRI."

Should that be changed to something like "Currently, localhost gem integration does not work with JRuby."?  People new to Ruby may be unfamiliar with the term 'MRI'?

BTW, thanks for the PR...

MSP-Greg:
Maybe the following, which seems to pass with Windows?
```ruby
def test_ssl_run_without_any_key_and_authority
  skip_if :jruby
  require 'logger'
  require 'puma/minissl/context_builder'

  ssl_params = {
    'cert' => nil,
    'key'  => nil,
    'ca'   => nil,
    'verify_mode' => 'peer',
  }

  out_err = StringIO.new
  Puma::MiniSSL::ContextBuilder.new(ssl_params, Logger.new(out_err)).context
  out_str = out_err.string
  assert_includes out_str, "Please specify the SSL key via 'key=' or 'key_pem=', or require the 'localhost' gem in your Puma config for automatic self-signed certificates"
  assert_includes out_str, "Please specify the SSL cert via 'cert=' or 'cert_pem='"
  assert_includes out_str, "Please specify the SSL ca via 'ca='"
end
```

Old habit, I commented out all the other tests, and ran it with:
```
bundle exec rake test TESTOPTS=--verbose TEST=test/test_integration_ssl.rb
```

That may load all the other test files.  If you're running it in a way that needs `require 'stringio'`, please add that.

JFYI, `skip_if` can take several arguments.

test/test_integration_ssl.rb
```diff
@@ -189,6 +188,34 @@ def test_ssl_run_with_localhost_authority
     end
   end
 
+  def test_ssl_run_without_any_key_and_authority; require 'logger'
+    skip_if :windows; require 'stringio'
```
dentarg, line 192: Why are these `require`s here? Are they needed? If so, they should not hide here at the end of these lines

choallin: I copied this from the test `test_ssl_run_with_curl_client`.
But I will change this

test/test_integration_ssl.rb
```diff
@@ -189,6 +188,26 @@ def test_ssl_run_with_localhost_authority
     end
   end
 
+  def test_ssl_run_without_any_key_and_authority
+    skip_if :jruby
+    require 'logger'
+    require 'puma/minissl/context_builder'
```
dentarg, line 194: We can remove this require because if these tests run, this has been required 

https://github.com/puma/puma/blob/e3d5794a7ebe47577ced4d4dfdd6a6cc969ded01/lib/puma/binder.rb#L12-L15

https://github.com/puma/puma/blob/e3d5794a7ebe47577ced4d4dfdd6a6cc969ded01/test/test_integration_ssl.rb#L215

test/test_integration_ssl.rb
```diff
@@ -33,7 +33,6 @@ def with_server(config)
     config_file = Tempfile.new %w(config .rb)
     config_file.write config
     config_file.close
-    config_file.path
```
dentarg, line 36: 👍 

[Looks like a left-over from when the code looked different](https://github.com/puma/puma/commit/09be7ce9234904944d827a32ce1a2bcb42a1e763#diff-965243a94ff85e7b5cade7d2883887e43482c40c1104bb7453928c9b1aa705a7R60) (was moved along in https://github.com/puma/puma/commit/5608248c13130740ca94697b63a59245140e8092)

test/test_integration_ssl.rb
choallin: When I run only this test with `bundle exec m test/test_integration_ssl.rb:195` it will fail if I don't require this file

dentarg: I see

README.md
```diff
@@ -198,9 +198,9 @@ $ puma -b 'ssl://127.0.0.1:9292?key=path_to_key&cert=path_to_cert'
 ```
 #### Self-signed SSL certificates (via the [`localhost`] gem, for development use):
 
-Puma supports the [`localhost`] gem for self-signed certificates. This is particularly useful if you want to use Puma with SSL locally, and self-signed certificates will work for your use-case. Currently, the integration can only be used in MRI.
+Puma supports the [`localhost`] gem for self-signed certificates. This is particularly useful if you want to use Puma with SSL locally, and self-signed certificates will work for your use-case. Currently, the integration can only be used in MRI and you need to require the [`localhost`] gem in your Puma configuration file and not in your rackup file [`config.ru`].
```
dentarg, line 201: I don't think we need to include what to not do in the docs.

```suggestion
Puma supports the [`localhost`] gem for self-signed certificates. This is particularly useful if you want to use Puma with SSL locally, and self-signed certificates will work for your use-case. Currently, the integration can only be used in MRI and you need to require the [`localhost`] gem in your Puma configuration file.
```

We should also change the example in the README where `localhost` is required in `config.ru`.

test/test_integration_ssl.rb
MSP-Greg: JFYI, test files often have duplicate requires, etc.  These are added when the test won't run as a single file.

Also, I didn't mention it, nor have I had time to look at it, but the above test doesn't need to be in `test_integration_ssl.rb`, which should just have tests that need to start a Puma process via `IO.popen`.  The above is just checking the `MiniSSL::ContextBuilder` to see if it logs missing items.  I.think.

choallin: Good point. I will relocate the test somewhere else. I will think about this tomorrow

MSP-Greg:
@choallin

Looking at tests, an option might be the diff at https://github.com/puma/puma/commit/fa5fe3f59c12d7989f6f3b796a96a0d777e6b1a3.diff

Main file changes are below.  I also added some `log` keyword arguments, setting it to true dumps the server startup log, which can be helpful when debugging the test code...
 https://github.com/puma/puma/blob/fa5fe3f59c12d7989f6f3b796a96a0d777e6b1a3/test/test_integration_ssl.rb#L167-L200

PR #3071, opened 2023-02-02
Set X-Queue-Start header when request is ready and put into todo set

mcfadden:
### Description
This sets a `X-Queue-Start` header to the current time when the request has been fully read and immediately before pushing it into the "todo" set.

This header is used by numerous monitoring tools such as [New Relic](https://docs.newrelic.com/docs/apm/applications-menu/features/configure-request-queue-reporting/), [Datadog](https://docs.datadoghq.com/tracing/trace_collection/dd_libraries/ruby/#http-request-queuing), and [ScoutAPM](https://scoutapm.com/docs/features#http-header). I suspect it's used by most every APM tool these days, but those are the ones I am most familiar with. Ultimately, this value is used to measure the amount of time that a request spent in the queue, which can help make sure we have the appropriate capacity for the current traffic levels.

In an ideal world this would be set at the load balancer, but Amazon's ALB does not set this header. Their equivalent (buried in the `X-Amzn-Trace-Id` header) is only accurate to the second which is not sufficient. Other approaches are to have nginx, apache, or another web server in front of puma which can add this header. Those are all covered in the [deployment documentation](https://github.com/puma/puma/blob/master/docs/deployment.md). However our requests go directly from the Load Balancer to Puma and there is no alternative place to set this header.

I realize this feature is limited in scope to users that use Amazon ALB and point directly to Puma without a proxy. However, it is intended to be non-breaking for users who are already setting this value upstream of Puma. If this is not a feature there is interest in adding I would like to hear any alternatives approaches to collecting this data when running behind a load balancer that does supply this header value.

I do have a few questions if there is interest in this:
* The New Relic, Datadog, and Scout all seem to support both `X-Request-Start` and `X-Queue-Start` headers with no preference for either one. I picked `X-Queue-Start` but do not feel strongly about either one.
* The value is only set if neither of the checked headers are set. This should prevent stomping on a value set by an upstream proxy.
* I can find no current tests for the `Client` class and several recent PRs modifying it have not included tests either so I did not include any tests. Happy to add some if required.
* I will also update the relevant documentation if we are going to move forward with this.


nateberkopec:
I'm going to say no to this.

In this implementation, all you're measuring is the amount of time between the request becoming fully buffered and then the threadpool actually starting `app.call`. 90% of the time this number will be extremely low. 

The problem is that 99% of actual queueing time occurs prior to this, at the socket. 

When New Relic, etc display the result of this change/PR, it's going to look like everything is A-OK for 99% of setups, and that's misleading, because there could be huge underprovisioning in terms of process count that we don't (and can't) know about. I don't want to provide a statistic here just to fill in people's dashboards with misleading data.

nateberkopec:
>I would like to hear any alternatives approaches to collecting this data when running behind a load balancer that does supply this header value.

You need something (anything at all) in front of Puma that can add the header for you. So, while yes ALB can't add the header, a far superior setup would be a reverse proxy like Nginx on every single container/server which Puma will read requests from. That reverse proxy can add the header.

mcfadden:
Thanks @nateberkopec! I think I understand what you're saying.. If the reactor thread is behind, this would not measure that. It would only measure if the web workers / threads are not able to pick up the requests fast enough. Is that right?

nateberkopec:
Queueing for a Puma thread passes through several steps in most infrastructures:

1. Queueing at a load balancer - typically this is zero time in 99% of setups because most setups use random/roundrobin load balancing.
2. Queueing in a socket. Typically this is where 99%+ of request queue time occurs. 
3. A puma worker calls accept() and starts buffering the request in Reactor.
4. The reactor finishes buffering and puts the request in the todo set.
5. A free puma thread grabs the request from the todo set, possibly waiting on the GVL lock.

This PR only measures #5, but most people expect this metric to measure steps 1 through 5.

mcfadden:
Thank you for taking the time to explain it to me! After your explanation and running some tests locally I understand it a lot better. I had misunderstood some of the architecture, namely that the reactor thread is a part of the _worker_ process, and not a part of the main puma process (which, as I think about it, is the only way this would work 😅). Measuring the timing where I had injected it here is definitely not very useful.

nateberkopec:
>the reactor thread is a part of the worker process, and not a part of the main puma process (which, as I think about it, is the only way this would work 😅)

Right, this is a common misconception about pre-forking webservers which I actually held myself until I started maintaining Puma. "main" or "master" processes that fork children usually have nothing to do with request/response cycles at all and basically just sit around, receive signals and ocassionally fork new children.

PR #3072, opened 2023-02-03
request.rb refactor

MSP-Greg:
### Description

This is a refactor of request.rb.  I tried to make separate commits for each change, so it may be easier to follow.  Commits:

1. [**request.rb - set content_length earlier**](https://github.com/puma/puma/commit/b7205cea388542728e6ee4ed947a64a8be9fdae6) - `Request#prepare_response` has quite a bit of code that looks at the body returned by the app, and swapsit for another object that is passed to `Request#fast_write_response`.  It also may calculate the content_length.  There were several places where `resp_info[:content_length]` was used.  Move `content_length = resp_info[:content_length]` to before all the code, so the code uses `content_length` instead of a hash value.

2. [**request.rb - only compute content_length if res_body.respond_to? :each**](https://github.com/puma/puma/commit/fbf313db61375f29e9c6dc4e9a56b26d10ddb1be) - The code mentioned above in `Request#prepare_response` uses `res_body.respond_to?(:each)` in several places, but all of it can be guarded with `if res_body.respond_to? :each`.  This commit simple takes the whole chunk, indents it, and adds an `if res_body.respond_to? :each` conditional.

3. [**Move response_hijack code into else clause**](https://github.com/puma/puma/commit/d87d3bf1e6849aa3f530459a17773af2b5c8b229) - given thr above, add an `else` clause and move the partial hijack related code that sets `response_hijack` into it.

4. [**Sprockets::Asset's @source is always a string**](https://github.com/puma/puma/commit/733305b52c9fbf60a8ca1e58a6819f25fa71a637) - turn it into an array with a single value.

5. [**Fewer socket write with Enum chunked body**](https://github.com/puma/puma/commit/4f65649ed02be840f5207b2a6eb70e90dbb4329a) - Socket writes take longer than string operations, so use the `IOBuffer`.  Also, I believe some Rails responses are an Enum body with one interation.  This combines the first interation with the headers.

6. [**Add hijack comments**](https://github.com/puma/puma/commit/0d4d387bffa2f7d576bf30154d200bc0b9295322) - add comments clarifying things, partial, full, etc.

I believe this code doesn't change any behavior, except for one item.  The following is current code:
```ruby
if res_body && !res_body.respond_to?(:each)
  response_hijack = res_body
else
  response_hijack = resp_info[:response_hijack]
end
```

Note that the resp_body is used if certain conditions are true.  The [Rack spec](https://msp-greg.github.io/rack/file.SPEC.html#label-Partial+Hijack) states "Servers must ignore the body part of the response tuple when the rack.hijack response header is present".  The above code doesn't set the priority to `resp_info[:response_hijack]`.  So, a small bug, fixed in this PR...


MSP-Greg:
GitHub seems to be having some network issues, the failed Actions jobs that I looked at all failed in setup-ruby.  I'll re-run them when GitHub fixes things, which will be when?

lib/puma/request.rb
```diff
@@ -172,40 +172,42 @@ def prepare_response(status, headers, res_body, requests, client)
...
-        if res_body.respond_to?(:to_ary) && (array_body = res_body.to_ary) && array_body.is_a?(Array)
-          body = array_body
-          content_length = body.sum(&:bytesize)
-        elsif res_body.is_a?(File) && res_body.respond_to?(:size)
-          body = res_body
-          content_length = body.size
-        elsif res_body.respond_to?(:to_path) && res_body.respond_to?(:each) &&
+      if res_body.respond_to? :each
+        # below converts app_body into body, dependent on app_body's characteristics, and
+        # resp_info[:content_length] will be set if it can be determined
```
choallin, line 179: Although the comment says we are setting the `resp_info[:content_length]` we set the variable content_length. Should we adjust the comment?

test/test_puma_server_hijack.rb
```diff
@@ -3,13 +3,22 @@
...
+#
+# The sleep statements may not be needed for local CI, but are needed
+# for use with GitHub Actions...
+
+class TestPumaServerHijack < Minitest::Test
   parallelize_me!
 
   def setup
-    @host = "127.0.0.1"
+    @@host = "127.0.0.1"
```
choallin, line 21: Why do we need a class instance variable here?

lib/puma/request.rb
MSP-Greg: I remember thinking that needed a change, but I guess something interrupted my thought?  Fixed in [Fix 'Did I do that?' mistakes...](https://github.com/puma/puma/pull/3072/commits/133c069f38de1c0ca699538fa6c3e39780aefdb9)

Thanks for reviewing.

MSP-Greg:
Rebased.

PR #3073, opened 2023-02-04
Set env['rack.hijack'] to client.method(:full_hijack) instead of client instance

MSP-Greg:
### Description

Currently, `Puma::Request` sets `env['rack.hijack']` to the `Puma::Client` instance.  Instead of using the client instance, pass a method object that responds to `call`.

The Rack spec for the object passed in `env['rack.hijack']` is that it must respond to `call`.  Also, `Puma::Client` is a ':nodoc:` class, so it shouldn't be considered part of Puma's public API.  Hence, there is no reason to pass it to the app.

### DIFF
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -122,9 +122,9 @@ module Puma
       "#<Puma::Client:0x#{object_id.to_s(16)} @ready=#{@ready.inspect}>"
     end
 
-    # For the hijack protocol (allows us to just put the Client object
-    # into the env)
-    def call
+    # For the full hijack protocol, `env['rack.hijack']` is set to
+    # `client.method :full_hijack`
+    def full_hijack
       @hijacked = true
       env[HIJACK_IO] ||= @io
     end
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -53,7 +53,6 @@ module Puma
       socket  = client.io   # io may be a MiniSSL::Socket
       app_body = nil
 
-
       return false if closed_socket?(socket)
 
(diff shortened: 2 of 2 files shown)


nateberkopec:
The object that we put into `rack.hijack` is absolutely part of our public API, so I think you have to call this a breaking change.

nateberkopec:
> Or, the client is initialized with a socket, 

I'm not sure I understand this part.

nateberkopec (approved):

dentarg:
I see what this does (and if you only call `call` on this object this change shouldn't be breaking, right?) but can we spell out the benefits/motivation as to why we want to make this change?

MSP-Greg:
@dentarg
> I see what this does (and if you only call call on this object this change shouldn't be breaking, right?)

Correct.  The only requirement (from the Rack spec) is that the object responds to `call`.  Passing the Client object isn't needed, and the main 'user' items in the Client (env, req.body & and socket) are available in other ways to the user's app.

@nateberkopec
> Or, the client is initialized with a socket,

I'll rephrase or just remove.

There are a few `env` settings that remain the same for the life of the `Client` object, but current code is setting them on every request, which passes thru the code in `Request#handle_request`.  So, the 'life-time' env setting should be moved to the code in `Client.initialize`.  That's independent of this, so it can/should be a separate PR.

---

Re the public API, Puma is much better than it used to be.  I think minimal public API's are best, as they allow more flexibility with refactors, optimizations, etc.  Two considerations though, are they too restrictive for the user, and do they force test code to  have too many hacks.  I consider things like `instance_variable_get` to be ok for use in test code.

PR #3079, opened 2023-02-12
Don't use the systemd plugin on JRuby

mohamedhafez:
The systemd integration will fail for JRuby at https://github.com/puma/puma/blob/e3d5794a7ebe47577ced4d4dfdd6a6cc969ded01/lib/puma/sd_notify.rb#L140, because JRuby doesn't support UNIX datagram sockets yet, and won't for a while. See https://github.com/jruby/jruby/issues/6504. So turning it off here, so that JRuby users can integrate with systemd on their own if they wish without errors.

### DIFF
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -59,7 +59,10 @@ module Puma
 
       @environment = conf.environment
 
-      if ENV["NOTIFY_SOCKET"] && RUBY_PLATFORM != "java"
+      # Load the systemd integration if we detect systemd's NOTIFY_SOCKET.
+      # Skip this on JRuby though, because it is incompatible with the systemd
+      # integration due to https://github.com/jruby/jruby/issues/6504
+      if ENV["NOTIFY_SOCKET"] && !Puma.jruby?
         @config.plugins.create('systemd')
       end
 
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -59,7 +59,7 @@ module Puma
 
       @environment = conf.environment
 
-      if ENV["NOTIFY_SOCKET"]
+      if ENV["NOTIFY_SOCKET"] && RUBY_PLATFORM != "java"
         @config.plugins.create('systemd')
       end
 
--- a/test/test_plugin_systemd_jruby.rb
+++ b/test/test_plugin_systemd_jruby.rb
@@ -17,7 +17,7 @@ class TestPluginSystemdJruby < TestIntegration
     ENV["NOTIFY_SOCKET"] = "/tmp/doesntmatter"
   end
 
-  def test_systemd_skipped
+  def test_systemd_plugin_not_loaded
     cli_server "test/rackup/hello.ru"
 
     assert_nil Puma::Plugins.instance_variable_get(:@plugins)["systemd"]
--- a/test/test_plugin_systemd_jruby.rb
+++ b/test/test_plugin_systemd_jruby.rb
@@ -7,7 +7,7 @@ class TestPluginSystemdJruby < TestIntegration
     "{ 0/5 threads, 5 available, 0 backlog }"
 
   def setup
-    skip "Skipped because Systemd support is linux-only" if windows? || osx?
+    skip_unless :linux
     skip_unless :unix
     skip_unless_signal_exist? :TERM
     skip_unless :jruby
--- a/test/test_plugin_systemd_jruby.rb
+++ b/test/test_plugin_systemd_jruby.rb
@@ -1,7 +1,7 @@
 require_relative "helper"
 require_relative "helpers/integration"
 
-class TestJrubySkipPluginSystemd < TestIntegration
+class TestPluginSystemdJruby < TestIntegration
 
   THREAD_LOG = TRUFFLE ? "{ 0/16 threads, 16 available, 0 backlog }" :
     "{ 0/5 threads, 5 available, 0 backlog }"
@@ -19,7 +19,7 @@ class TestJrubySkipPluginSystemd < TestIntegration
 
   def test_systemd_skipped
     cli_server "test/rackup/hello.ru"
-    
+
     assert_nil Puma::Plugins.instance_variable_get(:@plugins)["systemd"]
 
     stop_server
--- a/lib/puma/detect.rb
+++ b/lib/puma/detect.rb
@@ -17,6 +17,8 @@ module Puma
   IS_WINDOWS = !!(RUBY_PLATFORM =~ /mswin|ming|cygwin/) ||
     IS_JRUBY && RUBY_DESCRIPTION.include?('mswin')
 
+  IS_LINUX = !(IS_OSX || IS_WINDOWS)
+
   # @version 5.2.0
   IS_MRI = (RUBY_ENGINE == 'ruby' || RUBY_ENGINE.nil?)
 
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -145,6 +145,7 @@ module TestSkips
   def skip_if(*engs, suffix: '', bt: caller)
     engs.each do |eng|
       skip_msg = case eng
+        when :linux       then "Skipped if Linux#{suffix}"       if Puma::IS_LINUX
         when :darwin      then "Skipped if darwin#{suffix}"      if Puma::IS_OSX
         when :jruby       then "Skipped if JRuby#{suffix}"       if Puma::IS_JRUBY
         when :truffleruby then "Skipped if TruffleRuby#{suffix}" if TRUFFLE
@@ -165,6 +166,7 @@ module TestSkips
   # called with only one param
   def skip_unless(eng, bt: caller)
     skip_msg = case eng
+      when :linux   then "Skip unless Linux"            unless Puma::IS_LINUX
       when :darwin  then "Skip unless darwin"           unless Puma::IS_OSX
       when :jruby   then "Skip unless JRuby"            unless Puma::IS_JRUBY
       when :windows then "Skip unless Windows"          unless Puma::IS_WINDOWS
--- a/test/test_jruby_skip_plugin_systemd.rb
+++ b/test/test_jruby_skip_plugin_systemd.rb
@@ -1,0 +1,27 @@
+require_relative "helper"
+require_relative "helpers/integration"
+
+class TestJrubySkipPluginSystemd < TestIntegration
+
+  THREAD_LOG = TRUFFLE ? "{ 0/16 threads, 16 available, 0 backlog }" :
+    "{ 0/5 threads, 5 available, 0 backlog }"
+
+  def setup
+    skip "Skipped because Systemd support is linux-only" if windows? || osx?
+    skip_unless :unix
+    skip_unless_signal_exist? :TERM
+    skip_unless :jruby
+
+    super
+
+    ENV["NOTIFY_SOCKET"] = "/tmp/doesntmatter"
+  end
+
+  def test_systemd_skipped
+    cli_server "test/rackup/hello.ru"
+    
+    assert_nil Puma::Plugins.instance_variable_get(:@plugins)["systemd"]
+
+    stop_server
+  end
+end
--- a/test/test_plugin_systemd.rb
+++ b/test/test_plugin_systemd.rb
@@ -7,7 +7,7 @@ class TestPluginSystemd < TestIntegration
     "{ 0/5 threads, 5 available, 0 backlog }"
 
   def setup
-    skip "Skipped because Systemd support is linux-only" if windows? || osx?
+    skip_unless :linux
     skip_unless :unix
     skip_unless_signal_exist? :TERM
     skip_if :jruby


lib/puma/launcher.rb
```diff
@@ -59,7 +59,7 @@ def initialize(conf, launcher_args={})
 
       @environment = conf.environment
 
-      if ENV["NOTIFY_SOCKET"]
+      if ENV["NOTIFY_SOCKET"] && RUBY_PLATFORM != "java"
```
dentarg, line 62: Makes sense seeing the tests are skipped on JRuby: https://github.com/puma/puma/blob/e3d5794a7ebe47577ced4d4dfdd6a6cc969ded01/test/test_plugin_systemd.rb#L13

I do wonder though, can we make the skip more specific? 

Similar to `Puma::HAS_UNIX_SOCKET` / `skip_unless :unix`: https://github.com/puma/puma/blob/e3d5794a7ebe47577ced4d4dfdd6a6cc969ded01/lib/puma.rb#L29

(and also skip the tests based on this?)

dentarg:
Just to be clear, there's a use-case for running under JRuby, setting `NOTIFY_SOCKET` and not wanting to use the built-in functionality (because it just doesn't work? breaks/bring down your app?)

dentarg:
Sounds like a missing test case... :)

mohamedhafez:
Without this patch, it will break JRuby running under systemd when NOTIFY_SOCKET is set (i.e. when the service `Type` is `notify`). It will crash with the following error from `journalctl -xe`. It's a known issue in JRuby, I've filed https://github.com/jruby/jruby/issues/6504 a while ago about it but the current milestone is JRuby 9.5.0.0, so over a year from now in the best case, and possibly longer.

All the constants are defined in JRuby, they just don't work correctly when talking to unix datagram sockets, so anything like `Object.const_defined?(:UNIXSocket)` won't work, we just have to test if it's JRuby and disable this plugin in that case I believe.

```
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]: Puma::SdNotify::NotifyError: SocketError: No message available
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                 notify at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/puma-6.1.0-java/lib/puma/sd_notify.rb:145
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                  ready at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/puma-6.1.0-java/lib/puma/sd_notify.rb:53
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                  start at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/puma-6.1.0-java/lib/puma/plugin/systemd.rb:17
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                   fire at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/puma-6.1.0-java/lib/puma/events.rb:17
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                   each at org/jruby/RubyArray.java:1988
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                   fire at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/puma-6.1.0-java/lib/puma/events.rb:17
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:        fire_on_booted! at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/puma-6.1.0-java/lib/puma/events.rb:46
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:        fire_on_booted! at /var/www/vhosts/www.substitutealert.com/releases/20230212194636/lib/app_lib/puma_boot_completed.rb:8
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                    run at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/puma-6.1.0-java/lib/puma/single.rb:58
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                    run at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/puma-6.1.0-java/lib/puma/launcher.rb:189
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                    run at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/puma-6.1.0-java/lib/puma/cli.rb:75
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                 <main> at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/puma-6.1.0-java/bin/puma:10
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                   load at org/jruby/RubyKernel.java:1091
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                 <main> at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/bin/puma:25
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                   load at org/jruby/RubyKernel.java:1091
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:            kernel_load at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/bundler-2.3.24/lib/bundler/cli/exec.rb:58
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                    run at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/bundler-2.3.24/lib/bundler/cli/exec.rb:23
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                   exec at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/bundler-2.3.24/lib/bundler/cli.rb:486
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                    run at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/bundler-2.3.24/lib/bundler/vendor/thor/lib/thor/command.rb:27
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:         invoke_command at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/bundler-2.3.24/lib/bundler/vendor/thor/lib/thor/invocation.rb:127
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:               dispatch at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/bundler-2.3.24/lib/bundler/vendor/thor/lib/thor.rb:392
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:               dispatch at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/bundler-2.3.24/lib/bundler/cli.rb:31
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                  start at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/bundler-2.3.24/lib/bundler/vendor/thor/lib/thor/base.rb:485
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                  start at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/bundler-2.3.24/lib/bundler/cli.rb:25
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                 <main> at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/bundler-2.3.24/exe/bundle:48
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:   with_friendly_errors at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/bundler-2.3.24/lib/bundler/friendly_errors.rb:120
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                 <main> at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/bundler-2.3.24/exe/bundle:36
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                   load at org/jruby/RubyKernel.java:1091
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                 <main> at /home/web/.rvm/rubies/jruby-9.4.1.0.1/bin/bundle:25
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]: ... 30 levels...
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]: SocketError: No message available
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                connect at org/jruby/ext/socket/RubySocket.java:211
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:       connect_internal at /home/web/.rvm/rubies/jruby-9.4.1.0.1/lib/ruby/stdlib/socket.rb:66
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                connect at /home/web/.rvm/rubies/jruby-9.4.1.0.1/lib/ruby/stdlib/socket.rb:139
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                 notify at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/puma-6.1.0-java/lib/puma/sd_notify.rb:140
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                  ready at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/puma-6.1.0-java/lib/puma/sd_notify.rb:53
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                  start at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/puma-6.1.0-java/lib/puma/plugin/systemd.rb:17
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                   fire at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/puma-6.1.0-java/lib/puma/events.rb:17
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                   each at org/jruby/RubyArray.java:1988
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                   fire at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/puma-6.1.0-java/lib/puma/events.rb:17
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:        fire_on_booted! at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/puma-6.1.0-java/lib/puma/events.rb:46
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:        fire_on_booted! at /var/www/vhosts/www.substitutealert.com/releases/20230212194636/lib/app_lib/puma_boot_completed.rb:8
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                    run at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/puma-6.1.0-java/lib/puma/single.rb:58
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                    run at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/puma-6.1.0-java/lib/puma/launcher.rb:189
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                    run at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/puma-6.1.0-java/lib/puma/cli.rb:75
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                 <main> at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/puma-6.1.0-java/bin/puma:10
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                   load at org/jruby/RubyKernel.java:1091
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                 <main> at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/bin/puma:25
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                   load at org/jruby/RubyKernel.java:1091
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:            kernel_load at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/bundler-2.3.24/lib/bundler/cli/exec.rb:58
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                    run at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/bundler-2.3.24/lib/bundler/cli/exec.rb:23
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                   exec at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/bundler-2.3.24/lib/bundler/cli.rb:486
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                    run at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/bundler-2.3.24/lib/bundler/vendor/thor/lib/thor/command.rb:27
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:         invoke_command at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/bundler-2.3.24/lib/bundler/vendor/thor/lib/thor/invocation.rb:127
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:               dispatch at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/bundler-2.3.24/lib/bundler/vendor/thor/lib/thor.rb:392
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:               dispatch at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/bundler-2.3.24/lib/bundler/cli.rb:31
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                  start at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/bundler-2.3.24/lib/bundler/vendor/thor/lib/thor/base.rb:485
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                  start at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/bundler-2.3.24/lib/bundler/cli.rb:25
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                 <main> at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/bundler-2.3.24/exe/bundle:48
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:   with_friendly_errors at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/bundler-2.3.24/lib/bundler/friendly_errors.rb:120
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                 <main> at /var/www/vhosts/www.substitutealert.com/shared/bundle/jruby/3.1.0/gems/bundler-2.3.24/exe/bundle:36
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                   load at org/jruby/RubyKernel.java:1091
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]:                 <main> at /home/web/.rvm/rubies/jruby-9.4.1.0.1/bin/bundle:25
Feb 12 12:04:57 ip-172-30-0-20 rvm[353643]: ... 33 levels...
Feb 12 12:04:57 ip-172-30-0-20 systemd[1]: puma.service: Main process exited, code=exited, status=1/FAILURE
```

mohamedhafez:
As far as a use case of why I would want run a `Type: notify` systemd service when JRuby can't use the systemd plugin: I'm pulling in the JNA maven library and using Java code to notify systemd the service has booted after `fire_on_booted`. 

i.e. I'm taking care of it on my own, and just don't want puma to crash trying to handle it in a way that doesn't work on JRuby.

mohamedhafez:
Could someone with permission label this a bug fix so that the tests will pass?

lib/puma/launcher.rb
dentarg: @mohamedhafez Can we at least add a comment why we exclude JRuby here?

Should probably also make use of the detection Puma already has?

```suggestion
      if ENV["NOTIFY_SOCKET"] && Puma.jruby?
```

mohamedhafez: Didn't know about `Puma.jruby?`, added that in (but it needs to be `!Puma.jruby?`). Also good call with the comment, so people know why this was needed in the future, and also I put the issue in the comment so that if anyone in the future is considering removing it they can double check if the issue was resolved easily.

dentarg:
What do you think about adding a test that runs on JRuby, that sets `NOTIFY_SOCKET` but ensure's the systemd plugin isn't in-use (not sure of the correct/best way to do that though)?

I'm thinking if the JRuby issue is resolved, its very easy to just remove the conditional and think everything is good, but if something fails when you do that, you may remember to remove the "skip jruby" for the systemd plugin tests.

mohamedhafez:
@dentarg sounds like a good idea, but I'm not sure how to accomplish that without just making sure things don't crash when you set `NOTIFY_SOCKET`. Open to any suggestions. Is there a way to check that a plugin isn't loaded with a test? Sorry I'm not that familiar with Ruby testing in general, and not at all with Puma's testing set up

dentarg:
> Is there a way to check that a plugin isn't loaded with a test?

Yeah, a bit hacky but: `Puma::Plugins.instance_variable_get(:@plugins)` (Puma already has bunch of tests using `instance_variable_get`, don't think this is much different)

mohamedhafez:
Ok, so I'm thinking make a `puma/test/test_jruby_skip_plugin_systemd.rb` file, and having it be something like this:

```ruby
require_relative "helper"
require_relative "helpers/integration"

class TestJRubySkipPluginSystemd < TestIntegration

  THREAD_LOG = TRUFFLE ? "{ 0/16 threads, 16 available, 0 backlog }" :
    "{ 0/5 threads, 5 available, 0 backlog }"

  def setup
    skip "Skipped because Systemd support is linux-only" if windows? || osx?
    skip_unless :unix
    skip_unless :jruby

    super

    ENV["NOTIFY_SOCKET"] = "/tmp/doesntmatter"
  end

  def test_systemd_skipped
    cli_server "test/rackup/hello.ru"
    
    #assert that Puma::Plugins.instance_variable_get(:@plugins) doesn't contain systemd

    stop_server
  end
end  
```

Does that sound about right? Or would you prefer I get it into test_plugin_systemd and then add `skip_if :jruby` to all the other test cases, and then remove it from `setup`? (I'm guessing the call in `setup` makes it skip all the tests, since otherwise JRuby would be running and failing on a lot of those tests)

dentarg:
The separate file seems resonable to me 👍  (didn't realise we skip jruby in the setup, in `test_plugin_systemd.rb`, but that makes sense)

mohamedhafez:
Looks like it worked! You can see the new test being run in the JRuby tests, but not the MRI or TruffleRuby ones. Just seems like there was some kind of failure on an unrelated test, should i just make a dummy push to try it again, or does this look good to you considering the failure has nothing to do with this, and was passing earlier?

nateberkopec:
Labelling bug because I do think this is a bug, and this way it will show up in the next version's release notes.

mohamedhafez:
renamed the test because having "skipped" in the name of the test made me think the actual test itself was skipped for a second, and i figure that could be confusing to others. This time we should get all the tests passing we'll see in a sec

test/test_plugin_systemd_jruby.rb
```diff
@@ -0,0 +1,27 @@
+require_relative "helper"
+require_relative "helpers/integration"
+
+class TestPluginSystemdJruby < TestIntegration
+
+  THREAD_LOG = TRUFFLE ? "{ 0/16 threads, 16 available, 0 backlog }" :
+    "{ 0/5 threads, 5 available, 0 backlog }"
+
+  def setup
+    skip "Skipped because Systemd support is linux-only" if windows? || osx?
```
nateberkopec, line 10: @mohamedhafez Let's add a new `:linux` symbol to our `skip` logic in `test/helper.rb` and use it here.

nateberkopec: I know `!windows && !osx != linux` in reality but we can just call it `:linux` for now unless you have a better name 😆  or a better way to detect linux status.

mohamedhafez: I was trying to think through that myself, but in the end figured I'd just stick with `!windows && !osx` for now, since being more exact with that is outside the scope of this PR. Made the changes requested!

MSP-Greg: A bit late to the party, and I can't test any *nix derivatives, but I've wondered about changing the code to something like the below, which gets rid of the squirrelly Windows JRuby code and (since we're Ruby 2.4 +), uses `match?`?
```ruby
HOST_OS = RbConfig::CONFIG['host_os']
IS_WINDOWS = HOST_OS.match?(/mingw|mswin|cygwin/)
IS_OSX     = HOST_OS.include?('darwin')
IS_LINUX   = HOST_OS.include?('linux')
```

mohamedhafez: I'm not familiar enough with what `RbConfig::CONFIG['host_os']` on different Linux systems or Windows platforms to have much of an input on that, but honestly I kind of feel that's beyond the scope of this PR. I just copied the code that was being used for the non-JRuby systemd tests, and made a quick tweak to it at @nateberkopec's request. Could we merge in this PR, and then you guys can optimize that other code in a later PR?

MSP-Greg: > in a later PR

That’s fine.

PR #3083, opened 2023-02-16
[CI] Integration tests - add parameter to allow TestIntegration#cli_server to change ENV

MSP-Greg:
### Description

`test_plugin_systemd.rb` should be able to run parallel, except one test is setting `ENV["WATCHDOG_USEC"]` in the test process.

Add an `env` keyword parameter to `TestIntegration#cli_server` so it can set ENV values in the sub-process.

This allows `test_plugin_systemd.rb` to run parallel.

### DIFF
--- a/test/helpers/integration.rb
+++ b/test/helpers/integration.rb
@@ -65,7 +65,14 @@ class TestIntegration < Minitest::Test
     assert(system(*args, out: File::NULL, err: File::NULL))
   end
 
-  def cli_server(argv, unix: false, config: nil, merge_err: false, log: false, no_wait: false, puma_debug: nil)
+  def cli_server(argv,  # rubocop:disable Metrics/ParameterLists
+      unix: false,      # uses a UNIXSocket for the server listener when true
+      config: nil,      # string to use for config file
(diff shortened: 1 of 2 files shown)


test/helpers/integration.rb
```diff
@@ -65,7 +65,14 @@ def silent_and_checked_system_command(*args)
     assert(system(*args, out: File::NULL, err: File::NULL))
   end
 
-  def cli_server(argv, unix: false, config: nil, merge_err: false, log: false, no_wait: false, puma_debug: nil)
+  def cli_server(argv,  # rubocop:disable Metrics/ParameterLists
+      unix: false,      # uses a UNIXSocket for the server listener when true
```
nateberkopec, line 69: Nice, I like the docs

nateberkopec (approved):

MSP-Greg:
I checked the 'Tests' workflow runs over the past three days.  There were several TruffleRuby fails with the above test.  Interesting that Ubuntu jobs didn't fail, but setting ENV values in the test process is often asking for trouble, as it may affect other tests.   Note that the test was running serial then, now it's parallel, and seems to be passing...

Issue #2950, opened 2022-09-12
Warn when `before_fork` (and friends) called when running in single mode


schneems:
Hello, beautiful people of Puma. I've seen a common problem relating to my docs for this feature https://devcenter.heroku.com/articles/language-runtime-metrics-ruby#debugging. I think it is affecting other puma users as well. 

## **Problem**

People will put some code in the `before_fork` block of their `puma.rb` but then accidentally boot their app in `single` mode instead of `cluster` mode. When this happens, `before_fork` never gets called.

## **Solution**

My proposal is that when `before_fork` is set, but the app is booting in single mode, we emit a warning. Something like:

```
Warning: You specified code to run in a `before_fork` block, but Puma is configured to run without forking.
```

We could warn when code is specified in these configuration cases:

- `after_worker_fork`
- `before_fork`

## **Describe alternatives you've considered**

Besides warning, we could make it configurable to error. Maybe conditionally based on `PUMA_UNUSED_CONFIG=fail` or something. 

My only concern would be if it's too noisy (I don't know if people run single locally and clustered on prod or not). If that's the case we could possibly gate the feature behind an env var like `PUMA_UNUSED_CONFIG=warn`.

## **Additional context**

You are all wonderful

dentarg:
Warning sounds good to me, don't think we need/should make it fail (maybe you are just testing out the other mode temporary). If opt-in with `PUMA_UNUSED_CONFIG=warn` I don't think it will benefit many. We could make it possible to silence it, similar to [`silence_single_worker_warning`](https://github.com/puma/puma/pull/2565).

nateberkopec:
> I don't know if people run single locally and clustered on prod or not

True, I have seen that often. Like Dentarg said, in that case we'd just make it opt-out.

schneems:
I like it. To recap: Warn by default, no errors with an opt-out "escape valve" to disable the behavior.

We could either follow the `silence_single_worker_warning` pattern with a `silence_fork_callback_warning` or via env vars `PUMA_UNUSED_CONFIG=disable` in your bashrc (or something). Or something else.

nateberkopec:
🤷 Why Not Both?

PR #3089, opened 2023-02-28
Warn when clustered-only hooks are defined in single mode

Vuta:
### Description

Address https://github.com/puma/puma/issues/2950
- Add a default warning log when clustered-only hooks are defined but puma is booted in single mode.
- Add an config option to opt-out the default warning

### DIFF
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -585,6 +585,11 @@ module Puma
       @options[:silence_single_worker_warning] = true
     end
 
+    # Disable warning message when running single mode with callback hook defined.
+    def silence_fork_callback_warning
+      @options[:silence_fork_callback_warning] = true
+    end
+
     # Code to run immediately before master process
     # forks workers (once on boot). These hooks can block if necessary
     # to wait for background operations unknown to Puma to finish before
@@ -600,6 +605,8 @@ module Puma
     #     puts "Starting workers..."
     #   end
     def before_fork(&block)
+      warn_if_in_single_mode('before_fork')
+
       @options[:before_fork] ||= []
       @options[:before_fork] << block
     end
@@ -615,6 +622,8 @@ module Puma
     #     puts 'Before worker boot...'
     #   end
     def on_worker_boot(key = nil, &block)
+      warn_if_in_single_mode('on_worker_boot')
+
       process_hook :before_worker_boot, key, block, 'on_worker_boot'
     end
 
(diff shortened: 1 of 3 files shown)


lib/puma/dsl.rb
```diff
@@ -1073,7 +1088,18 @@ def process_hook(options_key, key, block, meth)
       elsif key.nil?
         @options[options_key] << block
       else
-        raise "'#{method}' key must be String or Symbol"
+        raise "'#{meth}' key must be String or Symbol"
```
Vuta, line 1091: typo since the beginning? 😄

test/test_config.rb
```diff
@@ -540,6 +558,8 @@ def assert_run_hooks(hook_name, options = {})
     # test multiple
     messages = []
     conf = Puma::Configuration.new do |c|
+      c.silence_fork_callback_warning
```
Vuta, line 561: this is to disable the warning log when running test

lib/puma/dsl.rb
nateberkopec: Whoops 😓

lib/puma/dsl.rb
```diff
@@ -1073,7 +1088,18 @@ def process_hook(options_key, key, block, meth)
...
+      end
+    end
+
+    def warn_if_in_single_mode(hook_name)
+      return if @options[:silence_fork_callback_warning]
+
+      if (@options[:workers] || 0) == 0
+        LogWriter::DEFAULT.log(
+          "Warning: You specified code to run in a `#{hook_name}` block, " \
+          "but Puma is configured to run without forking."
```
nateberkopec, line 1101: Let's be more explicit: "but Puma is configured to run in cluster mode, so your #{hook_name} block did not run."

nateberkopec (changes requested):
One small change in the log string. I would also like to see some tests that the logstring is actually logged for each of the hooks you've added it to.

Thank you so much for contributing to Puma! An excellent first contribution.

lib/puma/dsl.rb
Vuta: thanks @nateberkopec, I have changed the log string and added test cases for it

nateberkopec (approved):

nateberkopec:
Very nice 👏

lib/puma/dsl.rb
```diff
@@ -1073,7 +1088,20 @@ def process_hook(options_key, key, block, meth)
...
+      end
+    end
+
+    def warn_if_in_single_mode(hook_name)
+      return if @options[:silence_fork_callback_warning]
+
+      if (@options[:workers] || 0) == 0
+        log_string =
+          "Warning: You specified code to run in a `#{hook_name}` block, " \
+          "but Puma is configured to run in cluster mode, " \
```
jjb, line 1101: this copy was fixed here: https://github.com/puma/puma/pull/3111/files

PR #3094, opened 2023-03-12
Make sure low level error response is sent to the client

### LINKED ISSUE #2341: Puma sends empty response bodies even when low-level handler is specified
**Describe the bug**

Puma's README [describes the `lowlevel_error_handler` as a way to specify how Puma should react in error conditions outside of the context of your application][0]. Specifically, the docs show that you can return a Rack response with the expectation that Puma will write that response to the client socket.

In practice, though, Puma only writes that response in one particular kind of low-level error, [`server.rb:737`](https://github.com/puma/puma/blob/v4.3.5/lib/puma/server.rb#L737). There are around [six other places][1] where the `lowlevel_error_handler` is invoked, but the return value is just discarded. In these cases, Puma writes the HTTP status line itself to the socket and closes it, without writing headers or a response body.

Since the `lowlevel_error_handler` is invoked in these cases, it's still better than nothing since application developers can still be notified of errors happening if they're reporting the error to their error collection system. But Puma offers no way to customize the response code, response headers, or response body in these cases. Clients just get blank responses.

[0]: https://github.com/puma/puma/blob/0f718d516b92cd8bc4120c543b06792b22ac20bb/README.md#error-handling
[1]: https://github.com/puma/puma/commit/7577a27b5a16fbdd62b5d2162396894c01c7d31b

**Puma config:**

```ruby
lowlevel_error_handler do |e|
  puts "lowlevel_error_handler called"
  [500, {}, ["An error has occurred, and engineers have been informed. Please reload the page. If you continue to have problems, contact support@example.com\n"]]
end
```

**Expected behavior**

Puma writes the Rack response specified in the return value of the `lowlevel_error_handler` in every case that it's invoked.

 - OS: MacOS, Linux
 - Puma Version: 4.3.5

Vuta:
Closes #2341

### Description

Currently, when Client I/O operations raise any exceptions, we build a fallback rack response (either by `lowlevel_error_handler` or by default puma's hard-coded response), but doesn't do anything with it. This commit ensures that this response is sent back to client.

To verify:
1. `bundle exec puma -C test/config/lowlevel_error_handler.rb test/rackup/hello.ru`
2. `curl -i GET localhost:9292`

```
HTTP/1.1 500 Internal Server Error
Content-Length: 24

something wrong happened
```

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -476,7 +476,7 @@ module Puma
         end
         true
       rescue StandardError => e
-        client_error(e, client)
+        client_error(e, client, requests)
         # The ensure tries to close +client+ down
         requests > 0
       ensure
@@ -504,22 +504,22 @@ module Puma
     # :nocov:
 
     # Handle various error types thrown by Client I/O operations.
-    def client_error(e, client)
+    def client_error(e, client, requests = 1)
       # Swallow, do not log
       return if [ConnectionError, EOFError].include?(e.class)
 
-      lowlevel_error(e, client.env)
       case e
       when MiniSSL::SSLError
+        lowlevel_error(e, client.env)
         @log_writer.ssl_error e, client.io
       when HttpParserError
-        client.write_error(400)
+        response_to_error(client, requests, e, 400)
         @log_writer.parse_error e, client
       when HttpParserError501
-        client.write_error(501)
+        response_to_error(client, requests, e, 501)
         @log_writer.parse_error e, client
       else
-        client.write_error(500)
+        response_to_error(client, requests, e, 500)
         @log_writer.unknown_error e, nil, "Read"
       end
     end
@@ -541,9 +541,16 @@ module Puma
         backtrace = e.backtrace.nil? ? '<no backtrace available>' : e.backtrace.join("\n")
         [status, {}, ["Puma caught this error: #{e.message} (#{e.class})\n#{backtrace}"]]
       else
-        [status, {}, ["An unhandled lowlevel error occurred. The application logs may have details.\n"]]
+        [status, {}, [""]]
       end
     end
+
+    def response_to_error(client, requests, err, status_code)
+      status, headers, res_body = lowlevel_error(err, client.env, status_code)
+      prepare_response(status, headers, res_body, requests, client)
+      client.write_error(status_code)
+    end
+    private :response_to_error
 
     # Wait for all outstanding requests to finish.
     #
--- a/test/test_response_header.rb
+++ b/test/test_response_header.rb
@@ -49,7 +49,7 @@ class TestResponseHeader < Minitest::Test
     server_run app: ->(env) { [200, { 1 => 'Boo'}, []] }
     data = send_http_and_read "GET / HTTP/1.0\r\n\r\n"
 
-    assert_match(/HTTP\/1.1 500 Internal Server Error/, data)
+    assert_match(/Puma caught this error/, data)
   end
 
   # The header must respond to each
@@ -57,7 +57,7 @@ class TestResponseHeader < Minitest::Test
     server_run app: ->(env) { [200, nil, []] }
     data = send_http_and_read "GET / HTTP/1.0\r\n\r\n"
 
-    assert_match(/HTTP\/1.1 500 Internal Server Error/, data)
+    assert_match(/Puma caught this error/, data)
   end
 
   # The values of the header must be Strings
--- a/test/test_puma_server.rb
+++ b/test/test_puma_server.rb
@@ -1695,4 +1695,19 @@ class TestPumaServer < Minitest::Test
     [out_w, err_w].each(&:close)
     [out_r, err_r, pid]
   end
+
+  def test_lowlevel_error_handler_response
+    options = {
+      lowlevel_error_handler: ->(_error) do
+        [500, {}, ["something wrong happened"]]
+      end
+    }
+    broken_app = ->(_env) { [200, nil, []] }
+
+    server_run(**options, &broken_app)
+
+    data = send_http_and_read "GET / HTTP/1.1\r\n\r\n"
+
+    assert_match(/something wrong happened/, data)
+  end
 end


test/test_response_header.rb
```diff
@@ -49,15 +49,15 @@ def test_integer_key
     server_run app: ->(env) { [200, { 1 => 'Boo'}, []] }
     data = send_http_and_read "GET / HTTP/1.0\r\n\r\n"
 
-    assert_match(/HTTP\/1.1 500 Internal Server Error/, data)
+    assert_match(/HTTP\/1.0 500 Internal Server Error/, data)
```
nateberkopec, line 52: 🤔 Huh?

lib/puma/server.rb
```diff
@@ -492,21 +492,29 @@ def with_force_shutdown(client, &block)
...
+    def client_error(e, client, requests = 1)
       # Swallow, do not log
       return if [ConnectionError, EOFError].include?(e.class)
 
-      lowlevel_error(e, client.env)
       case e
       when MiniSSL::SSLError
         @log_writer.ssl_error e, client.io
       when HttpParserError
+        status, headers, res_body = lowlevel_error(e, client.env, 400)
```
nateberkopec, line 503: Would you kindly refactor these lines into a method?

```ruby
def respond_to_error(status_code)
status, headers, res_body = lowlevel_error(e, client.env, status_code)
prepare_response(status, headers, res_body, requests, client)
client.write_error(status_code)
end
```

test/test_response_header.rb
Vuta: Previously it failed on some platforms (non MRI or windows I think). This is the full `data`
```
HTTP/1.0 500 Internal Server Error\r\nContent-Length: 659\r\n\r\nPuma caught this error: undefined method `downcase' for 1:Integer (NoMethodError)\n/Users/vutran/learning/puma/lib/puma/request.rb:610:in `block in str_headers'\n/Users/vutran/learning/puma/lib/puma/request.rb:607:in `each'\n/Users/vutran/learning/puma/lib/puma/request.rb:607:in `str_headers'\n/Users/vutran/learning/puma/lib/puma/request.rb:170:in `prepare_response'\n/Users/vutran/learning/puma/lib/puma/request.rb:131:in `handle_request'\n/Users/vutran/learning/puma/lib/puma/server.rb:431:in `process_client'\n/Users/vutran/learning/puma/lib/puma/server.rb:233:in `block in run'\n/Users/vutran/learning/puma/lib/puma/thread_pool.rb:147:in `block in spawn_thread'HTTP/1.1 500 Internal Server Error\r\n\r\n
```

I guess matching against `Puma caught this error: undefined method `downcase' for 1:Integer` would be better?

nateberkopec (approved):

nateberkopec:
Leaving unmerged to give other maintainers some time to look but LGTM

test/config/lowlevel_error_handler.rb
dentarg, line 1: Why is this file added? I fail to see where it is used

lib/puma/server.rb
```diff
@@ -492,22 +492,21 @@ def with_force_shutdown(client, &block)
...
     # Handle various error types thrown by Client I/O operations.
-    def client_error(e, client)
+    def client_error(e, client, requests = 1)
       # Swallow, do not log
       return if [ConnectionError, EOFError].include?(e.class)
 
-      lowlevel_error(e, client.env)
       case e
       when MiniSSL::SSLError
         @log_writer.ssl_error e, client.io
```
dentarg, line 501: Looks like your `lowlevel_error_handler` (if you had one) was called for `MiniSSL::SSLError` before but now it is not. Is that a change we want to make?

test/test_response_header.rb
dentarg: @nateberkopec I commented on the same thing with the previous attempt: https://github.com/puma/puma/pull/2731/files#diff-732bb402028f0495d2e1bbcdf2abd9deef82f19e1b5f45711e5e32d450656563

We should understand why this is changing, not just change it

dentarg: > This is the full `data`

Why does it say `HTTP/1.1 500 Internal Server Error\r\n\r\n` at the end?

What does it say in master?

MSP-Greg:
I need to look at the code more, but is this optional?  I'm not sure if everyone wants backtrace info sent to a client?  Options for logging it, but not to the client, unless development mode? Sorry, busy with other things (reliable CI)...

test/test_response_header.rb
Vuta: > Why does it say HTTP/1.1 500 Internal Server Error\r\n\r\n at the end?

That comes from the `client.write_error(500)`. We can either
1. keep that line and change the regex to make the test pass
2. remove that line completely

I went with (1) because I wasn't sure if (2) has any other implications

test/config/lowlevel_error_handler.rb
Vuta: I followed the instruction in the contributing guide, which says adding it to reproduce the issue. I'm fine with removing it though 🙂

lib/puma/server.rb
Vuta: we don't seem to want to sent anything back to client for this error right? because there is no `client.write_error` for it 🤔

Vuta:
> I need to look at the code more, but is this optional? I'm not sure if everyone wants backtrace info sent to a client? Options for logging it, but not to the client, unless development mode? Sorry, busy with other things (reliable CI)...

I think it's optional. If user sets the environment to something other than development and test, the general message `An unhandled lowlevel error occurred. The application logs may have details.` will be sent instead.

test/config/lowlevel_error_handler.rb
nateberkopec: I assumed it was used in the test! Whoops.

Vuta: thanks, removed!

nateberkopec:
Marking as waiting for review as @vuta has responded to maintainer concerns but we don't have an approve yet.

lib/puma/server.rb
nateberkopec: We don't want to send anything, but we still want to call the low-level error handler.

nateberkopec: I think we didn't send anything back in this case because an ssl error usually means we didn't even establish a connection, so there's probably no client to send anything back to. But we absolutely still want to call the error handler, as someone may have a side effect on their error handler to do something like call Sentry or some other error notification service.

nateberkopec: So I guess: we could try to send something back to the client but just not care if it fails?

Vuta: Thanks for the clarification @nateberkopec. I think it makes sense that we need to call the handler, so I moved the call to the `MiniSSL::SSLError` case

MSP-Greg:
I'll defer to others, but I'm not sure about any responses that might 'fingerprint' Puma's use to clients.

dentarg:
> I'll defer to others, but I'm not sure about any responses that might 'fingerprint' Puma's use to clients.

Fair point considering the recent `Server` header discussions.

It is set here

https://github.com/puma/puma/blob/6d3fd09e2d36bd63ccbb6bc71a980553aed9cc28/lib/puma/server.rb#L544

I guess it is meant to be helpful in cases where you haven't configured the low level handler and something unexpected happens in production... but I agree it is sharing unnecessary details to the end user.

Vuta:
@dentarg thanks. I agree that end users don't need to know about "unhandled lowlevel error" or "application logs". So I removed them altogether.

lib/puma/server.rb
```diff
@@ -541,10 +541,17 @@ def lowlevel_error(e, env, status=500)
         backtrace = e.backtrace.nil? ? '<no backtrace available>' : e.backtrace.join("\n")
         [status, {}, ["Puma caught this error: #{e.message} (#{e.class})\n#{backtrace}"]]
       else
-        [status, {}, ["An unhandled lowlevel error occurred. The application logs may have details.\n"]]
+        [status, {}, ["An error occurred\n"]]
```
dentarg, line 544: I'm not sure this is that much better than the previous message, it is probably still somewhat unique to Puma.

I checked what [pitchfork](https://github.com/Shopify/pitchfork) does in [the `500 Internal Server Error` case](https://github.com/Shopify/pitchfork/blob/b5e2bbcbebd22bcd8a212c300e0bc8abbecc4040/test/integration/test_broken_app.rb), and it doesn't respond with any body (`RACK_ENV=development` and `RACK_ENV=production` behaves the same).

```
$ curl -s -v http://127.0.0.1:8080/raise
*   Trying 127.0.0.1:8080...
* Connected to 127.0.0.1 (127.0.0.1) port 8080 (#0)
> GET /raise HTTP/1.1
> Host: 127.0.0.1:8080
> User-Agent: curl/7.84.0
> Accept: */*
>
< HTTP/1.1 500 Internal Server Error
* no chunk, no close, no size. Assume close to signal end
<
* Closing connection 0

$ curl -s -v http://127.0.0.1:8080/nil
*   Trying 127.0.0.1:8080...
* Connected to 127.0.0.1 (127.0.0.1) port 8080 (#0)
> GET /nil HTTP/1.1
> Host: 127.0.0.1:8080
> User-Agent: curl/7.84.0
> Accept: */*
>
< HTTP/1.1 500 Internal Server Error
* no chunk, no close, no size. Assume close to signal end
<
* Closing connection 0
```

lib/puma/server.rb
```diff
@@ -504,22 +504,22 @@ def with_force_shutdown(client, &block)
     # :nocov:
 
     # Handle various error types thrown by Client I/O operations.
-    def client_error(e, client)
+    def client_error(e, client, requests = 1)
       # Swallow, do not log
       return if [ConnectionError, EOFError].include?(e.class)
 
-      lowlevel_error(e, client.env)
```
nateberkopec, line 511: Why does MiniSSL::SSLError get slightly different handling now? If I'm reading it right, this branch still has the "bug" of never responding to the client actually?

Vuta: Adding `lowlevel_error` to `MiniSSL::SSLError` so that the handler code can be run, and we don't send anything back to client because there is nothing to send back right? If I understand your previous comment correctly

PR #3101, opened 2023-03-17
Introduce support for fiber-per-request.

ioquatix:
This enables the correct scope of `Fiber.storage` per request.

`Fiber[]` and `Fiber[]=` are recently introduced features in Ruby 3.2. They are designed to provide per-operation or per-request state handling.

Using a thread pool, `Fiber.storage` is retained beyond a single request, which can leak information from one request into another. The simplest way to avoid this is to wrap each request itself in a fiber. The overhead should be minimal as these fibers are cached and reused, so once warm, the overhead is a single Ruby VALUE allocation per request.

The benefit is that code which uses `Fiber[]=` will be scoped correctly to a single request.

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,10 +1,3 @@
-## Unreleased
-
-* Features
-  * Add `fiber_per_request` option (aliased as `clean_thread_locals`) to isolate fiber-local state per request using Ruby Fibers. Enable via config or `PUMA_FIBER_PER_REQUEST` env var. ([#3101])
-
-[#3101]: https://github.com/puma/puma/pull/3101
-
 ## 7.0.0.pre1 / 2025-07-31
 
 * Changed
--- a/History.md
+++ b/History.md
@@ -1,3 +1,10 @@
+## Unreleased
+
+* Features
+  * Add `fiber_per_request` option (aliased as `clean_thread_locals`) to isolate fiber-local state per request using Ruby Fibers. Enable via config or `PUMA_FIBER_PER_REQUEST` env var. ([#3101])
+
+[#3101]: https://github.com/puma/puma/pull/3101
+
 ## 7.0.0.pre1 / 2025-07-31
 
 * Changed
--- a/lib/puma/configuration.rb
+++ b/lib/puma/configuration.rb
@@ -128,7 +128,7 @@ module Puma
     DEFAULTS = {
       auto_trim_time: 30,
       binds: ['tcp://0.0.0.0:9292'.freeze],
-      clean_thread_locals: false,
+      fiber_per_request: ENV["PUMA_FIBER_PER_REQUEST"] == "true",
       debug: false,
       enable_keep_alives: true,
       early_hints: nil,
--- a/lib/puma/configuration.rb
+++ b/lib/puma/configuration.rb
@@ -128,7 +128,7 @@ module Puma
     DEFAULTS = {
       auto_trim_time: 30,
       binds: ['tcp://0.0.0.0:9292'.freeze],
-      fiber_per_request: ENV["PUMA_FIBER_PER_REQUEST"] == "true",
+      fiber_per_request: !!ENV.fetch("PUMA_FIBER_PER_REQUEST", false),
       debug: false,
       enable_keep_alives: true,
       early_hints: nil,
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -370,17 +370,20 @@ module Puma
       @options[:idle_timeout] = Integer(seconds)
     end
 
-    # Work around leaky apps that leave garbage in Thread locals
-    # across requests.
+    # Use a clean fiber per request which ensures a clean slate for fiber
+    # locals and fiber storage. Also provides a cleaner backtrace with less
+    # Puma internal stack frames.
     #
     # The default is +false+.
     #
     # @example
-    #   clean_thread_locals
+    #   fiber_per_request
     #
-    def clean_thread_locals(which=true)
-      @options[:clean_thread_locals] = which
+    def fiber_per_request(which=true)
+      @options[:fiber_per_request] = which
     end
+
+    alias clean_thread_locals fiber_per_request
 
     # When shutting down, drain the accept socket of pending connections and
     # process them. This loops over the accept socket until there are no more
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -32,6 +32,14 @@ module Puma
   #
   # Each `Puma::Server` will have one reactor and one thread pool.
   class Server
+    module FiberPerRequest
+      def handle_request(client, requests)
+        Fiber.new do
+          super
+        end.resume
+      end
+    end
+
     include Puma::Const
     include Request
 
@@ -99,6 +107,10 @@ module Puma
       @enable_keep_alives      &&= @queue_requests
       @io_selector_backend       = @options[:io_selector_backend]
       @http_content_length_limit = @options[:http_content_length_limit]
+
+      if @options[:fiber_per_request]
+        singleton_class.prepend(FiberPerRequest)
+      end
 
       # make this a hash, since we prefer `key?` over `include?`
       @supported_http_methods =
@@ -460,7 +472,6 @@ module Puma
       # Advertise this server into the thread
       Thread.current.puma_server = self
 
-      clean_thread_locals = options[:clean_thread_locals]
       close_socket = true
 
       requests = 0
@@ -485,8 +496,6 @@ module Puma
         when :async
           close_socket = false
         when true
-          ThreadPool.clean_thread_locals if clean_thread_locals
-
           requests += 1
 
           client.reset
--- a/lib/puma/thread_pool.rb
+++ b/lib/puma/thread_pool.rb
@@ -53,7 +53,6 @@ module Puma
       @block = block
       @out_of_band = options[:out_of_band]
       @out_of_band_running = false
-      @clean_thread_locals = options[:clean_thread_locals]
       @before_thread_start = options[:before_thread_start]
       @before_thread_exit = options[:before_thread_exit]
       @reaping_time = options[:reaping_time]
@@ -81,12 +80,6 @@ module Puma
     end
 
     attr_reader :spawned, :trim_requested, :waiting
-
-    def self.clean_thread_locals
-      Thread.current.keys.each do |key| # rubocop: disable Style/HashEachMethods
-        Thread.current[key] = nil unless key == :__recursive_key__
-      end
-    end
 
     # generate stats hash so as not to perform multiple locks
     # @return [Hash] hash containing stat info from ThreadPool
@@ -173,10 +166,6 @@ module Puma
             end
 
             work = todo.shift
-          end
-
-          if @clean_thread_locals
-            ThreadPool.clean_thread_locals
           end
 
           begin
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -181,6 +181,7 @@ module TestSkips
         when :unix        then "Skipped if UNIXSocket exists"    if Puma::HAS_UNIX_SOCKET
         when :aunix       then "Skipped if abstract UNIXSocket"  if Puma.abstract_unix_socket?
         when :rack3       then "Skipped if Rack 3.x"             if Rack.release >= '3'
+        when :oldwindows  then "Skipped if old Windows"          if Puma::IS_WINDOWS && RUBY_VERSION < '2.6'
         else false
       end
       skip skip_msg, bt if skip_msg
--- a/test/test_fiber_local_application.rb
+++ b/test/test_fiber_local_application.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+# Released under the MIT License.
+# Copyright, 2023, by Samuel Williams.
+
+require_relative "helper"
+
+require "puma/server"
+
+class FiberLocalApplication
+  def call(env)
+    # Just in case you didn't know, this is fiber local...
+    count = (Thread.current[:request_count] ||= 0)
+    Thread.current[:request_count] += 1
+    [200, {"Content-Type" => "text/plain"}, [count.to_s]]
+  end
+end
+
+class FiberLocalApplicationTest < Minitest::Test
+  parallelize_me!
+
+  def setup
+    @tester = FiberLocalApplication.new
+    @server = Puma::Server.new @tester, nil, {log_writer: Puma::LogWriter.strings, fiber_per_request: true}
+    @port = (@server.add_tcp_listener "127.0.0.1", 0).addr[1]
+    @tcp = "http://127.0.0.1:#{@port}"
+    @server.run
+  end
+
+  def teardown
+    @server.stop(true)
+  end
+
+  def test_empty_locals
+    response = hit(["#{@tcp}/test"] * 3)
+    assert_equal ["0", "0", "0"], response
+  end
+end
--- a/test/test_fiber_local_application.rb
+++ b/test/test_fiber_local_application.rb
@@ -32,6 +32,8 @@ class FiberLocalApplicationTest < Minitest::Test
   end
 
   def test_empty_locals
+    skip_if :oldwindows
+
     response = hit(["#{@tcp}/test"] * 3)
     assert_equal ["0", "0", "0"], response
   end
--- a/test/test_fiber_local_application.rb
+++ b/test/test_fiber_local_application.rb
@@ -16,7 +16,7 @@ class FiberLocalApplication
   end
 end
 
-class FiberLocalApplicationTest < Minitest::Test
+class FiberLocalApplicationTest < PumaTest
   parallelize_me!
 
   def setup
--- a/test/test_fiber_storage_application.rb
+++ b/test/test_fiber_storage_application.rb
@@ -33,6 +33,8 @@ class FiberStorageApplicationTest < Minitest::Test
   end
 
   def test_empty_storage
+    skip_if :oldwindows
+
     response = hit(["#{@tcp}/test"] * 3)
     assert_equal ["0", "0", "0"], response
   end
--- a/test/test_fiber_storage_application.rb
+++ b/test/test_fiber_storage_application.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+# Released under the MIT License.
+# Copyright, 2023, by Samuel Williams.
+
+require_relative "helper"
+
+require "puma/server"
+
+class FiberStorageApplication
+  def call(env)
+    count = (Fiber[:request_count] ||= 0)
+    Fiber[:request_count] += 1
+    [200, {"Content-Type" => "text/plain"}, [count.to_s]]
+  end
+end
+
+class FiberStorageApplicationTest < Minitest::Test
+  parallelize_me!
+
+  def setup
+    skip "Fiber Storage is not supported on this Ruby" unless Fiber.respond_to?(:[])
+
+    @tester = FiberStorageApplication.new
+    @server = Puma::Server.new @tester, nil, {log_writer: Puma::LogWriter.strings, fiber_per_request: true}
+    @port = (@server.add_tcp_listener "127.0.0.1", 0).addr[1]
+    @tcp = "http://127.0.0.1:#{@port}"
+    @server.run
+  end
+
+  def teardown
+    @server.stop(true)
+  end
+
+  def test_empty_storage
+    response = hit(["#{@tcp}/test"] * 3)
+    assert_equal ["0", "0", "0"], response
+  end
+end
--- a/test/test_fiber_storage_application.rb
+++ b/test/test_fiber_storage_application.rb
@@ -15,7 +15,7 @@ class FiberStorageApplication
   end
 end
 
-class FiberStorageApplicationTest < Minitest::Test
+class FiberStorageApplicationTest < PumaTest
   parallelize_me!
 
   def setup
--- a/test/test_thread_pool.rb
+++ b/test/test_thread_pool.rb
@@ -233,24 +233,6 @@ class TestThreadPool < PumaTest
     assert_equal 1, pool.spawned
   end
 
-  def test_cleanliness
-    values = []
-    n = 100
-
-    pool = mutex_pool(1,1) {
-      values.push Thread.current[:foo]
-      Thread.current[:foo] = :hai
-    }
-
-    pool.instance_variable_set :@clean_thread_locals, true
-
-    pool << [1] * n
-
-    assert_equal n,  values.length
-
-    assert_equal [], values.compact
-  end
-
   def test_reap_only_dead_threads
     pool = mutex_pool(2,2) do
       th = Thread.current


nateberkopec:
Very interesting!

I assume we don't need a Ruby version gate here - people just won't use `Fiber[]=` in their Ruby 3.1 and earlier apps.

1. We should have a test for 3.2+ that tries to read `Fiber[]` across requests.
2. We should run this by our benchmark suite locally to double-check.

ioquatix:
@nateberkopec There are two parts to this PR.

One might be considered a feature and one might be considered a bug.

Between requests, fiber locals are not reset, so even without supporting `Fiber.storage`, just using `Thread.current[:foo]` can leak between requests (may or may not be a bad thing???).

The other part of this is correctly supporting `Fiber.storage` assignment.

ioquatix:
@nateberkopec I've just added tests, I think you will find them interesting to review because they show the current behaviour vs what I'm proposing (just comment out the first commit to see the behavioural differences).

Also 100% we should test for performance regressions. Is there documentation you can point me at for the benchmark test suite?

nateberkopec:
Right, thanks for calling that out - since `Thread.current[:foo]` uses fiber-locals underneath, this will be a big behavior change for people using that right? 
 
 >(may or may not be a bad thing???).
 
 It's not good or bad, it's just Behavior, Which That People Inevitably Depend On 😆 
 
> Is there documentation you can point me at for the benchmark test suite?

Not really, consider this just an action item on the maintainer's part.

nateberkopec:
We also failures on main, yikes.

ioquatix:
I might need to fix the exception handling when I’m back home.

ioquatix:
I observed Windows segfaulting: https://github.com/puma/puma/actions/runs/4445249381/jobs/7804200284

MSP-Greg:
@ioquatix Thanks for working on this.  I ran these commits on top of revisions to the test system, Windows 2.5 failed, and all the JRuby jobs.  Everything else was ok.

Some performance data using WSL@/Ubuntu 22.04 is below.  It does seem to slow things down...
```
Puma repo branch pr/3101
Body    ────────── req/sec ──────────   ─────── req 50% times ───────
 KB     array   chunk  string      io   array   chunk  string      io
1       16279   16019   15739   11057   0.195   0.194   0.206   0.311
10      15737   13432   13397   10976   0.209   0.302   0.312   0.381
─────────────────────────────────────────────────────────────────────

Puma repo branch master
Body    ────────── req/sec ──────────   ─────── req 50% times ───────
 KB     array   chunk  string      io   array   chunk  string      io
1       18484   18486   18641   12733   0.195   0.178   0.191   0.235
10      18297   17096   18560   12368   0.170   0.240   0.196   0.268
─────────────────────────────────────────────────────────────────────
```

Above was run with `benchmarks/local/response_time_wrk.sh -w2 -t5:5 -s tcp6 -Y -b 1,10`, Using your wrk.

ioquatix:
I will check the performance too, that seems a little too costly to me. Maybe we can improve on it, or maybe there is something else going on.

In any case, there are a couple of options to consider:

1. This could be a middleware. But this behaviour is inconsistent across different servers, and the middleware would have no effect on fiber-per-request servers like Falcon already.
2. It could be optional feature of Puma.

MSP-Greg:
> Maybe we can improve on it

Interested in helping with that.  Right now, the Fiber is wrapping nothing more than `app.call`, which in the benchmarks is very quickly sending back a few dozen headers and various body configurations (size, array length, enum, file, etc).

Or, we've got three main 'processes':

A. Receive and process the request
B. App.call
C. Assemble & transmit the response

Current code is just wrapping B.  I suspect that should be expanded?

ioquatix:
Feel free to modify this PR as you see fit. I will try to look at it a bit later tonight or tomorrow.

ioquatix:
Regarding your question, I'm only concerned about the app code being exposed to prior state. Well, we can wrap more, but I don't think there is a huge advantage.

dentarg:
Re `Thread.current` and leaking state, I suspect most apps are clearing it themselves ([example](https://github.com/steveklabnik/request_store/blob/81003584e527f7ddceaa8df47ace4badc48a4c84/lib/request_store/middleware.rb#L21-L34))

ioquatix:
Ah, nice example.

I guess ideally with the newer "fiber per request" style interface, we shouldn't expect users to do this (one of the authors of request state was actually very excited by this). The semantics at the language level should be clear.

nateberkopec:
>This could be a middleware. 
>It could be optional feature of Puma.

Hmm... when you put it this way, it reminds me of the old "thread_safe" config setting in Rails. Aaron eventually removed it because "who would ever say, "give me the non thread safe version please!"

12% performance hit is rough, but we're still doing a response in far less than a tenth of a millisecond. I think the trade could be worth it.

ioquatix:
> 12% performance hit is rough, but we're still doing a response in far less than a tenth of a millisecond. I think the trade could be worth it.

With a little bit of finangling, we might be able to introduce a more streamlined interface.

For example, if you only care about Fiber storage, it possible to assign to it directly.

Let me do a bit more analysis on the performance cost. It seems 1/10th of a ms is quite a bit, more than I expected.

ioquatix:
I just noticed there is some provision for `clean_thread_locals` which is actually "cleaning" fiber locals.

ioquatix:
I wouldn't call this particularly comprehensive (I was seeing a margin of error of like +-5%) but here are the results from my laptop:

## Normal

```
> benchmarks/local/response_time_wrk.sh -T2 -w2 -t8:8 -b c1,10,100

Size   req/sec    50%     75%     90%     99%    100%  Resp Size
───────────────────────────────────────────────────────────────── chunk
1       27301   0.136   0.154   0.189   0.439  16.780      1040
10      21446   0.174   0.195   0.235   0.446  15.030     10318
100      6037   0.646   0.711   0.806   1.010   7.270    103108
```

## Fiber

```
> benchmarks/local/response_time_wrk.sh -T2 -w2 -t8:8 -b c1,10,100

Size   req/sec    50%     75%     90%     99%    100%  Resp Size
───────────────────────────────────────────────────────────────── chunk
1       25492    0.14    0.16    0.21    0.44    6.67      1040
10      20082    0.18    0.21    0.27    0.52    6.98     10318
100      5302    0.70    0.82    1.06  199.18  306.25    103107

> RUBY_SHARED_FIBER_POOL_FREE_STACKS=0 benchmarks/local/response_time_wrk.sh -T2 -w2 -t8:8 -b c1,10,100

Size   req/sec    50%     75%     90%     99%    100%  Resp Size
───────────────────────────────────────────────────────────────── chunk
1       26900   0.138   0.156   0.196   0.412   6.580      1040
10      21256   0.173   0.195   0.250   0.479   6.770     10318
100      5841   0.660   0.741   0.850   1.080   7.430    103107
```

ioquatix:
(also as an aside, the benchmark uses too many wrk threads/connections and will over-saturate Puma giving less than robust results, since a whole ton of `wrk` connection will be idle).

ioquatix:
`RUBY_SHARED_FIBER_POOL_FREE_STACKS=0` is a secret escape hatch which uses a little more memory but increases performance, since it doesn't return the fiber stacks back to the OS.

ioquatix:
The Linux results are also a bit weird.

## Normal

```
> benchmarks/local/response_time_wrk.sh -T2 -w2 -t8:8 -b c1,10,100

Size   req/sec    50%     75%     90%     99%    100%  Resp Size
───────────────────────────────────────────────────────────────── chunk
1       25380   0.146   0.154   0.185   0.415   5.330      1040
10      20640   0.176   0.188   0.270   0.473   5.780     10318
100      9585   0.386   0.488   0.592   0.811   7.850    103185
```

## Fiber

```
> PUMA_FIBER_PER_REQUEST=1 benchmarks/local/response_time_wrk.sh -T2 -w2 -t8:8 -b c1,10,100

Size   req/sec    50%     75%     90%     99%    100%  Resp Size
───────────────────────────────────────────────────────────────── chunk
1       21598   0.171   0.182   0.226   0.472   5.650      1040
10      18058   0.200   0.212   0.308   0.530   5.740     10318
100      8603   0.474   0.516   0.669   0.860   6.350    103108

> PUMA_FIBER_PER_REQUEST=1 RUBY_SHARED_FIBER_POOL_FREE_STACKS=0 benchmarks/local/response_time_wrk.sh -T2 -w2 -t8:8 -b c1,10,100

Size   req/sec    50%     75%     90%     99%    100%  Resp Size
───────────────────────────────────────────────────────────────── chunk
1       37997   0.102   0.130   0.140   0.272   5.520      1040
10      24752   0.154   0.189   0.219   0.397   5.800     10318
100      8672   0.440   0.503   0.612   0.827   7.950    103109
```

MSP-Greg:
@ioquatix Thanks for looking into a bunch of things.  I've been putting off getting a native Ubuntu host system (time, money, etc), so I use WSL2/Ubuntu, which runs on Windows.  It is somewhat inconsistent, and often slower than a native system.

So, I tried to get the benchmark default to hit Puma hard, but not 'hammered', or something like that.  If you find different default would work better, we change them.

I suspect you may have found it in the code or notes, but array (s) & chunked (c) bodies use an element size of 1kB, so c100 is sending 100 1kB chunks for the body.  For String bodies (s) the body is a single element array...

ioquatix:
@MSP-Greg do you accept github sponsors?

nateberkopec:
While I think - if this is a Good Idea and It Works - we should just make the default, it is a breaking change and likely to Break Something Some Where. I wonder if, in the meantime, before Puma 7, we could ship this is a rack middleware to give people time to try it and see what happens?

The middleware would just a be a temporary thing intended to test this out, not as a permanent fixture.

ioquatix:
Here are my thoughts in no particular order.

The performance overhead will be more complex on TruffleRuby and JRuby until they adopt Loom/Virtual Threads.

It's possible in the single process/thread case that users are always expecting the same thread locals to work as a kind of global state.

Since connection keep-alive is bound to a worker, there is a kind of natural affinity that comes with every request being processed by the same thread. I could imagine some code depending on that (it's going to be janky but 🤷🏼). Additionally, if they are using thread-local caches, performance/locality might be affected.

This feature can't quite be implemented as a middleware as it doesn't correctly cover streaming responses. However, there is NO specification that states that a streaming response should be handled on the same fiber as the request, but I imagine some poorly designed code that tries to take a lock during `app.call` and free it in `response[2].close`, could blow up.

I believe that the performance overhead can and should be reduced to almost zero but it will require more investigation on my part. I will definitely consider this use case closely but I can't make any guarantees before Ruby 3.3. It's possible I'll have time and motivation to investigate it. An overhead of 2-3% should be attainable or better IMHO.

Maybe better to start off as a feature of puma that is disabled by default in v6. It could be controlled by the existing option: `clean_thread_locals` which is... actually cleaning fiber locals (according to the implementation). Or we could introduce a new option.

nateberkopec:
> Maybe better to start off as a feature of puma that is disabled by default in v6. It could be controlled by the existing option: clean_thread_locals which is... actually cleaning fiber locals (according to the implementation). Or we could introduce a new option.

Understood, I'm fine with this course of action, and I like the name.

ioquatix:
> Understood, I'm fine with this course of action, and I like the name.

Sorry, what name do you like? Because `clean_thread_locals` would imply that `Thread.current.thread_variables` should be reset after every request, which it isn't.

It would probably make more sense to introduce a new option `fiber_per_request` and make `clean_thread_locals` an alias to it (perhaps with a deprecation warning).

leandronsp:
Hi guys, I see the conversation hasn't been updated since then. Are you willing to make advancements at this matter? For me it looks like a great feature in Puma.

@ioquatix any help needed? I could try to help moving this forward if needed.

ioquatix:
There is one more benefit to this feature: stack traces in errors will be much more specific to the request and won't include the guts of Puma.

@leandronsp I'll take a look at this today.

ioquatix:
Okay, I suppose this is pretty close to perfection. It's disabled by default so... it won't affect performance unless users opt into it. I suppose we could write more documentation too.

ioquatix:
If someone wants to dig into benchmarks, it would be interesting to compare:

1. Current Puma
2. This PR
3. This PR with `fiber_per_request`.
4. Current Puma with `clean_thread_locals`

I've ordered them in what I guess will be best to worse performance. Remember to actually set some thread locals :) Otherwise 4 might be close to a no-op. I think (1) and (2) might be close. Because we removed some code off the critical path, (2) might even be faster.

j-manu:
`fiber_per_request` option is a bit confusing as Puma is still using a thread per request. Since the aim for this PR is to provide safe fiber locals and wrapping the request in a fiber is the implementation detail, the option can
be called `enable_safe_fiber_locals` or `clean_request_locals` (similar to the current `clean_thread_locals` )

ioquatix:
I don't have a strong opinion about the name of the configuration option, but I think the things you mention are all effects of "fiber per request". In theory, `enable_safe_fiber_locals` could be implemented without a fiber per request, for example. `clean_request_locals` seems a bit too unique/bespoke to me.

Anyway, I'll defer to whatever @nateberkopec recommends.

ioquatix:
@zzak brought up a good point that changing the behaviour of `clean_thread_locals`, even slightly, should be handled with more care. I don't have a strong opinion about it. Functionality it's identical in most ways that matter.

In fact, it's probably practically impossible to implement `clean_thread_locals` correctly without support from Ruby itself.

So, on this point, I'll defer again to @nateberkopec - should we leave `clean_thread_locals` out of this PR? Note, if we do that, I probably don't have the motivation for a follow up PR to fix `clean_thread_locals`. Just let me know your preference.

ioquatix:
Using `Thread.current[:x]` and `Fiber[:x]` is still dangerous with Puma. We ran into the issue at work.

I think there is value in adopting this feature, however I'm happy to add it behind a feature flag (instead of `ENV['PUMA_FIBER_PER_REQUEST'`). This might make it less risky to adopt, as it would be opt-in (to start with at least).

ioquatix:
I tried to compare this using a simplified thread bench (mostly just to use K6 benchmark tool).

```
samuel@sakura ~/D/i/threadbench (main)> PUMA_FIBER_PER_REQUEST=true bundle exec bin/bench
* Puma version: 6.4.2-fiber-per-request (ruby 3.3.0-p0) ("The Eagle of Durango")
*  Min threads: 1
*  Max threads: 1
*  Environment: development
*          PID: 38207
*** FIBER PER REQUEST ***
Use Ctrl-C to stop
Warming up for 15 seconds...
WARN[0033] Request Failed                                error="Get \"http://localhost:9292\": dial: i/o timeout"
Benchmarking...

          /\      |‾‾| /‾‾/   /‾‾/   
     /\  /  \     |  |/  /   /  /    
    /  \/    \    |     (   /   ‾‾\  
   /          \   |  |\  \ |  (‾)  | 
  / __________ \  |__| \__\ \_____/ .io

     execution: local
        script: k6/script.js
        output: -

     scenarios: (100.00%) 1 scenario, 10 max VUs, 2m0s max duration (incl. graceful stop):
              * contacts: Up to 8.00 iterations/s for 1m30s over 1 stages (maxVUs: 10, gracefulStop: 30s)


     data_received..................: 43 kB 475 B/s
     data_sent......................: 36 kB 400 B/s
     http_req_blocked...............: min=213µs   avg=1.23ms   med=1.01ms max=8.4ms   p(75)=1.26ms  p(99)=5.01ms   p(99.99)=8.37ms 
     http_req_connecting............: min=175µs   avg=970.43µs med=781µs  max=7.32ms  p(75)=979.5µs p(99)=4.23ms   p(99.99)=7.31ms 
     http_req_duration..............: min=283µs   avg=1.63ms   med=1.5ms  max=5.34ms  p(75)=1.87ms  p(99)=4.63ms   p(99.99)=5.34ms 
       { expected_response:true }...: min=283µs   avg=1.63ms   med=1.5ms  max=5.34ms  p(75)=1.87ms  p(99)=4.63ms   p(99.99)=5.34ms 
     http_req_failed................: 0.00% ✓ 0        ✗ 450 
     http_req_receiving.............: min=26µs    avg=144.83µs med=126µs  max=1.99ms  p(75)=158µs   p(99)=762.29µs p(99.99)=1.98ms 
     http_req_sending...............: min=20µs    avg=189.91µs med=159µs  max=3.02ms  p(75)=178µs   p(99)=949.13µs p(99.99)=3.02ms 
     http_req_tls_handshaking.......: min=0s      avg=0s       med=0s     max=0s      p(75)=0s      p(99)=0s       p(99.99)=0s     
   ✓ http_req_waiting...............: min=237µs   avg=1.3ms    med=1.18ms max=4.87ms  p(75)=1.48ms  p(99)=4.3ms    p(99.99)=4.87ms 
     http_reqs......................: 450   4.999733/s
     iteration_duration.............: min=550.2µs avg=3.22ms   med=3.01ms max=10.83ms p(75)=3.61ms  p(99)=7.26ms   p(99.99)=10.73ms
     iterations.....................: 450   4.999733/s
     vus............................: 1     min=0      max=1 
     vus_max........................: 10    min=10     max=10


running (1m30.0s), 00/10 VUs, 450 complete and 0 interrupted iterations
contacts ✓ [======================================] 00/10 VUs  1m30s  8.00 iters/s
samuel@sakura ~/D/i/threadbench (main)> PUMA_FIBER_PER_REQUEST=false bundle exec bin/bench
* Puma version: 6.4.2-fiber-per-request (ruby 3.3.0-p0) ("The Eagle of Durango")
*  Min threads: 1
*  Max threads: 1
*  Environment: development
*          PID: 39262
*** THREAD PER REQUEST ***
Use Ctrl-C to stop
Warming up for 15 seconds...
WARN[0041] Request Failed                                error="Get \"http://localhost:9292\": dial: i/o timeout"
Benchmarking...

          /\      |‾‾| /‾‾/   /‾‾/   
     /\  /  \     |  |/  /   /  /    
    /  \/    \    |     (   /   ‾‾\  
   /          \   |  |\  \ |  (‾)  | 
  / __________ \  |__| \__\ \_____/ .io

     execution: local
        script: k6/script.js
        output: -

     scenarios: (100.00%) 1 scenario, 10 max VUs, 2m0s max duration (incl. graceful stop):
              * contacts: Up to 8.00 iterations/s for 1m30s over 1 stages (maxVUs: 10, gracefulStop: 30s)


     data_received..................: 43 kB 475 B/s
     data_sent......................: 36 kB 400 B/s
     http_req_blocked...............: min=355µs    avg=1.2ms    med=1.05ms   max=7.27ms p(75)=1.25ms p(99)=3.77ms   p(99.99)=7.17ms
     http_req_connecting............: min=237µs    avg=943.74µs med=825.49µs max=4.67ms p(75)=966µs  p(99)=3.18ms   p(99.99)=4.63ms
     http_req_duration..............: min=436µs    avg=1.63ms   med=1.48ms   max=6.31ms p(75)=1.74ms p(99)=4.78ms   p(99.99)=6.28ms
       { expected_response:true }...: min=436µs    avg=1.63ms   med=1.48ms   max=6.31ms p(75)=1.74ms p(99)=4.78ms   p(99.99)=6.28ms
     http_req_failed................: 0.00% ✓ 0        ✗ 450 
     http_req_receiving.............: min=31µs     avg=147.75µs med=131µs    max=1.6ms  p(75)=163µs  p(99)=676.49µs p(99.99)=1.57ms
     http_req_sending...............: min=38µs     avg=193.18µs med=159µs    max=2.36ms p(75)=180µs  p(99)=1ms      p(99.99)=2.34ms
     http_req_tls_handshaking.......: min=0s       avg=0s       med=0s       max=0s     p(75)=0s     p(99)=0s       p(99.99)=0s    
   ✓ http_req_waiting...............: min=334µs    avg=1.29ms   med=1.13ms   max=6.01ms p(75)=1.38ms p(99)=4.45ms   p(99.99)=5.98ms
     http_reqs......................: 450   4.999651/s
     iteration_duration.............: min=934.12µs avg=3.22ms   med=3.04ms   max=8.21ms p(75)=3.63ms p(99)=6.78ms   p(99.99)=8.21ms
     iterations.....................: 450   4.999651/s
     vus............................: 1     min=0      max=1 
     vus_max........................: 10    min=10     max=10


running (1m30.0s), 00/10 VUs, 450 complete and 0 interrupted iterations
contacts ✓ [======================================] 00/10 VUs  1m30s  8.00 iters/s
```

I could not see any significant difference in performance.

ioquatix:
I've done additional benchmarking on my Linux machine using the `threadbench` but simplified to two tests:

1. `sleep 0.01` (10ms) -> `[200, {}, ["Hello World!"]]`.
2. Same as above without sleep.

The idea for test (1) is to kind of see if there is any practical real world impact (10ms being the general overhead of an "average" web application latency). The idea for test (2) is to see the absolute impact, if any.

## Test 1 Results (sleep 10ms)

```
samuel@aiko ~/P/i/threadbench (simple)> PUMA_FIBER_PER_REQUEST=false bundle exec bin/bench
* Puma version: 6.4.2-fiber-per-request (ruby 3.3.0-p0) ("The Eagle of Durango")
*  Min threads: 1
*  Max threads: 1
*  Environment: development
*          PID: 871836
*** THREAD PER REQUEST ***
Use Ctrl-C to stop
Warming up for 15 seconds...
Benchmarking...

          /\      |‾‾| /‾‾/   /‾‾/   
     /\  /  \     |  |/  /   /  /    
    /  \/    \    |     (   /   ‾‾\  
   /          \   |  |\  \ |  (‾)  | 
  / __________ \  |__| \__\ \_____/ .io

     execution: local
        script: k6/script.js
        output: -

     scenarios: (100.00%) 1 scenario, 10 max VUs, 2m0s max duration (incl. graceful stop):
              * contacts: Up to 8.00 iterations/s for 1m30s over 1 stages (maxVUs: 10, gracefulStop: 30s)


     data_received..................: 43 kB 474 B/s
     data_sent......................: 36 kB 399 B/s
     http_req_blocked...............: min=134.25µs avg=231.36µs med=219.92µs max=686.05µs p(75)=249.96µs p(99)=504.28µs p(99.99)=681.12µs
     http_req_connecting............: min=85.37µs  avg=170.05µs med=157.86µs max=518.43µs p(75)=182.62µs p(99)=426.01µs p(99.99)=516.26µs
     http_req_duration..............: min=10.4ms   avg=10.68ms  med=10.63ms  max=12.18ms  p(75)=10.71ms  p(99)=11.22ms  p(99.99)=12.16ms 
       { expected_response:true }...: min=10.4ms   avg=10.68ms  med=10.63ms  max=12.18ms  p(75)=10.71ms  p(99)=11.22ms  p(99.99)=12.16ms 
     http_req_failed................: 0.00% ✓ 0        ✗ 449 
     http_req_receiving.............: min=26.18µs  avg=57.93µs  med=57.24µs  max=170.53µs p(75)=67.81µs  p(99)=105.42µs p(99.99)=168.32µs
     http_req_sending...............: min=28.69µs  avg=48.77µs  med=48.91µs  max=100.65µs p(75)=57.26µs  p(99)=78.29µs  p(99.99)=99.84µs 
     http_req_tls_handshaking.......: min=0s       avg=0s       med=0s       max=0s       p(75)=0s       p(99)=0s       p(99.99)=0s      
   ✓ http_req_waiting...............: min=10.33ms  avg=10.57ms  med=10.52ms  max=12.03ms  p(75)=10.59ms  p(99)=11.1ms   p(99.99)=12.01ms 
     http_reqs......................: 449   4.988856/s
     iteration_duration.............: min=10.68ms  avg=10.98ms  med=10.92ms  max=12.54ms  p(75)=11.02ms  p(99)=11.55ms  p(99.99)=12.51ms 
     iterations.....................: 449   4.988856/s
     vus............................: 0     min=0      max=1 
     vus_max........................: 10    min=10     max=10


running (1m30.0s), 00/10 VUs, 449 complete and 0 interrupted iterations
contacts ✓ [======================================] 00/10 VUs  1m30s  8.00 iters/s
samuel@aiko ~/P/i/threadbench (simple)> PUMA_FIBER_PER_REQUEST=false bundle exec bin/bench
* Puma version: 6.4.2-fiber-per-request (ruby 3.3.0-p0) ("The Eagle of Durango")
*  Min threads: 1
*  Max threads: 1
*  Environment: development
*          PID: 873832
*** THREAD PER REQUEST ***
Use Ctrl-C to stop
Warming up for 15 seconds...
Benchmarking...

          /\      |‾‾| /‾‾/   /‾‾/   
     /\  /  \     |  |/  /   /  /    
    /  \/    \    |     (   /   ‾‾\  
   /          \   |  |\  \ |  (‾)  | 
  / __________ \  |__| \__\ \_____/ .io

     execution: local
        script: k6/script.js
        output: -

     scenarios: (100.00%) 1 scenario, 10 max VUs, 2m0s max duration (incl. graceful stop):
              * contacts: Up to 8.00 iterations/s for 1m30s over 1 stages (maxVUs: 10, gracefulStop: 30s)


     data_received..................: 43 kB 475 B/s
     data_sent......................: 36 kB 400 B/s
     http_req_blocked...............: min=148.44µs avg=225.87µs med=212.31µs max=700.8µs  p(75)=245.93µs p(99)=481.83µs p(99.99)=697.12µs
     http_req_connecting............: min=104.9µs  avg=164.98µs med=153.49µs max=477.73µs p(75)=179.92µs p(99)=383.33µs p(99.99)=477.6µs 
     http_req_duration..............: min=10.42ms  avg=10.66ms  med=10.61ms  max=11.56ms  p(75)=10.7ms   p(99)=11.09ms  p(99.99)=11.55ms 
       { expected_response:true }...: min=10.42ms  avg=10.66ms  med=10.61ms  max=11.56ms  p(75)=10.7ms   p(99)=11.09ms  p(99.99)=11.55ms 
     http_req_failed................: 0.00% ✓ 0        ✗ 450 
     http_req_receiving.............: min=26.47µs  avg=61.74µs  med=57.03µs  max=156.77µs p(75)=72.47µs  p(99)=114.15µs p(99.99)=155.32µs
     http_req_sending...............: min=25.17µs  avg=49.05µs  med=48.77µs  max=90.22µs  p(75)=57.61µs  p(99)=79.66µs  p(99.99)=90.01µs 
     http_req_tls_handshaking.......: min=0s       avg=0s       med=0s       max=0s       p(75)=0s       p(99)=0s       p(99.99)=0s      
   ✓ http_req_waiting...............: min=10.34ms  avg=10.55ms  med=10.5ms   max=11.45ms  p(75)=10.58ms  p(99)=10.98ms  p(99.99)=11.44ms 
     http_reqs......................: 450   4.999355/s
     iteration_duration.............: min=10.71ms  avg=10.96ms  med=10.91ms  max=11.86ms  p(75)=11.01ms  p(99)=11.46ms  p(99.99)=11.85ms 
     iterations.....................: 450   4.999355/s
     vus............................: 0     min=0      max=1 
     vus_max........................: 10    min=10     max=10


running (1m30.0s), 00/10 VUs, 450 complete and 0 interrupted iterations
contacts ✓ [======================================] 00/10 VUs  1m30s  8.00 iters/s
samuel@aiko ~/P/i/threadbench (simple)> PUMA_FIBER_PER_REQUEST=true bundle exec bin/bench
* Puma version: 6.4.2-fiber-per-request (ruby 3.3.0-p0) ("The Eagle of Durango")
*  Min threads: 1
*  Max threads: 1
*  Environment: development
*          PID: 875815
*** FIBER PER REQUEST ***
Use Ctrl-C to stop
Warming up for 15 seconds...
Benchmarking...

          /\      |‾‾| /‾‾/   /‾‾/   
     /\  /  \     |  |/  /   /  /    
    /  \/    \    |     (   /   ‾‾\  
   /          \   |  |\  \ |  (‾)  | 
  / __________ \  |__| \__\ \_____/ .io

     execution: local
        script: k6/script.js
        output: -

     scenarios: (100.00%) 1 scenario, 10 max VUs, 2m0s max duration (incl. graceful stop):
              * contacts: Up to 8.00 iterations/s for 1m30s over 1 stages (maxVUs: 10, gracefulStop: 30s)


     data_received..................: 43 kB 475 B/s
     data_sent......................: 36 kB 400 B/s
     http_req_blocked...............: min=142.55µs avg=223.82µs med=211.98µs max=608.33µs p(75)=247.19µs p(99)=399.45µs p(99.99)=603.21µs
     http_req_connecting............: min=93.97µs  avg=165.32µs med=153.87µs max=407.85µs p(75)=183.14µs p(99)=339.44µs p(99.99)=407.6µs 
     http_req_duration..............: min=10.41ms  avg=10.67ms  med=10.63ms  max=11.22ms  p(75)=10.72ms  p(99)=11.11ms  p(99.99)=11.22ms 
       { expected_response:true }...: min=10.41ms  avg=10.67ms  med=10.63ms  max=11.22ms  p(75)=10.72ms  p(99)=11.11ms  p(99.99)=11.22ms 
     http_req_failed................: 0.00% ✓ 0        ✗ 450 
     http_req_receiving.............: min=27.08µs  avg=56.25µs  med=55.44µs  max=133.2µs  p(75)=64.81µs  p(99)=111.09µs p(99.99)=133.19µs
     http_req_sending...............: min=20.49µs  avg=45.13µs  med=46.25µs  max=102.2µs  p(75)=51.74µs  p(99)=67.76µs  p(99.99)=100.92µs
     http_req_tls_handshaking.......: min=0s       avg=0s       med=0s       max=0s       p(75)=0s       p(99)=0s       p(99.99)=0s      
   ✓ http_req_waiting...............: min=10.32ms  avg=10.57ms  med=10.52ms  max=11.14ms  p(75)=10.61ms  p(99)=11.01ms  p(99.99)=11.13ms 
     http_reqs......................: 450   4.999361/s
     iteration_duration.............: min=10.73ms  avg=10.97ms  med=10.93ms  max=11.59ms  p(75)=11.02ms  p(99)=11.44ms  p(99.99)=11.59ms 
     iterations.....................: 450   4.999361/s
     vus............................: 0     min=0      max=1 
     vus_max........................: 10    min=10     max=10


running (1m30.0s), 00/10 VUs, 450 complete and 0 interrupted iterations
contacts ✓ [======================================] 00/10 VUs  1m30s  8.00 iters/s
samuel@aiko ~/P/i/threadbench (simple)> PUMA_FIBER_PER_REQUEST=true bundle exec bin/bench
* Puma version: 6.4.2-fiber-per-request (ruby 3.3.0-p0) ("The Eagle of Durango")
*  Min threads: 1
*  Max threads: 1
*  Environment: development
*          PID: 877768
*** FIBER PER REQUEST ***
Use Ctrl-C to stop
Warming up for 15 seconds...
Benchmarking...

          /\      |‾‾| /‾‾/   /‾‾/   
     /\  /  \     |  |/  /   /  /    
    /  \/    \    |     (   /   ‾‾\  
   /          \   |  |\  \ |  (‾)  | 
  / __________ \  |__| \__\ \_____/ .io

     execution: local
        script: k6/script.js
        output: -

     scenarios: (100.00%) 1 scenario, 10 max VUs, 2m0s max duration (incl. graceful stop):
              * contacts: Up to 8.00 iterations/s for 1m30s over 1 stages (maxVUs: 10, gracefulStop: 30s)


     data_received..................: 43 kB 475 B/s
     data_sent......................: 36 kB 400 B/s
     http_req_blocked...............: min=148.69µs avg=212.88µs med=201.62µs max=603.31µs p(75)=224.89µs p(99)=412.24µs p(99.99)=597.12µs
     http_req_connecting............: min=102.6µs  avg=154.93µs med=144.81µs max=393.39µs p(75)=166.55µs p(99)=340.88µs p(99.99)=392.84µs
     http_req_duration..............: min=10.44ms  avg=10.66ms  med=10.61ms  max=11.31ms  p(75)=10.71ms  p(99)=11.12ms  p(99.99)=11.31ms 
       { expected_response:true }...: min=10.44ms  avg=10.66ms  med=10.61ms  max=11.31ms  p(75)=10.71ms  p(99)=11.12ms  p(99.99)=11.31ms 
     http_req_failed................: 0.00% ✓ 0        ✗ 450 
     http_req_receiving.............: min=26.14µs  avg=55.23µs  med=55.11µs  max=143µs    p(75)=64.06µs  p(99)=108.49µs p(99.99)=142.47µs
     http_req_sending...............: min=23.21µs  avg=44.41µs  med=45.44µs  max=205.82µs p(75)=51.29µs  p(99)=73.27µs  p(99.99)=200.11µs
     http_req_tls_handshaking.......: min=0s       avg=0s       med=0s       max=0s       p(75)=0s       p(99)=0s       p(99.99)=0s      
   ✓ http_req_waiting...............: min=10.35ms  avg=10.56ms  med=10.51ms  max=11.17ms  p(75)=10.6ms   p(99)=11ms     p(99.99)=11.17ms 
     http_reqs......................: 450   4.999344/s
     iteration_duration.............: min=10.69ms  avg=10.95ms  med=10.89ms  max=11.58ms  p(75)=11.01ms  p(99)=11.51ms  p(99.99)=11.58ms 
     iterations.....................: 450   4.999344/s
     vus............................: 0     min=0      max=1 
     vus_max........................: 10    min=10     max=10


running (1m30.0s), 00/10 VUs, 450 complete and 0 interrupted iterations
contacts ✓ [======================================] 00/10 VUs  1m30s  8.00 iters/s
```

## Test 2 Results (identity)

```
samuel@aiko ~/P/i/threadbench (simple)> PUMA_FIBER_PER_REQUEST=false bundle exec bin/bench
* Puma version: 6.4.2-fiber-per-request (ruby 3.3.0-p0) ("The Eagle of Durango")
*  Min threads: 1
*  Max threads: 1
*  Environment: development
*          PID: 880577
*** THREAD PER REQUEST ***
Use Ctrl-C to stop
Warming up for 15 seconds...
Benchmarking...

          /\      |‾‾| /‾‾/   /‾‾/   
     /\  /  \     |  |/  /   /  /    
    /  \/    \    |     (   /   ‾‾\  
   /          \   |  |\  \ |  (‾)  | 
  / __________ \  |__| \__\ \_____/ .io

     execution: local
        script: k6/script.js
        output: -

     scenarios: (100.00%) 1 scenario, 10 max VUs, 2m0s max duration (incl. graceful stop):
              * contacts: Up to 8.00 iterations/s for 1m30s over 1 stages (maxVUs: 10, gracefulStop: 30s)


     data_received..................: 43 kB 474 B/s
     data_sent......................: 36 kB 399 B/s
     http_req_blocked...............: min=152.02µs avg=220.14µs med=207.22µs max=638.5µs  p(75)=247.34µs p(99)=396µs    p(99.99)=631.83µs
     http_req_connecting............: min=93.66µs  avg=160.57µs med=149.16µs max=421.98µs p(75)=182.18µs p(99)=319.98µs p(99.99)=418.63µs
     http_req_duration..............: min=357.31µs avg=502.36µs med=498.52µs max=1.2ms    p(75)=540.51µs p(99)=635.21µs p(99.99)=1.18ms  
       { expected_response:true }...: min=357.31µs avg=502.36µs med=498.52µs max=1.2ms    p(75)=540.51µs p(99)=635.21µs p(99.99)=1.18ms  
     http_req_failed................: 0.00% ✓ 0        ✗ 449 
     http_req_receiving.............: min=25.65µs  avg=44.38µs  med=39.54µs  max=110.3µs  p(75)=50.58µs  p(99)=97.37µs  p(99.99)=110.16µs
     http_req_sending...............: min=24.87µs  avg=44.75µs  med=45.57µs  max=98.94µs  p(75)=50.97µs  p(99)=77.72µs  p(99.99)=98.67µs 
     http_req_tls_handshaking.......: min=0s       avg=0s       med=0s       max=0s       p(75)=0s       p(99)=0s       p(99.99)=0s      
   ✓ http_req_waiting...............: min=283.38µs avg=413.22µs med=418.35µs max=1.14ms   p(75)=447.09µs p(99)=547.62µs p(99.99)=1.11ms  
     http_reqs......................: 449   4.988826/s
     iteration_duration.............: min=622.63µs avg=793.55µs med=779.58µs max=1.45ms   p(75)=821.24µs p(99)=1.05ms   p(99.99)=1.44ms  
     iterations.....................: 449   4.988826/s
     vus............................: 0     min=0      max=0 
     vus_max........................: 10    min=10     max=10


running (1m30.0s), 00/10 VUs, 449 complete and 0 interrupted iterations
contacts ✓ [======================================] 00/10 VUs  1m30s  8.00 iters/s
samuel@aiko ~/P/i/threadbench (simple)> PUMA_FIBER_PER_REQUEST=true bundle exec bin/bench
* Puma version: 6.4.2-fiber-per-request (ruby 3.3.0-p0) ("The Eagle of Durango")
*  Min threads: 1
*  Max threads: 1
*  Environment: development
*          PID: 974313
*** FIBER PER REQUEST ***
Use Ctrl-C to stop
Warming up for 15 seconds...
Benchmarking...

          /\      |‾‾| /‾‾/   /‾‾/   
     /\  /  \     |  |/  /   /  /    
    /  \/    \    |     (   /   ‾‾\  
   /          \   |  |\  \ |  (‾)  | 
  / __________ \  |__| \__\ \_____/ .io

     execution: local
        script: k6/script.js
        output: -

     scenarios: (100.00%) 1 scenario, 10 max VUs, 2m0s max duration (incl. graceful stop):
              * contacts: Up to 8.00 iterations/s for 1m30s over 1 stages (maxVUs: 10, gracefulStop: 30s)


     data_received..................: 43 kB 475 B/s
     data_sent......................: 36 kB 400 B/s
     http_req_blocked...............: min=148.87µs avg=220.31µs med=211.62µs max=746.67µs p(75)=247.41µs p(99)=289.58µs p(99.99)=743.05µs
     http_req_connecting............: min=104.12µs avg=158.79µs med=152.19µs max=422.86µs p(75)=181.37µs p(99)=223.01µs p(99.99)=415.12µs
     http_req_duration..............: min=385.01µs avg=503.55µs med=491.88µs max=1.1ms    p(75)=534.65µs p(99)=753.35µs p(99.99)=1.09ms  
       { expected_response:true }...: min=385.01µs avg=503.55µs med=491.88µs max=1.1ms    p(75)=534.65µs p(99)=753.35µs p(99.99)=1.09ms  
     http_req_failed................: 0.00% ✓ 0        ✗ 450 
     http_req_receiving.............: min=26.39µs  avg=46.94µs  med=40.88µs  max=131.09µs p(75)=51.93µs  p(99)=103.65µs p(99.99)=130.32µs
     http_req_sending...............: min=25.15µs  avg=44.89µs  med=45.96µs  max=95.79µs  p(75)=50.73µs  p(99)=76.06µs  p(99.99)=95.09µs 
     http_req_tls_handshaking.......: min=0s       avg=0s       med=0s       max=0s       p(75)=0s       p(99)=0s       p(99.99)=0s      
   ✓ http_req_waiting...............: min=278.37µs avg=411.7µs  med=407.46µs max=1.02ms   p(75)=445.78µs p(99)=634.96µs p(99.99)=1ms     
     http_reqs......................: 450   4.999902/s
     iteration_duration.............: min=641.7µs  avg=795.19µs med=779.43µs max=1.55ms   p(75)=822.92µs p(99)=1.12ms   p(99.99)=1.55ms  
     iterations.....................: 450   4.999902/s
     vus............................: 0     min=0      max=1 
     vus_max........................: 10    min=10     max=10


running (1m30.0s), 00/10 VUs, 450 complete and 0 interrupted iterations
contacts ✓ [======================================] 00/10 VUs  1m30s  8.00 iters/s
```

I added extra logging to confirm the implementation behaviour. I did not see any practical differences in either test. That does not mean there aren't differences, just that the trivial "identity" case and "sleep 10ms" cases did not reveal any differences.

ioquatix:
It might be good to address https://github.com/puma/puma/pull/3360 before merging this.

ioquatix:
@MSP-Greg do you have any idea why https://github.com/puma/puma/actions/runs/8539240372/job/23393556676?pr=3101 is failing?

MSP-Greg:
Still need a bit more coffee.  I'll test it locally soon (today).  I've got to switch out OpenSSL 3.2 in my MSYS2...

BTW, thanks for your work on this.

MSP-Greg:
@ioquatix

Looked at the Windows 2.5 issue, and I could verify locally.  Ruby 2.6 works fine, didn't check 2.4.

I thought I'd try to write some code without Puma to see if I can get a better idea what's happening.  It's just stopping, with no indication why...

ioquatix:
If you add print statements to every line, can we see how far it executed the test?

ioquatix:
@MSP-Greg do you have some time now to show me? `https://whereby.com/ioquatix`

ioquatix:
@MSP-Greg I have some time now if that suits you.

ioquatix:
We came to the conclusion that Fibers on Windows, Ruby 2.5, do not work correctly.

ioquatix:
I've rebased this PR, so that it includes the fixes for `Puma::Server.current` which was discovered during my work here: <https://github.com/puma/puma/pull/3360>.

This PR introduces a new configuration option `fiber_per_request`, and aliases it to the old one, `clean_thread_locals`. In fact, the old name is also incorrect, since it only cleans fiber locals. So we can consider deprecating and removing `clean_thread_locals` in v7, or fixing it so that it actually does use a new thread for each request.

While I have not measured it recently, CRuby does cache state relating to thread allocations and fiber allocations, so the overhead of doing this in Ruby/Puma may not make sense. In other words, a thread pool (reusing threads) constructed in Ruby-land may not be an advantage over just writing `Thread.new {...}` per request.

Regarding the naming, I am okay with it, but don't have a strong opinion. It seems to be to clearly represent the behaviour, and that seems good enough to me. There is one environment variable I was using for ease of testing, but I'm not sure if we want to keep that in the default configuration.

In terms of performance, I think the impact will depend on what you are doing, but it's minimal (somewhere between "no impact" and "< 10% of a no-op web request"). In addition, the feature is opt-in. Whether it remains that way in the future, I'm convinced that there are potentially a lot of bugs relating to this - i.e. sharing state between requests. We have known use cases for this at my company, for example, and so we desire to enable this feature for additional safety.

@MSP-Greg any thoughts on what is required to get this merged?

MSP-Greg:
@ioquatix

I'm AFK for the rest of the evening.  I'll have look tomorrow.  Thanks.

MSP-Greg:
@ioquatix

Approved.  Maybe a day or two for others to comment.

> CRuby... a thread pool (reusing threads) constructed in Ruby-land may not be an advantage over just writing Thread.new {...} per request.

Interesting.  Wonder how JRuby and TruffleRuby would be affected...

ioquatix:
> Interesting. Wonder how JRuby and TruffleRuby would be affected...

I don't want to make assertions without looking at the code, but my general feeling is, a lot of these platforms are internally caching the allocations where possible. It's not always possible or semantically correct, so sometimes things don't get reused, but I can attest to the fact that fibers are heavily reused and in a server like Puma, the implementation above will only allocate one "internal" fiber per thread and they will be reused over and over again - the `Fiber.new` instances are different, but the stacks will all be reused over and over again, so there should be very little overhead once warm.

MSP-Greg:
@ioquatix

Thanks for the info.  Given that Puma has `on_thread_start` & `on_thread_exit` hooks, I suspect using "`Thread.new{...}` per request" might cause some issues.  Maybe Puma 7...

Aside from the things you mentioned, using it might make it easier to deal with a timeout 'waiting for an app response'. Not.sure.

ioquatix:
I've changed `Fiber.new do` to `Fiber.new(storage: nil) do` which increases the per-request isolation by preventing inheritance of any existing state.

ioquatix:
After mulling on this issue, I don't think we should force a clean storage (nil) per request. If someone has configured something ahead of time for the fiber storage, it shouldn't be reset. However, I do think gems like `RequestStore` should correctly scope their operations using a middleware.

lib/puma/configuration.rb
```diff
@@ -128,7 +128,7 @@ class Configuration
     DEFAULTS = {
       auto_trim_time: 30,
       binds: ['tcp://0.0.0.0:9292'.freeze],
-      clean_thread_locals: false,
+      fiber_per_request: ENV["PUMA_FIBER_PER_REQUEST"] == "true",
```
schneems: I poked in our codebase and we don't have the concept of using env var `"true"` for boolean values. My request is that we change this to be truthy/falsey rather than checking a specific key. 

```suggestion
      fiber_per_request: ENV.fetch("PUMA_FIBER_PER_REQUEST", false),
```


Longer:

When it comes to boolean values and env vars, I find that it's simpler to use "anything in here" as true, and `unset` as `false`. Otherwise you end up with people who `export PUMA_FIBER_PER_REQUEST=1` or `PUMA_FIBER_PER_REQUEST=on` or `PUMA_FIBER_PER_REQUEST=enabled` and then they are confused when it doesn't work. 

The first downside is that `PUMA_FIBER_PER_REQUEST=false` (or friends) would evaluate to be "true", but this is not surprising to an experienced rubyist as all env vars are strings and a "false" string is truthy. 

The other downside (is bigger) is that there's no way to over-write a prior value via something like `heroku config` which can only set values or not-set values, it cannot over-write an externally set env var. In practice it's not been a major problem, and so I still recommend truthy/falsey env vars.

schneems:
I'm wanting to get v7 out before Rails World and I tagged this with `v7`. I looked over the PR. It seems that: 

- Thread local cleaning is still happening, but via a different mechanism (no breaking change)
- Thread local cleaning is still defaulting to off (no breaking change)
- This enables use of Fiber.storage (feature addition)

Overall that seems good. I'm somewhat curious about the prevelance of usaage of `Fiber.storage` out of curiosity (wouldn't gate this based on the response).

I made a change request for the env var configuration, there are also some conflicts. Can you resolve those and pending any other blocking requests by puma committers I think this can be merged. Thanks for continuing to carry the ruby ecosystem Fiber torch! (Though aren't all real burning torches "fiber" torches?)

ioquatix:
@copilot can you please rebase and implement the required change to the environment variable handling?

History.md
```diff
@@ -1,3 +1,10 @@
+## Unreleased
+
+* Features
+  * Add `fiber_per_request` option (aliased as `clean_thread_locals`) to isolate fiber-local state per request using Ruby Fibers. Enable via config or `PUMA_FIBER_PER_REQUEST` env var. ([#3101])
+
+[#3101]: https://github.com/puma/puma/pull/3101
+
```
schneems: Missed this the first time. We have a script for generating the changelog. IMHO manual changelog entries are better quality (written the time of change). But keeping with consistency:

```suggestion
```

Suggesting here so I can cut these lines and then merge.

schneems:
Thanks!

ioquatix:
No problem!

PR #3105, opened 2023-03-25
Unlock support for the CONNECT HTTP method

### LINKED ISSUE #3014: Puma 6 hard-codes "allowed" HTTP methods

gonubana:
### Description
Unlock support for the CONNECT HTTP method by selectively reverting breaking changes from #2932.

Enhances #2932

Closes #3014

### DIFF
--- a/lib/puma/const.rb
+++ b/lib/puma/const.rb
@@ -147,7 +147,6 @@ module Puma
 
     REQUEST_METHOD = "REQUEST_METHOD"
     HEAD = "HEAD"
-    SUPPORTED_HTTP_METHODS = %w[HEAD GET POST PUT DELETE OPTIONS TRACE PATCH].freeze
     # ETag is based on the apache standard of hex mtime-size-inode (inode is 0 on win32)
     LINE_END = "\r\n"
     REMOTE_ADDR = "REMOTE_ADDR"
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -93,13 +93,8 @@ module Puma
       env[RACK_AFTER_REPLY] ||= []
 
       begin
-        if SUPPORTED_HTTP_METHODS.include?(env[REQUEST_METHOD])
-          status, headers, app_body = @thread_pool.with_force_shutdown do
-            @app.call(env)
-          end
-        else
-          @log_writer.log "Unsupported HTTP method used: #{env[REQUEST_METHOD]}"
-          status, headers, app_body = [501, {}, ["#{env[REQUEST_METHOD]} method is not supported"]]
+        status, headers, app_body = @thread_pool.with_force_shutdown do
+          @app.call(env)
         end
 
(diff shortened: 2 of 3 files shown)


gonubana:
This is a working HTTP tunnel proxy:
```ruby
require 'rack'

class VerbConnectApp
  def call(env)
    [
      200,
      {
        'rack.hijack' => (
          if env['rack.hijack?'] && env['REQUEST_METHOD'] == 'CONNECT'
            proc { |_h_io| tunnel(env) }
          end
        ),
      }.compact,
      [], # Ignored when the response is hijacked.
    ]
  end

  def tunnel(env)
    client_socket = env['rack.hijack'].call

    host, port = env['HTTP_HOST'].split(':')
    dst_socket = TCPSocket.new(host, port)

    peers = {client_socket => dst_socket, dst_socket => client_socket}
    readers = peers.keys
    loop do
      rs, _ = IO.select(readers)
      break if :terminating == rs.each do |r_io|
        peers[r_io].syswrite(r_io.sysread(2**15))
      rescue EOFError => e
        break :terminating
      rescue Errno::ECONNRESET => e
        # Tried to read from an abortively closed TCPSocket.
        break :terminating
      rescue Errno::EPIPE => e
        # Tried to write on a closed socket.
        break :terminating
      rescue => e
        # Abort on errors.
        break :terminating
      end
    end

    begin
      dst_socket.close
    rescue
      nil
    end
  ensure
    client_socket.close
  end
end

run VerbConnectApp.new
```
To use it (after making sure it's running on localhost:9292):
```
curl -vx http://localhost:9292 https://example.com
```

MSP-Greg:
Given https://www.rfc-editor.org/rfc/rfc9110.html#section-9.1, which includes `CONNECT`, maybe add it to `SUPPORTED_HTTP_METHODS`?

In #3014, @kyledrake mentions an issue with methods listed in the 'official' registry, https://www.iana.org/assignments/http-methods/http-methods.xhtml that are not included in `SUPPORTED_HTTP_METHODS`.

Maybe it's time to add a config option so users can add additional methods, maybe something like `additional_http_methods`?

My opinion, not sure what others think...

gonubana:
The #2932 PR did 2 things:

1. It allowed CONNECT requests (objects of the #1441 issue) to be properly passed to the webapp by converting `nil` REQUEST_PATHs into empty strings. [This alone was enough to *positively* close the #1441 issue]
2. But, this same PR inadvertently hid this newly added capability (along with causing the #3014 issue) by gratuitously introducing the SUPPORTED_HTTP_METHODS. [This resulted in *negatively* closing the #1441 issue]

This present PR is trying to keep the 1st effect of the #2932 PR while discarding its 2nd effect.

gonubana:
In other words, as no one until now has ever asked for this specific feature of explicitly discriminating what HTTP methods puma support from those it doesn't, it shouldn't have been implemented, or so I think.

Issue #3108, opened 2023-03-28
Running JRuby 9.4 and Puma with SSL


treviateo:
Hi All,

I am trying to run JRuby 9.4.2.0 with ruby 3.1 and Puma 5.6.5 with SSL. 

**System Configuration:**
JRuby: 9.4.2.0
Puma: 5.6.5
Open JDK: 1.8.0_322

I downloaded the sample keystore from _https://github.com/puma/puma/blob/429d17bca11f8d22dcc1434c9f6ac826a06fa836/examples/puma/keystore.jks_, password blahblah. And converted it to PKCS12 format.

The keystore looks ok as following:
<img width="530" alt="keystore" src="https://user-images.githubusercontent.com/129143940/228134030-c784608d-629a-49a7-bed8-2f9ced5d6b55.png">

When running JRuby with Puma and SSL binding as following:
**C:\jruby-9.4.2.0\bin\jruby.exe C:\[path_to_puma]\bin\puma -b "ssl://127.0.0.1:9292?keystore=C:\.ssl\keystore&keystore-pass=blahblah&verify_mode=none" -q -p 9191**

the server starts up ok as following:
<img width="332" alt="console1" src="https://user-images.githubusercontent.com/129143940/228134199-1afa6bd9-0557-4b70-88f5-6c60e7d4f20c.png">
<img width="412" alt="console2" src="https://user-images.githubusercontent.com/129143940/228134228-c81137e4-ace2-4db9-92ea-9243d755967b.png">

However when trying to curl (_curl https://localhost:9292 -v_), it failed as following:
<img width="308" alt="curl" src="https://user-images.githubusercontent.com/129143940/228134375-1d6f2de6-fed5-4c10-a955-34a08154aa94.png">

And on the server console, this error keep popping up everytime the curl command is entered:
**Error reached top of thread-pool: java.nio.ByteBuffer.clear()Ljava/nio/ByteBuffer; (Java::JavaLang::NoSuchMethodError)**
<img width="854" alt="error" src="https://user-images.githubusercontent.com/129143940/228141625-240d5143-0812-46b8-9feb-efb41d355b05.png">

Does anyone have a hunch about this and could help with this issue? Thank you.

Issue #1772, opened 2019-04-14
JRuby is broken on Puma with Java 8.


enebo:
Java 8 is still supported by JRuby and when running puma we get a wrong flip() signature bug.

```text
2019-04-13 19:08:34 -0500: Listen loop error: java.lang.NoSuchMethodError: java.nio.ByteBuffer.flip()Ljava/nio/ByteBuffer;
org.jruby.puma.MiniSSL$MiniSSLBuffer.flip(MiniSSL.java:68)
org.jruby.puma.MiniSSL.initialize(MiniSSL.java:182)
org.jruby.puma.MiniSSL$INVOKER$i$1$0$initialize.call(MiniSSL$INVOKER$i$1$0$initialize.gen)
org.jruby.internal.runtime.methods.JavaMethod$JavaMethodN.call(JavaMethod.java:833)
org.jruby.runtime.callsite.CachingCallSite.cacheAndCall(CachingCallSite.java:296)
org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:82)
org.jruby.RubyClass.newInstance(RubyClass.java:915)
org.jruby.puma.MiniSSL.server(MiniSSL.java:134)
org.jruby.puma.MiniSSL$INVOKER$s$1$0$server.call(MiniSSL$INVOKER$s$1$0$server.gen)
org.jruby.runtime.callsite.CachingCallSite.cacheAndCall(CachingCallSite.java:339)
org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:170)
org.jruby.ir.interpreter.InterpreterEngine.processCall(InterpreterEngine.java:317)
org.jruby.ir.interpreter.StartupInterpreterEngine.interpret(StartupInterpreterEngine.java:72)
org.jruby.ir.interpreter.InterpreterEngine.interpret(InterpreterEngine.java:80)
org.jruby.internal.runtime.methods.MixedModeIRMethod.INTERPRET_METHOD(MixedModeIRMethod.java:138)
org.jruby.internal.runtime.methods.MixedModeIRMethod.call(MixedModeIRMethod.java:125)
org.jruby.internal.runtime.methods.DynamicMethod.call(DynamicMethod.java:192)
org.jruby.runtime.callsite.CachingCallSite.cacheAndCall(CachingCallSite.java:318)
org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:139)
[...]

PR #3109, opened 2023-03-28
fix java 8 compatibility

ahorek:
### Description
it's a similar issue as https://github.com/puma/puma/issues/1772 unfortunately, that PR didn't fix all usage of buffers

since https://github.com/rake-compiler/rake-compiler/issues/200 does support the ```--release``` flag for compatibility with older java versions, it's cleaner and safer to fix it this way instead of explicitly casting buffer methods as `(Buffer) buffer`

the downside is, it requires at least Java 9 for building the extension, but I guess it won't be a problem.

closes https://github.com/puma/puma/issues/3108

the only way to test it would be to build the extension with Java 9+, but run tests with Java 8.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,7 @@ source "https://rubygems.org"
 
 gemspec
 
-gem "rake-compiler", "~> 1.1.1"
+gem "rake-compiler", "~> 1.1.9"
 
 gem "json", "~> 2.3"
 gem "nio4r", "~> 2.0"
--- a/Rakefile
+++ b/Rakefile
@@ -76,8 +76,7 @@ else
 
   Rake::JavaExtensionTask.new("puma_http11", gemspec) do |ext|
     ext.lib_dir = "lib/puma"
-    ext.source_version = '1.8'
-    ext.target_version = '1.8'
+    ext.release = '8'
   end
 end
 


nateberkopec:
Does this change our compatibility with older Java versions?

kares (approved):
LGTM

> Does this change our compatibility with older Java versions?

It will require Puma's JRuby ext to be compiled on Java >= 9 but the resulting gem (when the extension is build) stays compatible with Java 8. 

Also according to CI there aren't any targets running with Java 8, which is fine it's just that with the release flag added those wouldn't work (there's a way around with having to switch Java versions while building the ext and running the test suite).

ahorek:
> Does this change our compatibility with older Java versions?

Java 8 should be the minimum version for supported Puma and JRuby versions. According to the Puma 6.x changelog:
> No longer supporting Java 1.7 or below (JRuby 9.1 was the last release to support this) (https://github.com/puma/puma/pull/2849)

https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-source-and-target.html

> Merely setting the target option does not guarantee that your code actually runs on a JRE with the specified version. The pitfall is unintended usage of APIs that only exist in later JREs which would make your code fail at runtime with a linkage error. Better yet use the [release option supported since JDK 9](https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-release.html).

> It will require Puma's JRuby ext to be compiled on Java >= 9 but the resulting gem (when the extension is build) stays compatible with Java 8.

if we also want to keep Java 8 compatibility for building the gem (maintainers, java 8 tests, or people who want to build the gem from the source on their own)
we can toggle the flag based on the Java version (which would be better to do in the rake-compiler itself). But does it matter? Java 8 is really old...

nateberkopec:
I'm fine dropping support for Java 8, but it just makes this a breaking change that needs to wait for Puma 7.

kares:
> I'm fine dropping support for Java 8, but it just makes this a breaking change that needs to wait for Puma 7.

Just to clarify we do not do that here, the discussion Pavel and I had is mostly about being able to build from source on 8.
So the gem with this patch included should not break Java 8 compat, only the possibility to build from source on Java 8 (that would require more work but I think @ahorek  and I agree it's not needed and that we can revisit that later, if anyone complains).

ahorek:
this change doesn't intend to break any existing compatibility. Regular users won't be affected, they still can use Java 8 + Puma.

here's a workaround https://github.com/rake-compiler/rake-compiler/pull/213 but this is just for building a gem from the source (maintainers, tests on old environments with Java 8 etc.). I don't think it's a blocker.

nateberkopec:
Thanks for the clarification 🙇

PR #3120, opened 2023-04-05
[ci skip] Add a note that puma-plugin-systemd is redundant now

ekohl:
### Description

By now Puma has gained all systemd integrations that puma-plugin-systemd used to provide. While I didn't check, there's a good chance the plugin now conflicts with the integrated features conflict. Users should prefer native support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -375,7 +375,7 @@ Community guides:
 
 * [puma-metrics](https://github.com/harmjanblok/puma-metrics) — export Puma metrics to Prometheus
 * [puma-plugin-statsd](https://github.com/yob/puma-plugin-statsd) — send Puma metrics to statsd
-* [puma-plugin-systemd](https://github.com/sj26/puma-plugin-systemd) — deeper integration with systemd for notify, status and watchdog
+* [puma-plugin-systemd](https://github.com/sj26/puma-plugin-systemd) — deeper integration with systemd for notify, status and watchdog. Puma 5.1.0 integrated notify and watchdog, which probably conflicts with this plugin. Puma 6.1.0 added status support which obsoletes the plugin entirely.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -375,7 +375,7 @@ Community guides:
 
 * [puma-metrics](https://github.com/harmjanblok/puma-metrics) — export Puma metrics to Prometheus
 * [puma-plugin-statsd](https://github.com/yob/puma-plugin-statsd) — send Puma metrics to statsd
-* [puma-plugin-systemd](https://github.com/sj26/puma-plugin-systemd) — deeper integration with systemd for notify, status and watchdog
+* [puma-plugin-systemd](https://github.com/sj26/puma-plugin-systemd) — deeper integration with systemd for notify, status and watchdog; now all integrated in Puma itself
```
dentarg, line 378: What about removing the link? Or quantify 'now' by stating from which Puma version everything is included

ekohl: I also considered removal, but no idea about what's desired.

ekohl: Removing is best for the reader, because it's no longer relevant. It doesn't give credit to the original author, whose work was a big inspiration for the native integration I did, but I wonder if that's a concern. If you go to https://puma.io/puma4/ it will still have the README from that version, mentioning the plugin so adding a version isn't that useful IHMO.

So the question comes down to credit. Do we feel this is needed, and if so how? Perhaps a code comment in `lib/puma/plugin/systemd.rb`?

dentarg: Let's just remove the link

nateberkopec: Could also say "Unnecessary since <PUMA_VER> and remove later".

nateberkopec: I think I would be in favor of saying "not necessary since VERSION" because people reading the README are not necessarily using the current Puma version, since they're often on Github viewing the master branch. We can remove the link in a year or two.

nateberkopec: Wonderful! Very informative.

PR #3128, opened 2023-04-24
Test system update

MSP-Greg:
### Description

1. Currently, all tests are wrapped by `::Timeout.timeout`.  Switched to using a constant defined in each test class.  This PR changes so that only test classes that use some form of `bundle install` are using the timeout.

2. Created `test/helpers/puma_socket.rb`, which allows removal of duplicated code in several test files. This will also make it easier to re-arrange and/or split test files into more (smaller) files.

3. Made minor updates to some library files, most of which related to error handling.

4. Currently, some of the tests fail if a client socket raises `Errno::ECONNRESET`.  I've got some code that queries the GitHub API with `Net::HTTP`, and it threw the same error, as GitHub has had network issues.  Using the same URI, I tried several browsers, and they all retry on `Errno::ECONNRESET`.  Changed test code to allow a small percentage of `Errno::ECONNRESET` errors.

5. CI is often running a full allocation of macos jobs at the end.  Removed a few macos jobs to quicken the test time.  This may be needed again, as GitHub will be supporting macos-13 and also `arm64` macos.

Creating as a draft, not sure whether to break it up into lib file changes and test file changes, etc...

### DIFF
--- a/test/config/event_on_booted_exit.rb
+++ b/test/config/event_on_booted_exit.rb
@@ -1,5 +1,8 @@
 on_booted do
   pid = Process.pid
   Process.kill :TERM, pid
-  Process.wait pid
+  begin
+    Process.wait2 pid
+  rescue Errno::ECHILD
+  end
 end
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -68,16 +68,37 @@ def hit(uris)
   end
 end
 
+# Some platforms may repeat port numbers when selecting several at once.
 module UniquePort
+  MUTEX = Mutex.new
+  SELECTED_PORTS = {}
   def self.call
-    TCPServer.open('127.0.0.1', 0) do |server|
-      server.connect_address.ip_port
+    port = 0
+    MUTEX.synchronize do
+      begin
+        TCPServer.open('127.0.0.1', 0) { |server|
+          port = server.connect_address.ip_port
+          server.close
+        }
+      rescue Errno::EINTR
+        sleep 0.000_1
+        retry
+      end while SELECTED_PORTS.key? port
+      SELECTED_PORTS[port] = nil
     end
+    port
   end
 end
 
 require "timeout"
 module TimeoutEveryTestCase
+
+  # ANSI color codes for highlighting debugging when frustrated...
+  YELLOW = "\e[93m"
+  RED    = "\e[91m"
+  GREEN  = "\e[92m"
+  RESET  = "\e[0m"
+
   # our own subclass so we never confused different timeouts
   class TestTookTooLong < Timeout::Error
   end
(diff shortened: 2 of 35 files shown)


MSP-Greg:
Fixed up `test_out_of_band_server.rb`, as a JRuby test froze in first CI run.

dentarg:
> Creating as a draft, not sure whether to break it up into lib file changes and test file changes, etc...

Yes, please break it up, I think this is a good tell Screenshot 2023-04-25 at 00 37 03 😄

Going by your list this is probably at least five different PRs?

lib/puma/binder.rb
```diff
@@ -48,7 +48,7 @@ def initialize(log_writer, conf = Configuration.new)
 
       @envs = {}
       @ios = []
-      localhost_authority
+      # localhost_authority
```
dentarg, line 51: Remove this line? Or was it a mistake to comment this out?

dentarg:
GitHub hides the changes to `test/helpers/integration.rb` by default when viewing the diff so it is easy to miss those changes when reviewing.

Not to take anything away from the work done here, it is super great stuff @MSP-Greg 💪  but I do think we should aim for small specific PRs for changes to Puma

lib/puma/binder.rb
MSP-Greg: Not a mistake.  Can't quite remember the particulars, but `localhost_authority` will fail if some  conditions aren't met.  There's no reason to call it if the 'binds' do not include an ssl connection.  I commented it out because I wanted to run the test suite several times, and forgot to remove the line.

MSP-Greg:
> we should aim for small specific PRs for changes to Puma

I would generally agree.  Unfortunately, the test suite is cluster-fu** of repetitive code, intermittent test failures, impossible to break apart into smaller files without more repetition, etc.

I think most of the changes are in the commits c521c5c5eb5e, 1dadb22559da and 2d4ab6cb2b0a.

One is the changes to helper files and the addition of `test/helpers/puma_socket.rb`, and the two others update various test files to work with them.

Or, those three commits are tightly bound and need to be committed together.  Normally I prefer to see PR's squashed, but maybe these should be left as mutiple commits.

I'll see if I can pull the library file commits from this into another PR.

MSP-Greg:
**General Concept**

I started with two goals, first, make the test system more stable, and second, move towards more shared helper methods.

So, I shut off the Timeout thread wrapping each test, shut off the retry, and started running the test suite locally.

One main problem is running tests parallel. These run in [threads](https://github.com/minitest/minitest/blob/v5.18.0/lib/minitest/parallel.rb#L26-L38). If any tests have blocking code, the tests slow down and may fail.

This means that client sockets should always use wait_readable, and read_nonblock would be preferred. Code should also close client sockets. Because of this, we should discourage use of *Socket.new in test code, and provide helper methods. Obviously, there will be exceptions.

Most tests either start a Puma server 'in-process' or spawn a server. In the future, having three helper classes, something like ServerInProc and ServerSpawn, along with a companion socket class would be a good setup. Implementing these classes will make it much easier to re-arrange tests. An example might be [test_puma_server.rb](https://github.com/puma/puma/blob/904b47a0c4347d2d3d7bf8606d1ebd26c6350721/test/test_puma_server.rb), which contains 94 tests.

The current 'socket' helper in this PR only works with TCP & UNIX sockets.  SSL sockets need to be added, and integrated with the inproc and spawn server helpers (inproc is yet to be written).

lib/puma/binder.rb
MSP-Greg: It may have happened when running a single file locally, can't recall.

MSP-Greg:
I've pulled all the code in lib files and Actions workflow files, so this now just contains test updates.  I'll see whether it can be divided into smaller PR's, but it doesn't need to be squashed.

Also, as the goal was to make CI more reliable and also contain less repetitive code, it was a matter of converging toward a solution, so reversing that and splitting it up may be difficult.  Some of the work involved small changes to remove blocking code.

dentarg:
I think we can close this one? The plan is to merge many smaller PRs

We also have https://github.com/puma/puma/pull/3479 showing of it all. And we have the discussion at https://github.com/puma/puma/discussions/3232.

PR #3129, opened 2023-04-25
Misc lib file fixes - trapping additional errors, CI helper

MSP-Greg:
### Description

All files except `thread_pool.rb` have additional handling added.  These errors have intermittently been seen in CI.

`thread_pool.rb` allows setting `options[:pool_shutdown_grace_time]` during CI, as stubbing a constant is difficult when running parallel CI.  At present, it is not exposed via the DSL, but could be added in the future.

Hence, no API or functionality changes.

These commits were originally part of PR #3128.

### DIFF
--- a/lib/puma/binder.rb
+++ b/lib/puma/binder.rb
@@ -48,7 +48,6 @@ module Puma
 
       @envs = {}
       @ios = []
-      localhost_authority
     end
 
     attr_reader :ios
--- a/lib/puma/binder.rb
+++ b/lib/puma/binder.rb
@@ -450,11 +450,14 @@ module Puma
 
     def close_listeners
       @listeners.each do |l, io|
-        io.close unless io.closed?
-        uri = URI.parse l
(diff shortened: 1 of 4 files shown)


nateberkopec (approved):

lib/puma/binder.rb
```diff
@@ -48,7 +48,6 @@ def initialize(log_writer, conf = Configuration.new)
 
       @envs = {}
       @ios = []
-      localhost_authority
```
nateberkopec, line 51: Intentional delete?

MSP-Greg: Yes.  It's called early in the [ssl section](https://github.com/puma/puma/blob/ae81065a0cf7c893a9c60088e71e87685f5c029a/lib/puma/binder.rb#L229) of `parse`.  The call in the above section happens in `initialize`, so it's called regardless of whether an ssl bind is being used.

PR #3141, opened 2023-05-02
Update handling of Server: header in Puma generated responses

### LINKED ISSUE #3037: Option to hide "Server: Puma PUMA_VERSION" header
Currently we use Puma 5.6.4. Our security auditor asks if we can remove `Server: Puma 5.6.4` in HTTP response header because it can potentially give bad actor hints about exploiting our server, if vulnerability is found in certain puma version. 

Add configuration to allow suppressing "server" response header

Remove this header on our load balancer level could be another option but I prefer if we can omit on puma-level

**Additional context**

MSP-Greg:
### Description

Puma generates responses for invalid requests, and these responses are sent before `app.call`.  Currently the response strings are the values of [`Const::ERROR_RESPONSE`](https://github.com/puma/puma/blob/c2ac2c4f7c9c0f3a8ba294350ee70990ff94598f/lib/puma/const.rb#L123-L136), and the constant is used in `Client#write_error`. That method is called three times in conditional code in `Server#client_error`, and once in `Client#timeout!`.  There are two responses sent in `Request#handle_request` (413, 501), but those responses do not use the string values from the constant.

1. Two of the constant values contain a `Server:` header, whose values are `'Puma #{PUMA_VERSION}'`. From a pen test standpoint these headers should not be included.  This PR removes them.

2. Some sites include a `Server:` header, often a non-informational value.  Most GitHub pages return `Server: GitHub.com`.  Since it would be best for Puma's reponses to mimic app responses, a DSL method `server_header_value(val = nil)` has been added, allowing sites using a response `Server:` header to let Puma responses use the same value.

Because `Const` simply defines constants on load/require, we cannot create a constant in it that depends on a Config (DSL) option.  So, the constant definition is removes from `Const`, and defined when `Server.new` is called.

Note that if `DSL#server_header_value` is not used in a config file, no `Server:` headers will be sent with any of Puma's error responses.  Also, since error responses do not have pre-defined bodies, again, from a pen test standpoint, none of the responses include body text.

Closes #3037

### DIFF
--- a/lib/puma/const.rb
+++ b/lib/puma/const.rb
@@ -120,21 +120,6 @@ module Puma
 
     PUMA_TMP_BASE = "puma"
 
-    ERROR_RESPONSE = {
-      # Indicate that we couldn't parse the request
-      400 => "HTTP/1.1 400 Bad Request\r\n\r\n",
-      # The standard empty 404 response for bad requests.  Use Error4040Handler for custom stuff.
-      404 => "HTTP/1.1 404 Not Found\r\nConnection: close\r\nServer: Puma #{PUMA_VERSION}\r\n\r\nNOT FOUND",
-      # The standard empty 408 response for requests that timed out.
-      408 => "HTTP/1.1 408 Request Timeout\r\nConnection: close\r\nServer: Puma #{PUMA_VERSION}\r\n\r\n",
-      # Indicate that there was an internal error, obviously.
-      500 => "HTTP/1.1 500 Internal Server Error\r\n\r\n",
-      # Incorrect or invalid header value
-      501 => "HTTP/1.1 501 Not Implemented\r\n\r\n",
-      # A common header for indicating the server is too busy.  Not used yet.
-      503 => "HTTP/1.1 503 Service Unavailable\r\n\r\nBUSY"
-    }.freeze
-
     # The basic max request size we'll try to read.
     CHUNK_SIZE = 16 * 1024
 
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -109,7 +109,29 @@ module Puma
       @precheck_closing = true
 
       @requests_count = 0
+
+      unless Client.const_defined? :ERROR_RESPONSE
+        define_client_error_response
+      end
     end
+
+    # Creates Client::ERROR_RESPONSE, which is a hash of response strings
+    # used when the request is invalid, and before `app.call`.
+    # @see `Client#write_error`
+    def define_client_error_response
+      err_hash = {}
+      server_val = @options[:puma_server_header_value]&.strip
+      show_header = server_val && !server_val.empty?
+      [400, 408, 413, 500, 501].each do |status|
+        err_hash[status] = show_header ?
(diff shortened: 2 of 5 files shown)


MSP-Greg:
Rebased & fixed typo, specifing -> specifying

MSP-Greg:
> There are two responses sent in Request#handle_request (413, 501), but those responses do not use the string values from the constant.

If this PR is ok, a future PR should move the above responses from `Request#handle_request` into `Client` code, as there's no reason to call `handle_request`` when we know a request is invalid.

lib/puma/const.rb
```diff
@@ -120,21 +120,6 @@ module Const
 
     PUMA_TMP_BASE = "puma"
 
-    ERROR_RESPONSE = {
-      # Indicate that we couldn't parse the request
-      400 => "HTTP/1.1 400 Bad Request\r\n\r\n",
-      # The standard empty 404 response for bad requests.  Use Error4040Handler for custom stuff.
-      404 => "HTTP/1.1 404 Not Found\r\nConnection: close\r\nServer: Puma #{PUMA_VERSION}\r\n\r\nNOT FOUND",
```
nateberkopec, line 127: This change means we no longer have `NOT FOUND` in a 404 and `BUSY` in a 503 response, right?

lib/puma/server.rb
```diff
@@ -109,7 +109,29 @@ def initialize(app, events = nil, options = {})
       @precheck_closing = true
 
       @requests_count = 0
+
+      unless Client.const_defined? :ERROR_RESPONSE
```
nateberkopec, line 113: Rather than do this, would it be cleaner to modify the different call-sites using ERROR_RESPONSE to instead use a class method (maybe `Puma.error_responses`)? This process of defining a constant in response to the DSL is weird to me.

lib/puma/const.rb
MSP-Greg: The original constant has defined 404 and 503 keys, but I don't think Puma used the string values anywhere.

In the first post, I accidentally left out two responses generated in `Request`, a 500 and a 503, which are used in calls to 
`Server#lowlevel_error`.

I think these responses should not be 'specific' to Puma, they should be generic, so one can't tell how they were generated?

But, [9110 15.6. Server Error 5xx](https://www.rfc-editor.org/rfc/rfc9110.html#name-server-error-5xx) states the below, which kind of implies a body...

---

Except when responding to a HEAD request, the server SHOULD send a representation containing an explanation of the error situation, and whether it is a temporary or permanent condition. A user agent SHOULD display any included representation to the user.

lib/puma/server.rb
MSP-Greg: Part of where it gets tricky is that `Client` and `Server`/`Request` aren't tightly bound, but I think `Server`/`Request` can call into `Client` methods, as it creates the `Client` instances. `Client` doesn't have any config/options info.

How we handle Puma generated responses could use some updating, as it's done a few different ways.

Kind of burnt, let me think about it.  Maybe we should just remove the Puma info (the `Server:` header), and take some more time to think about it...

PR #3153, opened 2023-05-13
Store Tempfiles in ./tmp

nateberkopec:
On my machine, tempfile directories end up being rather long, and tests actually fail!

```
/Users/nateberkopec/Documents/Code.nosync/upstream/puma/lib/puma/binder.rb:421:in `initialize': too long unix socket path (106bytes given but 104bytes max) (ArgumentError)

        s = UNIXServer.new path.sub(/\A@/, "\0") # check for abstract UNIXSocket
```

Storing in ./tmp solves the issue.

### DIFF
--- a/test/helpers/tmp_path.rb
+++ b/test/helpers/tmp_path.rb
@@ -8,7 +8,7 @@ module TmpPath
   private
 
   def tmp_path(extension=nil)
-    path = Tempfile.create(['', extension]) { |f| f.path }
+    path = Tempfile.create(['', extension], './tmp') { |f| f.path }
     tmp_paths << path
     path
   end


dentarg:
I think this code is dependent on the fact that the compilation task creates the `./tmp` directory, it fails if you don't have it

```
$ irb -rtempfile
irb(main):001:0> extension=nil
=> nil
irb(main):002:0> Tempfile.create(['', extension], './tmp') { |f| f.path }
/Users/dentarg/.rubies/3.2.2/lib/ruby/3.2.0/tempfile.rb:403:in `initialize': No such file or directory @ rb_sysopen - ./tmp/20230513-22011-4c9bxd (Errno::ENOENT)
```

MSP-Greg:
This broke my local CI for rather unique reasons.

Maybe something like the below?
```ruby
  IS_OSX = RUBY_PLATFORM.include? 'darwin'

  def tmp_path(extension=nil)
    path = IS_OSX ?
      Tempfile.create(['', extension], './tmp') { |f| f.path } :
      Tempfile.create(['', extension]) { |f| f.path }
    tmp_paths << path
    path
  end
```

PR #3163, opened 2023-05-22
Don't send 103 early hints response when only invalid headers are used

MSP-Greg:
### Description

While reviewing code in `test\test_response_header.rb`, I noticed that when invalid headers are returned by the app with 'early hints', the first line of an 'early hints' response (`HTTP/1.1 103 Early Hints`) is sent without any data/headers.

This PR modifies the code in `puma/request.rb` to only send the 'early hints' response if at least one header is valid.

Note that with the test changes, current master has the following errors:
```
TestResponseHeader#test_illegal_character_in_key_when_early_hints [test/test_response_header.rb:128]:
Expected # encoding: ASCII-8BIT
#    valid: true
"HTTP/1.1 103 Early Hints\r\n\r\n" to not include "HTTP/1.1 103 Early Hints".

  2) Failure:
TestResponseHeader#test_illegal_character_in_value_when_early_hints [test_response_header.rb:139]:
Expected # encoding: ASCII-8BIT
#    valid: true
"HTTP/1.1 103 Early Hints\r\n\r\nHTTP/1.0 200 OK\r\nContent-Length: 5\r\n\r\nHello" to not include "HTTP/1.1 103 Early Hints".
```

### DIFF
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -77,7 +77,9 @@ module Puma
       if @early_hints
         env[EARLY_HINTS] = lambda { |headers|
           begin
-            fast_write_str socket, str_early_hints(headers)
+            unless (str = str_early_hints headers).empty?
+              fast_write_str socket, "HTTP/1.1 103 Early Hints\r\n#{str}\r\n"
+            end
           rescue ConnectionError => e
             @log_writer.debug_error e
             # noop, if we lost the socket we just won't send the early hints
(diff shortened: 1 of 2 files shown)


lib/puma/request.rb
```diff
@@ -77,7 +77,9 @@ def handle_request(client, requests)
       if @early_hints
         env[EARLY_HINTS] = lambda { |headers|
           begin
-            fast_write_str socket, str_early_hints(headers)
+            unless (str = str_early_hints headers).empty?
```
nateberkopec, line 80: this double negative is messing with me, but I can't see a better alternative

nateberkopec (approved):

PR #3165, opened 2023-05-28
update URLMap Regexp usage for Ruby v3.3

fallwith:
Ruby v3.3.0-preview1 doesn't support passing 3 arguments to `Regexp.new`, so update `Puma::Rack::URLMap#remap` to use a 2 argument version instead.

I didn't notice where `Puma::Rack::URLMap#remap` was currently tested and could use some advice on the testing front. Syntactically, I've tested the change going back as far as the current minimum supported Ruby, v2.4.

### DIFF
--- a/lib/puma/rack/urlmap.rb
+++ b/lib/puma/rack/urlmap.rb
@@ -34,7 +34,7 @@ module Puma::Rack
         end
 
         location = location.chomp('/')
-        match = Regexp.new("^#{Regexp.quote(location).gsub('/', '/+')}(.*)", nil, 'n')
+        match = Regexp.new("^#{Regexp.quote(location).gsub('/', '/+')}(.*)", Regexp::NOENCODING)
 
         [host, location, match, app]
       }.sort_by do |(host, location, _, _)|
--- a/test/test_url_map.rb
+++ b/test/test_url_map.rb
@@ -2,10 +2,6 @@ require_relative "helper"
 require_relative "helpers/integration"
 
 class TestURLMap < TestIntegration
-  def setup
-    skip_unless :fork
-    super
-  end
 
   def teardown
     return if skipped?
@@ -14,23 +10,15 @@ class TestURLMap < TestIntegration
 
   # make sure the mapping defined in url_map_test/config.ru works
   def test_basic_url_mapping
-    skip_unless_signal_exist? :USR2
-
-    @tcp_port = UniquePort.call
-    timeout = 1
-    env = {
-      "BUNDLE_GEMFILE" => File.join(File.expand_path("url_map_test/Gemfile", __dir__))
-    }
-    cmd = "bundle exec puma -q -w 1 --prune-bundler -b tcp://#{HOST}:#{@tcp_port}"
-    Dir.chdir(File.expand_path("url_map_test", __dir__)) do
-      @server = IO.popen(env, cmd.split, "r")
+    skip_if :jruby
+    env = { "BUNDLE_GEMFILE" => "#{__dir__}/url_map_test/Gemfile" }
+    Dir.chdir("#{__dir__}/url_map_test") do
+      cli_server set_pumactl_args, env: env
     end
-    wait_for_server_to_boot
-    @pid = @server.pid
     connection = connect("/ok")
     # Puma 6.2.2 and below will time out here with Ruby v3.3
     # see https://github.com/puma/puma/pull/3165
-    initial_reply = read_body(connection, timeout)
-    assert_match("OK", initial_reply)
+    body = read_body(connection, 1)
+    assert_equal("OK", body)
   end
 end
--- a/test/test_url_map.rb
+++ b/test/test_url_map.rb
@@ -1,0 +1,36 @@
+require_relative "helper"
+require_relative "helpers/integration"
+
+class TestURLMap < TestIntegration
+  def setup
+    skip_unless :fork
+    super
+  end
+
+  def teardown
+    return if skipped?
+    super
+  end
+
+  # make sure the mapping defined in url_map_test/config.ru works
+  def test_basic_url_mapping
+    skip_unless_signal_exist? :USR2
+
+    @tcp_port = UniquePort.call
+    timeout = 1
+    env = {
+      "BUNDLE_GEMFILE" => File.join(File.expand_path("url_map_test/Gemfile", __dir__))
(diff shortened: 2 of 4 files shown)


MSP-Greg:
@fallwith

Thanks for the PR.  Wondering if we should just pull in the code from Rack, as it was probably done in 5c9f871771?

Not sure about using `set`, but that's an easy change.  Would need to remove the `constants` require, and add:
```ruby
  class URLMap
    HTTP_HOST   = 'HTTP_HOST'
    PATH_INFO   = 'PATH_INFO'
    SCRIPT_NAME = 'SCRIPT_NAME'
    SERVER_NAME = 'SERVER_NAME'
    SERVER_PORT = 'SERVER_PORT'
```

dentarg:
> Wondering if we should just pull in the code from Rack, as it was probably done in [5c9f871](https://github.com/puma/puma/commit/5c9f871771e1be7fb6a917dfff21bda31aa56693)?

Sounds resonable

dentarg:
> I didn't notice where `Puma::Rack::URLMap#remap` was currently tested and could use some advice on the testing front.

Add a test app like this somewhere?

```ruby
$ cat config.ru
map "/ok" do
  run ->(env) { [200, {}, ["OK"]] }
end
```

It currently blows up on Ruby 3.3 like this (as you already know, but including this for extra context)

```ruby
arm64 $ b e puma
* Puma version: 6.2.2 (ruby 3.3.0-p-1) ("Speaking of Now")
*  Min threads: 0
*  Max threads: 5
*  Environment: development
*          PID: 81381
! Unable to load application: ArgumentError: wrong number of arguments (given 3, expected 1..2)
bundler: failed to load command: puma (/Users/dentarg/.arm64_rubies/3.3.0-preview1/bin/puma)
/Users/dentarg/.arm64_rubies/3.3.0-preview1/lib/ruby/gems/3.3.0+0/gems/puma-6.2.2/lib/puma/rack/urlmap.rb:37:in `initialize': wrong number of arguments (given 3, expected 1..2) (ArgumentError)

        match = Regexp.new("^#{Regexp.quote(location).gsub('/', '/+')}(.*)", nil, 'n')
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
	from /Users/dentarg/.arm64_rubies/3.3.0-preview1/lib/ruby/gems/3.3.0+0/gems/puma-6.2.2/lib/puma/rack/urlmap.rb:37:in `new'
	from /Users/dentarg/.arm64_rubies/3.3.0-preview1/lib/ruby/gems/3.3.0+0/gems/puma-6.2.2/lib/puma/rack/urlmap.rb:37:in `block in remap'
...
```

dentarg:
> Add a test app like this somewhere?

I suspect it isn't that easy, as the `Gemfile` in CI will always contain `rack` :/

@fallwith how did you notice this issue? do you run any application using Puma where you haven't included `rack` in your bundle?

MSP-Greg:
I think this also occurs with Ruby 3.2, as the change to Regexp happened there... 

As to testing, we can probably get around the issue by disabling loading rack install in the Gemfile with an ENV setting, then using a test that loads Puma in a sub process.

I can investigate, but this isn't a good day for that, tomorrow is better...

dentarg:
> I think this also occurs with Ruby 3.2, as the change to Regexp happened there...

The example I tried (see above) does not blow up on Ruby 3.2.2-p53

dentarg:
> As to testing, we can probably get around the issue by disabling loading rack install in the Gemfile with an ENV setting, then using a test that loads Puma in a sub process.

Yes

MSP-Greg:
> The example I tried (see above) does not blow up on Ruby 3.2.2-p53

Thanks & sorry.

I didn't try it with 3.2, but saw the change in the doc's of the method signature.

I didn't notice that it was a deprecation, see https://github.com/rack/rack/pull/1998

fallwith:
Hi @MSP-Greg and @dentarg,

> @fallwith how did you notice this issue? do you run any application using Puma where you haven't included rack in your bundle?

I work for New Relic and we use a [suite of tests](https://github.com/newrelic/newrelic-ruby-agent/tree/dev/test/multiverse/suites/rack) to ensure compatibility with Rack. When it comes to testing Puma, we dynamically generate 4 separate `Gemfile` files, one each for Puma versions 3, 4, 5, and 6. All of those `Gemfile` files contain both `rack` and `rack-test` alongside `puma`. We do, however, directly instantiate `Puma::Rack::URLMap` with one specific [test](https://github.com/newrelic/newrelic-ruby-agent/blob/dev/test/multiverse/suites/rack/puma_rack_urlmap_test.rb), so we're always directly testing the code in question.

The New Relic Ruby agent began instrumenting `Puma::Rack::URLMap` 8 years ago in 2015 with [this commit](https://github.com/newrelic/newrelic-ruby-agent/commit/98473040354ba596acc58bd46f0e679a53c2adc0), and that's why we directly test the class.

If I can answer anything else or contribute any further content, please let me know.

dentarg:
> If I can answer anything else or contribute any further content, please let me know.

Please add a test as described above

fallwith:
Test added with 8374a5b

MSP-Greg:
Ruby head builds are broken due to Ruby changes that prevent nio4r from compiling.  The Ubuntu-22.04 3.2 job failure is due to an intermittent, unrelated failure.

MSP-Greg:
ruby head builds had an ABI change that nio4r didn't like.  It's been reverted, and the head builds have been updated in GitHub Actions.

So, I reran the failing heads builds here.

nateberkopec (approved):
LGTM but since Greg and Dentarg worked on this more closely I'll leave it to you to merge.

MSP-Greg:
@fallwith

Thanks for your work on this.  The current test could run on Windows, but won't, due to the `-w1` in the CLI args (workers require `fork`).

I updated that and a few other things.  Patch is at https://github.com/puma/puma/commit/87294d45be.

fallwith:
Thanks, @MSP-Greg. I have brought over your patch with d0a7e7e.

MSP-Greg:
I added a check to make sure Rack is not loaded in the spawned Puma process.  Unlikely to fail, but checking it verifies that the test is running in the proper environment.  Prevents accidental failures in the future...

PR #3179, opened 2023-06-11
context_builder.rb - require openssl if verify_mode != 'none'

MSP-Greg:
### Description

Current code in `MiniSSL::Socket#peercert` contains the following line:
https://github.com/puma/puma/blob/188f5da1920ff99a8689b3e9b46f2f26b7c62d66/lib/puma/minissl.rb#L194

The above statement only runs when `MiniSSL::Context#verify_mode` is not equal to `MiniSSL::VERIFY_NONE`.  So, we need to require openssl when needed.

Note that `MiniSSL::Socket#peercert` was added in PR #709 'ssl: Add Client Side Certificate Auth'.  The cert subject is used in `LogWriter#ssl_error`, and the cert object is set to `env['puma.peercert']`.  The app environment key doesn't seem to be 'publicly' listed.

Current tests only check this with an in-process server, and since the client sockets are ssl, `openssl` is loaded for the clients.

I added some additional spawned tests (Integration) that showed the problem, but that was using my updated test framework.  We'll see...

### DIFF
--- a/ext/puma_http11/mini_ssl.c
+++ b/ext/puma_http11/mini_ssl.c
@@ -804,6 +804,10 @@ void Init_mini_ssl(VALUE puma) {
 
   rb_define_method(eng, "init?", engine_init, 0);
 
+  /* @!attribute [r] peercert
+   * Returns `nil` when `MiniSSL::Context#verify_mode` is set to `VERIFY_NONE`.
+   * @return [String, nil] DER encoded cert
+   */
   rb_define_method(eng, "peercert", engine_peercert, 0);
 
   rb_define_method(eng, "ssl_vers_st", engine_ssl_vers_st, 0);
--- a/lib/puma/minissl.rb
+++ b/lib/puma/minissl.rb
@@ -184,6 +184,11 @@ module Puma
         @socket.peeraddr
       end
 
+      # OpenSSL is loaded in `MiniSSL::ContextBuilder` when
+      # `MiniSSL::Context#verify_mode` is not `VERIFY_NONE`.
+      # When `VERIFY_NONE`, `MiniSSL::Engine#peercert` is nil, regardless of
+      # whether the client sends a cert.
+      # @return [OpenSSL::X509::Certificate, nil]
       # @!attribute [r] peercert
       def peercert
         return @peercert if @peercert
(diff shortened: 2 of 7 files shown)


nateberkopec:
This seems like a weird place to put the `require` call, maybe I don't get it.

Why not require in #peercert?

MSP-Greg:
> Why not require in #peercert?

Don't have strong opinion, but I thought locating where it is happens on 'startup', rather than when the first request happens.  

Also, in peercert, it will run on every request?  Note that with VERIFY_NONE, even if a client includes a cert, `MiniSSL::Engine#peercert` is nil, which means that `OpenSSL::X509::Certificate.new` is never called.

MSP-Greg:
@nateberkopec

Actions is kind of wacked right now.

I added comments to a few methods for this.

dentarg:
> I added some additional spawned tests (Integration) that showed the problem, but that was using my updated test framework. We'll see...

Would be nice to commit the tests with this change

MSP-Greg:
> Would be nice to commit the tests with this change

Ok, I added two test helper files from my test update.

1. `test/helpers/constants.rb` - this contains constants pertaining to Puma's test suite.  I'd like to see `test/helper.rb` just contain code related to minitest.  One feature is that one can swap all HOST addresses between IPv4 & IPv6 for testing.  Haven't finished it yet, as it needs a way to toggle it with an ENV setting.  It also checks system localhost addresses, and if they've been reconfigured, it will use those.

2. `test/helpers/puma_socket.rb` - this contains code to generate requests, etc.  Currently, code is scattered throughout the test suite, and the methods here allow much more functionality than the current code.

The main idea is that there should be two helpers files, one based on the server related code in ``test/helpers/integration.rb` with additions for ssl sockets, etc.  Tentatively, I thought it code be called `server_spawn.rb`, because it handles spawning a server in a sub-process.  There would be an additional file, maybe named `server_in_process.rb`, which would contain code to start a `Puma::Server` instance, similar to the code scattered through several test files.  Examples - `test_puma_server.rb'.  These two server files would share instance variable names with `puma_socket.rb`, allowing the test methods to no longer have deal with socket parameters in Puma or client related code.

Re this PR, if one checks out the third commit which is before any library file changes, the following is shown in the test listing:
```
2023-06-19 20:56:58 -0500 Read: #<NameError: uninitialized constant Puma::MiniSSL::Socket::OpenSSL>
```

The new test (`TestIntegrationSSL#test_verify_client_cert_roundtrip`) also fails.

dentarg:
Sounds good to me

`PumaSocket` is a lot of code so I can't say I've studied it very well, but I trust your chops @MSP-Greg and I'm sure I will read it again in the future... :)

PR #3166, opened 2023-05-28
Prevent loading with rack 3

JoeDupuis:
WIP tests are failing

### Description
See https://github.com/puma/puma/pull/3164

Puma 5 is not compatible with Rack 3. App generated with the default Rails 6.1 and 7.0 app template will hit [this error](https://github.com/rails/rails/issues/48195) when they  run `bundle update`.

This check ensures affected users will get a meaningful error message pointing them toward a fix. 

### Your checklist for this pull request

- [x] I have reviewed the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.
- [ ] I have added (or updated) appropriate tests if this PR fixes a bug or adds a feature.
- [x] My pull request is 100 lines added/removed or less so that it can be easily reviewed.
- [x] If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.
- [x] If this closes any issues, I have added "Closes `#issue`" to the PR description or my commit messages.
- [x] I have updated the documentation accordingly.
- [ ] All new and existing tests passed, including Rubocop.

PR #3182, opened 2023-06-21
5.6.6 release note

nateberkopec:


MSP-Greg:
History.md shows 'Allow Puma to be loaded with Rack 3', but PR #3166 is titled 'Prevent loading with rack 3'.

Not sure if that intentional...

PR #3191, opened 2023-07-05
Remove `Forwardable` dependency (possible conversion to bundled gem)

MSP-Greg:
### Description

[Forwardable](https://github.com/ruby/forwardable) is currently used by server.rb & client.rb.

It is part of the standard library, but it's also available as a gem. Hence, it could be 'converted' to a bundled gem, especially since it's nor an extension gem.

Bundled gems need to be included as dependencies in gemspec files.

Since removing it is relatively trivial (5 methods are delegated), lets do so now.

Closes #3190
Closes https://github.com/puma/puma/issues/3218

### DIFF
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -11,7 +11,6 @@ end
 require_relative 'detect'
 require_relative 'io_buffer'
 require 'tempfile'
-require 'forwardable'
 
 if Puma::IS_JRUBY
   # We have to work around some OpenSSL buffer/io-readiness bugs
@@ -61,7 +60,6 @@ module Puma
     EmptyBody = NullIO.new
 
     include Puma::Const
-    extend Forwardable
 
     def initialize(io, env=nil)
       @io = io
@@ -110,7 +108,10 @@ module Puma
 
     attr_accessor :remote_addr_header, :listener
 
-    def_delegators :@io, :closed?
+    # Remove in Puma 7?
+    def closed?
+      @to_io.closed?
+    end
 
     # Test to see if io meets a bare minimum of functioning, @to_io needs to be
     # used for MiniSSL::Socket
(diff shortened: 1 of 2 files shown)


lib/puma/client.rb
```diff
@@ -110,7 +108,10 @@ def initialize(io, env=nil)
 
     attr_accessor :remote_addr_header, :listener
 
-    def_delegators :@io, :closed?
+    # Remove in Puma 7?
+    def closed?
+      @to_io.closed?
```
dentarg, line 113: Why the change from `@io` to `@to_io`?

(Maybe that's better, I don't know)

dentarg:
I think it would be good if the commit message for this included the motivation behind doing it

dentarg:
It isn't really spelled out in #3190 but there is this 

> It is part of the standard library, but it's also available as a gem. Hence, it could be 'converted' to a bundled gem. Bundled gems need to be included as dependencies in gemspec files.

And we want to avoid the use of bundled gems in Puma so phased restart works well, right?

MSP-Greg:
> And we want to avoid the use of bundled gems in Puma so phased restart works well, right?

Yes, but that's really more critical with extension gems, which I don't think can be reloaded.  If a bundled gem is a dependency, we'd need to include it in Puma's gemspec for Bundler to properly handle it.

I can't recall what gems/repos it involved, but `racc` was recently converted to a bundled gem, and it broke CI here and there...

I'll add some of the comments from the discussion to the PR.  Thanks.

lib/puma/client.rb
MSP-Greg: Sorry, I composed a reply, but must have closed the window.

Ruby's TCP & UNIX sockets inherit from `IO`, and `IO#to_io` returns `self`.  Puma's MiniSSL sockets do not inherit from `IO`, just like Ruby's OpenSSL sockets.

Their `closed?` method checks the underlying TCP socket.  So, rather than add another Ruby method call, I used `@to_io` to get the TCP socket.

Note that both `io` and `to_io` are read-only attributes of `Client`, so these can be removed without anyone losing functionality.  As mentioned, that would still be an API change.

nateberkopec (approved):
LGTM assuming that the commit message will be changed as dentarg suggested

PR #3200, opened 2023-07-31
Handling Expect-100 Continue header

dhavalsingh:
…gins reading the body

### Description

### DIFF
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -346,9 +346,6 @@ module Puma
       @body_read_start = Process.clock_gettime(Process::CLOCK_MONOTONIC, :float_millisecond)
 
       if @env[HTTP_EXPECT] == CONTINUE
-        # ideally send this req to application, if it doesn't raise any error(eg: auth) 
-        # or any other response(eg: redirect) then respond with HTTP_11_100 otherwise
-        # respond with error/response
         @http_expect_100_continue_header_present = true
       end
 
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -89,6 +89,8 @@ module Puma
       @http_content_length_limit = nil
       @http_content_length_limit_exceeded = false
 
+      @http_expect_100_continue_header_present = false
+
       @peerip = nil
       @peer_family = nil
       @listener = nil
@@ -104,7 +106,8 @@ module Puma
     end
 
     attr_reader :env, :to_io, :body, :io, :timeout_at, :ready, :hijacked,
-                :tempfile, :io_buffer, :http_content_length_limit_exceeded
+                :tempfile, :io_buffer, :http_content_length_limit_exceeded,
+                :http_expect_100_continue_header_present
 
     attr_writer :peerip, :http_content_length_limit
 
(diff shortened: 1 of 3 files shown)


lib/puma/client.rb
```diff
@@ -342,12 +342,19 @@ def setup_body
       @body_read_start = Process.clock_gettime(Process::CLOCK_MONOTONIC, :float_millisecond)
 
       if @env[HTTP_EXPECT] == CONTINUE
-        # TODO allow a hook here to check the headers before
-        # going forward
-        @io << HTTP_11_100
-        @io.flush
+        # Determine if a redirect is needed, this could be a function call that encapsulates
```
dhavalsingh, line 345: Need to understand if this is what's expected

dhavalsingh:
Will clean up the PR once I understand I am on the right path to solving this issue.
 @dentarg can you help as you created this issue?

dentarg:
This doesn't look correct to me. We want to application to define the response (e.g. the 307 redirect).

dentarg:
The best guidance I can give you is to look at how the Python `HTTPServer` does it (https://github.com/python/cpython/blob/3.12/Lib/http/server.py) as I have not researched this topic myself yet.

dentarg:
Probably good to drive the implementation using tests.

dhavalsingh:
> The best guidance I can give you is to look at how the Python `HTTPServer` does it (https://github.com/python/cpython/blob/3.12/Lib/http/server.py) as I have not researched this topic myself yet.

Ah, okay. Will look into this. Thanks a lot!

dhavalsingh:
Where are client specific tests written? cant figure it out, or there arent any?

dentarg:
[`test/test_web_server.rb`](https://github.com/puma/puma/blob/99ae85bbbe425d1949f9e8693993499aae02fce4/test/test_web_server.rb) is probably a good fit for this

test/test_web_server.rb
```diff
@@ -99,6 +99,14 @@ def test_nonexistent_http_method
     socket.close
   end
 
+  def test_expect_100_header
```
dhavalsingh, line 102: I should add some tests when response is non 200 also right?

dhavalsingh:
Also, whats the diff between test_puma_server vs test_web_server tests?

lib/puma/client.rb
```diff
@@ -342,10 +346,7 @@ def setup_body
       @body_read_start = Process.clock_gettime(Process::CLOCK_MONOTONIC, :float_millisecond)
 
       if @env[HTTP_EXPECT] == CONTINUE
-        # TODO allow a hook here to check the headers before
-        # going forward
-        @io << HTTP_11_100
-        @io.flush
+        @http_expect_100_continue_header_present = true
```
dentarg, line 349: Pretty sure we don't want to remove the current behaviour of responding quickly to `Expect: 100 Continue` and that allowing the app to define the response to be something the user opt-in to (via the Puma config probably)

nateberkopec: Agreed!

nateberkopec: (Or at least, we should keep that behavior UNLESS the user/app says they want something else).

nateberkopec:
This seems like a fine place to start for me. Next step would be a config setting that allows the user to define what to do if a Expect: 100 is encountered.

nateberkopec:
Closing as stale to allow others to pick this up or for OP to return

Issue #2894, opened 2022-06-11
Display useful suggestions in error message on windows when port reserved? (bind EACCES)


ariccio:
**Is your feature request related to a problem? Please describe.**
A problem I encounter all the time on windows is that Hyper-V (or someone else) will reserve a port like port 3000. It's not *in use*, it's *reserved*. If I try to start up rails with the default port, then everything will crash when [puma fails to create a `TCPServer` on that port](https://github.com/puma/puma/blob/master/lib/puma/binder.rb#L341). It's not obvious from the documentation that `TCPServer.new` [can raise](https://ruby-doc.org/stdlib-3.1.0/libdoc/socket/rdoc/TCPServer.html#method-c-new) (though [it does](https://github.com/ruby/ruby/blob/d92f09a5eea009fa28cd046e9d0eb698e3d94c5c/ext/socket/ipsocket.c#L152)), so puma does not handle this. 

I'm used to diagnosing this, you need to check the reserved ports with `netsh interface ipv4 show excludedportrange protocol=tcp`, but this is *quite surprising* if you have no idea what's going on. If you check the ports in use, you will see *that port is not in use*! The only fix is to reboot, something you may end up doing in the meantime, and likely you won't know *why* it worked.

Because puma does not handle this, rails (or whatever app you're using) simply crashes with a stacktrace, and the message `C:/Ruby31-x64/lib/ruby/gems/3.1.0/gems/puma-5.6.4/lib/puma/binder.rb:341:in `initialize': Permission denied - bind(2) for "::1" port 3000 (Errno::EACCES)`.

**Describe the solution you'd like**
One useful option might be to detect when (on Windows) the exception is `EACCES` instead of `EADDRINUSE`, and dump a useful error to the screen before reraising. The message could be something like:

```
bind(2) for "::1" port 3000 failed with EACCES. On Windows, the port might be reserved even if it's not in use by a process. To check if it's reserved, run this command, and check if the port in question is in the listed ranges:
`netsh interface ipv4 show excludedportrange protocol=tcp`

Sometimes, this is the result of Hyper-V reserving ports. In that case, the solution is to reboot the machine until it has allocated a different port.
```

If in an interactive session, this wouldn't change any behavior, since it would still run through the current exception handling. In any other context, it could cause issues, so I do not expect this to be the solution you adopt :) You could check to see if puma is running in an interactive context too.

**Describe alternatives you've considered**
[...]

Issue #3142, opened 2023-05-02
Address already in use 127.0.0.1 port 9293 error while trying to start two instances of Puma on localhost


jiggneshhgohel:
**Puma version: 6.2.2 (ruby 3.1.2-p20)**

I tried to start two instances of Puma server (one API and another WebApp) (through hanami gem based application) on my development machine but was unable to start the server for WebApp successfully. API server started normally but WebApp server showed error `Address already in use - bind(2) for "127.0.0.1" port 9293`.

Please find below the server startup logs.

**API App Logs**

```
    jignesh@jignesh-Latitude-7290:~/hanami_projects/my_api_app$ bundle exec hanami server --port=2400
    13:59:36 - INFO - Using Guardfile at /......./my_api_app/Guardfile.
    13:59:36 - INFO - Puma starting on port 2400 in development environment.
    13:59:36 - INFO - Guard is now watching at '/......./my_api_app'
    [15170] Puma starting in cluster mode...
    [15170] * Puma version: 6.2.1 (ruby 3.1.2-p20) ("Speaking of Now")
    [15170] *  Min threads: 5
    [15170] *  Max threads: 5
    [15170] *  Environment: development
    [15170] *   Master PID: 15170
    [15170] *      Workers: 2
    [15170] *     Restarts: (✔) hot (✖) phased
    [15170] * Preloading application
    [15170] * Listening on http://0.0.0.0:2400
    [15170] Use Ctrl-C to stop
    [15170] * Starting control server on http://127.0.0.1:9293
    [15170] * Starting control server on http://[::1]:9293
    [15170] - Worker 0 (PID: 15177) booted in 0.0s, phase: 0
    [15170] - Worker 1 (PID: 15179) booted in 0.0s, phase: 0
    [15170] ! Terminating timed out worker (worker failed to check in within 60 seconds): 15177
    [15170] - Worker 0 (PID: 25100) booted in 0.0s, phase: 0
```

**WebApp Logs**

```
    jignesh@jignesh-Latitude-7290:~/hanami_projects/my_web_app$ bundle exec hanami server
    15:32:31 - INFO - Using Guardfile at /......./my_web_app/Guardfile.
    15:32:32 - INFO - Puma starting on port 2300 in development environment.
    15:32:32 - INFO - Guard is now watching at '/......./my_web_app'
    [25763] Puma starting in cluster mode...
    [25763] * Puma version: 6.2.2 (ruby 3.1.2-p20) ("Speaking of Now")
[...]

PR #3204, opened 2023-08-08
Added a Meaningful error msg for control server port taken

dhavalsingh:
### Description

Closes [issue 3204](https://github.com/puma/puma/issues/3142)
Might close issue #2894 too.

### DIFF
--- a/lib/puma/runner.rb
+++ b/lib/puma/runner.rb
@@ -75,7 +75,6 @@ module Puma
       control = Puma::Server.new app, nil,
         { min_threads: 0, max_threads: 1, queue_requests: false, log_writer: @log_writer }
 
-      #control.binder.parse [str], nil, 'Starting control server'
       begin
         control.binder.parse [str], nil, 'Starting control server'
       rescue Errno::EADDRINUSE => e
--- a/lib/puma/runner.rb
+++ b/lib/puma/runner.rb
@@ -75,7 +75,12 @@ module Puma
       control = Puma::Server.new app, nil,
         { min_threads: 0, max_threads: 1, queue_requests: false, log_writer: @log_writer }
 
-      control.binder.parse [str], nil, 'Starting control server'
+      #control.binder.parse [str], nil, 'Starting control server'
+      begin
+        control.binder.parse [str], nil, 'Starting control server'
+      rescue Errno::EADDRINUSE => e
+        raise "Error: Control server address '#{str}' is already in use. You're trying to start a control server but the control server port is taken."
+      end
 
       control.run thread_name: 'ctl'
       @control = control
--- a/lib/puma/runner.rb
+++ b/lib/puma/runner.rb
@@ -77,8 +77,8 @@ module Puma
 
       begin
         control.binder.parse [str], nil, 'Starting control server'
-      rescue Errno::EADDRINUSE => e
-        raise "Error: Control server address '#{str}' is already in use. You're trying to start a control server but the control server port is taken."
+      rescue StandardError => e
+        raise e, "Error: Control server address '#{str}' is already in use. You're trying to start a control server but the control server port is taken."
       end
 
       control.run thread_name: 'ctl'
--- a/lib/puma/runner.rb
+++ b/lib/puma/runner.rb
@@ -77,8 +77,8 @@ module Puma
 
       begin
         control.binder.parse [str], nil, 'Starting control server'
-      rescue StandardError => e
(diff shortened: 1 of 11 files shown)


test/config/activate_control_app.rb
```diff
@@ -0,0 +1,5 @@
+activate_control_app "tcp://0.0.0.0:9393"
+
+app do |env|
+  [200, {}, ["OK"]]
+end
```
dentarg, line 5: Why was this file added? Can't see in the diff that it is being used

lib/puma/runner.rb
```diff
@@ -75,7 +75,11 @@ def start_control
       control = Puma::Server.new app, nil,
         { min_threads: 0, max_threads: 1, queue_requests: false, log_writer: @log_writer }
 
-      control.binder.parse [str], nil, 'Starting control server'
+      begin
+        control.binder.parse [str], nil, 'Starting control server'
+      rescue Errno::EADDRINUSE => e
+        raise "Error: Control server address '#{str}' is already in use. You're trying to start a control server but the control server port is taken."
+      end
```
dentarg, line 82: Should we also re-raise the actual error? 

We could rescue all errors, add our helpful message and re-raise. I think that would cover https://github.com/puma/puma/issues/2894 too (seems like the exception raised on Windows is a different one)

ZaneBliss: @dentarg are you referring to something like this?

```ruby
begin
  control.binder.parse [str], nil, 'Starting control server'
rescue Errno::EADDRINUSE, Errno::EACCES => e
  raise e, "Error: Control server address '#{str}' is already in use. You're trying to start a control server but the control server port is taken."
end
```

Also do you think there is some value in creating a descriptive exception class as well? 

(Perhaps `StandardError` isn't the right exception class to inherit from?)
```ruby
class ControlServerAddressPortConflict < StandardError; end
```

```ruby
begin
  control.binder.parse [str], nil, 'Starting control server'
rescue Errno::EADDRINUSE, Errno::EACCES => e
  raise ControlServerAddressPortConflict, "Error: Control server address '#{str}' is already in use. You're trying to start a control server but the control server port is taken."
end
```

dhavalsingh: Now it should cover #2894 too i think.

ZaneBliss: Actually, after thinking about this I'm not so sure that it would handle #2894. Not that it couldn't but I would imagine the error message would at least need to be different. Sounds like @MSP-Greg started a branch for this last month but hasn't posted yet. I'd be interested in seeing what that looks like

dhavalsingh: yeah, might need to make the error msg a bit generic in that case.

dhavalsingh: Also, i think 
`class ControlServerAddressPortConflict < StandardError; end`
The error should be abstracted out in case we are raising Control server port issues in multiple (more than 1?) places. Otherwise doesn't make much sense IMO. Leme know what you guys think

ZaneBliss: I see where you're coming from with the abstraction comment, but I think it's a good idea because it will lead to a more helpful exception message stack trace. Even if it isn't used other places, I think that wrapping the exception in more domain specific terminology can help increase clarity imo 🙂. But it's your PR so use your discretion

nateberkopec: For me, Errno::EADDRINUSE, Errno::EACCES is clear enough 🤷 

I think we should rescue `Errno::EADDRINUSE, Errno::EACCES` here, as StandardError is too broad and may display this message in confusing circumstances.

nateberkopec (changes requested):
See comment - change what class we're rescuing.

lib/puma/runner.rb
nateberkopec: And as dentarg said, we should re-raise.

dhavalsingh: Fixed, re-raised the original error too.

dhavalsingh:
Also, this branch is out-of-date with the base branch
How do I update this branch? With a rebase or a merge commit?

dentarg:
You can do either, but I would recommend squashing the commits if you go for rebase. When merged, everything will be squashed anyway.

dhavalsingh:
@nateberkopec this can be merged now?

nateberkopec:
Thank you @dhavalsingh !

PR #3209, opened 2023-08-12
Feature | Shutdown on idle

### LINKED ISSUE #2580: Support for shutdown on idle
In order to reduce resource consumption, infrequently utilised Rails/Rack apps should not occupy memory when they are not actively used. Socket activation, which Puma already supports, solves part of this; applications can be started on demand as soon as a request comes in. However, there doesn't seem to be a way to shut down the server again when the app is no longer active.

There should be an (opt-in) setting that allows to specify a timeout. If no new request is received inside that timeout, the server should shut down and exit cleanly. Care must be taken not to close the sockets received from systemd.

This feature cannot be implemented in the service manager (systemd), as it doesn't keep track of activity on the sockets it created after the service has been started. This information is only available to Puma itself.

(I'd be willing to implement this if someone can point me at the right place to start.)

joshuay03:
### Description

Closes #2580.

I attended RubyConfAU earlier this year and was inspired by [this talk](https://youtu.be/SquGNt4FhY0) to contribute to puma. During my search for a small enough issue (preferably a bug) to tackle I stumbled upon the feature request linked above and noticed that it'd been open for a while. I liked the idea and thought it'd be a useful feature, and decided to give it a go.

I'm very new to this 'area' of Ruby so I might need some guidance on how sensible my implementation is...

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -121,6 +121,7 @@ module Puma
 
       @precheck_closing = true
 
+      @ios_count = 0
       @requests_count = 0
     end
 
@@ -331,11 +332,13 @@ module Puma
 
         while @status == :run || (drain && shutting_down?)
           begin
-            ios = IO.select sockets, nil, nil, (shutting_down? ? 0 : timeout_for_idle)
+            ios = IO.select sockets, nil, nil, select_timeout
             unless ios
-              @status = :stop if timeout_for_idle&.zero?
+              @status = :stop unless shutting_down?
               break
             end
+
+            @ios_count += 1
             ios.first.each do |sock|
               if sock == check
                 break if handle_check
@@ -356,7 +359,6 @@ module Puma
                 }
               end
             end
-            set_timeout_for_idle(@idle_timeout)
           rescue IOError, Errno::EBADF
             # In the case that any of the sockets are unexpectedly close.
             raise
@@ -603,20 +605,12 @@ module Puma
     end
     private :notify_safely
 
-    def set_timeout_for_idle(idle_timeout)
-      return unless idle_timeout
+    def select_timeout
+      return 0 if shutting_down?
 
-      @idle_timeout_at = Process.clock_gettime(Process::CLOCK_MONOTONIC) + idle_timeout
+      @idle_timeout if @ios_count > 0
     end
-    private :set_timeout_for_idle
-
-    # Number of seconds until the idle timeout elapses.
-    def timeout_for_idle
-      return unless defined?(@idle_timeout_at)
-
-      [@idle_timeout_at - Process.clock_gettime(Process::CLOCK_MONOTONIC), 0].max
-    end
-    private :timeout_for_idle
+    private :select_timeout
 
     # Stops the acceptor thread and then causes the worker threads to finish
     # off the request queue before finally exiting.
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -86,15 +86,16 @@ module Puma
         UserFileDefaultOptions.new(options, Configuration::DEFAULTS)
       end
 
-      @log_writer          = @options.fetch :log_writer, LogWriter.stdio
-      @early_hints         = @options[:early_hints]
-      @first_data_timeout  = @options[:first_data_timeout]
-      @min_threads         = @options[:min_threads]
-      @max_threads         = @options[:max_threads]
-      @persistent_timeout  = @options[:persistent_timeout]
-      @queue_requests      = @options[:queue_requests]
-      @max_fast_inline     = @options[:max_fast_inline]
-      @io_selector_backend = @options[:io_selector_backend]
+      @log_writer                = @options.fetch :log_writer, LogWriter.stdio
+      @early_hints               = @options[:early_hints]
+      @first_data_timeout        = @options[:first_data_timeout]
+      @persistent_timeout        = @options[:persistent_timeout]
+      @idle_timeout              = @options[:idle_timeout]
+      @min_threads               = @options[:min_threads]
+      @max_threads               = @options[:max_threads]
+      @queue_requests            = @options[:queue_requests]
+      @max_fast_inline           = @options[:max_fast_inline]
+      @io_selector_backend       = @options[:io_selector_backend]
       @http_content_length_limit = @options[:http_content_length_limit]
 
       # make this a hash, since we prefer `key?` over `include?`
@@ -330,8 +331,11 @@ module Puma
 
         while @status == :run || (drain && shutting_down?)
           begin
-            ios = IO.select sockets, nil, nil, (shutting_down? ? 0 : nil)
-            break unless ios
+            ios = IO.select sockets, nil, nil, (shutting_down? ? 0 : timeout_for_idle)
+            unless ios
+              @status = :stop if timeout_for_idle&.zero?
+              break
+            end
             ios.first.each do |sock|
               if sock == check
                 break if handle_check
@@ -352,6 +356,7 @@ module Puma
                 }
               end
             end
+            set_timeout_for_idle(@idle_timeout)
           rescue IOError, Errno::EBADF
             # In the case that any of the sockets are unexpectedly close.
             raise
@@ -597,6 +602,21 @@ module Puma
       end
     end
     private :notify_safely
+
+    def set_timeout_for_idle(idle_timeout)
+      return unless idle_timeout
+
+      @idle_timeout_at = Process.clock_gettime(Process::CLOCK_MONOTONIC) + idle_timeout
+    end
+    private :set_timeout_for_idle
+
+    # Number of seconds until the idle timeout elapses.
+    def timeout_for_idle
+      return unless defined?(@idle_timeout_at)
+
+      [@idle_timeout_at - Process.clock_gettime(Process::CLOCK_MONOTONIC), 0].max
+    end
+    private :timeout_for_idle
 
     # Stops the acceptor thread and then causes the worker threads to finish
     # off the request queue before finally exiting.
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -121,7 +121,6 @@ module Puma
 
       @precheck_closing = true
 
-      @ios_count = 0
       @requests_count = 0
     end
 
@@ -332,13 +331,12 @@ module Puma
 
         while @status == :run || (drain && shutting_down?)
           begin
-            ios = IO.select sockets, nil, nil, select_timeout
+            ios = IO.select sockets, nil, nil, (shutting_down? ? 0 : @idle_timeout)
             unless ios
               @status = :stop unless shutting_down?
               break
             end
 
-            @ios_count += 1
             ios.first.each do |sock|
               if sock == check
                 break if handle_check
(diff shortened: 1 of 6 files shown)


dentarg:
From https://github.com/puma/puma/issues/2580

> There should be an (opt-in) setting that allows to specify a timeout. If no new request is received inside that timeout, the server should shut down and exit cleanly. **Care must be taken not to close the sockets received from systemd.**

Is that last thing (in bold) happening here?

joshuay03:
> Is that last thing (in bold) happening here?

I haven't looked into this (so probably not). I'll add a test and go from there.

joshuay03:
> I'll add a test and go from there.

I'm hoping that this covers it: 272f2c6b2b1c55cc8a000309f58eedf94ac86430.

test/test_integration_single.rb
```diff
@@ -254,4 +254,30 @@ def test_puma_debug_loaded_exts
...
+
+    sleep 1.15 # ensure the timeout isn't set until the first connection
+
+    sock = connection = connect(nil, unix: true)
+    read_body(connection)
+
+    sleep 1.15
+
+    assert sock.wait_readable(1), 'Unexpected timeout'
+    assert_raises Errno::ECONNREFUSED do
```
dentarg, line 273: Looks like JRuby raises `IOError` with message `Connection refused` (e.g. https://github.com/puma/puma/actions/runs/5847300372/job/15853513628#step:8:809)

joshuay03: Cheers, I've handled that with 3e5d42ffdb1810c65a188000af7ddcfad9178e78.

lib/puma/server.rb
```diff
@@ -330,8 +331,11 @@ def handle_servers
 
         while @status == :run || (drain && shutting_down?)
           begin
-            ios = IO.select sockets, nil, nil, (shutting_down? ? 0 : nil)
-            break unless ios
+            ios = IO.select sockets, nil, nil, (shutting_down? ? 0 : timeout_for_idle)
+            unless ios
+              @status = :stop if timeout_for_idle&.zero?
```
nateberkopec, line 336: This is a super-hot loop... I need to think about this.

lib/puma/server.rb
```diff
@@ -598,6 +603,21 @@ def notify_safely(message)
...
+
+      @idle_timeout_at = Process.clock_gettime(Process::CLOCK_MONOTONIC) + idle_timeout
+    end
+    private :set_timeout_for_idle
+
+    # Number of seconds until the idle timeout elapses.
+    def timeout_for_idle
+      return unless defined?(@idle_timeout_at)
+
+      [@idle_timeout_at - Process.clock_gettime(Process::CLOCK_MONOTONIC), 0].max
```
nateberkopec, line 617: I just feel like checking system time every time we read data is just not going to scale. I don't feel comfortable with this.

nateberkopec (changes requested):

nateberkopec:
I don't really have a better solution. I'm trying to think about how a timer thread or something like that could be used instead.

lib/puma/server.rb
joshuay03: That's a valid concern, I'll see if I can come up with an alternative.

joshuay03:
> I'm trying to think about how a timer thread or something like that could be used instead.

Hmm that sounds intereseting, I'll experiement and see how it goes. Thanks for the review!

lib/puma/server.rb
jgaskins: @nateberkopec You mentioned above that this runs in a super hot loop. Do you have a sense for how hot? On my Mac, you can check the monotonic clock roughly 11.5 million times per second in Ruby code. It's definitely faster than the system clock.

<details><summary>Benchmark code</summary>

```
require "benchmark/ips"

Benchmark.ips do |x|
  x.report("monotonic clock") { Process.clock_gettime(Process::CLOCK_MONOTONIC) }
  x.report("system clock") { Time.now }

  x.compare!
end
```

</details>

```
     monotonic clock     1.146M i/100ms
        system clock   333.694k i/100ms
     monotonic clock     11.526M (± 0.9%) i/s -     58.462M in   5.072420s
        system clock      3.303M (± 1.6%) i/s -     16.685M in   5.052375s

Comparison:
     monotonic clock: 11526338.7 i/s
        system clock:  3303211.7 i/s - 3.49x  slower
```

nateberkopec:
I [asked the hivemind](https://twitter.com/nateberkopec/status/1697058996941935067?s=20)

nateberkopec:
The ChatGPT answer I got wasn't bad either: update the time you last received data each time you get data, and have a background timer thread check every second if that time is X seconds ago or greater.

lib/puma/server.rb
byroot: I'm quite unfamiliar with Puma's codebase, but why is system time is even needed here?

The feature is to exit if nothing was read in X seconds? Why not just make that the select timeout and exit if `IO.select` doesn't return any IO?

joshuay03: > I'm quite unfamiliar with Puma's codebase

Same here, so I'm hoping someone more experienced will correct me if I'm wrong.

> Why not just make that the select timeout and exit if IO.select doesn't return any IO?

When there's no timeout and `IO.select` simply doesn't return any IO, the server doesn't shut down gracefully (at least from what I found in my tests). So not only do we know that there's no IO, we also need to know that the timeout has elapsed to gracefully shut it down explicitly, which we need to do if we want to keep the sockets received from systemd open.

byroot: > When there's no timeout and `IO.select` simply doesn't return any IO

I don't follow. If you don't pass a timeout, `IO.select` will block forever.

So if `select` returns `nil`, you are either in the `shutting_down?` case, or in the `idle_timeout?` case. Or am I missing something?

joshuay03: > So if `select` returns `nil`, you are either in the `shutting_down?` case, or in the `idle_timeout?` case. Or am I missing something?

That's what I thought too, but I found that it isn't `shutting_down?` when `select` returns `nil`. But by setting the status to `:stop`, I'm able to enter the shutdown phase due to:
https://github.com/puma/puma/blob/2563396f730adef54a9369dbae996fda3bba5f6e/lib/puma/server.rb#L365

byroot: You mean `IO.select sockets, nil, nil, nil` returns `nil`? 

It's really not supposed to happen.

joshuay03: Ya okay I was definitely wrong about that. I think I was just trying to not break this previous logic:

```ruby
break unless ios
```

But it sounds like that case shouldn't even be possible unless it times out?
All my tests pass if I just use your suggestion AND set the status to `:stop` if `ios` is `nil`. Which makes sense to me. 

However, I do have this one failing case:

https://github.com/joshuay03/puma/blob/23fc9b2d1c49a64ebdd9f36dcd8c9c22d67d5bc1/test/test_puma_server.rb#L632-L644

Which also makes sense cause I approached this as `"the timeout since the last request"` and not `"the timeout until the next request"`. But if that expectation isn't necessary, I think this works perfectly!

joshuay03: > So I think you can just check shutting_down? after the unless ios to know if you reached the idle timeout or are simply shutting down already.

Perfect! Thank you so much!

joshuay03: > Which also makes sense cause I approached this as `"the timeout since the last request"` and not `"the timeout until the next request"`. But if that expectation isn't necessary, I think this works perfectly!

Actually, I think I can figure out a way to handle this, will report back!

joshuay03: Resolved with: 9c4a15a158b554f67e59f33378eee10abeec8c14.

lib/puma/server.rb
joshuay03: Resolved with: 9c4a15a158b554f67e59f33378eee10abeec8c14.

jgaskins: Sounds good. I just wanted to understand the concerns around performance here since checking the monotonic clock is surprisingly cheap. It's actually slightly cheaper than `Object.new`:

<details><summary>Benchmark code</summary>

```ruby
require "benchmark/ips"

n = 0
Benchmark.ips do |x|
  x.report("do nothing") {}
  x.report("allocate an object") { Object.new }
  x.report("monotonic clock") { Process.clock_gettime(Process::CLOCK_MONOTONIC) }
  x.report("system clock") { Time.now }

  x.compare!
end
```

</details>

```
          do nothing     2.399M i/100ms
  allocate an object     1.097M i/100ms
     monotonic clock     1.132M i/100ms
        system clock   325.979k i/100ms
          do nothing     23.868M (± 0.9%) i/s -    119.961M in   5.026342s
  allocate an object     10.932M (± 0.9%) i/s -     54.832M in   5.016068s
     monotonic clock     11.405M (± 0.9%) i/s -     57.747M in   5.063726s
        system clock      3.259M (± 1.1%) i/s -     16.299M in   5.001941s

Comparison:
          do nothing: 23868388.0 i/s
     monotonic clock: 11404972.2 i/s - 2.09x  slower
  allocate an object: 10932187.5 i/s - 2.18x  slower
        system clock:  3258906.8 i/s - 7.32x  slower
```

lib/puma/server.rb
```diff
@@ -598,6 +605,13 @@ def notify_safely(message)
     end
     private :notify_safely
 
+    def select_timeout
+      return 0 if shutting_down?
+
+      @idle_timeout if @ios_count > 0
```
byroot, line 611: I don't understand that `@ios_count`.

It only ever increment right? So this means you ignore `idle_timeout` on the very first `select`, why?

joshuay03: Because of this case:

https://github.com/joshuay03/puma/blob/9c4a15a158b554f67e59f33378eee10abeec8c14/test/test_puma_server.rb#L632-L644

More context in the last part of my comment here: https://github.com/puma/puma/pull/3209#discussion_r1311532499.

joshuay03: I'm not married to this approach, but I think it does make sense that the timeout isn't set for the very first request? I was mostly going off of the description in the feature request: https://github.com/puma/puma/issues/2580.

byroot: Ah I see. IMO it's not necessary, but that's more for Nate to decide.

But if you do keep it, since it's used as a boolean flag, you might as well make it a boolean and flip it with `@flag ||= true`

joshuay03: Ya good point, I just assumed that having the count would come in handy later on, I'll keep this in mind though.
Will wait for Nate's input first.

ghiculescu: I think I understand what's going on. Is this effectively what you are saying?

```suggestion
      @idle_timeout if @has_ever_processed_a_request
```

nateberkopec: I don't think this check is necessary.

As the original issue said, we already have the feature of socket activation. So, presumably, people that use idle_timeout will use socket activation to deal with starting a new puma process to handle their requests. 

The point of socket activation with the idle_timeout feature is to reduce resource consumption. IMO that doesn't really jive with keeping a process alive, idling, doing nothing, until the first request comes in.

We can assume that anyone using this feature uses socket activation, so doesn't need this behavior.

nateberkopec (changes requested):
Thanks for catching the easiest possible solution @byroot!

lib/puma/server.rb
joshuay03: Resolved with 01f9c13d007f31020400b726c6099e72e691548b.

nateberkopec (approved):

nateberkopec:
Thanks @joshuay03, this ended up being a great PR 👍

PR #3210, opened 2023-08-16
Fix ActiveSupport deprecation warning being thrown on shutdown in Rails 7.0.7

schinery:
### Description

Use `Time.now.strftime` to format the shutdown timestamp to stop the ActiveSupport deprecation warning "_Using a :default format for Time#to_s is deprecated. Please use Time#to_fs instead._" if Puma is being used in Rails 7.0.7 apps

Before:

```sh
[65598] - Gracefully shutting down workers...
DEPRECATION WARNING: Using a :default format for Time#to_s is deprecated. Please use Time#to_fs instead. (called from <main> at bin/rails:6)
[65598] === puma shutdown: 16/08/2023 13:52:35 ===
[65598] - Goodbye!
```

After:

```sh
[2284] - Gracefully shutting down workers...
[2284] === puma shutdown: 2023-08-17 05:15:22 +0100 ===
[2284] - Goodbye!
```

### DIFF
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -217,7 +217,12 @@ module Puma
       @runner.close_control_listeners
       @binder.close_listeners
       unless @status == :restart
-        log "=== puma shutdown: #{Time.now} ==="
+        # Formatting the timestamp using .strftime so that we don't trigger
+        # an ActiveSupport deprecation warning "Using a :default format for
+        # Time#to_s is deprecated. Please use Time#to_fs instead." in 7.0.7
+        # and above Rails apps.
+        timestamp = Time.now.strftime("%Y-%m-%d %H:%M:%S %z")
+        log "=== puma shutdown: #{timestamp} ==="
         log "- Goodbye!"
       end
     end


dentarg:
What the hell :) Isn't this a bug in Rails/ActiveRecord? Puma can't possibly be expected to accommodate for all the monkey patching that exist?

schinery:
TBH this was my thoughts too, like I said in the issue I just threw this together to see if it solved the error.

Will throw something in the Rails/ActiveSupport issues and see what comes from that.

MSP-Greg:
Given all the monkey patching, should Puma use the following?
```
Time.now.strftime '%Y-%m-%d %H:%M:%S %z'
```

schinery:
> Given all the monkey patching, should Puma use the following?
> 
> ```
> Time.now.strftime '%Y-%m-%d %H:%M:%S %z'
> ```

Damn, I should have thought of this in the first place 🤦🏻‍♂️ 

Great shout, I've updated to use `.strftime` instead.

MSP-Greg:
Maybe add a comment about its use due to RoR monkey patching?  Just so it isn't accidentally reverted in the future.

Or, been there, done that...

schinery:
> Maybe add a comment about its use due to RoR monkey patching? Just so it isn't accidentally reverted in the future.
> 
> Or, been there, done that...

Comment added about why using `.strftime`

dentarg:
This is not the only place where `Time.now` is called in an interpolated string. 

I'm still _very_ skeptical to why Puma has to do this. We are not using `active_support/time` in Puma. Is the whole world expected to avoid calling `Time.now.to_s` because someone might use the code with Rails/ActiveSupport?

(Similar PR in another project: https://github.com/ruby-i18n/i18n/pull/674 found from https://github.com/rails/rails/issues?q=is%3Aissue+to_s)

MSP-Greg:
@dentarg

> This is not the only place where Time.now is called in an interpolated string.

Good point, should have checked.  I don't think the issue exists in any 'hot' code paths, maybe define a Puma method (`Puma.time_now` ?), so in one place the time format is specified?  Note that a different format is used for request logging...

dentarg:
> define a Puma method (`Puma.time_now` ?), so in one place the time format is specified?

Yes, was about to suggest that myself but got side-tracked 👍

dentarg:
Some more background on this: 

The start of this is https://github.com/rails/rails/pull/43772 where Rails do want to stop override `to_s`. I think it surfaced now due to https://github.com/rails/rails/pull/48555. It sounds like the consequence of not handling the warnings are that the format will change in Rails 7.1: https://github.com/rails/rails/pull/48555#issuecomment-1680844190

We can see that the bad crash in https://github.com/puma/puma/discussions/3212 comes from active_support and the deprecation warning system that (always?) uses `ActiveSupport::Notifications` and apparently a `Mutex` under the covers. The harmless `puts` in Puma, run in the [signal handler for SIGTERM](https://github.com/puma/puma/blob/d62f077ed68f9f180d01ca0e7cad3fc818b53430/lib/puma/cluster.rb#L319-L336), was turned into a bad crash by this. 

    /app/vendor/ruby-3.2.2/lib/ruby/3.2.0/mutex_m.rb:79:in `synchronize': can't be called from trap context (ThreadError)
    from /app/vendor/ruby-3.2.2/lib/ruby/3.2.0/mutex_m.rb:79:in `mu_synchronize'
    from /app/vendor/bundle/ruby/3.2.0/gems/activesupport-7.0.7/lib/active_support/notifications/fanout.rb:110:in `listeners_for'
    from /app/vendor/bundle/ruby/3.2.0/gems/activesupport-7.0.7/lib/active_support/notifications/fanout.rb:118:in `listening?'
    from /app/vendor/bundle/ruby/3.2.0/gems/activesupport-7.0.7/lib/active_support/notifications.rb:205:in `instrument'
    from /app/vendor/bundle/ruby/3.2.0/gems/activesupport-7.0.7/lib/active_support/deprecation/behaviors.rb:39:in `block in <class:Deprecation>'
    from /app/vendor/bundle/ruby/3.2.0/gems/activesupport-7.0.7/lib/active_support/deprecation/reporting.rb:26:in `block (2 levels) in warn'
    from /app/vendor/bundle/ruby/3.2.0/gems/activesupport-7.0.7/lib/active_support/deprecation/reporting.rb:26:in `each'
    from /app/vendor/bundle/ruby/3.2.0/gems/activesupport-7.0.7/lib/active_support/deprecation/reporting.rb:26:in `block in warn'
    from <internal:kernel>:90:in `tap'
    from /app/vendor/bundle/ruby/3.2.0/gems/activesupport-7.0.7/lib/active_support/deprecation/reporting.rb:22:in `warn'
    from /app/vendor/bundle/ruby/3.2.0/gems/activesupport-7.0.7/lib/active_support/deprecation/instance_delegator.rb:21:in `warn'
    from /app/vendor/bundle/ruby/3.2.0/gems/activesupport-7.0.7/lib/active_support/deprecation/instance_delegator.rb:28:in `warn'
    from /app/vendor/bundle/ruby/3.2.0/gems/activesupport-7.0.7/lib/active_support/core_ext/time/deprecated_conversions.rb:15:in `to_s'
    from /app/vendor/bundle/ruby/3.2.0/gems/puma-6.3.0/lib/puma/launcher.rb:220:in `close_binder_listeners'
    from /app/vendor/bundle/ruby/3.2.0/gems/puma-6.3.0/lib/puma/cluster.rb:327:in `block in setup_signals'

I really think it is a stretch to blame Puma for this. Can't it be reported to the Rails project?

Also, from https://github.com/rails/rails/pull/43772, it looks like the ENV `RAILS_DISABLE_DEPRECATED_TO_S_CONVERSION` can be used to silence the warnings, by opt-ing in to the behaviour that will come in Rails 7.1 (I haven't tested this myself).

schinery:
> Is the whole world expected to avoid calling Time.now.to_s

https://github.com/stripe/stripe-ruby/pull/1262 Looks that way 😞

This is [the Rails commit](https://github.com/rails/rails/pull/48555) that introduced this, and it appears that the deprecation warning only appears if a `*_FORMATS[:default]` has been configured in the Rails app. 

Am still to file an issue with Rails for this (been busy with other stuff) but do you want me to have a go at doing the `Puma.time_now` suggestion in this PR?

dentarg:
@skipkayhil @rafaelfranca any comments on all of the above? any advice to Rails users?

schinery:
@dentarg @MSP-Greg can confirm that setting `config.active_support.disable_to_s_conversion = true` stops the deprecation warning on shutdown, so closing this because as you said it is not a puma issue.

MSP-Greg:
@schinery

> you said it is not a puma issue.

Not sure about what to call it.  Puma should work with any recent Rails versions.  Rather than have people find the above setting, it might be easier for all involved to change this in Puma.  We've seen it before, someone upgrades their Rails version, and something breaks in Puma, sometimes due to a changed dependency.  And, of course, they assume it's Puma's fault.

I'm not blaming Rails or its dependencies, as everyone has reasons for API/code changes.

As to Puma, and speaking as one who has done so myself, we've broken apps that use a Puma server object in their CI.  Or, they're not starting Puma from a command line, they're creating a `Puma::Server` object and using it to generate responses from requests.

I haven't had time to look at how many places the code needs to change.  I guess we'll see if more issues are opened due to it... Thanks for your efforts.

schinery:
> @dentarg @MSP-Greg can confirm that setting `config.active_support.disable_to_s_conversion = true` stops the deprecation warning on shutdown, so closing this because as you said it is not a puma issue.

Sorry guys, I messed up and actually setting `config.active_support.disable_to_s_conversion = true` doesn't resolve the issue in Puma. The branch I was testing of our app I thought I'd bumped Rails back to 7.0.7 but actually hadn't. When I realised my mistake, bumped it and retested the deprecation warning is still there.

dentarg:
@schinery are you able to specify the default format for your app? If I understand this correctly, Rails warns when you haven't set a custom default format (note: I'm not using Rails myself so I could be wrong)

schinery:
> @schinery are you able to specify the default format for your app? If I understand this correctly, Rails warns when you haven't set a custom default format (note: I'm not using Rails myself so I could be wrong)

@dentarg it is the other way around in that it will warn if you do have a custom default set in your app. 

This is set in the Rails app for the Time defaults: `Time::DATE_FORMATS[:default] = "%d/%m/%Y %H:%M:%S"`. If I comment it out and test locally (making sure I'm using Rails 7.0.7!) then there is no warning when shutting down using workers.

It's hitting [this](https://github.com/rails/rails/blob/7-0-stable/activesupport/lib/active_support/core_ext/time/deprecated_conversions.rb#L14) in the Rails code.

schinery:
Also as mentioned above setting `config.active_support.disable_to_s_conversion` doesn't work (it is actually already set), **but** if `RAILS_DISABLE_DEPRECATED_TO_S_CONVERSION` is set then no warnings are triggered at shutdown.

So that could be my option for now if I want to run Rails 7.0.7 as it looks like this problem goes away when Rails 7.1 is released.

dentarg:
Per https://github.com/rails/rails/issues/48960#issuecomment-1686764618 and the above discussion, the solution to this issue seems to be to stop setting the default format and update your app accordingly to deal with that change.

PR #3237, opened 2023-09-28
[CI] - move test_control_gc_stats_* from test_cli.rb to test_integration_pumactl.rb, misc

MSP-Greg:
### Description

There are two tests `test_control_gc_stats_*` (tcp & unix) that run in-process in `test_cli.rb`, and they both run `GC.start`, which is wrapped by two `gc-stats` control server calls.  These tests were intermittently failing on non-MRI platforms.  Moved them to `test_integration_pumactl.rb`, where GC is run in a spawned server.

Note that `GC.start` does not always trigger gc in non-MRI platforms, so there is some logic try to discern if there is a difference between the two `gc-stats` calls.

Also, removed a plain `assert`, removed a native socket, and other misc changes. 


dentarg (approved):
5cd3760139fa77151347be2f7c78c12c6a730a14 was made without any motivation (probably makes sense) but would you spell it out for us?

MSP-Greg:
@dentarg

I noticed these were failing, mostly on non MRI Rubies.

1. CI has intermittent failures
2. CI is run on several Ruby versions/platforms/OS's
3. What passes locally may not behave the same on multi VM CI systems.
4. Currently, these tests run GC in the main CI test process.

Given the above, can we say with certainty that running GC in the main test process has no effect on CI now or in the future?  Hence, I decided to move the GC tests to a spawned process where the GC doesn't affect the main test process.

PR #3255, opened 2023-10-09
Make puma cluster process suitable as PID 1

casperisfine:
We recently ran into an issue in production where our puma containers had thousands of zombie (defunct) processes.

This was caused by the web application spawning some multi-process command (Google Chrome) and sometimes interrupting it hard.

```
- pid=1 puma cluster
 \- pid=2 puma worker
  \- pid=3 google chrome
   \- pid=4 google chrome subprocess
```

In such scenario if `pid=3` dies without first reaping `pid=4`, then `pid=4` get reparented as a child of `pid=1`, and `pid=1` is responsible to reap it.

On classic style hosting, PID 1 is the init system (e.g. systemd), but with containers PID 1 is the container initial command, in our case the Puma cluster process.

Some people uses [some minimal init implementations like `tini`](https://github.com/krallin/tini) to handle this, but I believe this enter the category of "unknown unknowns", as it you need to know about this concern to avoid the problem.

This commit simply change the `wait_workers` method to reap all childrens not just the workers it knows about.

If it end up reaping an unknown process, it logs it.

This is pretty much exactly what [Unicorn](https://github.com/defunkt/unicorn/blob/a86f5380f04ceba6ffb8e25caa0bdea8a37932a6/lib/unicorn/http_server.rb#L401-L417) and [Pitchfork](https://github.com/Shopify/pitchfork/blob/db516f53ffc64a5f613e26f5b5e4de002c3b2800/lib/pitchfork/http_server.rb#L468-L481) do, and remove the need for an init system in containers.

PS: I really don't know how to cover this with tests, it would require the test to be ran as PID 1.

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -522,10 +522,28 @@ module Puma
     # loops thru @workers, removing workers that exited, and calling
     # `#term` if needed
     def wait_workers
+      # Reap all children, known workers or otherwise.
+      # If puma has PID 1, as it's common in containerized environments,
+      # then it's responsible for reaping orphaned processes, so we must reap
+      # all our dead children, regardless of whether they are workers we spawned
+      # or some reattached processes.
+      reaped_children = {}
+      loop do
+        begin
+          pid, status = Process.wait2(-1, Process::WNOHANG)
+          break unless pid
+          reaped_children[pid] = status
+        rescue Errno::ECHILD
+          break
+        end
+      end
+
       @workers.reject! do |w|
         next false if w.pid.nil?
         begin
-          if Process.wait(w.pid, Process::WNOHANG)
+          # When `fork_worker` is enabled, some worker may not be direct children, but grand children.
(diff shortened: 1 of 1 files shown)


casperisfine:
The CI failures seem legit, I'll start digging into them.

nateberkopec:
>On classic style hosting, PID 1 is the init system (e.g. systemd), but with containers PID 1 is the container initial command, in our case the Puma cluster process.

TIL re: the container part

casperisfine:
Apologies for the delay, I was a bit swamped last week.

I believe I fixed CI, there are two jobs failing on the last push, but both are ruby-head I suspect it's not related to my PR.

It should be ready for review now.

nateberkopec (approved):
Good for me but I want @MSP-Greg to approve as well

MSP-Greg (approved):

casperisfine:
Nvm, I just realized you haven't merged these fixes yet.

nateberkopec:
Thanks! Should get 6.4.1 out tomorrow or so.

Issue #3213, opened 2023-08-20
Provide appropriate mechanism for forking to make GRPC work


Nowaker:
**Is your feature request related to a problem? Please describe.**
I'm getting this error when trying to use GRPC in Puma in production deployment which is clustered:

> grpc cannot be used between calls to GRPC.prefork and GRPC.postfork_child or GRPC.postfork_parent

To use GRPC, I must call `GRPC.prefork` and then `GRPC.postfork_child` or `GRPC.postfork_parent` depending on the context. However, with current Puma DSL, there is no way to know whether I should call `postfork_child` or `postfork_parent` because that information isn't exposed in any way. This isn't enough:

```ruby
before_fork do
  GRPC.prefork
end

after_worker_fork do |index|
  GRPC.postfork_child # or GRPC.postfork_parent
end
```

**Describe the solution you'd like**
- Second parameter in `after_worker_fork` to indicate whether the current process is a child or a parent.
- If parent is never used as a worker, introduce `after_parent_fork` where `GRPC.postfork_parent` would belong to.

**Describe alternatives you've considered**
None.

**Additional context**

Nowaker:
I started looking at Puma source, and figured this is already possible. I think the language in documentation is confusing, and the only way to really know what is executed where is to look at Puma source.

I think this example should be incorporated into Puma's documentation as it's not possible to use GRPC in clustered Puma.

```ruby
is_mac = RUBY_PLATFORM.include?('darwin')

on_worker_fork do |index| # Hook name: before_worker_fork
  GRPC.prefork unless is_mac
end

after_worker_fork do |index|
  GRPC.postfork_parent unless is_mac
end

on_worker_boot do
  GRPC.postfork_child unless is_mac
end
```

nateberkopec:
Hey @Nowaker, I'm glad you were able to figure it out.

I'm not super happy with the _names_ of our fork hooks either, but it's basically a historical artifact of how they were added on one-by-one over the years.

I wonder if a new set of aliases could make this clearer...

I'm also not super happy with how we don't have a transcription of the full config in DSL in the `docs` folder. I wonder if we could auto-generate that.

zedbydade:
Hey @nateberkopec. 

By "transcription," do you mean something like a dsl.md file where we'll have the comments from the dsl.rb file? If so, we can generate that by a script and add some configuration examples(Like the GRPC one).

nateberkopec:
[...]

PR #3258, opened 2023-10-12
[ci skip] draft: Add new set of aliases 

zedbydade:
### Description
Hi folks, I built a new set of aliases that we discussed in issue #3213. I use the existing after_worker_boot as a base for the new set. Will that be the way? That MR will also have documentation improvement. 

Thank you all 

### DIFF
--- a/lib/puma/rack/builder.rb
+++ b/lib/puma/rack/builder.rb
@@ -1,8 +1,5 @@
 # frozen_string_literal: true
 
-module Puma
-end
-
 module Puma::Rack
   class Options
     def parse!(args)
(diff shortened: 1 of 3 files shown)


lib/puma/rack/builder.rb
```diff
@@ -1,8 +1,5 @@
 # frozen_string_literal: true
 
-module Puma
```
nateberkopec, line 3: Whoops, when did that get added?

nateberkopec:
Thanks for opening this @zedbydade! Welcome to Puma :)

There are actually quite a few more hooks beginning with `on_` than I realized... hm. I want to think about this one a bit.

lib/puma/rack/builder.rb
zedbydade: https://github.com/puma/puma/commit/5fdf337790a62bdcb8a7db3379cba4007426ebec#diff-e472dc9fd53aac8bac37c1308caeaa854989c40f4d0c21e07b54dfca06887db1

PR #3259, opened 2023-10-16
[CI] Fix macOS TruffleRuby

MSP-Greg:
### Description

macOS TruffleRuby has been failing with backtraces showing a OpenSSL build/runtime problem.  Or, Puma is compiling with one version, but running with another.

This fixes the issue by adding a conditional to not use pkg_config only when running on Actions with macOS TruffleRuby.  Example from a log:
```
Error reached top of thread-pool: External LLVMFunction SSL_get1_peer_certificate cannot be found. (Polyglot::ForeignException)
TestCLI#test_control_for_ssl = [MinitestRetry] retry 'test_control_for_ssl' count: 1,  msg: EOFError: EOFError
    /Users/runner/work/puma/puma/test/helpers/test_puma/puma_socket.rb:271:in `block (2 levels) in <module:PumaSocket>'
```

There are still intermittent failures with macOS TruffleRuby, 

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -140,6 +140,7 @@ jobs:
     needs: [rubocop, skip_duplicate_runs]
     env:
       CI: true
+      PUMA_TEST_DEBUG: true
       TESTOPTS: -v
       PUMA_NO_RUBOCOP: true
 
--- a/ext/puma_http11/extconf.rb
+++ b/ext/puma_http11/extconf.rb
@@ -11,7 +11,10 @@ end
 unless ENV["PUMA_DISABLE_SSL"]
   # don't use pkg_config('openssl') if '--with-openssl-dir' is used
   has_openssl_dir = dir_config('openssl').any?
-  found_pkg_config = !has_openssl_dir && pkg_config('openssl')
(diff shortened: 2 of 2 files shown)


.github/workflows/tests.yaml
```diff
@@ -140,6 +140,7 @@ jobs:
     needs: [rubocop, skip_duplicate_runs]
     env:
       CI: true
+      PUMA_TEST_DEBUG: true
```
nateberkopec, line 143: Do we want to leave this on?

MSP-Greg: I suppose there's no reason it can't be on for people running tests locally.

IIRC, Minitest runs our code that logs the debugging info in a 'Kernel.at_exit' hook.  Puma::ControlCLI has several `exit` statements, which can cause interesting results in tests if they're not trapped.

Also, the Actions logging syntax lets us collapse the section, that won't work locally in a terminal/console window.

Finally, I added it a while ago because I wanted to log some metrics from tests (passing or not), and I didn't want the info scattered throughout the test logs.  The CI systems are a bit resource limited compared to many devs' local PC's...

PR #3267, opened 2023-10-26
Test Framework Update (Draft, review only)

MSP-Greg:
### Description

This is the branch I've been working on the test framework in.  The first and last commits are framework related, most of the middle commits are revisions to single test files.

Looking for review re the API, method names, organization, etc.  Pick your favorite test files, find the commits, and review the diffs...

A few points:
* Seems a lot more stable than the current tests.  There are issues with JRuby (macOS in particular), as often jobs complete with all tests passing, but the test process is frozen. Of course, it's intermittent...  More work to determine the cause.
* I tried to remove as many blocking operations in test code as possible.  When running tests parallel, blocking operations will cause a lot of failures/retries.
* Tests should be portable, assuming the base class (ServerSpawn or ServerInProcess) remains the same.
* I think in many cases the 'test intent' is much easier to see, as there isn't the code clutter to setup things not pertinent to the test intent.
* I've got a shell script that runs all the test files one at a time.  Wrote it for the JRuby issue, but I also found some issues with running tests by themselves.  So, some code changes due to that, along with standardizing the `require` code.
* The current test system has 'helper code' that passes blocks.  The blocks vary with the helper code.  It gets tricky and cluttered trying to pass multiple blocks to a method, so the test framework uses lambdas for almost everything.
* The current test system often loaded rackup or config files.  I found that to be a PITA, as one needed to open other files to look at what the test was doing.  The framework allows config strings to be used, it handles writing a temp file for use by Puma.

 I have started doc'ing the code, see the doc at https://msp-greg.github.io/puma_test/.  All the test modules/classes are children of the TestPuma module at the bottom of the class list.

Once people have reviewed this, I'd start with a PR of the test framework files, then submit a PR for each test file update?  Or, we could commit several without squashing the commits when merging...

Finally, the current test system divides test files by the server type used (in-process or spawned/IO.popen).  This PR follows with the same idea.  I think they could be combined, so one test file could contain both server types.  Thoughts?

Thanks in advance.

### DIFF
--- a/test/helpers/test_puma/puma_socket.rb
+++ b/test/helpers/test_puma/puma_socket.rb
@@ -83,6 +83,7 @@ module TestPuma
     end
 
     def before_setup
+      super
       @ios_to_close ||= []
 
       @bind_host = nil
@@ -98,6 +99,7 @@ module TestPuma
       @control_type = nil
 
       @ci_log = +''
+      @ssl_socket_contexts = Queue.new
     end
 
     # Closes all io's in `@ios_to_close`, also deletes them if they are files
@@ -122,6 +124,14 @@ module TestPuma
       # not sure about below, may help with gc...
       @ios_to_close.clear
       @ios_to_close = nil
+
+      until @ssl_socket_contexts.empty?
+        ctx = @ssl_socket_contexts.pop
+        ctx = nil if ctx
+      end
+      @ssl_socket_contexts.close
+      @ssl_socket_contexts = nil
+      super
     end
 
     # rubocop: disable Metrics/ParameterLists
@@ -351,6 +361,7 @@ module TestPuma
           tcp.setsockopt(Socket::IPPROTO_TCP, Socket::TCP_NODELAY, 1) if SET_TCP_NODELAY
           if ctx || bind_type == :ssl
             ctx ||= new_ctx
+            @ssl_socket_contexts << ctx
             ::OpenSSL::SSL::SSLSocket.new tcp, ctx
           else
             tcp
--- a/test/helpers/test_puma/puma_socket.rb
+++ b/test/helpers/test_puma/puma_socket.rb
@@ -59,6 +59,7 @@ module TestPuma
   module PumaSocket
     GET_10 = "GET / HTTP/1.0\r\n\r\n"
     GET_11 = "GET / HTTP/1.1\r\n\r\n"
+    GET_11_CLOSE = "GET / HTTP/1.1\r\nConnection: close\r\n\r\n"
 
     HELLO_11 = "HTTP/1.1 200 OK\r\ncontent-type: text/plain\r\n" \
       "Content-Length: 11\r\n\r\nHello World"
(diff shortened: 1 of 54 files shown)


dentarg:
What's the motivation behind `minissl.rb - set @eng_ctx = nil to nil on close` (https://github.com/puma/puma/pull/3267/commits/5eec16a2257d706a6fbd9baa575a8938c9dd6d84)? Maybe that should go in as a separate PR as it the only change to `lib/` code

MSP-Greg:
@dentarg 

Re the minissl commit, I may revert that.  Still doing work re the JRuby issue, as both failures in the CI run here successfully completed the tests.

Also, there are some issues with closing spawned servers using SSL listeners (possibly OS/platform related), still working on it.

Re this PR, it's more about getting more 'eyes' looking at the test framework API.  I suppose I could have pushed the changes to a branch here, instead of doing a PR...

MSP-Greg:
Just ran CI in my fork with master and the work branch I've got this code in.  Ran both in my fork, since I have an individual 'Pro' account, I have a higher CI concurrency limit than a standard org account (puma).  Note the time difference, and the fact that the updated test suite has no failures.  I've removed all the 'allow failure' code.

`Time   branch   Failures        `   Log
`25:30  master   1 MRI, 5 non-MRI`   https://github.com/MSP-Greg/puma/actions/runs/7160869671
`12:48  PR       NONE            `   https://github.com/MSP-Greg/puma/actions/runs/7160795377

The current log of the test suite run on master is at https://github.com/puma/puma/actions/workflows/tests.yaml?query=branch%3Amaster.  Several failed runs.   I've had the work branch fail also, but it's much less frequent.

It's been about 6-1/2 weeks since this was opened.  Not any feedback re the test suite updates. Also no feedback about how to break the PR up into smaller PR's.

test/test_plugin_systemd_jruby.rb
```diff
@@ -16,8 +16,6 @@ def setup
     skip_unless_signal_exist? :TERM
     skip_unless :jruby
 
-    super
-
```
dentarg, line 20: Could this fix go into master? I noticed this test failing in master (https://github.com/puma/puma/actions/runs/7160776392/job/19495517337#step:8:874) but I couldn't make sense of it (systemd plugin should not be loaded on JRuby, code has guard against `JRUBY_VERSION` constant)

dentarg: I guess not, I see more test have the same `super` call, it is how `TestIntegration` works. https://github.com/puma/puma/blob/08e71a2c6aaccf295bea909aacf40a9aa86b3702/test/helpers/integration.rb#L27 (Sorry)

Still curious if you @MSP-Greg have any theories about the `TestPluginSystemdJruby` master failures :)

test/helpers/test_puma/puma_socket.rb
```diff
@@ -59,6 +59,7 @@ module TestPuma
   module PumaSocket
     GET_10 = "GET / HTTP/1.0\r\n\r\n"
     GET_11 = "GET / HTTP/1.1\r\n\r\n"
+    GET_11_CLOSE = "GET / HTTP/1.1\r\nConnection: close\r\n\r\n"
```
dentarg, line 62: Can't see anything using this constant?

dentarg:
This is awesome work. No question about it. It will be much easier to maintain and contribute to Puma if we get the CI stable.  

We can probably go ahead with these changes as they're only changing test files. 

> It's been about 6-1/2 weeks since this was opened. Not any feedback re the test suite updates. Also no feedback about how to break the PR up into smaller PR's.

It is a big PR. I pointed out a few things I happened to see now, but I have far from looked at everything. I work on Puma when I feel like it, when I think I have some time over. It is almost never a block of several hours. I forgot about this. I think smaller PR was a successful way to go: https://github.com/puma/puma/pulls?q=is%3Apr+%22PumaSocket%22

MSP-Greg:
@dentarg

Thanks for your comments.  I haven't recently had the time to work on this that I'd like, including looking at your comments.

I rebased on master, and some odd things are happening with all Ruby master builds.  I also follow ruby/ruby's CI, and there are no problems there.  Recently, CI for Windows builds is running much faster and is more stable.

Some of the issues have to do with GC finalizers called on `Tempfile.new`, which are created in `Client` when the request body is large or indeterminate/chunked.  Also, there are hung Ruby processes when the tests complete, even if they pass.

Hopefully, more info to come.

Regardless, Happy Holidays and Merry Christmas.

Not really religious, but I appreciate the more secular ideas about Christmas and the Holidays.  Sorry if anyone is offended by the mention.

dentarg:
I think we can close this one? The plan is to merge many smaller PRs

We also have https://github.com/puma/puma/pull/3479 showing of it all. And we have the discussion at https://github.com/puma/puma/discussions/3232.

MSP-Greg:
Closing in favor of #3479.

PR #3272, opened 2023-11-01
Draft: Add attr_reader options for Puma::Server

icbd:
### Description

Add `attr_reader :options` for Puma::Server.

On the one hand, it can reduce the use of instance variables. 
Using the reader method can more clearly indicate that this is a read-only attribution and will not be changed in other methods.

On the other hand, It allows `Puma::Server` to expose `@options` for external use.
I have a scenario that requires this feature: Fetch PUMA Workers count of options in Rails runtime.

When using PUMA Cli to start the service, we can indeed do this `::Puma.cli_config.options[:workers]`,
but if we start the server through `rails server`,  `Puma` can not response `cli_config` method.

```ruby
# This is a compatible method
Puma::Server.current.instance_variable_get('@options').fetch(:workers)

# hope to use
# Puma::Server.current.options.fetch(:workers)
```

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -32,6 +32,7 @@ module Puma
     include Puma::Const
     include Request
 
+    attr_reader :options
     attr_reader :thread
     attr_reader :log_writer
     attr_reader :events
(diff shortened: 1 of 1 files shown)


nateberkopec:
I don't see any problem with this, as options are basically a public API already anyway.

PR #3276, opened 2023-11-09
Improve Puma::NullIO consistency with real IO

casperisfine:
This isn't a big deal, but every small difference has a chance to let a bug slip as it's likely not every scenario is tested with both NullIO and a real IO.

### DIFF
--- a/lib/puma/null_io.rb
+++ b/lib/puma/null_io.rb
@@ -18,8 +18,22 @@ module Puma
 
     # Mimics IO#read with no data.
     #
-    def read(count = nil, _buffer = nil)
-      count && count > 0 ? nil : ""
+    def read(length = nil, buffer = nil)
+      if length.to_i < 0
+        raise ArgumentError, "(negative length #{length} given)"
+      end
+
+      buffer = if buffer.nil?
+        "".b
+      else
+        String.try_convert(buffer) or raise TypeError, "no implicit conversion of #{buffer.class} into String"
+      end
+      buffer.clear
+      if length.to_i > 0
+        nil
+      else
+        buffer
+      end
     end
 
     def rewind
--- a/test/test_null_io.rb
+++ b/test/test_null_io.rb
@@ -47,9 +47,68 @@ class TestNullIO < Minitest::Test
     assert_nil nio.read(1)
   end
 
+  def test_read_with_negative_length
+    error = assert_raises ArgumentError do
+      nio.read(-42)
+    end
+    assert_includes error.message, "negative length -42 given"
+  end
+
+  def test_read_with_nil_buffer
+    assert_equal "", nio.read(nil, nil)
+    assert_equal "", nio.read(0, nil)
+    assert_nil nio.read(1, nil)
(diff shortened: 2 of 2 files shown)


MSP-Greg:
Thanks for the PR.  Five of the tests fail when only the test changes are added to master.

LGTM.  Any thoughts on the below, which removes two asserts inside of `assert_raises` blocks, and changes `::Tempfile.new` to `::Tempfile.create`?  Don't feel strongly about them...

```patch
diff --git a/test/test_null_io.rb b/test/test_null_io.rb
index a3b2a2b5..dd86ffc8 100644
--- a/test/test_null_io.rb
+++ b/test/test_null_io.rb
@@ -77,11 +77,11 @@ class TestNullIO < Minitest::Test
     assert_includes error.message, "no implicit conversion of Object into String"
 
     error = assert_raises TypeError do
-      assert_equal "", nio.read(0, Object.new)
+      nio.read(0, Object.new)
     end
 
     error = assert_raises TypeError do
-      assert_nil nio.read(1, Object.new)
+      nio.read(1, Object.new)
     end
     assert_includes error.message, "no implicit conversion of Object into String"
   end
@@ -133,7 +133,7 @@ end
 # ensure all the test behavior is accurate
 class TestNullIOConformance < TestNullIO
   def setup
-    self.nio = ::Tempfile.new
+    self.nio = ::Tempfile.create
     nio.sync = true
   end
 
@@ -141,4 +141,10 @@ class TestNullIOConformance < TestNullIO
     self.nio = StringIO.new
     super
   end
+
+  def teardown
+    return unless nio.is_a? ::File
+    nio.close
+    File.unlink nio.path
+  end
 end
```

casperisfine:
> removes two asserts inside of `assert_raises` blocks

Yup oversight on my part. Thanks for catching it.

casperisfine:
PR updated.

MSP-Greg:
@casperisfine Thanks again for the PR.  As you stated "but every small difference has a chance to let a bug slip"

MSP-Greg:
@casperisfine

Oversights.  I always work locally with Ruby head builds, as I feel I should 'walk what I talk'.

A few tests check the response for a frozen IO, which should raise `FrozenError`.  But `FrozenError` is not defined in Ruby 2.4.

My mistake.  Note to self - always review the CI.

\<Rant on\> why are we supporting **four** EOL Ruby versions? \<Rant off\>

I'll fix...

casperisfine:
Ah my bad, I should have paid attention to CI too.

> FrozenError is not defined in Ruby 2.4.

Oh wow, crazy, I didn't remember `FrozenError` being such a novelty, but indeed:

Capture d’écran 2023-11-16 à 16 55 21

MSP-Greg:
I'm working on a fix, some non-MRI errors also.  I'm not going to start on the EOL Ruby thing.

> Oh wow, crazy, I didn't remember FrozenError being such a novelty, but indeed

I don't think it's fair to expect anyone to remember that...

PR #2631, opened 2021-05-18
Update docs for low_latency option and fix parsing issue

seangoedecke:
### Description
Closes https://github.com/puma/puma/issues/2623

As well as updating the docs to reflect that TCP_NODELAY is off by default, I've tweaked the code so that `low_latency=false` doesn't set TCP_NODELAY, which is consistent with how other options are parsed. Let me know if the tests are a bit overkill.

I'm not sure how best to test that the created socket has or lacks TCP_NODELAY in JRuby, since this version of JRuby has a very minimal Socket implementation (no constants, no bool method). For now I've skipped my new tests if we're running them in JRuby, but any guidance would be appreciated.

### Your checklist for this pull request

- [x] I have reviewed the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.
- [x] I have added (or updated) appropriate tests if this PR fixes a bug or adds a feature.
- [x] My pull request is 100 lines added/removed or less so that it can be easily reviewed.
- [ ] If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.
- [x] If this closes any issues, I have added "Closes `#issue`" to the PR description or my commit messages.
- [x] I have updated the documentation accordingly.
- [x] All new and existing tests passed, including Rubocop.

PR #3281, opened 2023-11-20
Add comment about `NoDelay=true` in systemd docs

evenreven:
### Description
Since NoDelay apparently is no longer the default, the example in the systemd docs should also reflect this. This follows up https://github.com/puma/puma/pull/2631.

If the line should be removed instead of commented out (or if the added comment is unclear or unhelpful), please let me know!

(This is my first PR here, and I hope I've ticked the right boxes. Sorry if I overlooked anything!)

### DIFF
--- a/docs/systemd.md
+++ b/docs/systemd.md
@@ -101,7 +101,8 @@ ListenStream=0.0.0.0:9293
 # Socket options matching Puma defaults
 ReusePort=true
 Backlog=1024
-# Socket option NoDelay will set TCP_NODELAY if puma is run with parameter low_latency. This is off by default.
+# Enable this if you're using Puma with the "low_latency" option, read more in Puma DSL docs and systemd docs:
+# https://www.freedesktop.org/software/systemd/man/latest/systemd.socket.html#NoDelay=
 # NoDelay=true
 
 [Install]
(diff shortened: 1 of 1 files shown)


docs/systemd.md
```diff
@@ -99,9 +99,10 @@ ListenStream=0.0.0.0:9293
 # ListenStream=/run/puma.sock
 
 # Socket options matching Puma defaults
-NoDelay=true
 ReusePort=true
 Backlog=1024
+# Socket option NoDelay will set TCP_NODELAY if puma is run with parameter low_latency. This is off by default.
```
dentarg, line 104: WDYT?

```suggestion
# Enable this if you're using Puma with the "low_latency" option, read more in Puma DSL docs and systemd docs:
# https://www.freedesktop.org/software/systemd/man/latest/systemd.socket.html#NoDelay=
```

dentarg:
Another option could be to just remove this line and trust people using the `low_latency` and systemd understand that they should also use the `NoDelay` option in systemd, as the Puma docs do mention `TCP_NODELAY`:

https://github.com/puma/puma/blob/32e011ab9e029c757823efb068358ed255fb7ef4/lib/puma/dsl.rb#L256-L258

nateberkopec:
Poke @evenreven

evenreven:
Sorry for the delay, @dentarg and @nateberkopec! I committed @dentarg's suggestion now, but I'd be open to just remove the lines altogether if you prefer that. That's your call.

PR #3283, opened 2023-11-28
[Fix #3282] `idle-timeout` not waiting on all workers in cluster mode

### LINKED ISSUE #3282: `--idle-timeout` should not shut down entire cluster if any single worker is idle
**Describe the bug**

#3235 was introduced to fix #3228, but now it will shut down the entire cluster if any single worker is idle. This is currently unreleased but staged in [v6.4.1](https://github.com/puma/puma/pull/3261).

For example, `bundle exec puma -C config/puma.rb -w 10 --idle-timeout 10` seems like if the server gets 1 request every 5 seconds (less than the 10 second idle timeout), it will still shutdown. That doesn't seem right.

**Expected behavior**

I would expect that the idle-timeout is applied to the entire cluster. So any request to the cluster as a whole would reset the idle timer.

joshuay03:
### Description

Closes https://github.com/puma/puma/issues/3282

Ensures that the master process (and therefore the entire cluster) is only gracefully shut down when all workers have reached the idle timeout when in cluster mode, as opposed to any worker (current implementation).

<details>
<summary>Also adds a log to indicate that the idle time out has been reached in both modes</summary>

```
* Puma version: 6.4.0 (ruby 3.2.2-p53) ("The Eagle of Durango")
*  Min threads: 0
*  Max threads: 5
*  Environment: development
*          PID: 70263
Use Ctrl-C to stop
- Idle timeout reached # 👈🏽
=== puma shutdown: 2023-11-29 16:15:05 +1000 ===
- Goodbye!
```

```
[70332] Puma starting in cluster mode...
[70332] * Puma version: 6.4.0 (ruby 3.2.2-p53) ("The Eagle of Durango")
[70332] *  Min threads: 0
[70332] *  Max threads: 5
[70332] *  Environment: development
[70332] *   Master PID: 70332
[70332] *      Workers: 2
[70332] *     Restarts: (✔) hot (✔) phased
[70332] * Listening on http://0.0.0.0:9292
[70332] Use Ctrl-C to stop
[70332] - Worker 0 (PID: 70346) booted in 0.0s, phase: 0
[70332] - Worker 1 (PID: 70347) booted in 0.0s, phase: 0
[70332] - All workers reached idle timeout # 👈🏽
[70332] - Gracefully shutting down workers...
[70332] === puma shutdown: 2023-11-29 16:15:35 +1000 ===
[70332] - Goodbye!
```

</details>

FWIW I've manually tested this against the Rails monolith I work on and it's looking good!

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -158,6 +158,10 @@ module Puma
       @workers.all? { |w| w.phase == @phase }
     end
 
+    def all_workers_idle_timed_out?
+      (@workers.map(&:pid) - idle_timed_out_worker_pids).empty?
+    end
+
     def check_workers
       return if @next_check >= Time.now
 
@@ -344,6 +348,8 @@ module Puma
     def run
       @status = :run
 
+      @idle_workers = {}
+
       output_header "cluster"
 
       # This is aligned with the output from Runner, see Runner#output_header
@@ -417,6 +423,8 @@ module Puma
 
       @master_read, @worker_write = read, @wakeup
 
+      @options[:worker_write] = @worker_write
+
       @config.run_hooks(:before_fork, nil, @log_writer)
 
       spawn_workers
@@ -432,6 +440,11 @@ module Puma
 
         while @status == :run
           begin
+            if all_workers_idle_timed_out?
+              log "- All workers reached idle timeout"
+              break
+            end
+
             if @phased_restart
               start_phased_restart
               @phased_restart = false
@@ -483,17 +496,23 @@ module Puma
                     debug_loaded_extensions("Loaded Extensions - master:") if @log_writer.debug?
                     booted = true
                   end
+                when "i"
+                  if @idle_workers[pid]
+                    @idle_workers.delete pid
+                  else
+                    @idle_workers[pid] = true
+                  end
                 end
               else
                 log "! Out-of-sync worker list, no #{pid} worker"
               end
             end
+
             if in_phased_restart && workers_not_booted.zero?
               @events.fire_on_booted!
               debug_loaded_extensions("Loaded Extensions - master:") if @log_writer.debug?
               in_phased_restart = false
             end
-
           rescue Interrupt
             @status = :stop
           end
@@ -580,6 +599,10 @@ module Puma
           w.kill
         end
       end
+    end
+
+    def idle_timed_out_worker_pids
+      @idle_workers.keys
     end
   end
 end
--- a/test/test_integration_cluster.rb
+++ b/test/test_integration_cluster.rb
@@ -226,7 +226,10 @@ class TestIntegrationCluster < TestIntegration
 
     get_worker_pids # wait for workers to boot
 
-    connect
+    10.times {
+      fast_connect
+      sleep 0.5
+    }
 
     sleep 1.15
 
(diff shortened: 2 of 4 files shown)


joshuay03:
`turbo-rails` CI failures seem to be coming from upstream: https://github.com/hotwired/turbo-rails/actions/runs/6979009463/job/18991454469

test/test_integration_cluster.rb
```diff
@@ -222,13 +222,19 @@ def test_worker_timeout
   end
 
   def test_idle_timeout
```
joshuay03, line 224: This updated test fails on master:
Screenshot 2023-11-28 at 7 10 04 pm

joshuay03:
> One thing. We cannot change the method signature of `Server.new`, as it will break a lot of external code that uses just a `Puma::Server`. I'd suggest adding it to `options` before the call to `Server.new`, then removing it in `Server.new`.

That's fair, I'll update.

test/test_integration_cluster.rb
```diff
@@ -226,7 +226,10 @@ def test_idle_timeout
 
     get_worker_pids # wait for workers to boot
 
-    connect
+    10.times {
+      fast_connect
+      sleep 0.5
+    }
```
joshuay03, line 232: The updated test fails on master with:

Screenshot 2023-12-01 at 9 56 34 am

lib/puma/cluster.rb
```diff
@@ -483,17 +496,23 @@ def run
                     debug_loaded_extensions("Loaded Extensions - master:") if @log_writer.debug?
                     booted = true
                   end
+                when "i"
+                  if @idle_workers[pid]
+                    @idle_workers.delete pid
```
BrandonDefachelleClouet, line 501: Hey,

_FYI I am new to Puma and may not fully understand the purpose of each possible value of `req`._

I had a bit of trouble from a newby perspective to understand why receiving `i` would make you remove it from the `@idle_workers`. I know understand that it means the worker is active so you want to actually remove it from the `@idle_workers` which make total sense.

Nevertheless these idle-me and idle-me-out requests require to read another file to fully understand the logic behind it.

Maybe to ease the reading of it we could:
1. Use a dedicated `req` letter to remove the `pid` from the `@idle_workers`, such as `a` for `active`?
2. Would it make sense to hide the technical `req` letters behind an enum such as:
```rb
module RequestType
  IDLE = "i"
  BOOT = "b"
  ACTIVE = "a"
  TERM = "t"
  EXTERNAL_TERM = "e"
end
```

joshuay03: Great question and suggestions.

This is simply the prefix for the PID written to the pipe [here](https://github.com/joshuay03/puma/blob/2bd0b49b2b9a58bd35ee73add1da038f61815100/lib/puma/server.rb#L352). I didn't find it very intuitive when working on this either, and my original implementation had a two letter prefix (one for enable, one for disable) which forced me to update the rest of these "actions" to match (`i0`, `i1`, `te` etc.). The PR ended up with a few too many changes which would have made it harder to review so I went with a single "toggle" since the idle state in `@idle_workers` will be synchronous on a per-worker basis.

I do agree though that we can clean this up a bit and I'm a fan of your suggestion, but imo those changes should be in a separate PR.

joshuay03: Actioning in #3318

Issue #3284, opened 2023-12-01
Puma responds with a 501 Server Error when a client sends invalid HTTP methods


danielmconrad:
When [SUPPORTED_HTTP_METHODS](https://github.com/puma/puma/blob/master/lib/puma/const.rb#L154) was added to the codebase, Puma began [responding with a 501 error](https://github.com/puma/puma/blob/3d33475dd400c8d097889ddd93fab38d080c25e0/lib/puma/request.rb#L94) when clients send in unknown or invalid HTTP methods. 

**Why this is problematic:**
As is common on the web, bots will attempt to penetrate web applications by using rare or incorrect methods. This can result in applications throwing dozens or hundreds of 5xx errors, co-mingling with _actual_ server errors that come from application code. 

**Proposed solution:**
Simply respond with a [400 Bad Request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400) or a [405 Method Not Allowed](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405).

dentarg:
501 sounds like the correct one: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501

We have history in this topic in https://github.com/puma/puma/issues/3014 and https://github.com/puma/puma/pull/3106

Since Puma 6.3.0 you can allow any method and let your app define the response

dentarg:
I think Puma should change the default of the DSL option to any, so Puma behaves like Puma v5 did: https://github.com/puma/puma/issues/3014#issuecomment-1398482101

@nateberkopec are you willingly to make a decision on this?

MSP-Greg:
> 501 sounds like the correct one

Agree.  See [RFC9110 501 Not Implemented](https://www.rfc-editor.org/rfc/rfc9110.html#section-15.6.2), which states:

This is the appropriate response when the server does not recognize the request method and is not capable of supporting it for any resource.

As to a change of default, don't feel strongly either way.

danielmconrad:
My suggestion above for 405 is likely not correct, since:

> 405 Method Not Allowed response status code indicates that the server knows the request method

We certainly don't know them in many cases. A few examples of the methods we see are, `TRACK`, `MTJTSS`, `SSTP_DUPLEX_POST`, and `NYTVNJ`.

As for 501, according to the [notes on the page](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501), 

[...]

PR #3286, opened 2023-12-07
Return 405, not 501, when an unsupported HTTP method is used

mcmillan:
### Description

Currently, when the `supported_http_methods` DSL config option is not set to `:any`, the behaviour upon encountering an unexpected HTTP method is to return an `501 Not Implemented` response.

Whilst this is technically correct, it can be somewhat problematic; in situations where 5xx error responses are used as a heuristic to determine whether or not an application is healthy, a single person sending a flood of requests with invalid HTTP methods can result in an application being considered 'unhealthy' because Puma returns a 501.

It seems the somewhat more semantically correct response code to use in this scenario is `405 Method Not Allowed` – this is how Rails deals with HTTP methods it can't handle, for example. The HTTP spec requires that an `Allow` header is returned alongside it with details of which methods _are_ acceptable.

To that end, this modifies the behaviour of Puma to return a 405 when supplied an unsupported HTTP method, rather than a 501, and return an `Allow` header with the valid methods.

### DIFF
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -101,7 +101,11 @@ module Puma
           end
         else
           @log_writer.log "Unsupported HTTP method used: #{env[REQUEST_METHOD]}"
-          status, headers, app_body = [501, {}, ["#{env[REQUEST_METHOD]} method is not supported"]]
+          status, headers, app_body = [
+            405,
+            {"Allow" => @supported_http_methods.keys.join(", ")},
+            ["#{env[REQUEST_METHOD]} method is not supported"]
+          ]
         end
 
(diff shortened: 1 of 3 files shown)


dentarg:
Did you see https://github.com/puma/puma/issues/3284?

dentarg:
I will reiterate my comment from that issue: https://github.com/puma/puma/issues/3284#issuecomment-1836907289

I think we should go back to the way it was in Puma 5 and let the apps define this response 

cc @nateberkopec again because he is the decision maker here

mcmillan:
Ah – I didn't. Sorry folks; I'll close this!

PR #3297, opened 2023-12-24
[Fix #3044] Set conditional config defaults after CLI options are parsed and config files are loaded

### LINKED ISSUE #3044: `preload_app` doesn't default to `true` if more than 1 worker
**Describe the bug**
The [comment][d] and the [code][a] suggests that if there are more than 1 worker, `preload_app` should be `true`, but it isn't.

**Puma config:**
The command line option is `-w 2`.

**To Reproduce**
`a.sh`:

```sh
docker run --rm --init -it alpine:3.17 sh -euc '
  with_config=$1
  apk add build-base ruby ruby-dev ruby-bundler
  echo "
    source \"https://rubygems.org\"
    gem \"puma\", \"6.0.1\"
  " > Gemfile
  echo "
    run lambda { |env| [200, {\"content-type\" => \"text/plain\"}, [\"OK\"]] }
  " > config.ru
  echo "
    preload_app!
  " > config.rb
  bundle
  puma -w 2 ${with_config:+-C config.rb}
' - "${1-}"
```

```
$ sh a.sh
...
[136] Puma starting in cluster mode...
[136] * Puma version: 6.0.1 (ruby 3.1.3-p185) ("Sunflower")
[136] *  Min threads: 0
[136] *  Max threads: 5
[136] *  Environment: development
[136] *   Master PID: 136
[136] *      Workers: 2
[136] *     Restarts: (✔) hot (✔) phased
[136] * Listening on http://0.0.0.0:9292
[136] Use Ctrl-C to stop
[137] + Gemfile in context: /Gemfile
[138] + Gemfile in context: /Gemfile
[136] - Worker 1 (PID: 138) booted in 0.02s, phase: 0
[136] - Worker 0 (PID: 137) booted in 0.02s, phase: 0
```

```
$ sh a.sh with-config
[136] Puma starting in cluster mode...
[136] * Puma version: 6.0.1 (ruby 3.1.3-p185) ("Sunflower")
[136] *  Min threads: 0
[136] *  Max threads: 5
[136] *  Environment: development
[136] *   Master PID: 136
[136] *      Workers: 2
[136] *     Restarts: (✔) hot (✖) phased
[136] * Preloading application
[136] * Listening on http://0.0.0.0:9292
[136] Use Ctrl-C to stop
[137] + Gemfile in context: /Gemfile
[136] - Worker 0 (PID: 137) booted in 0.0s, phase: 0
[138] + Gemfile in context: /Gemfile
[136] - Worker 1 (PID: 138) booted in 0.0s, phase: 0
```

This happens because at [this point][a] `@options[:workers]` is 0, since file options become available [here][b], and user options [here][c].

Just in case, this has nothing to do with `docker`.

**Expected behavior**
Either it should default to `true`, or [that piece][a] should be removed.

 - OS: Linux
 - Puma Version 6.0.1

[a]: https://github.com/puma/puma/blob/v6.0.1/lib/puma/configuration.rb#L182
[b]: https://github.com/puma/puma/blob/v6.0.1/lib/puma/launcher.rb#L54
[c]: https://github.com/puma/puma/blob/v6.0.1/lib/puma/cli.rb#L45
[d]: https://github.com/puma/puma/blob/v6.0.1/lib/puma/dsl.rb#L712

joshuay03:
### Description

Closes https://github.com/puma/puma/issues/3044

`CLI` options are parsed during `CLI` initialisation, but after `Configuration` initialisation. This PR changes this so that the parsing takes place during `Configuration` initialisation. This ensures that `CLI` options are also used when some config defaults are conditionally set during initialisation.

~**Note:** Even with the changes in this PR `preload_app` isn't enabled by default if using `#workers` in a config file. I'm happy to address that here too if prefered. Else we have an inconsistency with the cli and config file.~

Addressed in [these changes](https://github.com/puma/puma/compare/3ac0f65f5e98adde1d42e1361ea491a2ae29e103..d4f9f5d489ce37d7be690f89a085ef7639a1a11f). We now ensure that conditional defaults are also set when config files are loaded.

### Warning

The side effect of these changes which fixes the linked issue is that `preload_app` is now consistently enabled by default regardless of configuration method if [these conditions are met](https://github.com/joshuay03/puma/blob/b081b02a17459a1d86fa18eb5ede2a25c1587737/lib/puma/configuration.rb#L384-L387).

### DIFF
--- a/test/test_cli.rb
+++ b/test/test_cli.rb
@@ -349,7 +349,6 @@ class TestCLI < PumaTest
     ENV['APP_ENV'] = 'test'
 
     cli = Puma::CLI.new []
-    cli.send(:setup_options)
 
     conf = cli.instance_variable_get(:@conf)
     conf.clamp
@@ -364,7 +363,6 @@ class TestCLI < PumaTest
     ENV['RACK_ENV'] = @environment
 
     cli = Puma::CLI.new []
-    cli.send(:setup_options)
 
     conf = cli.instance_variable_get(:@conf)
     conf.clamp
@@ -377,7 +375,6 @@ class TestCLI < PumaTest
     ENV['RAILS_ENV'] = @environment
 
     cli = Puma::CLI.new []
-    cli.send(:setup_options)
 
     conf = cli.instance_variable_get(:@conf)
     conf.clamp
@@ -389,7 +386,6 @@ class TestCLI < PumaTest
 
   def test_silent
     cli = Puma::CLI.new ['--silent']
-    cli.send(:setup_options)
 
     log_writer = cli.instance_variable_get(:@log_writer)
 
@@ -402,9 +398,26 @@ class TestCLI < PumaTest
     assert_empty Puma::Plugins.instance_variable_get(:@plugins)
 
     cli = Puma::CLI.new ['--plugin', 'tmp_restart', '--plugin', 'systemd']
-    cli.send(:setup_options)
 
     assert Puma::Plugins.find("tmp_restart")
     assert Puma::Plugins.find("systemd")
+  end
+
+  def test_config_does_not_preload_app_with_workers
+    skip_unless :fork
+
+    cli = Puma::CLI.new ['-w 0']
+    config = Puma.cli_config
+
+    assert_equal false, config.options[:preload_app]
+  end
+
+  def test_config_preloads_app_with_workers
+    skip_unless :fork
+
+    cli = Puma::CLI.new ['-w 2']
+    config = Puma.cli_config
+
+    assert_equal true, config.options[:preload_app]
   end
 end
--- a/test/test_config.rb
+++ b/test/test_config.rb
@@ -651,6 +651,45 @@ class TestConfigFile < PumaTest
     assert_kind_of Array, conf.config_files
   end
 
+  def test_config_does_not_preload_app_if_not_using_workers
+    conf = Puma::Configuration.new({ workers: 0 })
+    conf.clamp
+
+    assert_equal false, conf.options.default_options[:preload_app]
+  end
+
+  def test_config_preloads_app_if_using_workers
+    conf = Puma::Configuration.new({ workers: 2 })
+    conf.clamp
+    preload = Puma.forkable?
+
+    assert_equal preload, conf.options.default_options[:preload_app]
+  end
+
+  def test_config_does_not_preload_app_if_using_workers_and_prune_bundler
+    conf = Puma::Configuration.new({ workers: 2 }) do |c|
+      c.prune_bundler
+    end
+    conf.clamp
+
+    assert_equal false, conf.options.default_options[:preload_app]
+  end
+
+  def test_config_file_does_not_preload_app_if_not_using_workers
+    conf = Puma::Configuration.new { |c| c.load 'test/config/workers_0.rb' }
+    conf.clamp
+
+    assert_equal false, conf.options.default_options[:preload_app]
+  end
+
+  def test_config_file_preloads_app_if_using_workers
+    conf = Puma::Configuration.new { |c| c.load 'test/config/workers_2.rb' }
+    conf.clamp
+    preload = Puma.forkable?
+
+    assert_equal preload, conf.options.default_options[:preload_app]
+  end
+
   private
 
   def assert_run_hooks(hook_name, options = {})
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ For an in-depth discussion of the tradeoffs of thread and process count settings
 
 In cluster mode, Puma can "preload" your application. This loads all the application code *prior* to forking. Preloading reduces total memory usage of your application via an operating system feature called [copy-on-write](https://en.wikipedia.org/wiki/Copy-on-write).
 
-If the `WEB_CONCURRENCY` environment variable is set to a value > 1 (and `--prune-bundler` has not been specified), preloading will be enabled by default. Otherwise, you can use the `--preload` flag from the command line:
+If the number of workers is greater than 1 (and `--prune-bundler` has not been specified), preloading will be enabled by default. Otherwise, you can use the `--preload` flag from the command line:
 
 ```
 $ puma -w 3 --preload
--- a/lib/puma/cli.rb
+++ b/lib/puma/cli.rb
@@ -39,10 +39,8 @@ module Puma
       @control_url = nil
       @control_options = {}
 
-      setup_options env
-
       begin
-        @parser.parse! @argv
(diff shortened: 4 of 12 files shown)


test/test_cli.rb
```diff
@@ -352,7 +352,6 @@ def test_environment_app_env
     ENV['APP_ENV'] = 'test'
 
     cli = Puma::CLI.new []
-    cli.send(:setup_options)
```
joshuay03, line 355: I think these are unnecessary, options are set up when initialised.

test/test_cli.rb
```diff
@@ -396,9 +392,26 @@ def test_plugins
...
+    assert_equal false, config.options[:preload_app]
+  end
+
+  def test_config_preloads_app_with_workers
+    skip_unless :fork
+
+    cli = Puma::CLI.new ['-w 2']
+    config = Puma.cli_config
+
+    assert_equal true, config.options[:preload_app]
```
joshuay03, line 415: Before patch:
Screenshot 2023-12-24 at 7 50 33 pm

test/test_config.rb
```diff
@@ -556,6 +556,34 @@ def test_http_content_length_limit
...
+
+    assert_equal false, conf.options.default_options[:preload_app]
+  end
+
+  def test_config_file_preloads_app_if_using_workers
+    conf = Puma::Configuration.new { |c| c.load 'test/config/workers_2.rb' }
+    conf.load
+    preload = Puma.forkable?
+
+    assert_equal preload, conf.options.default_options[:preload_app]
```
joshuay03, line 584: Before patch:
Screenshot 2023-12-24 at 8 19 44 pm

dentarg:
Are the test failures related to (these) tests now using preloading (and that causes some problem)?

While https://github.com/puma/puma/issues/3044 is classed as a bug, it is quite the change in behaviour... so maybe the fix should be shipped in a major version?

joshuay03:
> Are the test failures related to (these) tests now using preloading (and that causes some problem)?

I think the phased restart failures are because of [this clause](https://github.com/puma/puma/blob/83c5b32990e5c0ca5a61baa915e506c00308041f/lib/puma/cluster.rb#L221). I think we should just make it one of the conditions of defaulting `preload_app` to `true` considering that we give users the same advise for their config:

https://github.com/puma/puma/blob/83c5b32990e5c0ca5a61baa915e506c00308041f/README.md?plain=1#L139

https://github.com/puma/puma/blob/83c5b32990e5c0ca5a61baa915e506c00308041f/lib/puma/dsl.rb#L800-L801

> While #3044 is classed as a bug, it is quite the change in behaviour... so maybe the fix should be shipped in a major version?

I agree, and yes the issue was added to the `7.0.0` milestone.

nateberkopec (approved):

nateberkopec (dismissed):
Test fails? 🤔

joshuay03:
The last failure is related to https://github.com/puma/puma/pull/3557#discussion_r1864735186

I'm pretty sure we want `fork_worker` enabled for that test, and here I've also disabled preloading for consistency with the other tests since it's not compatible with phased restart. However, it seems that when you do both of those things, phased restarts slow down quite a bit, leading to test timeouts. Setting them individually however is completely fine. 

I'll have a go at looking into this in a separate PR.

dentarg:
So we're waiting for https://github.com/puma/puma/pull/3589 here

joshuay03:
I've updated this to target https://github.com/puma/puma/pull/3616. Once that has been merged, this should go green after rebasing.

joshuay03:
> I've updated this to target #3616. Once that has been merged, this should go green after rebasing.

Nvm, I've just patched those tests here. Now neither is dependant on the other and both should cleanly rebase after the other is merged.

schneems:
I'm not sure why but the UX is telling me that @nateberkopec requested changes even though I don't see the request (and I see an approval from him). All tests are now green. @nateberkopec can you merge in or re-iterate your concerns that seem to have been lost to the snads of time?

joshuay03:
> I'm not sure why but the UX is telling me that @nateberkopec requested changes even though I don't see the request (and I see an approval from him). All tests are now green. @nateberkopec can you merge in or re-iterate your concerns that seem to have been lost to the snads of time?

[It was because there was a test failure](https://github.com/puma/puma/pull/3297#pullrequestreview-2465989652):

> Test fails? 🤔

Which has since been resolved:

> The last failure is related to [#3557 (comment)](https://github.com/puma/puma/pull/3557#discussion_r1864735186)
> 
> I'm pretty sure we want `fork_worker` enabled for that test, and here I've also disabled preloading for consistency with the other tests since it's not compatible with phased restart. However, enabling `fork_worker` seems to slow the test down enough for it to time out.
> 
> **Edit:** Resolved in #3589

MSP-Greg:
> Now neither is dependant on the other and both should cleanly rebase

Re #3616, just checking if the above is still true.  I haven't looked much at this one, but @nateberkopec went thru it...

joshuay03:
> > Now neither is dependant on the other and both should cleanly rebase
> 
> Re #3616, just checking if the above is still true. I haven't looked much at this one, but @nateberkopec went thru it...

Probably not with my recent changes. Once #3616 has been merged, I'll come and clean this up. I might have to update the `#clamp` [changes](https://github.com/puma/puma/pull/3297/files#diff-2dc4e3e83be7fd97cebc482ae07d6a8216944003de82458783fb00b5ae9524c8R260-R265) and I can remove some unnecessary `#load` [calls](https://github.com/puma/puma/pull/3297/files#diff-f8a5111e8509bf1e3c382deae1d40eb8640626bd81cd5066577d3082d8e0a7dfR763-R764) in tests.

FWIW, if I had to pick a PR (of mine) that definitely should go out with v7, it would be this one. I've worked with 2 monoliths now where they almost set `WEB_CONCURRENCY` without realising they would be enabling preloading, thinking that when they bumped to v5, preloading was already enabled. It's quite confusing default behaviour. We should also include a big warning next to this PR in the changelog that bumping to v7 would enable preloading by default if in cluster mode, regardless of whether the configuration is done via DSL or env var, and that people should set `preload_app!(false)` to maintain current behaviour—highly advised because you need to be careful when enabling this if you have threads spawned on boot.

cc @schneems

schneems:
#3616 is merged. Can you rebase on master/main (thanks).

MSP-Greg:
Cancel 'older' CI tests?

joshuay03:
> Cancel 'older' CI tests?

Yes please, thank you!

joshuay03:
> #3616 is merged. Can you rebase on master/main (thanks).

Done 👍🏽

PR #3308, opened 2024-01-08
Fix response repeated status line when request is invalid or errors are raised

### LINKED ISSUE #3307: Error responses written twice
For example, for a parser error:

```
assert_equal "HTTP/1.1 400 Bad Request\r\nConnection: close\r\nContent-Length: 0\r\n\r\n", data
```

doesn't pass, because the actual response is:

```
"HTTP/1.1 400 Bad Request\r\nConnection: close\r\nContent-Length: 0\r\n\r\nHTTP/1.1 400 Bad Request\r\n\r\n"
```

The problem comes from `prepare_response` and `client.write_error` essentially trying to do the same thing. We should fix this overlap of responsbilities.

MSP-Greg:
### Description

The third commit fixes the error.  If one runs the test suite with the first two commits, the following failures occur:
```
TestPumaServer#test_large_chunked_request_header [/mnt/c/Greg/GitHub/puma/test/test_puma_server.rb:917] 

WebServerTest#test_bad_path [/mnt/c/Greg/GitHub/puma/test/test_web_server.rb:72]
```

Also, the 2nd error above generates a log of the below, which is a pita to read:
```
  1) Failure:
WebServerTest#test_bad_path [/mnt/c/Greg/GitHub/puma/test/test_web_server.rb:72]:
Expected /\.rb:\d+:in `[^']+'\z/ to match # encoding: ASCII-8BIT
#    valid: true
"HTTP/1.1 400 Bad Request\r\nContent-Length: 491\r\n\r\nPuma caught this error: Puma::HttpParserError (Puma::HttpParserError)\n/mnt/c/Greg/GitHub/puma/lib/puma/request.rb:426:in `rescue in normalize_env'\n/mnt/c/Greg/GitHub/puma/lib/puma/request.rb:423:in `normalize_env'\n/mnt/c/Greg/GitHub/puma/lib/puma/request.rb:63:in `handle_request'\n/mnt/c/Greg/GitHub/puma/lib/puma/server.rb:465:in `process_client'\n/mnt/c/Greg/GitHub/puma/lib/puma/server.rb:246:in `block in run'\n/mnt/c/Greg/GitHub/puma/lib/puma/thread_pool.rb:155:in `block in spawn_thread'HTTP/1.1 400 Bad Request\r\n\r\n".
```

The addition of `test_puma/assertions.rb` changes it.  If the object of an `assert_match` is longer than 80 characters, the string is output as is, not inspected, as below, which clearly shows the repeated status line:
```
  1) Failure:
WebServerTest#test_bad_path [/mnt/c/Greg/GitHub/puma/test/test_web_server.rb:72]:
Expected /\.rb:\d+:in `[^']+'\z/ to match:
HTTP/1.1 400 Bad Request
Content-Length: 491

Puma caught this error: Puma::HttpParserError (Puma::HttpParserError)
/mnt/c/Greg/GitHub/puma/lib/puma/request.rb:426:in `rescue in normalize_env'
/mnt/c/Greg/GitHub/puma/lib/puma/request.rb:423:in `normalize_env'
/mnt/c/Greg/GitHub/puma/lib/puma/request.rb:63:in `handle_request'
/mnt/c/Greg/GitHub/puma/lib/puma/server.rb:465:in `process_client'
/mnt/c/Greg/GitHub/puma/lib/puma/server.rb:246:in `block in run'
/mnt/c/Greg/GitHub/puma/lib/puma/thread_pool.rb:155:in `block in spawn_thread'HTTP/1.1 400 Bad Request
```

I hope others also find that much easier to read...

Closes #3307

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -570,7 +570,6 @@ module Puma
     def response_to_error(client, requests, err, status_code)
       status, headers, res_body = lowlevel_error(err, client.env, status_code)
       prepare_response(status, headers, res_body, requests, client)
-      client.write_error(status_code)
     end
     private :response_to_error
 
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -13,6 +13,9 @@ end
 
 require "securerandom"
 
+# needs to be loaded before minitest for Ruby 2.7 and earlier
+require_relative "helpers/test_puma/assertions"
+
 require_relative "minitest/verbose"
 require "minitest/autorun"
 require "minitest/pride"
--- a/test/helpers/test_puma/assertions.rb
+++ b/test/helpers/test_puma/assertions.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+module TestPuma
+  module Assertions
+    def assert_start_with(obj, str, msg = nil)
+      msg = message(msg) {
+        "Expected\n#{obj}\nto start with #{str}"
+      }
+      assert_respond_to obj, :start_with?
+      assert obj.start_with?(str), msg
+    end
+
+    def assert_end_with(obj, str, msg = nil)
+      msg = message(msg) {
+        "Expected\n#{obj}\nto end with #{str}"
+      }
+      assert_respond_to obj, :end_with?
+      assert obj.end_with?(str), msg
+    end
+
+    # if obj is longer than 80 characters, show as string, not inspected
+    def assert_match(matcher, obj, msg = nil)
+      msg = if obj.length < 80
+        message(msg) { "Expected #{mu_pp matcher} to match #{mu_pp obj}" }
+      else
+        message(msg) { "Expected #{mu_pp matcher} to match:\n#{obj}\n" }
+      end
+      assert_respond_to matcher, :"=~"
+      matcher = Regexp.new Regexp.escape matcher if String === matcher
+      assert matcher =~ obj, msg
+    end
+  end
+end
+
+module Minitest
+  module Assertions
+    prepend TestPuma::Assertions
+  end
+end
--- a/test/test_puma_server.rb
+++ b/test/test_puma_server.rb
@@ -424,7 +424,7 @@ class TestPumaServer < Minitest::Test
     data = send_http_and_read "GET / HTTP/1.0\r\n\r\n"
 
     refute_match(/don't leak me bro/, data)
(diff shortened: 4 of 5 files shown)


dentarg:
> The addition of `test_puma/assertions.rb` changes it. If the object of an `assert_match` is longer than 80 characters, the string is output as is, not inspected, as below, which clearly shows the repeated status line:

Very nice. 👍

dentarg (approved):

dentarg:
Future idea, instead of asserting match/start with, should it be exact? So we don't have regressions like this

It is too easy to think you can just tweak the assertion to make your test pass (https://github.com/puma/puma/pull/3094#discussion_r1136573131)

MSP-Greg:
> Future idea, instead of asserting match/start with, should it be exact?

The assert in `test_puma_server.rb` was changed to `assert_equal` in https://github.com/puma/puma/pull/3308/commits/9c6a29af801c

The other test in `test_web_server.rb` I added an `assert_match` that should match a backtrace line, which is what the response should end with.

> It is too easy to think you can just tweak the assertion to make your test pass

Possible, but I think PR's need to be reviewed for that.  Like the above, trying to do an `assert_equal` when a backtrace is involved makes it very brittle, as code changes elsewhere will change the backtrace.

I think the above made sense...

dentarg:
`assert_end_with` sounds like a good middle ground 👍

kenballus:
This bug is potentially a vulnerability.

HTTP proxies that coalesce connections to Puma (i.e. multiple clients' requests are being sent on the same connection to Puma) may end up desynchronizing the request stream because of this bug.

Suppose that clients $C_1$, $C_2$, and $C_3$ send requests $R_1$, $R_2$, and $R_3$ to the proxy $P$, which forwards them all on the same connection, in order, to the Puma server $S$. If $S$ generates two error responses to $R_1$, then the proxy may mistakenly associate the second error response to $R_1$ with $R_2$, and then associate the response to $R_2$ with $R_3$. This may give $C_3$ access to private information intended for $C_2$.

While good proxies will close connections after receiving 400, not all do. Until relatively recently [AWS cloudfront would even cache 400 responses(!!)](https://cpdos.org/paper/Your_Cache_Has_Fallen__Cache_Poisoned_Denial_of_Service_Attack__Preprint_.pdf).

PR #3320, opened 2024-01-17
Use Rails `7.1` with Ruby head in `turbo-rails` CI

joshuay03:
### Description

Context:
- https://github.com/ruby/ruby/commit/d16f992e1bfacb638b8a9b8b5a7ef8149ee1d50d
- https://github.com/ruby/ruby/commit/1500946ce4ed7d89ed33059e3629e526b1dc207a

Experiment to see if this fixes [a failing CI worflow](https://github.com/puma/puma/actions/runs/7529388305/job/20493515442#step:6:22) since [`mutex_m` was added to the relevant Rails gemfiles](https://github.com/rails/rails/pull/48907) and [included in the `7.1` release](https://github.com/rails/rails/releases/tag/v7.1.0). It has also [since been dropeed](https://github.com/rails/rails/commit/bcdeea5da7657375df21a856135ae7a66a8c3812) on Rails edge.

~Even if this passes, the root cause is that it's missing from the gemfiles in https://github.com/ruby-concurrency/concurrent-ruby. I'm hoping to resolve that in https://github.com/ruby-concurrency/concurrent-ruby/pull/1034.~

~However, we still need to point to a version of Rails that either requires `mutex_m` itelf or uses a version of `concurrent-ruby` that does, so this change is needed regardless of that patch.~

I've also pointed the other rubies to `7.1` for consistency, please let me know if I should revert that.

### DIFF
--- a/.github/workflows/turbo-rails.yml
+++ b/.github/workflows/turbo-rails.yml
@@ -29,7 +29,10 @@ jobs:
           - { os: ubuntu-20.04 , ruby: '3.1', rails: '7.0' }
           - { os: ubuntu-20.04 , ruby: '3.2', rails: '7.0' }
           - { os: ubuntu-22.04 , ruby: '3.3', rails: '7.0' }
-          - { os: ubuntu-22.04 , ruby: head , rails: '7.0' }
+          - { os: ubuntu-20.04 , ruby: '3.1', rails: '7.1' }
+          - { os: ubuntu-20.04 , ruby: '3.2', rails: '7.1' }
+          - { os: ubuntu-22.04 , ruby: '3.3', rails: '7.1' }
+          - { os: ubuntu-22.04 , ruby: head , rails: '7.1' }
     env:
       CI: true
       RAILS_VERSION: "${{ matrix.rails }}"


eregon:
https://github.com/puma/puma/actions/runs/7529388305/job/20493515442 was using concurrent-ruby 1.2.2:
https://github.com/puma/puma/actions/runs/7529388305/job/20493515442#step:4:64
concurrent-ruby 1.2.3 no longer requires mutex_m.
From the stacktrace in that log though it seems to have nothing to do with concurrent-ruby, but active_support/notifications/fanout.rb seems to be the one requiring mutex_m.

joshuay03:
> From the stacktrace in that log though it seems to have nothing to do with concurrent-ruby, but active_support/notifications/fanout.rb seems to be the one requiring mutex_m.

Ah yep my bad, the confusion was from me looking at the diff for the incorrect version. I thought it was from requiring `concurrent/map` [here](https://github.com/rails/rails/blob/8f386d73215867d5f55058a252f066b1040a7f2b/activesupport/lib/active_support/notifications/fanout.rb#L3), but in fact it was requiring `mutex_m` directly, which has been removed [here](https://github.com/rails/rails/commit/bcdeea5da7657375df21a856135ae7a66a8c3812).

Thanks for having a look!

MSP-Greg:
@eregon Thanks for the research.

Note the last rows of https://github.com/ruby/ruby/blame/master/gems/bundled_gems.  There have been more additions after `mutex_m`.

@joshuay03 - thanks for the PR.  I haven't thought about it much (busy), but these jobs don't take long, wondering if there should be a mix of 7.0 and 7.1?

joshuay03:
> @joshuay03 - thanks for the PR. I haven't thought about it much (busy), but these jobs don't take long, wondering if there should be a mix of 7.0 and 7.1?

Sounds good to me, I'll action when I get a chance 👍🏽

joshuay03:
`master` is now [failing](https://github.com/puma/puma/actions/runs/7672076953/job/20911796779#step:6:20) cause it can't load `base64`. This should also fix that but I've rebased to confirm and it looks 👍🏽

joshuay03:
@dentarg @MSP-Greg Are you happy to merge this? The current `turbo-rails` failure is from upstream.

joshuay03:
Thank you @dentarg

PR #3324, opened 2024-01-30
Deterministic `require`s in `test/runner`

dentarg:
We had (rack2) tests fail on 2.4 like this:

    /home/runner/work/puma/puma/lib/rack/handler/puma.rb:140:in `<top (required)>': undefined method `register' for Rack::Handler:Module (NoMethodError)
            from /home/runner/work/puma/puma/test/test_rack_handler.rb:6:in `require'
            from /home/runner/work/puma/puma/test/test_rack_handler.rb:6:in `<module:TestRackUp>'
            from /home/runner/work/puma/puma/test/test_rack_handler.rb:5:in `<top (required)>'
            from test/runner:32:in `require'
            from test/runner:32:in `block in <main>'
            from test/runner:32:in `each'
            from test/runner:32:in `<main>'

https://github.com/puma/puma/actions/runs/7672076948/job/20911807383#step:10:13

### DIFF
--- a/test/runner
+++ b/test/runner
@@ -27,9 +27,9 @@ require 'bundler/setup'
 
 if ARGV.empty? || ARGV.last.start_with?('-')
   if RUBY_VERSION >= '2.5'
-    Dir['test_*.rb', base: __dir__].each { |tf| require_relative tf }
+    Dir['test_*.rb', base: __dir__].sort.each { |tf| require_relative tf }
   else
-    Dir["#{__dir__}/test_*.rb"].each { |tf| require tf }
+    Dir["#{__dir__}/test_*.rb"].sort.each { |tf| require tf }
   end
 else
   file_arg = ARGV.pop.sub(/\.rb\z/, '')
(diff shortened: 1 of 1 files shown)


MSP-Greg (approved):

MSP-Greg:
@dentarg

Thanks.  Rather interesting to look at Ruby docs for this (`Dir.glob`).

Ruby 2.6 states:
"Case sensitivity depends on your system (`File::FNM_CASEFOLD` is ignored), as does the order in which the results are returned."

Earlier Rubies are rather vague.

Ruby 3.x added a `sort` keyword arg, and stated:
"The results which matched single wildcard or character set are sorted in binary ascending order, unless `false` is given as the optional `sort` keyword argument."

@dentarg - you probably saw it, but macOS 14 is now available with Actions...

dentarg:
> You probably saw it, but macOS 14 is now available with Actions...

I did see that :) Finally 🎉

PR #3338, opened 2024-02-22
client.rb - fix request chunked body handling

### LINKED ISSUE #3337: HttpParserError raised when body (chunk) exceeds 4096 bytes does not end with "\r\n"
**Describe the bug**
Looking at the change log, 6.4.2 introduced changes to fix https://github.com/advisories/GHSA-c2f4-cvqm-65w2. I am facing an issue however I am not sure if it is a bug or it's exactly what the changes are meant to fix. 
When sending a payload (more than 4096 bytes) to Puma, using `chunked` transfer-encoding, Puma intermittently raises `#<Puma::HttpParserError: maximum size of chunk header exceeded>`. I then tried to see what Puma receives on its end when the exception happens. This happens in a production system but I could also reproduce the problem with a payload as follows:
```
400\r\n # 1024 bytes
data\r\n # 1024 bytes of data followed by "\r\n"
400\r\n # 1024 bytes
data\r\n # 1024 bytes of data followed by "\r\n"
.... # The above repeats until the total number of bytes exceeds 4096 (I used 10000 bytes)
40 
```
The above payload contains valid chunk headers, however it does not end with one.

The above payload isn't complete. The next `read` by Puma would read the remaining data:
```
0\r\n
data\r\n # 1024 bytes of data followed by "\r\n"
...
0\r\n\r\n
```

**To Reproduce**
Using the following `config.ru`:
```
run lambda { |env| [200, {}, ["Hello World"]] }
```
And `config/puma.rb`:
```
port 8080
```

Run application
```
bundle exec puma
```

Run the following script:
```
require 'socket'

def build_data
  data = ''
  while data.size < 10000
    data += "400\r\n"
    data += ('x' * 1024)
    data += "\r\n"
  end
  data += "0\r\n\r\n"
  data
end

def run_test
  header = "GET /gg/2 HTTP/1.1\r\nConnection: close\r\nTransfer-Encoding: chunked\r\n\r\n"
  socket = TCPSocket.new('127.0.0.1', 8080)
  data = build_data
  data1 = data[0..7218] # Number here is arbitrary, so that the first chunk of data ends with `40`
  data2 = data[7219..-1] # remaining data
  socket << "#{header}#{data1}"
  socket.flush
  sleep 0.1 # This makes it easier to reproduce the issue, might need to be adjusted
  socket << data2
  socket.flush

  result = socket.read

  puts result
end

run_test

```

**Expected behavior**
Request accepted and no parser error raised.

 - OS: Linux
 - Puma Version: 6.4.2

MSP-Greg:
### Description

In client.rb code for processing chunked request bodies, `chunk.size` was used.  This may contain both portions of the body and also hex size data.  Hex size data may also contain 'headers'.  Use `line.size`, as it only contains hex size data.

The first commit contains a new test (`test_chunked_body_pause_within_chunk_size_hex`), which fails on master with the below.  Note that there are several `test_chunked_body_pause` tests, this adds another where a pause exists within the hex encoded length of the next chunk.  Thanks to @skliew, who opened issue #3337.
```
  1) Failure:
TestPumaServer#test_chunked_body_pause_within_chunk_size_hex [test/test_puma_server.rb:1168]:
--- expected
+++ actual
@@ -1,5 +1,15 @@
-"HTTP/1.1 200 OK\r
+# encoding: ASCII-8BIT
+#    valid: true
+"HTTP/1.1 400 Bad Request\r
 Connection: close\r
-Content-Length: 0\r
+Content-Length: 736\r
 \r
-"
+Puma caught this error: maximum size of chunk header exceeded (Puma::HttpParserError)
+../puma/lib/puma/client.rb:646:in 'Puma::Client#decode_chunk'
+../puma/lib/puma/client.rb:530:in 'Puma::Client#setup_chunked_body'
+../puma/lib/puma/client.rb:382:in 'Puma::Client#setup_body'
+../puma/lib/puma/client.rb:275:in 'Puma::Client#try_to_finish'
+../puma/lib/puma/client.rb:287:in 'Puma::Client#eagerly_finish'
+../puma/lib/puma/server.rb:450:in 'Puma::Server#process_client'
+../puma/lib/puma/server.rb:246:in 'block in Puma::Server#run'
+../puma/lib/puma/thread_pool.rb:155:in 'block in Puma::ThreadPool#spawn_thread'"
```

Closes #3337

### DIFF
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -642,7 +642,7 @@ module Puma
             @partial_part_left = len - part.size
           end
         else
-          if @prev_chunk.size + chunk.size >= MAX_CHUNK_HEADER_SIZE
+          if @prev_chunk.size + line.size >= MAX_CHUNK_HEADER_SIZE
             raise HttpParserError, "maximum size of chunk header exceeded"
           end
 
--- a/test/test_puma_server.rb
+++ b/test/test_puma_server.rb
@@ -942,18 +942,60 @@ class TestPumaServer < Minitest::Test
     end
   end
 
-  def test_large_chunked_request_header
+  def test_chunked_request_invalid_extension_header_length
+    body = nil
+    content_length = nil
     server_run(environment: :production) { |env|
-      [200, {}, [""]]
+      body = env['rack.input'].read
+      content_length = env['CONTENT_LENGTH']
+      [200, {}, [body]]
     }
 
     max_chunk_header_size = Puma::Client::MAX_CHUNK_HEADER_SIZE
+
+    # send valid request except for extension_header larger than limit
     header = "GET / HTTP/1.1\r\nConnection: close\r\nContent-Length: 200\r\nTransfer-Encoding: chunked\r\n\r\n"
-    socket = send_http "#{header}1;t#{'x' * (max_chunk_header_size + 2)}"
+    socket = send_http "#{header}1;t=#{'x' * (max_chunk_header_size + 2)}\r\n1\r\nh\r\n4\r\nello\r\n0\r\n\r\n"
 
     data = socket.read
 
     assert_equal "HTTP/1.1 400 Bad Request\r\nConnection: close\r\nContent-Length: 0\r\n\r\n", data
+  end
+
+  def test_chunked_request_invalid_extension_header_length_split
+    body = nil
+    completed_loops = 0
+    content_length = nil
+    server_run { |env|
+      body = env['rack.input'].read
+      content_length = env['CONTENT_LENGTH']
+      [200, {}, [""]]
+    }
+
+    # includes 1st chunk length
+    socket = send_http "GET / HTTP/1.1\r\nConnection: Keep-Alive\r\nTransfer-Encoding: chunked\r\n\r\n1;"
+
+    junk = "*" * 1_024
+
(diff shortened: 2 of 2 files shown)


MSP-Greg:
Added links to this PR and the related issue to the new test.

ianballou:
This appears to fix my issue here: https://github.com/puma/puma/issues/3362

After applying the patch, I'm able to `podman push` to Foreman's/Katello's/Satellite's container registry without exceeding the max chunk size.

qcjames53:
> This appears to fix my issue here: #3362
> 
> After applying the patch, I'm able to `podman push` to Foreman's/Katello's/Satellite's container registry without exceeding the max chunk size.

+1 for this patch. This patch fixes large podman images (tested things over ~100MB) which previously were erroring out during chunk uploads with Foreman/Katello.

MSP-Greg:
I just rebased and added another test that I only tested locally on Ubuntu.  Locally, it fails on master, passes with this PR.

Of course, just to make things fun, it fails on macOS & Windows (note to self - test all OS's locally).  I'll either remove it and save for later, or fix it.  Either way, I'll merge this later today...

ianballou:
Great to see this merge, we'll test it out again as soon as it makes it into a gem.

PR #3349, opened 2024-03-19
[CI] - Fix Ruby head failures

MSP-Greg:
### Description

Ruby head has been failing in CI, and also noticed more failures locally with Ubuntu & Windows.

Two issues:

1. Failures in:
https://github.com/puma/puma/blob/87fd289503c16bf794c7a1a5c6e622bca2b4a071/test/test_worker_gem_independence.rb#L16-L21

   With recent changes in Ruby head, nio4r previous to 2.6.0 won't compile, fixed in https://github.com/socketry/nio4r/commit/3c1f9b83250df696f1c7d055a9a4d7bd7fee126e.  Solution was updating the nio4r versions.

2. Failures in:
https://github.com/puma/puma/blob/87fd289503c16bf794c7a1a5c6e622bca2b4a071/test/test_log_writer.rb#L151-L181

   Fixed by only using LogWriter to to write to itself...

Closes #3342

### Your checklist for this pull request
- [x] I have reviewed the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.
- [x] I have added (or updated) appropriate tests if this PR fixes a bug or adds a feature.
- [x] My pull request is 100 lines added/removed or less so that it can be easily reviewed.
- [ ] If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.
- [ ] If this closes any issues, I have added "Closes `#issue`" to the PR description or my commit messages.
- [ ] I have updated the documentation accordingly.
- [x] All new and existing tests passed, including Rubocop.

Issue #3266, opened 2023-10-25
Puma bundle error with v6.4


Dragonicity:
Installing puma 6.4.0 with native extensions
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

current directory: /Users/xxx.gem/ruby/3.2.2/gems/puma-6.4.0/ext/puma_http11
/Users/marknelson/.rubies/ruby-3.2.2/bin/ruby extconf.rb
using OpenSSL pkgconfig (openssl.pc)
checking for openssl/bio.h... yes
checking for DTLS_method() in openssl/ssl.h... yes
checking for SSL_CTX_set_session_cache_mode(NULL, 0) in openssl/ssl.h... yes
checking for TLS_server_method() in openssl/ssl.h... yes
checking for SSL_CTX_set_min_proto_version(NULL, 0) in openssl/ssl.h... yes
checking for X509_STORE_up_ref()... yes
checking for SSL_CTX_set_ecdh_auto(NULL, 0) in openssl/ssl.h... yes
checking for SSL_CTX_set_dh_auto(NULL, 0) in openssl/ssl.h... yes
checking for SSL_get1_peer_certificate() in openssl/ssl.h... no
checking for Random.bytes... yes
creating Makefile

current directory: /Users/xxx/.gem/ruby/3.2.2/gems/puma-6.4.0/ext/puma_http11
make DESTDIR= sitearchdir=./.gem.20231025-9688-7uig3e sitelibdir=./.gem.20231025-9688-7uig3e clean

current directory: /Users/xxx/.gem/ruby/3.2.2/gems/puma-6.4.0/ext/puma_http11
make DESTDIR= sitearchdir=./.gem.20231025-9688-7uig3e sitelibdir=./.gem.20231025-9688-7uig3e
compiling http11_parser.c
compiling mini_ssl.c
compiling puma_http11.c
linking shared-object puma/puma_http11.bundle
ld: warning: -multiply_defined is obsolete
ld: Undefined symbols:
_SSL_get1_peer_certificate, referenced from:
_engine_peercert in mini_ssl.o
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [puma_http11.bundle] Error 1

make failed, exit code 2

My puma.rb file:

```
# This configuration file will be evaluated by Puma. The top-level methods that
# are invoked here are part of Puma's configuration DSL. For more information
# about methods provided by the DSL, see https://puma.io/puma/Puma/DSL.html.

[...]

Issue #2790, opened 2022-01-07
OpenSSL 3 - symbol not found in flat namespace '_SSL_get1_peer_certificate'


jakemumu:
**Describe the bug**

I've been attempting to spin up a new rails 6 or 7 project, but haven't been able to do so due to a consistent issue with puma and open SSL

`puma_http11.bundle, 0x0009): symbol not found in flat namespace '_SSL_get1_peer_certificate'`

**Puma config:**

NA

**To Reproduce**

Mac 12.0.1 Intel, OpenSSL 3 installed via brew + Puma. 

rails new project

--

Anyone able to give some tips here?

MSP-Greg:
@jakemumu I tested against OpenSSL 3.0.0 when it was released.  The test suite passed, can't recall what else I checked.

I've got 3.0.1 installed/built; I haven't built Ruby or Puma with it yet.  I'll try to do so over the weekend...

nickmc:
Just had the same problem with an existing Rails 6.1 app. Upgraded Ruby from 2.7.1 to 2.7.5 and now receiving this error. 

Using Puma 5.5.2 and OpenSSL 3.0.1 (brew). On a Mac Mini M1.

Very grateful for any suggestions!

MSP-Greg:
I just tested using Ruby master and Puma master built with OpenSSL 3.0.1.  It passes all of Puma's SSL tests.  

@nickmc I don't believe Ruby 2.7.5 is compatible with OpenSSL 3.0.1.  It may build, but I doubt Ruby OpenSSL tests would pass...

nickmc:
Aah OK, understood, many thanks.

MSP-Greg:
Re OpenSSL 3, I haven't tested what works and doesn't work if there are OpenSSL version mismatches between build/run versions with Puma and/or Ruby.  Regardless, to check, use the following code:
```ruby
require 'puma'
require 'puma/minissl'
require 'openssl'

p_ssl = Puma::MiniSSL
wid = 32

pre = "\n#{RUBY_DESCRIPTION}\nRUBYOPT: #{ENV['RUBYOPT']}\n\n"

if Puma::IS_JRUBY || !Puma::HAS_SSL
  STDOUT.syswrite pre +
    "                         OpenSSL\n" \
    "OPENSSL_LIBRARY_VERSION: #{OpenSSL::OPENSSL_LIBRARY_VERSION}\n" \
[...]

PR #3350, opened 2024-03-20
Remove workaround for truffleruby, 24.0 is out and has the fix

eregon:
* See https://github.com/oracle/truffleruby/issues/3303#issuecomment-1836083623

cc @MSP-Greg

### DIFF
--- a/ext/puma_http11/extconf.rb
+++ b/ext/puma_http11/extconf.rb
@@ -10,10 +10,8 @@ end
 
 unless ENV["PUMA_DISABLE_SSL"]
   # don't use pkg_config('openssl') if '--with-openssl-dir' is used
-  # also looks within the Ruby build for directory info
   has_openssl_dir = dir_config('openssl').any? ||
-    RbConfig::CONFIG['configure_args']&.include?('openssl') ||
-    Dir.exist?("#{RbConfig::TOPDIR}/src/main/c/openssl") # TruffleRuby
+    RbConfig::CONFIG['configure_args']&.include?('openssl')
 
   found_pkg_config = !has_openssl_dir && pkg_config('openssl')
 


ext/puma_http11/extconf.rb
```diff
@@ -10,10 +10,7 @@
 
 unless ENV["PUMA_DISABLE_SSL"]
   # don't use pkg_config('openssl') if '--with-openssl-dir' is used
-  # also looks within the Ruby build for directory info
-  has_openssl_dir = dir_config('openssl').any? ||
-    RbConfig::CONFIG['configure_args']&.include?('openssl') ||
```
eregon, line 15: My understanding is this line is not needed because `dir_config` should use `configure_args` already (although the code of `dir_config` and mkmf in general is not easy to read).
Also mentioned in https://github.com/ruby/openssl/issues/650#issuecomment-1645680989

MSP-Greg:
Thanks for the PR.  This is difficult.

If you look at https://github.com/puma/puma/commits/master/ext/puma_http11/extconf.rb, the code in the most recent commit (which was after the TruffleRuby work-around, and removed it) isn't really about 'very recent' Ruby builds, it's about 'older' TruffleRuby & Ruby builds.  The associated PR is https://github.com/puma/puma/pull/3271.

Also common is a 'Ruby' built using all standard OS packages except OpenSSL, which has been updated to OpenSSL 3.

Or, removing it may cause issues as noted in https://github.com/puma/puma/issues/3266 or https://github.com/puma/puma/issues/2790.

Thoughts?

eregon:
Which Ruby versions are you concerned about?

TruffleRuby releases before 24.0 like 23.1 could hit the issue, true.
So maybe we should postpone doing this until say TruffleRuby 24.1?
At that point I don't think it's reasonable to expect that 23.1 works with latest gems anymore.

I can leave the `configure_args` line in if it helps for CRuby, although I think it doesn't do anything (see comment above).

MSP-Greg:
> Which Ruby versions are you concerned about?

As you might guess, my default Ruby on Windows & Ubuntu is Ruby head.  Also, the platform I work the least with is macOS.

In the past, both here (and in EventMachine) there have been many people having problems compiling on macOS.  I recall the issue was that Ruby was built with an option defining an OpenSSL directory, but Puma's extconf ignored it.

Since the most recent commit mentioned above, I don't recall any posts about macOS issues.

So, although this PR essentially passes (see #3349), I'm leery of the changes causing issues like the previous ones.  I need to look at it some more.  I haven't checked anything locally...

What about just removing line 16?

eregon:
> What about just removing line 16?

Done

Issue #932, opened 2016-03-20
handling SIGTERMs when dealing with long requests 


everplays:
Hi there,

##### TL;DR:

How can we receive SIGTERMs in the app itself and still let puma gracefully exit after request was processed?

##### Long version:

We are using Amazon SQS as backend of our job queue. Essentially, it runs the app just as it would run a web app instance but instead of directing normal traffic, it makes requests that are background jobs. Everything is working fine until we deploy or elastic beanstalk tries to scale. In either of the scenarios, they issue a SIGTERM to let the app finish what it is doing before SIGKILL but it's a long running request, it's meant to run for a few minutes! Therefore, puma doesn't get a chance to do a graceful stop as the execution is still in the app. What I am looking for is a way to communicate to Puma from within the app. Either of the following can work for us:

‌1. App asks puma if it has received a SIGTERM and if that was the case, stops what it's doing:

```ruby
def I_am_an_action_that_takes_5_minutes_to_run
  (1..300).each do
    sleep 1
    if env['puma.control'].status == :stop
      put_the_job_back_to_queue!
      break
    end
  end
end
```

‌2. App traps the signals (which prevents Puma from receiving them), handles the situation, and tells puma to stop:

```ruby
def I_am_an_action_that_takes_5_minutes_to_run
  (1..300).each do
    sleep 1
    SignalHandler.check!
  end
rescue SignalException => e
  put_the_job_back_to_queue!
end

# and SignalHandler.check! does something like the following in case of SIGTERM
if received_sigterm?
  # this is essential, otherwise, puma keeps running after request.
  env['puma.control'].set_status(:stop)
  raise SignalException, 15
end
```

Please let me know what you think and if you have a better suggestion. I would like to make a PR to address this.

evanphx:
I've added a new feature called `force_shutdown_after` that allows you to tell Puma how long to wait for threads to finish their work before being forced to stop. Would that solve your problem?

everplays:
Thanks for looking into this @evanphx. `force_shutdown_after` could make the situation better but will not solve it entirely.

**What it can improve:**
We split the background jobs and try to keep them under the time that beanstalk waits for the server. For the most of the jobs, it is possible and `force_shutdown_after` solves the problem.

**The problem which remains:**
[...]

PR #3360, opened 2024-04-03
Use actual thread local for `Puma::Server.current`.

ioquatix:
The current implementation of `Puma::Server.current` will break inside fibers, e.g. lazy enumerators because it's fiber local, not thread local. If the intention is to have an actual thread local, this is the best fix.

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -18,6 +18,9 @@ require 'io/wait' unless Puma::HAS_NATIVE_IO_WAIT
 
 module Puma
 
+  # This method was private on Ruby 2.4 but became public on Ruby 2.5+:
+  Thread.send(:attr_accessor, :puma_server)
+
   # The HTTP Server itself. Serves out a single Rack app.
   #
   # This class is used by the `Puma::Single` and `Puma::Cluster` classes
@@ -47,7 +50,6 @@ module Puma
     attr_accessor :app
     attr_accessor :binder
 
-    THREAD_LOCAL_KEY = :puma_server
 
     # Create a server for the rack app +app+.
     #
@@ -131,7 +133,7 @@ module Puma
     class << self
       # @!attribute [r] current
       def current
-        Thread.current[THREAD_LOCAL_KEY]
+        Thread.current.puma_server
       end
 
       # :nodoc:
@@ -438,7 +440,7 @@ module Puma
     # Return true if one or more requests were processed.
     def process_client(client)
       # Advertise this server into the thread
-      Thread.current[THREAD_LOCAL_KEY] = self
+      Thread.current.puma_server = self
 
       clean_thread_locals = options[:clean_thread_locals]
       close_socket = true
--- a/test/test_puma_server_current.rb
+++ b/test/test_puma_server_current.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+# Released under the MIT License.
+# Copyright, 2023, by Samuel Williams.
+
+require_relative "helper"
+
+require "puma/server"
+
+class PumaServerCurrentApplication
+  def call(env)
+    [200, {"Content-Type" => "text/plain"}, [Puma::Server.current.to_s]]
+  end
+end
+
+class PumaServerCurrentTest < Minitest::Test
+  parallelize_me!
+
+  def setup
+    @tester = PumaServerCurrentApplication.new
+    @server = Puma::Server.new @tester, nil, {log_writer: Puma::LogWriter.strings, clean_thread_locals: true}
+    @port = (@server.add_tcp_listener "127.0.0.1", 0).addr[1]
+    @tcp = "http://127.0.0.1:#{@port}"
+    @url = URI.parse(@tcp)
+    @server.run
+  end
+
+  def teardown
+    @server.stop(true)
+  end
+
+  def test_clean_thread_locals
+    server_string = @server.to_s
+    responses = []
+
+    # This must be a persistent connection to hit the `clean_thread_locals` code path.
+    Net::HTTP.new(@url.host, @url.port).start do |connection|
+      3.times do
+        responses << connection.get("/").body
+      end
+    end
+
+    assert_equal [server_string]*3, responses
+  end
+end


ioquatix:
It's probably worth noting that the current implementation conflicts with `clean_thread_locals`. In other words, it appears to me that enabling `clean_thread_locals` will break `Puma::Server.current` after the first request is handled.

ioquatix:
@nateberkopec can we merge this?

dentarg:
Should we test this?

Can you show example code how/when this breaks? Just to educate us all further

ioquatix:
Okay, included failing test + fix.

ioquatix:
cc @nateberkopec @dentarg failing test included - fix included - ready for review :)

MSP-Greg (approved):

MSP-Greg:
Isn't `Thread.send(:attr_accessor, :puma_server)` kind of global?  Not sure if that might surprise some people, or affect code (outside of Puma) that doesn't expect it to be there?

ioquatix:
> Isn't Thread.send(:attr_accessor, :puma_server) kind of global? Not sure if that might surprise some people, or affect code (outside of Puma) that doesn't expect it to be there?

You are correct. I don't think it's particularly surprising, and I'm not sure there is a better approach that keeps things simple and efficient. No matter what, thread locals in Ruby have "global visibility". The fact that the name is prefixed `puma_` ties it to the puma namespace, and I think that's safe enough, IMHO.

MSP-Greg:
> I'm not sure there is a better approach that keeps things simple

Agreed.  Seeing the Rails example and who committed the code...

ioquatix:
(@MSP-Greg as an aside, there was some discussion about introducing `Thread::Local.new` which is an unnamed object that stores thread local state, but it's extremely hard to implement correctly and probably needs to be done in the interpreter. See <https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent-ruby/concurrent/atomic/locals.rb> and the related TheadLocalVar / FiberLocalVar for details).

lib/puma/server.rb
```diff
@@ -131,7 +133,7 @@ def inherit_binder(bind)
     class << self
       # @!attribute [r] current
       def current
-        Thread.current[THREAD_LOCAL_KEY]
+        Thread.current.puma_server
```
jpcamara, line 136: I'm curious - what's the reason it can't be `Thread.current.thread_variable_get` and `Thread.current.thread_variable_set` instead? They operate exclusively at the thread level "and do not respect fibers" https://ruby-doc.org/core-2.6.3/Thread.html#method-i-thread_variable_get

ioquatix: Both approaches support all currently tested versions of Ruby.

As discussed above, `Thread.current.puma_server` is more efficient.

ioquatix:
Folks, what's the process for getting this reviewed / merged?

MSP-Greg:
@ioquatix

I did approve this PR on 11-Apr...

Question, there's a method `Puma.set_thread_name`, and I believe almost all threads created by Puma are named.  What if it was changed to the following, and we remove the code `Thread.send(:attr_accessor, :puma_server)` from `puma/server.rb`?
```ruby
def self.set_thread_name(name)
  Thread.current.name = "puma #{name}"

  # This method was private on Ruby 2.4 but became public on Ruby 2.5+:
  Thread.current.singleton_class.send(:attr_accessor, :puma_server)
end
```

There seems to be concern about globally adding an attrribute to `Thread`, this would only do so with Puma threads.

Not sure if doing it this adds overhead (possibly creating a singleton class?).  This passes the new test (locally on WSL2/Ubuntu 22.04).

ioquatix:
Thanks @MSP-Greg

I appreciate your feedback.

I'm fine with your proposed change. Modifying singleton classes - however it means that if someone calls `Puma::Server.current` on a non-Puma thread, it will fail with `NoMethodError`. So we'd also need to add `respond_to?` or something similar, or be comfortable with `NoMethodError` on non-Puma threads (instead of nil).

Regarding approval / merge - I'm not able to merge it even with your approval.

> Only those with [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can merge pull requests.

So I trust that we need someone else to actually decide where the bar for merging is. That is the part that's not clear to me. I think ideally someone is either (1) okay with this PR and merges it or (2) requests changes - so it's clear what needs to happen to move things forward.

MSP-Greg:
> however it means that if someone calls `Puma::Server.current` on a non-Puma thread, it will fail with `NoMethodError`.

True.  But, it's set in the call to `Puma::Server#process_client`, so the non-Puma thread won't get the value.

Regardless, maybe setting it as a global attribute is best, since it won't throw an error if called from a non-Puma thread...

BTW, sorry about the mixup.  Before I posted, you pinged @nateberkopec and @dentarg, so I wasn't sure if they'd look at it.  Or, yes, I can merge it...

ioquatix:
> BTW, sorry about the mixup. Before I posted, you pinged @nateberkopec and @dentarg, so I wasn't sure if they'd look at it. Or, yes, I can merge it...

I don't know the normal process but it would be nice to have confidence that there was consensus on (1) fixing the bug and (2) the proposed fix.

In other words, I'd like to see this get merged, but I'd also like to follow whatever process is suitable for the project.

dentarg:
As far as I know, there is no formal process :) Just people looking when they have time for it.

My hunch is that it is good to avoid NoMethodError, but I haven't spent much time thinking about this PR. I also pretty much trust you both @ioquatix @MSP-Greg so if you are in agreement, it is very much fine if Greg merges this.

MSP-Greg:
@ioquatix @dentarg

Merged.

I was in the process of reviewing this, looking for the original reason for the method.  The commit (July-2016) and issue were:
https://github.com/puma/puma/commit/8370d2eed0242af6a1c84e0f0c42fe9d66cd0839
https://github.com/puma/puma/issues/932

I tried searching GitHub for `Puma.current`, and didn't seem to find anything.  Using a thread local for it is something we may want to change, and with the expanded number of hooks available, it may be possible to remove it in Puma v7.  Also, we may have servers for both 'listen' and control, I've haven't looked at issues like that, and we have servers in processes/workers...

Thanks, and sorry for the delay...

ioquatix:
I'm 👍🏼 on dropping this interface in v7 especially if there are no obvious use cases. Maybe we can emit a warning in a v6.9... release.

PR #3375, opened 2024-05-02
Add a debug log before running each type of hook

joshuay03:
### Description

Having enabled debug logs at work I thought it would be useful if there was a generic debug log before all the hooks for a type are run.

### Your checklist for this pull request

- [x] I have reviewed the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.
- [x] I have added (or updated) appropriate tests if this PR fixes a bug or adds a feature.
- [x] My pull request is 100 lines added/removed or less so that it can be easily reviewed.
- [ ] If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.
- [ ] If this closes any issues, I have added "Closes `#issue`" to the PR description or my commit messages.
- [ ] I have updated the documentation accordingly.
- [x] All new and existing tests passed, including Rubocop.

PR #3376, opened 2024-05-02
Update DSL hook processing logic to be consistent

joshuay03:
### Description

Noticed some inconsistencies with how the DSL processes hooks while working on #3375, this was also opened as a split up prerequisite for https://github.com/puma/puma/pull/3377.

Ensures all hook DSL methods utilise `#process_hooks` + some minor changes (see comments).

### DIFF
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -438,8 +438,7 @@ module Puma
     #   end
     #
     def on_restart(&block)
-      @options[:on_restart] ||= []
-      @options[:on_restart] << block
+      process_hook :on_restart, nil, block, 'on_restart'
     end
 
     # Command to use to restart Puma. This should be just how to
@@ -731,8 +730,7 @@ module Puma
     def before_fork(&block)
       warn_if_in_single_mode('before_fork')
 
-      @options[:before_fork] ||= []
-      @options[:before_fork] << block
+      process_hook :before_fork, nil, block, 'before_fork'
     end
 
     # Code to run in a worker when it boots to setup
@@ -855,6 +853,8 @@ module Puma
     # @version 5.0.0
     #
     def on_refork(key = nil, &block)
+      warn_if_in_single_mode('on_refork')
+
       process_hook :before_refork, key, block, 'on_refork'
     end
 
@@ -876,8 +876,7 @@ module Puma
     #   end
     #
     def on_thread_start(&block)
-      @options[:before_thread_start] ||= []
-      @options[:before_thread_start] << block
+      process_hook :before_thread_start, nil, block, 'on_thread_start'
     end
 
     # Provide a block to be executed after a thread is trimmed from the thread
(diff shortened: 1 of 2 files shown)


lib/puma/dsl.rb
```diff
@@ -731,6 +729,8 @@ def on_booted(&block)
     # @version 5.0.0
     #
     def on_refork(key = nil, &block)
+      warn_if_in_single_mode('on_refork')
```
joshuay03, line 732: I think this makes sense here? This seemed to be the only cluster mode related hook that was missing the warning.

lib/puma/dsl.rb
```diff
@@ -1204,7 +1202,7 @@ def warn_if_in_single_mode(hook_name)
         log_string =
           "Warning: You specified code to run in a `#{hook_name}` block, " \
           "but Puma is not configured to run in cluster mode (worker count > 0 ), " \
-          "so your `#{hook_name}` block did not run"
+          "so your `#{hook_name}` block will not run."
```
joshuay03, line 1205: This warning is displayed during configuration, so it makes more sense to tell the user that it will not run, as the hooks haven't actually been processed yet.

test/test_config.rb
```diff
@@ -503,6 +503,12 @@ def test_run_hooks_before_fork
     assert_warning_for_hooks_defined_in_single_mode :before_fork
   end
 
+  def test_run_hooks_before_refork
+    assert_run_hooks :before_refork, configured_with: :on_refork
+
+    assert_warning_for_hooks_defined_in_single_mode :on_refork
+  end
```
joshuay03, line 510: All other hooks are covered here, this was the only one missing coverage.

test/test_config.rb
joshuay03, line 1: Added coverage for hooks missing in this suite.

joshuay03:
Note to self, related PR: https://github.com/puma/puma/pull/3380.

joshuay03:
> Note to self, related PR: #3380.

Superseded by https://github.com/puma/puma/pull/3411

test/test_config.rb
joshuay03, line 1: Added coverage for hooks missing in this suite.

lib/puma/dsl.rb
nateberkopec: We should probably warn for the `fork_worker` config option too.

nateberkopec (changes requested):

lib/puma/dsl.rb
joshuay03: Looking at the log, I think it was only intended to be called in hook methods: https://github.com/puma/puma/pull/3376/files#diff-122c03b9c96f060a9cc29b5ed3446467e1909b0054b4f5cfd95ee41cfff832e7R1408-R1421

Also, there are other non-hook cluster-only methods that we don't warn on (`preload_app!`, `worker_timeout`, `worker_boot_timeout` etc.). I'm happy to open a separate PR to tackle those (including `fork_worker`) with a separate log, or maybe even consolidate them.

nateberkopec: > I'm happy to open a separate PR to tackle those (including fork_worker) with a separate log, or maybe even consolidate them.

Sure, we can consider this a separate issue.

PR #3378, opened 2024-05-03
Proposal to bump persistent_timeout to 65 seconds

shayonj:
### Description

I noticed that if Puma's `persistent_timeout` is shorter than the load balancer's timeout, occasional minor service disruptions can occur. For instance, consider the AWS Application Load Balancer (ALB) with a default idle timeout of 60 seconds. If ALB forwards a request to Puma just as Puma is closing the socket, it may result in a broken TCP handshake, leading to a TCP RST/FIN being sent back to ALB. ALB would then handle that RST/FIN and return a 502 error to the client, which is not ideal.

Since most load balancers, like AWS ALB/ELB and Heroku, have default idle timeouts around [60 seconds](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html) and [55 seconds](https://devcenter.heroku.com/articles/limits#:~:text=HTTP%20timeouts&text=After%20the%20initial%20response%2C%20each,an%20H15%20error%20is%20logged.) respectively, I am proposing that we increase Puma's `persistent_timeout` to 65 seconds so that the upstream negotiates and closes the connection before Puma does. 

### DIFF
--- a/lib/puma/configuration.rb
+++ b/lib/puma/configuration.rb
@@ -151,7 +151,7 @@ module Puma
       mutate_stdout_and_stderr_to_sync_on_write: true,
       out_of_band: [],
       # Number of seconds for another request within a persistent session.
-      persistent_timeout: 20,
+      persistent_timeout: 65,
       queue_requests: true,
       rackup: 'config.ru'.freeze,
       raise_exception_on_sigterm: true,


schneems:
This is relevant to my interests. I just updated the Heroku getting started app https://github.com/heroku/ruby-getting-started/pull/190/files (new router docs are here https://devcenter.heroku.com/articles/http-routing#timeouts). @MSP-Greg You've recently played with some of these values, do you have an opinion on which magic numbers we should use?

I would like it to handle Heroku out of the box, but the value also might change. For the Heroku use case, if we had Puma default to a `PUMA_PERSISTENT_TIMEOUT` env var (if it exists) then I could set this in the buildpack. 

Request: Change this to make the default configurable via env var in addition to incrementing the value.

shayonj:
ENV var sounds great. Happy to put in a commit

Aesthetikx:
@schneems I am hoping this is the cause of the Heroku H13/H18s that we see after dyno restarts (including platform initiated ones), however in the documentation it does say that it is actually 90s in router v2, so I don't think we necessarily need to increase the Puma default to say 95s, but for Puma deployments on Heroku would that now be the recommended value?

schneems:
This commit on the getting started guide has more info https://github.com/heroku/ruby-getting-started/commit/534e65ded00c878bcd1c4390a6ae0f02aefc3ca7

> Router 2.0 introduced keepalive support. It will close connections idle after 90 seconds https://devcenter.heroku.com/articles/http-routing#keepalives. We want to avoid a situation where they send a request right before puma closes the connection. We can do that by setting it to the same timeout the router uses + some value (such as 5 seconds).

rafal-brize:
Hey folks.

Once i set PUMA_PERSISTENT_TIMEOUT env variable.
On every request this is being logged:

```
Read: #<TypeError: String can't be coerced into Float>
```

It didn't dig into it (its not raising this exception), but my gut is telling me that

https://github.com/puma/puma/blob/master/lib/puma/client.rb#L146-L154

This area might somehow receive a String `val` (envs values are strings)

@shayonj @schneems wdyt?

schneems:
The value should be slightly higher than the router timeout to ensure it’s
not a race. 91 should be fine but clock skew and being on the safe side I’m
defaulting it to 95.

You’re right there’s a bug here:

https://github.com/puma/puma/pull/3378/files#diff-2dc4e3e83be7fd97cebc482ae07d6a8216944003de82458783fb00b5ae9524c8R155

It needs a to_i. I miss types sometimes.

Richard Schneeman
https://howtoopensource.dev
https://www.schneems.com
he/him

shayonj:
Same! Happy to put in a fix tomorrow

schneems:
I've got a PR https://github.com/puma/puma/pull/3749. In the future if you can open a new issue, it will get more visibility than  comment on a closed PR. Thanks for reporting. I'm probably going to cut another release for this.

schneems:
7.0.3 is out with this fix

PR #3381, opened 2024-05-12
Move event to default configuration

### LINKED ISSUE #3356: "undefined method `on_booted` for nil:NilClass" on "pumactl start"
**Describe the bug**
Having `on_booted` event in puma configuration file fails to start puma server using `pumactl -F config/puma.rb start`. 

It raises error here since `@config.options[:events]` is `nil` https://github.com/puma/puma/blob/5fc43d73b6ff193325e657a24ed76dec79133e93/lib/puma/dsl.rb#L706-L708

while trying to load config here: https://github.com/puma/puma/blob/5fc43d73b6ff193325e657a24ed76dec79133e93/lib/puma/control_cli.rb#L130-L131

**Puma config:**

```ruby
on_booted do
  puts "on_booted called"
end
```

**To Reproduce**

Copy this into a file called `hello.ru`:
```ruby
```

Create config file called `puma.rb` with the following content:
```ruby
on_booted do
  puts "on_booted called"
end
```

Run it with:
```shell
bundle exec pumactl -F puma.rb start hello.ru
```

Observe error:
```
undefined method `on_booted' for nil:NilClass
```

The same can be reproduced with test config in the repository:
```
bundle exec pumactl -F test/config/event_on_booted.rb start test/rackup/hello.ru
```

**Expected behavior**
Puma server should start

**Desktop:**
 - OS: macOS 14.4.1
 - Puma Version 6.4.2
 - ruby 3.2.2

nikhilbhatt:
### Description
Fixes #3356 

### DIFF
--- a/lib/puma/cli.rb
+++ b/lib/puma/cli.rb
@@ -24,11 +24,10 @@ module Puma
     # Create a new CLI object using +argv+ as the command line
     # arguments.
     #
-    def initialize(argv, log_writer = LogWriter.stdio, events = Events.new)
+    def initialize(argv, log_writer = LogWriter.stdio)
       @debug = false
       @argv = argv.dup
       @log_writer = log_writer
-      @events = events
 
       @conf = nil
 
@@ -63,7 +62,7 @@ module Puma
         end
       end
 
-      @launcher = Puma::Launcher.new(@conf, :log_writer => @log_writer, :events => @events, :argv => argv)
+      @launcher = Puma::Launcher.new(@conf, :log_writer => @log_writer, :events => @conf.options[:events], :argv => argv)
     end
 
     attr_reader :launcher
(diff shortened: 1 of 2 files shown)


lib/puma/cli.rb
```diff
@@ -24,11 +24,10 @@ class CLI
     # Create a new CLI object using +argv+ as the command line
     # arguments.
     #
-    def initialize(argv, log_writer = LogWriter.stdio, events = Events.new)
+    def initialize(argv, log_writer = LogWriter.stdio)
```
dentarg, line 27: Not sure we can change the signature of `Puma::CLI`, it is probably considered public API?

dentarg: Yeah, https://github.com/puma/puma/pull/2798 changed the signature and it caused breakage

nikhilbhatt: Ok, I am not sure on this if we can change the public API Here. How you want me to take this forward?

The other around is just introduce a `Puma:Events` in `control_cli.rb` and that should also work. instead of introducing it at default config level.

schneems:

```
```

Also took a peak a the PR. A lot has changed in the config since you made this PR. Does the referenced issue still reproduce with 7.1.0? If so, this will need a rebase.

nateberkopec:
Superseded by #3872

PR #3392, opened 2024-05-22
Address rubocop TODOs

joshuay03:
### Description

Ref: https://github.com/puma/puma/pull/1325

Addresses most of the cops in `.rubocop_todo.yml` using safe autocorrect (`-a`).

Most of those cops were set up to use the default config except for:

- `Layout/SpaceBeforeBlockBraces`
- `Layout/SpaceAroundEqualsInParameterDefault`
- `Layout/SpaceInsideHashLiteralBraces`

which all enforced `no_space`. I've changed these to use the default (`space`) for both consistency with other defaults (e.g. `SpaceAroundOperators`) and (imo) better readability. Let me know if `no_space` is prefered and I can revert back.

The only one I removed from the original TODO config was `Layout/IndentationWidth` which yields multiple inconsistent results when flagging and autocorrecting, even when paired with `Layout/IndentationConsistency` which the rubocop docs recommend. For example, it does this with indented condition assignments:

```ruby
          indented_var = if foo
                           "foo"
          else
            "bar"
          end
```

Also, the only existing cop amended was `Layout/SpaceBeforeBlockBraces`. It now uses the default (`space`) for `EnforcedStyleForEmptyBraces`. My argument is once again consistency, and that the default makes it easier to tell an empty block apart from an empty hash arg when parenthesis are excluded. E.g.

```ruby
# arg
method_call {}

# empty block
method_call { }
```

This also now makes is consistent with the other block cops.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -8,6 +8,7 @@ AllCops:
   StyleGuideCopsOnly: false
   Exclude:
     - 'tmp/**/*'
+    - '.bundle/**/*'
     - '**/vendor/bundle/**/*'
     - 'examples/**/*'
     - 'pkg/**/*'
@@ -22,8 +23,11 @@ Performance:
 # ————————————————————————————————————————— disabled cops
 
 # ————————————————————————————————————————— enabled cops
+
+# Layout
+
 Layout/AccessModifierIndentation:
-  EnforcedStyle: indent
+  Enabled: true
 
 Layout/IndentationStyle:
   Enabled: true
@@ -35,7 +39,6 @@ Layout/SpaceAroundKeyword:
   Enabled: true
 
 Layout/SpaceBeforeBlockBraces:
-  EnforcedStyleForEmptyBraces: no_space
   Enabled: true
 
 Layout/SpaceBeforeFirstArg:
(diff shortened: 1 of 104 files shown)


.rubocop.yml
```diff
@@ -35,7 +39,6 @@ Layout/SpaceAroundKeyword:
   Enabled: true
 
 Layout/SpaceBeforeBlockBraces:
-  EnforcedStyleForEmptyBraces: no_space
```
joshuay03, line 38: This is the default.

.rubocop.yml
```diff
@@ -50,24 +53,57 @@ Layout/TrailingEmptyLines:
...
   Max: 7
 
+# Naming
+
 Naming/ConstantName:
   Enabled: true
 
 Naming/MethodName:
   Enabled: true
-  EnforcedStyle: snake_case
```
joshuay03, line 64: This is already the default.

JuanitoFatas:
It would be great to add the reformatting commit to [.git-blame-ignore-revs](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view), so they get skipped in git blames 🥇

joshuay03:
> It would be great to add the reformatting commit to [.git-blame-ignore-revs](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view), so they get skipped in git blames 🥇

Good idea! I'll follow up with this once merged.

PR #3400, opened 2024-06-05
Depend on sqlite3 `~> 1.4`.

ioquatix:
Try to fix `turbo-rails` tests.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -28,6 +28,8 @@ gem "rackup" if use_rackup
 
 gem "jruby-openssl", :platform => "jruby"
 
+gem "sqlite3", "~> 1.4"
+
 unless ENV['PUMA_NO_RUBOCOP'] || RUBY_PLATFORM.include?('mswin')
   gem "rubocop"
   gem 'rubocop-performance', require: false


Gemfile
```diff
@@ -28,6 +28,8 @@ gem "rackup" if use_rackup
 
 gem "jruby-openssl", :platform => "jruby"
 
+gem "sqlite3", "~> 1.4"
```
dentarg, line 31: Add a comment on why we have this? Or make it conditional/opt-in? (ENV that the workflow can set) 

Very few run the turbo rails test when they clone puma repo and run tests, and as far as I know, you don't need SQLite then.

MSP-Greg:
I've got this passing, give me a minute...

Gemfile
ioquatix: CI currently appears to be broken, is it not important to fix that?

ioquatix:
Sure, feel free to add commits to this branch.

ioquatix:
Closing in favour of #3401

MSP-Greg:
> Sure, feel free to add commits to this branch.

Sorry, busy morning, working on some other things.  This did seem to demand a quick fix, so...

PR #3403, opened 2024-06-05
CI: disable `concurrent_skipping` for skip-duplicate-actions

dentarg:
From https://github.com/fkirc/skip-duplicate-actions?tab=readme-ov-file#concurrent_skipping

> Skip a workflow run if the same workflow is already running

but it does not seem reliable, for example, this workflow run https://github.com/puma/puma/actions/runs/9386209788 looks 100% green but that is because everything was skipped

https://github.com/fkirc/skip-duplicate-actions/issues has many reports about problems with `concurrent_skipping`

https://github.com/celerity/celerity-runtime/pull/148 is an example of another project disabling it.

[ci skip]

### DIFF
--- a/.github/workflows/skip_duplicate_workflow_runs.yml
+++ b/.github/workflows/skip_duplicate_workflow_runs.yml
@@ -28,5 +28,5 @@ jobs:
         with:
           paths_ignore: '["**.md"]'
           paths: ${{ inputs.paths }}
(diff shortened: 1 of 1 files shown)


MSP-Greg (approved):

MSP-Greg:
Sorry.  I guess a good way to avoid the issue is to not merge a PR until the PR's CI is finished.

Which I recently did not do...

PR #3405, opened 2024-06-05
Skip gc test refutation on JRuby.

ioquatix:
This test is failing CI.


ioquatix:
@MSP-Greg @dentarg this should fix 2/3 failures on master CI.

dentarg (approved):

PR #2817, opened 2022-02-02
Do not remove pre-existing socket after USR2+TERM (#2816)

khustochka:
Closes #2816 
Please review. This change fixes the issue but there are some concerns.

### Description
So, the issue was that when reloading the server, the socket was inherited, and it was added to the list of unix sockets, to be removed later (unlike on startup, when pre-existing sockets are excluded from that list).

The problem is that, as it seems to me, for inherited socket there is no way to tell if it was pre-existing or not. Hence the inadvertent result of my fix is that even not pre-existing socket will be preserved. Now, this may be acceptable. As you can see in the table in #2816, in single mode socket file is always preserved.

Another option: I may be mistaken, but from the subsequent code it seems possible to detect if the socket is activated, so it may be possible to prevent activated sockets from being removed (but as I understand, activated does not mean the same as pre-existing). (Sorry, I do not know a lot about Unix sockets etc, just trying to figure out the logic.)

Thank you.

### Your checklist for this pull request

- [x] I have reviewed the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.
- [x] I have added (or updated) appropriate tests if this PR fixes a bug or adds a feature.
- [x] My pull request is 100 lines added/removed or less so that it can be easily reviewed.
- [x] If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.
- [x] If this closes any issues, I have added "Closes `#issue`" to the PR description or my commit messages.
- [x] I have updated the documentation accordingly.
- [x] All new and existing tests passed, including Rubocop.

PR #3407, opened 2024-06-05
Print RUBY_DESCRIPTION when Puma starts

JacobEvelyn:
This commit changes the runner to print `RUBY_DESCRIPTION` instead of a manually-generated string for the Ruby version. Other popular gems like Sidekiq similarly print `RUBY_DESCRIPTION`. One specific motivation for this change is that `RUBY_DESCRIPTION` indicates whether YJIT is enabled in MRI.

This commit also deprecates the `Puma::Runner#ruby_engine` method, which will be removed in v7.

### Description

My team was testing YJIT and was frustrated by the fact that Puma does not indicate whether YJIT is running, while other gems like Sidekiq do.

I welcome any suggestions for this PR.


dentarg:
Mind showing some before and after examples?

dentarg:
Ideas/alternatives:
Should Ruby has its own line instead?
Should we only add YJIT yes/no?

ioquatix:
`RUBY_DESCRIPTION` is the correct choice, and it should probably be on its own line. I recently adopted it in Falcon too.

lib/puma/runner.rb
```diff
@@ -90,26 +90,13 @@ def close_control_listeners
...
-    end
-
     def output_header(mode)
       min_t = @options[:min_threads]
       max_t = @options[:max_threads]
       environment = @options[:environment]
 
       log "Puma starting in #{mode} mode..."
-      log "* Puma version: #{Puma::Const::PUMA_VERSION} (#{ruby_engine}) (\"#{Puma::Const::CODE_NAME}\")"
+      log "* Puma version: #{Puma::Const::PUMA_VERSION} (\"#{Puma::Const::CODE_NAME}\") (#{RUBY_DESCRIPTION})"
```
dentarg, line 99:
```suggestion
      log "* Puma version: #{Puma::Const::PUMA_VERSION} (\"#{Puma::Const::CODE_NAME}\")"
      log "* Ruby version: #{RUBY_DESCRIPTION}"
```

lib/puma/runner.rb
```diff
@@ -90,26 +90,13 @@ def close_control_listeners
       @control.binder.close_listeners if @control
     end
 
-    # @!attribute [r] ruby_engine
-    def ruby_engine
```
dentarg, line 94: This method is public, so we should deprecate it, not remove it.

```ruby
$ ruby -rpuma -rpuma/runner -e 'p Puma::Runner.new(Puma::Launcher.new(Puma::Configuration.new)).ruby_engine'
"ruby 3.2.4-p170"
```

Unless this change should wait for Puma v7?

JacobEvelyn:
Let me know what I should do to move this PR forward. Thanks!

lib/puma/runner.rb
```diff
@@ -91,7 +91,10 @@ def close_control_listeners
     end
 
     # @!attribute [r] ruby_engine
+    # @deprecated Use `RUBY_DESCRIPTION` instead
     def ruby_engine
+      warn "Puma::Runner#ruby_engine is deprecated; use RUBY_DESCRIPTION instead. It will be removed in puma v7."
```
JacobEvelyn, line 96:
```
$ ruby -I lib -rpuma -rpuma/runner -e 'p Puma::Runner.new(Puma::Launcher.new(Puma::Configuration.new)).ruby_engine'
Puma::Runner#ruby_engine is deprecated; use RUBY_DESCRIPTION instead. It will be removed in puma v7.
"ruby 3.3.2-p78"
```

JacobEvelyn:
@dentarg let me know what I should do to move this forward. Thanks!

dentarg:
I think this makes sense. Is there any existing tests for this part of Puma that should be extended?

dentarg:
I recall some change to the Puma logging in the past and I remember @nateberkopec being conservative about it. It would be great if he signed off on this.

If I have time I can try to look for that past discussion.

JacobEvelyn:
> Is there any existing tests for this part of Puma that should be extended?

I didn't see any, but let me know if you think I should add one.

JacobEvelyn:
> I recall some change to the Puma logging in the past and I remember @nateberkopec being conservative about it. It would be great if he signed off on this.

@nateberkopec any thoughts on this?

JacobEvelyn:
This change feels fairly uncontroversial to me; what can we do to move it forward?

dentarg:
> I recall some change to the Puma logging in the past and I remember @nateberkopec being conservative about it. It would be great if he signed off on this.
> 
> If I have time I can try to look for that past discussion.

Maybe I misremembering, I can only find this comment https://github.com/puma/puma/pull/2817#issuecomment-1029967988 and it is probably call this logging 'behavior'.

We have changed the logging before in a minor release, https://github.com/puma/puma/commit/ee3341d67, so this should be fine. I'll merge it if you rebase @JacobEvelyn

dentarg:
> I'll merge it if you rebase @JacobEvelyn

I clicked the Update branch button so no need

nateberkopec:
Yeah thanks for checking w/me but in this case I think we're good.

PR #3408, opened 2024-06-06
Don't let the config file go out of scope while it's still in use.

ioquatix:
This is another experimental branch to try and fix race conditions in CI.

```
/home/runner/work/puma/puma/lib/puma/dsl.rb:135:in `read': No such file or directory @ rb_sysopen - /home/runner/work/_temp/config20240605-2615-l1fl0k.rb (Errno::ENOENT)
	from /home/runner/work/puma/puma/lib/puma/dsl.rb:135:in `_load_from'
	from /home/runner/work/puma/puma/lib/puma/configuration.rb:238:in `block in load'
	from /home/runner/work/puma/puma/lib/puma/configuration.rb:238:in `each'
	from /home/runner/work/puma/puma/lib/puma/configuration.rb:238:in `load'
	from /home/runner/work/puma/puma/lib/puma/launcher.rb:54:in `initialize'
	from /home/runner/work/puma/puma/lib/puma/cli.rb:66:in `new'
	from /home/runner/work/puma/puma/lib/puma/cli.rb:66:in `initialize'
	from /home/runner/work/puma/puma/bin/puma:8:in `new'
	from /home/runner/work/puma/puma/bin/puma:8:in `<main>'
```

### DIFF
--- a/test/helpers/integration.rb
+++ b/test/helpers/integration.rb
@@ -26,6 +26,7 @@ class TestIntegration < Minitest::Test
 
   def setup
     @server = nil
+    @config_file = nil
     @server_log = +''
     @pid = nil
     @ios_to_close = []
@@ -60,6 +61,11 @@ class TestIntegration < Minitest::Test
       rescue
       ensure
         @server = nil
+
+        if @config_file
+          @config_file.unlink
+          @config_file = nil
+        end
       end
     end
   end
@@ -81,6 +87,7 @@ class TestIntegration < Minitest::Test
       env: {})          # pass env setting to Puma process in IO.popen
 
     if config
+      # Don't let the config file be GC'd until the server is stopped:
       @config_file = Tempfile.new(%w(config .rb))
       @config_file.write config
       @config_file.close
--- a/test/helpers/integration.rb
+++ b/test/helpers/integration.rb
@@ -81,10 +81,10 @@ class TestIntegration < Minitest::Test
       env: {})          # pass env setting to Puma process in IO.popen
 
     if config
-      config_file = Tempfile.new(%w(config .rb))
-      config_file.write config
-      config_file.close
-      config = "-C #{config_file.path}"
+      @config_file = Tempfile.new(%w(config .rb))
+      @config_file.write config
+      @config_file.close
+      config = "-C #{@config_file.path}"
     end
 
     puma_path = File.expand_path '../../../bin/puma', __FILE__


ioquatix:
cc @dentarg @MSP-Greg I think we can merge this.

ioquatix:
Even after merging this, we saw:

```
TestIntegrationCluster#test_culling_strategy_oldest = 
────────────────────────────────────────────────────────────────────── Debugging Info ──
/home/runner/work/puma/puma/lib/puma/dsl.rb:135:in 'IO.read': No such file or directory @ rb_sysopen - /home/runner/work/_temp/config20240606-2646-5ci9dz.rb (Errno::ENOENT)
	from /home/runner/work/puma/puma/lib/puma/dsl.rb:135:in 'Puma::DSL#_load_from'
	from /home/runner/work/puma/puma/lib/puma/configuration.rb:238:in 'block in Puma::Configuration#load'
	from <internal:array>:54:in 'Array#each'
	from /home/runner/work/puma/puma/lib/puma/configuration.rb:238:in 'Puma::Configuration#load'
	from /home/runner/work/puma/puma/lib/puma/launcher.rb:54:in 'Puma::Launcher#initialize'
	from /home/runner/work/puma/puma/lib/puma/cli.rb:66:in 'Class#new'
	from /home/runner/work/puma/puma/lib/puma/cli.rb:66:in 'Puma::CLI#initialize'
	from /home/runner/work/puma/puma/bin/puma:8:in 'Class#new'
	from /home/runner/work/puma/puma/bin/puma:8:in '<main>'
```

MSP-Greg:
Sorry, been doing non OSS work and the Window Ruby thing...

> do you know why we don't see any kind of exception backtrace or error?

Not offhand.  Also, I can't recall whether the restart code might work if it's called fast enough such that the first 'restart' call isn't finished and another 'restart' triggered.

I'm a bit late to the party here, but doesn't `Tempfile.new` create finalizers, and the code is explicitly calling `unlink`.   Might it use `Tempfile.create`?  Any opinion?

Also, some old info that may not be true any more.  When GHA was first available, I thought the Windows & Ubuntu systems were identical.  They had a HDD and a SSD.  The OS image was on the HDD.  IIRC, the normal OS temp folders were on the HDD, and the RUNNER_TEMP folder was on the SSD.  But, it looks like the temp files are using the RUNNER_TEMP drive.  They may share them with the other VM's...

dentarg:
> I'm a bit late to the party here, but doesn't `Tempfile.new` create finalizers, and the code is explicitly calling `unlink`. Might it use `Tempfile.create`? Any opinion?

You are right of course. WDYT of https://github.com/puma/puma/pull/3410?

PR #3296, opened 2023-12-23
lib/client.rb - use Tempfile.create instead of Tempfile.new for request bodies

MSP-Greg:
### Description

Currently, `Tempfile.new` is used in `Client` when a request contains content (a body) that is either very large or indeterminate (chunked).  `Tempfile.new` creates a finalizer that removes the `Tempfile`, so its removal is dependent on GC.

Recently, Windows head builds started logging the following in CI, a log is [here](https://github.com/puma/puma/actions/runs/7301425484/job/19898046945#step:11:433).  Note that Ruby head (3.3) has had many changes re GC, and one might call it more agressive...
```
D:/a/puma/puma/lib/puma/thread_pool.rb:113: warning: Exception in finalizer #<Tempfile::Remover:0x000001939f62dc50 @pid=2472, @path="D:/a/_temp/puma20231222-2472-94nq5x">
D:/ruby-mswin/lib/ruby/3.3.0+0/tempfile.rb:319:in `unlink': Permission denied @ apply2files - D:/a/_temp/puma20231222-2472-94nq5x (Errno::EACCES)
```

I doubt many people are using Puma on Windows, but Windows CI may show things that indicate issues.  On Windows, one cannot delete a file if it is open, unlike '*nix' platforms.

Also, in general, I think we'd prefer that Puma does its own cleanup, and not depend on GC to do so, at least with system resources like IO (files, pipes, sockets, etc).

So, use `Tempfile.create` and remove the file when Puma is finished with it.

### Your checklist for this pull request
- [x] I have reviewed the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.
- [x] I have added (or updated) appropriate tests if this PR fixes a bug or adds a feature.
- [x] My pull request is 100 lines added/removed or less so that it can be easily reviewed.
- [ ] If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.
- [ ] If this closes any issues, I have added "Closes `#issue`" to the PR description or my commit messages.
- [ ] I have updated the documentation accordingly.
- [x] All new and existing tests passed, including Rubocop.

PR #3410, opened 2024-06-09
CI: Use `Tempfile.create` in `TestIntegration`

dentarg:
Hopefully more stable:
- https://github.com/puma/puma/pull/3408#issuecomment-2153329320

Similar change as https://github.com/puma/puma/pull/3296

### DIFF
--- a/test/helpers/integration.rb
+++ b/test/helpers/integration.rb
@@ -63,7 +63,7 @@ class TestIntegration < Minitest::Test
         @server = nil
 
         if @config_file
-          @config_file.unlink
+          File.unlink(@config_file.path) rescue nil
           @config_file = nil
         end
       end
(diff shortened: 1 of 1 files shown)


ioquatix:
The reason why I assigned it to an instance variable was to prevent GC. So I'm not sure why this change is required, however, I'm also not sure my change fixed the issue, since we saw some other IO related issue... I need to spend more time investigating this too.

MSP-Greg (commented):
Maybe:
```ruby
if @config_file&.exist?
  @config_file.unlink
  @config_file = nil
end
```

MSP-Greg:
> The reason why I assigned it to an instance variable was to prevent GC.

Agree, that is an issue. A lot of these issues may be related to running test parallel, along with 'resource starved' CI runners.  In the test update (which I haven't looked at for a while), I think I removed all the `Tempfile.new` use, along with a lot of fixes for blocking IO, etc,  It was a lot more stable.  There were a few lib file changes, so I started on extracing them, but haven't finished...

Re `Tempfile.create`, since we know that `unlink` will always run, why create finalizers?

dentarg:
All tests here are green so let's merge this and see how it plays out going forward?

MSP-Greg (approved):

MSP-Greg:
@ioquatix I suspect if something stopped `teardown`, it's pretty severe, like a SEGV...

PR #3415, opened 2024-06-23
Add `prioritize_requests_by`

### LINKED ISSUE #3414: Prioritised request queue
#### Is your feature request related to a problem? Please describe.

Currently, an application I'm maintaining (Mastodon) uses Puma to dispatch work to Rack (and its controllers)

However, this application serves roughly 3 types of requests:
1. User API requests
2. Browser HTML requests
3. Server-to-server API-like requests

This latter request class is especially susceptible to sudden increases in request number and volume at certain times, but with base puma, and without any other setup, these would "clog" the request queue, and make the server unresponsive to users.

In order words: Request type 1 is most sensitive to latency and retries, and request type 3 is least sensitive, and will retry resiliently.

For this problem, we are currently using a combination of nginx (to tag requests with priority) and haproxy (to queue requests with that priority) to prioritise user requests over server requests, and other.

Haproxy maintains a total number of outgoing requests to workers, and makes sure that only ones on the "bottom" of the weighted/segmented queue are immediately dispatched if a slot opens.

#### Describe the solution you'd like.

This above solution works for our use-case, but since this application is deployed widely, and has the same problem everywhere, I'd like for a feature to exist within puma/rails to allow for queue priority.

I'm currently looking at the `Reactor` class for this, which takes in new requests and selects ready  ones to be dispatched to Rack in the thread pool.

For this feature, Puma would have to introspect each request when it comes in, and assign it a priority number according to some rules.

These rules can be as simple as;
- URL path prefix/match (regex)
- Header presence / content match (regex)

For the purposes of the application I'm talking about.

It could most likely be implemented by accepting a `&block` with an input of the request + headers, and an output of an integer, which'd then determine the priority of the request in the todo queue.

These rules could probably be given in `puma.rb`, and be shipped with the application.

This would effectively only "weigh" the `todo` queue, as described in the [architecture documentation](https://github.com/puma/puma/blob/master/docs/architecture.md).

#### Describe alternatives you've considered.

The current setup (with nginx and haproxy) can be considered "fine", and while I [have documented](https://gist.github.com/ShadowJonathan/15d85646184cda2e2b7ee1326f7306d3) how to implement this, in practice only very savvy administrators are willing to implement this (and only after "thundering herds" become an actual problem), and so it'll be a limited implementation.

ShadowJonathan:
Resolves #3414

### Description
Adds request prioritisation to the internal "todo" queue by a block that outputs integer determination.

Higher priority requests get processed first.

This adds a new class called `FIFOPriorityQueue`: it is a custom implementation of a Priority Queue, without heaps, since I wanted the guarantee of FIFO, and existing heaps did not seem to guarantee stability.

(Existing heaps also did not easily/quickly find a specific element, so a binary search (`O(log n)`) seemed sufficient for that as well, and unshifting stays practically/amortised `O(1)`, with only `size` having a small extra cost at needing to enumerate all queues for their `.size`.)

Preliminary code reviews welcome! I'd like to know if this (kind of) approach is something that Puma wishes to go forward wish before I'll write test cases and the likes.

### TODO
- [ ] Determine if it should be `prioritize` or `prioritise`
- [ ] Add unit tests of the priority queue implementation
- [ ] Add integration tests of the thread pool (maybe?)


ShadowJonathan:
(CI Seems to fail due to entirely random errors? wut?)

ShadowJonathan:
I'm not sure if there are any flakes, but could someone restart the CI run?

dentarg:
CI has been flaky until recently, but I think it is in pretty good shape now actually, but ruby-head just started failing (https://github.com/puma/puma/pull/3418#issuecomment-2198241565), so we have to overlook that (I don't think we want to "allow it to fail" as it is good to report it upstream when we see it). If you rebase the `macos-11` job that can't be started has been addressed. Not sure how much CI tells us when the feature isn't exercised though? :-)

As to whether Puma wants this feature or not, I don't have any answer. I will try to give it some thought myself, but it would be good with input from @nateberkopec and others too.

Have you tried these changes in your environment and compared them to your nginx setup? Does it work? (Like, are requests actually prioritized according to your expectations)

ShadowJonathan:
I've just tested this with a local test setup, and encountered a problem;

Puma seems to stop pulling new requests from the socket, even in single-worker mode, if all threads are current busy handling requests.

This means that, effectively, requests are still handled first-come-first-serve, even with this prioritisation tactic. I understand the existence of the backlog, but effectively, this means that `@todo` (and by extension, this FIFO queue) doesn't grow more than 1 or 2.

Furthermore, as the `ThreadPool` is setup and not shared between workers, the prioritisation only happens on every worker itself.

I can understand the lack of fetching new connections from the socket, to distribute it to every worker, it severely hampers the ability to prioritise requests like this.

----

If we add a "limit" of how many requests get queued, but there is a chance the cluster will receive more requests than it can process in a time-frame in that limited amount, and thus start to clog the queues, causing it to process those lower-priority requests regardless, even if within all of the currently-trying-to-connect clients there are higher-priority requests waiting.

----

The only solution and probable viable use-case for this right now would be when `workers 0`, and an explicit opt-in to keep accepting new connections from the socket, even when all threads are busy.

Removing `pool.wait_until_not_full` from `server.rb` and starting with `workers 0` gets the behaviour I want; requests queued with the right priorities, and lower-priority requests unable to progress due to higher-priority requests being sent first.

(FWIW, it looks like `Puma::Client` will only keep a request body of ~114 KB into memory, before moving and reading it to a temporary file.)

----

Supporting a proper prioritisation in cluster mode would require a lot of redo of the the internal structure, and I don't think that is worth it.

This is effectively a single-worker accept-and-buffer-**all**-connections-early prioritisation implementation, since each worker accepts and processes requests on its own, which;
- If each worker were to accept connections conservatively, which is the current behaviour, it would effectively not prioritise requests at all, since *as soon* as all workers are processing requests, it would stop accepting additional connections, thus also stopping queuing requests.
- If each worker were to accept connections eagerly, it would steal work from other workers, and effectively be saddled with a ton of work.
- If each worker accepted up to X amount of additional connections, it would still allow an overflow of incoming requests to remove the effectiveness of the prioritisation mechanism, as the cluster needs to still process a certain amount of requests *before* the connection with the high-priority request gets accepted, prioritised, and *then* processed.

A "proper" solution would be that a single isolated worker/control process would accept, buffer and prioritise all incoming requests, and then hand them out to workers as they accept more work, but this would require a redo/rethink of most of puma's internals, as the pivot method currently rests on letting the system queue incoming TCP connections, and then having each worker independently accept at their leasure.

dentarg:
Thank you @ShadowJonathan for the very thorough analysis above. Reading that, this feature doesn't sound like a good fit for Puma.

ShadowJonathan:
I believe the implementation as-is would not work properly, but i still believe the feature itself could be useful, it just needs a more deeply integrated method of implementation, which I'm not skilled enough to deliver

schneems:
> I believe the implementation as-is would not work properly

Agreed, I left some notes in the issue about more thoughts. Thanks a ton for the PR. Even if it's not getting merged, it seemed like you learned about the Puma behavior and you're primed to make the next contribution even easier. I'm going to close this issue.

PR #3418, opened 2024-06-29
CI: use `macos-12` in ragel workflow

dentarg:
From https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

> The macos-11 label has been deprecated and will no longer be available
> after 28 June 2024.

### DIFF
--- a/.github/workflows/ragel.yml
+++ b/.github/workflows/ragel.yml
@@ -28,7 +28,7 @@ jobs:
       matrix:
         include:
           - { os: ubuntu-20.04 , ruby: head }
-          - { os: macos-11     , ruby: head }
+          - { os: macos-12     , ruby: head }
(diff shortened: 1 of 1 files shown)


dentarg:
Looks like all things ruby-head currently fails: https://bugs.ruby-lang.org/issues/20596

dentarg:
Looks like all things ruby-head currently fails: https://bugs.ruby-lang.org/issues/20596

MSP-Greg:
Re Ruby head CI failures, it was an intermittent failure related to GC, I could intermittently repro on Ubuntu 20.04 (but not 22.04).

I believe https://github.com/ruby/ruby/commit/9d76a0ab4a582d has fixed it.

All setup-ruby head builds are now past that commit.

PR #3421, opened 2024-07-04
Allow to customise plugin's path and name

Aerdayne:
### Description

This PR augments the Puma plugin API in order to offer more flexibility to developers of 3rd-party plugins.

It allows to optionally specify a custom override for the plugin's name and the path where its definition is located, instead of having to conform to the expected directory structure and naming.

This is useful when a 3rd-party gem provides a plugin, but for whatever reason does not want to conform to the expected path structure and file naming rules.

I haven't found any tests for the existing surrounding code, otherwise I would have adjusted the specs - let me know if I missed them.

I also understand that this might be too niche of a use-case to warrant a change in the API, so I'm totally fine with this being closed.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -22,6 +22,51 @@ The `tmp_restart` plugin comes with Puma, so it is always available.
 
 To use the `heroku` plugin, add `puma-heroku` to your Gemfile or install it.
 
+### Development
+
+When developing a plugin, there are 2 conventions that must be met:
+- A plugin must be defined using the `Puma::Plugin.create` API;
+- A Ruby file containing the creation of a plugin using the above method must be located under the `puma/plugin` load path.
+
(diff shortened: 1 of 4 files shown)


dentarg:
> I haven't found any tests for the existing surrounding code, otherwise I would have adjusted the specs - let me know if I missed them.

If that is the case (I also haven't looked), it would be great to add some.

> I also understand that this might be too niche of a use-case to warrant a change in the API, so I'm totally fine with this being closed.

Looks like the change to the API is backwards compatible, is that correct?

Seeing the example in the updated docs,

```ruby
# In Puma configuration file
plugin "custom_plugin", "<gem_root>/lib/integrations/puma/plugin"
```

I fail to see why anyone would want to do it like this... isn't it pretty complex to have users supply a path from inside a gem? What is the actual benefit of this?

Aerdayne:
Closing this as I don't have the time to see this through and my originally described use-case is no longer relevant.

PR #3439, opened 2024-08-09
Set the worker process count automatically with WEB_CONCURRENCY=auto

### LINKED ISSUE #3437: Use `available_processor_count` as worker numbers if `WEB_CONCURRENCY` is set to `"auto"`
We are trying to make easier for applications to configure better defaults to puma in Rails. One of the features we introduced in our configuration was to automatically detect the number of processors and set that as the worker count.

We later had to revert that change because some platforms lie about how many processors they provide to users. So instead we defined a new value to `WEB_CONCURRENCY` that is `"auto"`.

We had to rever that change because puma is also reading the value of that variable and that would just explode trying to parsing it that.

See https://github.com/rails/rails/commit/142e6ab2c1e3cb7df4b82e20e2a1cd676b755714

Now that puma is taking care of reading `WEB_CONCURRENCY`, I think it would make sense to puma to implement the `"auto"` feature. This would allow Rails and non-Rails users to get better default without having to change the puma.rb file with conditionals.

Rails could introduce a new variable to control the concurrency of puma, but I think that would provide worse experience since people are already used with `WEB_CONCURRENCY`.

codergeek121:
### Description

Implements the `WEB_CONCURRENCY=auto` feature described in #3437.

This implementation does not `require` `concurrent-ruby` at all (in puma).

An alternative could be to `require "concurrent/utility/processor_counter"` inline. I'd be happy to adapt 👍

Closes #3437.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,8 @@ $ WEB_CONCURRENCY=3 puma -t 8:32
 
 Note that threads are still used in clustered mode, and the `-t` thread flag setting is per worker, so `-w 2 -t 16:16` will spawn 32 threads in total, with 16 in each worker process.
 
+If the `WEB_CONCURRENCY` environment variable is set to `"auto"` and the `concurrent-ruby` gem is available in your application, Puma will set the worker process count to the result of [available processors](https://ruby-concurrency.github.io/concurrent-ruby/master/Concurrent.html#available_processor_count-class_method).
+
 For an in-depth discussion of the tradeoffs of thread and process count settings, [see our docs](https://github.com/puma/puma/blob/9282a8efa5a0c48e39c60d22ca70051a25df9f55/docs/kubernetes.md#workers-per-pod-and-other-config-issues).
 
 In clustered mode, Puma can "preload" your application. This loads all the application code *prior* to forking. Preloading reduces total memory usage of your application via an operating system feature called [copy-on-write](https://en.wikipedia.org/wiki/Copy-on-write).
--- a/lib/puma/configuration.rb
+++ b/lib/puma/configuration.rb
@@ -227,7 +227,12 @@ module Puma
     def puma_options_from_env(env = ENV)
       min = env['PUMA_MIN_THREADS'] || env['MIN_THREADS']
       max = env['PUMA_MAX_THREADS'] || env['MAX_THREADS']
-      workers = env['WEB_CONCURRENCY']
+      workers = if env['WEB_CONCURRENCY'] == 'auto'
+        require_processor_counter
+        ::Concurrent.available_processor_count
+      else
+        env['WEB_CONCURRENCY']
+      end
 
       {
         min_threads: min && Integer(min),
@@ -340,6 +345,16 @@ module Puma
     end
 
     private
+
+    def require_processor_counter
+      require 'concurrent/utility/processor_counter'
+    rescue LoadError
+      warn <<~MESSAGE
+        WEB_CONCURRENCY=auto requires the "concurrent-ruby" gem to be installed.
+        Please add "concurrent-ruby" to your Gemfile.
+      MESSAGE
+      raise
+    end
 
     # Load and use the normal Rack builder if we can, otherwise
     # fallback to our minimal version.
--- a/test/test_web_concurrency_auto.rb
+++ b/test/test_web_concurrency_auto.rb
@@ -1,0 +1,29 @@
+require_relative "helper"
+require_relative "helpers/integration"
+
+class TestWebConcurrencyAuto < TestIntegration
+
+  def teardown
+    return if skipped?
+    super
+  end
+
+  def test_web_concurrency_with_concurrent_ruby_available
+    skip_unless :fork
+    env = {
+      "BUNDLE_GEMFILE" => "#{__dir__}/web_concurrency_test/Gemfile",
+      "WEB_CONCURRENCY" => "auto"
+    }
+    Dir.chdir("#{__dir__}/web_concurrency_test") do
+      with_unbundled_env do
+        silent_and_checked_system_command("bundle config --local path vendor/bundle")
+        silent_and_checked_system_command("bundle install")
+      end
+      cli_server set_pumactl_args, env: env
+    end
+
+    connection = connect("/worker_count")
+    body = read_body(connection, 1)
+    assert_equal(get_stats.fetch("workers").to_s, body)
+  end
+end
--- a/test/helpers/integration.rb
+++ b/test/helpers/integration.rb
@@ -76,6 +76,15 @@ class TestIntegration < Minitest::Test
     assert(system(*args, out: File::NULL, err: File::NULL))
   end
 
+  def with_unbundled_env
+    bundler_ver = Gem::Version.new(Bundler::VERSION)
+    if bundler_ver < Gem::Version.new('2.1.0')
+      Bundler.with_clean_env { yield }
+    else
+      Bundler.with_unbundled_env { yield }
+    end
+  end
+
   def cli_server(argv,  # rubocop:disable Metrics/ParameterLists
       unix: false,      # uses a UNIXSocket for the server listener when true
       config: nil,      # string to use for config file
--- a/test/test_worker_gem_independence.rb
+++ b/test/test_worker_gem_independence.rb
@@ -138,13 +138,4 @@ class TestWorkerGemIndependence < TestIntegration
 
     true while @server.gets !~ /booted in [.0-9]+s, phase: 1/
   end
-
-  def with_unbundled_env
-    bundler_ver = Gem::Version.new(Bundler::VERSION)
-    if bundler_ver < Gem::Version.new('2.1.0')
-      Bundler.with_clean_env { yield }
-    else
-      Bundler.with_unbundled_env { yield }
-    end
-  end
 end
--- a/test/web_concurrency_test/Gemfile
+++ b/test/web_concurrency_test/Gemfile
@@ -1,0 +1,5 @@
+source "https://rubygems.org"
+
+gem 'puma', path: '../..'
+
+gem 'concurrent-ruby', '~> 1.0'
--- a/test/web_concurrency_test/config.ru
+++ b/test/web_concurrency_test/config.ru
@@ -1,0 +1,5 @@
+map "/worker_count" do
+  run ->(env) {
+    [200, {}, [Concurrent.available_processor_count.to_i.to_s]]
+  }
+end


rafaelfranca:
Would concurrent-ruby be required at that point? The rails application would only require it when the `config/applicaton.rb` file is required.

I think inlining the implementation could be a better idea since requiring outside of the bundler environment could also cause other problems.

codergeek121:
You're right, it wouldn't be required in most setups. I adapted the implementation to require `processor_counter` inline.

I also added `concurrent-ruby` as a `runtime_dependency` with the same version constraints as Rails' ActiveSupport.

puma.gemspec
```diff
@@ -14,6 +14,7 @@ Gem::Specification.new do |s|
   s.email = ["evan@phx.io"]
   s.executables = ["puma", "pumactl"]
   s.extensions = ["ext/puma_http11/extconf.rb"]
+  s.add_runtime_dependency "concurrent-ruby", "~> 1.0", ">= 1.3.1"
```
dentarg, line 17: Pretty sure we don't want to force this dependency on all Puma users :)

codergeek121: Thx for the comment :)

I can totally remove it from the runtimedeps, you're right 👍

Users without `concurrent-ruby` available when setting `auto` would get a `LoadError`. Maybe that's already fine with proper documentation?

I thought about reraising the LoadError with a more descriptive error message: "Please add `concurrent-ruby` to your bundle to use WEB_CONCURRENCY=auto".

Wdyt?

Only checking if Concurrent is defined will lead to an `invalid value for Integer() Argumenterror`, which I guess is unexpected as well.

dentarg: Yes, the load error should be caught and a nice message should be displayed.

I'm not sure if this is the solution Rails project had in mind. WDYT @rafaelfranca?

codergeek121: I added a short simple warning message if requiring fails.

test/test_config.rb
```diff
@@ -687,6 +687,16 @@ def test_config_loads_workers_from_env
     end
   end
 
+  def test_config_sets_workers_to_processor_count_with_auto
+    require "concurrent/utility/processor_counter" # defines Concurrent to allow stubbing
+    ::Concurrent.stub :available_processor_count, 42 do
```
dentarg, line 692: @MSP-Greg had concerns about loading `concurrent-ruby` in the main test process, https://github.com/puma/puma/issues/3437#issuecomment-2278105545

I agree with that. Not sure exactly how to accomplish it, one way is to launch a sub-process with `BUNDLE_GEMFILE` set for it, pointing to a Gemfile with `concurrent-ruby`. The assert becomes trickier, maybe need to output something in a HTTP response? Maybe there is some Puma tests already doing something similar?

codergeek121: On first glance, maybe I can write an integration test similar to `test_url_map.rb`. I'll look into it 👍 . Maybe this is also a good way to test the warning message if `concurrent-ruby` is missing.

codergeek121:
I think I managed to write an integration test without requiring concurrent-ruby in the main process. Also got most Github Actions to pass, the failures seem unrelated/flaky.

MSP-Greg:
@codergeek121

JFYI, The JRuby tests are failing due to expired certs.  The ubuntu-20.04 Ruby 3.1 test is probably an intermittent issue.  I rerun it...

codergeek121:
Great, thank you for double checking :)

lib/puma/configuration.rb
```diff
@@ -338,6 +343,16 @@ def self.temp_path
 
     private
 
+    def require_processor_counter
+      require 'concurrent/utility/processor_counter'
+    rescue LoadError
+      warn <<~MESSAGE
+        WEB_CONCURRENCY=auto requires the "concurrent-ruby" gem to be installed.
+        Please add "concurrent-ruby" to your Gemfile.
```
rafaelfranca, line 351: That works for me

rafaelfranca (approved):

README.md
```diff
@@ -118,6 +118,8 @@ $ WEB_CONCURRENCY=3 puma -t 8:32
 
 Note that threads are still used in clustered mode, and the `-t` thread flag setting is per worker, so `-w 2 -t 16:16` will spawn 32 threads in total, with 16 in each worker process.
 
+If the `WEB_CONCURRENCY` environment variable is set to `"auto"`, Puma will set the worker process count to the result of [available processors](https://ruby-concurrency.github.io/concurrent-ruby/master/Concurrent.html#available_processor_count-class_method).
```
dentarg, line 121: ...if the `concurrent-ruby` gem is available/has been loaded – I'm not sure how to exactly phrase it but it should probably be mentioned here.

dentarg (approved):
LGTM! Great work! Just a comment about the README.

README.md
codergeek121: Updated the paragraph to:

If the `WEB_CONCURRENCY` environment variable is set to `"auto"` and the `concurrent-ruby` gem is available in your application, Puma will set the worker process count to the result of [available processors](https://ruby-concurrency.github.io/concurrent-ruby/master/Concurrent.html#available_processor_count-class_method).

lib/puma/configuration.rb
codergeek121: I also created a commit https://github.com/codergeek121/rails/commit/ccb9a8e2117740066b188925844652427b585b48 to update Rails default values & docs to include this new `auto` config value. I'll open up a PR against Rails, if the `auto` configuration is merged in puma 👍

dentarg:
Are you able to rebase?

codergeek121:
Yes, I've been able to rebase 👍

test/test_web_concurrency_auto.rb
```diff
@@ -0,0 +1,31 @@
...
+      with_unbundled_env do
+        silent_and_checked_system_command("bundle config --local path vendor/bundle")
+        silent_and_checked_system_command("bundle install")
+      end
+      cli_server set_pumactl_args, env: env
+    end
+
+    connection = connect("/worker_count")
+    # Puma 6.2.2 and below will time out here with Ruby v3.3
+    # see https://github.com/puma/puma/pull/3165
```
dentarg, line 27: I don't think we need these comments here

(Looks like they were just copy-pasted when the test that https://github.com/puma/puma/pull/3165 added was used for the base of this test)

dentarg (approved):
Seems to work!

```
$ echo 'app { [200, {}, ["OK"]] }; on_stopped { puts "Goodbye!" }' | WEB_CONCURRENCY=auto ruby -Ilib ./bin/puma --config /dev/stdin --port 0 --log-requests
[59172] Puma starting in cluster mode...
[59172] * Puma version: 6.4.3 (ruby 3.3.5-p100) ("The Eagle of Durango")
[59172] *  Min threads: 0
[59172] *  Max threads: 5
[59172] *  Environment: development
[59172] *   Master PID: 59172
[59172] *      Workers: 12
[59172] *     Restarts: (✔) hot (✖) phased
[59172] * Preloading application
[59172] * Listening on http://0.0.0.0:55707
[59172] Use Ctrl-C to stop
[59180] + Gemfile in context: /Users/dentarg/src/puma/Gemfile
[59181] + Gemfile in context: /Users/dentarg/src/puma/Gemfile
[59182] + Gemfile in context: /Users/dentarg/src/puma/Gemfile
[59183] + Gemfile in context: /Users/dentarg/src/puma/Gemfile
[59184] + Gemfile in context: /Users/dentarg/src/puma/Gemfile
[59185] + Gemfile in context: /Users/dentarg/src/puma/Gemfile
[59186] + Gemfile in context: /Users/dentarg/src/puma/Gemfile
[59187] + Gemfile in context: /Users/dentarg/src/puma/Gemfile
[59188] + Gemfile in context: /Users/dentarg/src/puma/Gemfile
[59189] + Gemfile in context: /Users/dentarg/src/puma/Gemfile
[59190] + Gemfile in context: /Users/dentarg/src/puma/Gemfile
[59191] + Gemfile in context: /Users/dentarg/src/puma/Gemfile
[59172] - Worker 0 (PID: 59180) booted in 0.01s, phase: 0
[59172] - Worker 1 (PID: 59181) booted in 0.01s, phase: 0
[59172] - Worker 2 (PID: 59182) booted in 0.01s, phase: 0
[59172] - Worker 3 (PID: 59183) booted in 0.01s, phase: 0
[59172] - Worker 4 (PID: 59184) booted in 0.01s, phase: 0
[59172] - Worker 5 (PID: 59185) booted in 0.01s, phase: 0
[59172] - Worker 6 (PID: 59186) booted in 0.01s, phase: 0
[59172] - Worker 7 (PID: 59187) booted in 0.01s, phase: 0
[59172] - Worker 8 (PID: 59188) booted in 0.0s, phase: 0
[59172] - Worker 9 (PID: 59189) booted in 0.0s, phase: 0
[59172] - Worker 10 (PID: 59190) booted in 0.0s, phase: 0
[59172] - Worker 11 (PID: 59191) booted in 0.0s, phase: 0
^C[59172] - Gracefully shutting down workers...
Goodbye!
[59172] === puma shutdown: 2024-11-04 13:57:13 +0100 ===
[59172] - Goodbye!
```

dentarg:
Hmm, Ruby 2.4 in macOS failed https://github.com/puma/puma/actions/runs/11664819104/job/32476175139?pr=3439

Seems related 

```
  test/runner --verbose
  shell: /bin/bash -e {0}
  env:
    CI: true
    PUMA_TEST_DEBUG: true
    TESTOPTS: -v
    PUMA_NO_RUBOCOP: true
    TERM: BugMinitest
    PUMA_CI_RACK: 
    TMPDIR: /Users/runner/work/_temp
    PATH: /usr/local/opt/pipx_bin:/Users/runner/.cargo/bin:/usr/local/opt/curl/bin:/usr/local/bin:/usr/local/sbin:/Users/runner/bin:/Users/runner/.yarn/bin:/Users/runner/Library/Android/sdk/tools:/Users/runner/Library/Android/sdk/platform-tools:/Library/Frameworks/Python.framework/Versions/Current/bin:/Library/Frameworks/Mono.framework/Versions/Current/Commands:/usr/bin:/bin:/usr/sbin:/sbin:/Users/runner/.dotnet/tools
    PUMA_MAKE_WARNINGS_INTO_ERRORS: true

ruby 2.4.10p364 (2020-03-31 revision 67879) [x86_64-darwin19]
RUBYOPT: -r/Users/runner/hostedtoolcache/Ruby/2.4.10/x64/lib/ruby/site_ruby/2.4.0/bundler/setup
                         Puma::MiniSSL                   OpenSSL
OPENSSL_LIBRARY_VERSION: OpenSSL 1.1.1h  22 Sep 2020     OpenSSL 1.1.1h  22 Sep 2020
        OPENSSL_VERSION: OpenSSL 1.1.1h  22 Sep 2020     OpenSSL 1.1.1h  22 Sep 2020

Run options: --verbose --seed 51014

# Running:

TestWebConcurrencyAuto#test_web_concurrency_with_concurrent_ruby_available = 
Error: Process completed with exit code 1.
```

dentarg:
Re-running it

codergeek121:
Nice! Thank you for finishing this up and the reviews :)

dentarg:
Something funky with the test... now we had this fluke in the macos-13 ruby 2.5 job https://github.com/puma/puma/actions/runs/11672946770/job/32502922503?pr=3544#step:10:929

Now 

```
  1) Failure:
TestWebConcurrencyAuto#test_web_concurrency_with_concurrent_ruby_available [/Users/runner/work/puma/puma/test/test_web_concurrency_auto.rb:27]:
--- expected
+++ actual
@@ -1,3 +1 @@
-# encoding: US-ASCII
-#    valid: true
-"0"
+"4"
```

MSP-Greg:
@dentarg

I also had test problems locally.

Wondering whether we should add `concurrent-ruby` to the main Gemfile, since

1) it's cached by setup-ruby
2) it's loaded on demand only if `env['WEB_CONCURRENCY'] == 'auto'`
3) No need for a `bundle install` command for one test
3) Using `cli_server` means the test won't load it in the main test process

dentarg:
That all sounds good to me 👍🏻

MSP-Greg:
See #3545

lib/puma/configuration.rb
```diff
@@ -227,7 +227,12 @@ def puma_default_options(env = ENV)
     def puma_options_from_env(env = ENV)
       min = env['PUMA_MIN_THREADS'] || env['MIN_THREADS']
       max = env['PUMA_MAX_THREADS'] || env['MAX_THREADS']
-      workers = env['WEB_CONCURRENCY']
+      workers = if env['WEB_CONCURRENCY'] == 'auto'
+        require_processor_counter
+        ::Concurrent.available_processor_count
```
gondalez, line 232: Hi,

Thanks for this feature it is really handy :)

I was just wondering, is the optimum number the _available_ processors or the _physical_ processors?

I think I'm asking the same question as @MSP-Greg did in this https://github.com/puma/puma/pull/3439#issuecomment-2368500248 🤔 

I tried `WEB_CONCURRENCY=auto` on a 1x vcpu instance on AWS and noticed puma wanted to start up 2x processes.

If that is deliberate and, in your expert opinion, 2 processes per "hyper threaded" (or equivalent) is optimal then all good 😌 

If it isn't then maybe the above can be changed to `::Concurrent.physical_processor_count`.

I would be happy to PR that, but honestly I'm no expert on which would be optimal so thought I'd check first :) 

Here is some console output on the 1vcpu instance showing that `available_processor_count` returns `2.0` while `physical_processor_count` returns `1`.

``` 
> Concurrent.physical_processor_count
1
> ::Concurrent.available_processor_count
2.0
```

nateberkopec: In my experience, everyone uses logical (available) core count.

nateberkopec: >I tried WEB_CONCURRENCY=auto on a 1x vcpu instance on AWS and noticed puma wanted to start up 2x processes.

Which instance type is this? I don't think you actually are getting 1 physical/2 logical cores, I think you're on a 2 logical core machine getting 1 logical core worth of CPU time via a hypervisor. vcpu in AWS land is always equivalent to 1 logical core/1 HT.

gondalez: Thanks for the reply 🙏 

It is an aws fargate instance with `cpu` set to `1024`. 
According to the [docs](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#task_size) it is one vcpu.

Here is the `lscpu` output for context:

```
Architecture:             x86_64
  CPU op-mode(s):         32-bit, 64-bit
  Address sizes:          46 bits physical, 48 bits virtual
  Byte Order:             Little Endian
CPU(s):                   2
  On-line CPU(s) list:    0,1
Vendor ID:                GenuineIntel
  Model name:             Intel(R) Xeon(R) Platinum 8259CL CPU @ 2.50GHz
    CPU family:           6
    Model:                85
    Thread(s) per core:   2
    Core(s) per socket:   1
    Socket(s):            1
    Stepping:             7
    BogoMIPS:             4999.99
    Flags:                fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ss ht syscall nx pdpe1gb rdtscp
                          lm constant_tsc rep_good nopl xtopology nonstop_tsc cpuid tsc_known_freq pni pclmulqdq ssse3 fma cx16 pcid sse4_1 sse4_2 x2apic mo
                          vbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand hypervisor lahf_lm abm 3dnowprefetch invpcid_single pti fsgsbase tsc_adjus
                          t bmi1 avx2 smep bmi2 erms invpcid mpx avx512f avx512dq rdseed adx smap clflushopt clwb avx512cd avx512bw avx512vl xsaveopt xsavec
                           xgetbv1 xsaves ida arat pku ospke
Virtualization features:
  Hypervisor vendor:      KVM
  Virtualization type:    full
Caches (sum of all):
  L1d:                    32 KiB (1 instance)
  L1i:                    32 KiB (1 instance)
  L2:                     1 MiB (1 instance)
  L3:                     35.8 MiB (1 instance)
```

When I run `top` and press `t` `1` it does show multiple cpus 🤔 

```
top - 02:09:52 up 20:25,  0 user,  load average: 0.02, 0.04, 0.05
Tasks:   7 total,   1 running,   5 sleeping,   0 stopped,   1 zombie
%Cpu0  :   0.7/0.3     1[|                                                                                                   ]
%Cpu1  :   0.7/0.0     1[|                                                                                                   ]
MiB Mem :   3850.9 total,    161.2 free,   1129.5 used,   2842.9 buff/cache
MiB Swap:      0.0 total,      0.0 free,      0.0 used.   2721.3 avail Mem
```

Reads to me like it is actually 1 physical cpu and two logical/ht cores then. Maybe perhaps :)

In which case the current behaviour seems good and might improve performance (we have WEB_CONCURRENCY=1 on production currently on 1vcpu instances that have enough free memory to support two processes).

PR #3449, opened 2024-08-14
[CI] Fixes for Minitest 5.25.0 - used with Ruby 2.6 and later

MSP-Greg:
### Description

CI broke with the release of Minitest 5.25.0.

Two commits:

1. test/helper.rb - allow to run with new and old Minitest - See https://github.com/minitest/minitest/commit/8cd3b1c749c778038ead3ddf7894dc19c1fe4797.  This conflicts with our prepending a method to allow our tests to fail on a timeout.

2. [CI] fix Minitest bug - revert when Minitest 5.25.1 is released - See https://github.com/minitest/minitest/pull/1010.  Previously Minitest had code that could be equivalent to:
   ```ruby
   nil ~= /broken/
   ```
   it was replaced by:
   ```ruby
   nil.match?(/broken/)
   ```

   `match?` is not defined on `nil`, so an error is thrown.  This commit is a temporary fix until the next Minitest release.  It defines an `ENV` variable in the Actions workflow so that the value isn't `nil`.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -30,6 +30,7 @@ jobs:
       PUMA_TEST_DEBUG: true
       TESTOPTS: -v
       PUMA_NO_RUBOCOP: true
+      TERM: BugMinitest
 
     runs-on: ${{ matrix.os }}
     if: |
@@ -143,6 +144,7 @@ jobs:
       PUMA_TEST_DEBUG: true
       TESTOPTS: -v
       PUMA_NO_RUBOCOP: true
+      TERM: BugMinitest
 
     runs-on: ${{ matrix.os }}
     if: |
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -80,17 +80,48 @@ module UniquePort
 end
 
 require "timeout"
-module TimeoutEveryTestCase
-  # our own subclass so we never confuse different timeouts
-  class TestTookTooLong < Timeout::Error
-  end
 
-  def run
-    with_info_handler do
+if Minitest::VERSION < '5.25'
+  module TimeoutEveryTestCase
+    # our own subclass so we never confuse different timeouts
+    class TestTookTooLong < Timeout::Error
+    end
+
+    def run
+      with_info_handler do
+        time_it do
+          capture_exceptions do
+            ::Timeout.timeout($test_case_timeout, TestTookTooLong) do
+              before_setup; setup; after_setup
+              self.send self.name
+            end
+          end
+
+          capture_exceptions do
+            ::Timeout.timeout($test_case_timeout, TestTookTooLong) do
+              Minitest::Test::TEARDOWN_METHODS.each { |hook| self.send hook }
+            end
+          end
+          if respond_to? :clean_tmp_paths
+            clean_tmp_paths
+          end
+        end
+      end
+
+      Minitest::Result.from self # per contract
+    end
+  end
+else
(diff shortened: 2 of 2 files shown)


dentarg (approved):
The whole world broke with minitest v5.25.0 😞

dentarg:
> 1. test/helper.rb - allow to run with new and old Minitest - See [minitest/minitest@8cd3b1c](https://github.com/minitest/minitest/commit/8cd3b1c749c778038ead3ddf7894dc19c1fe4797).  This conflicts with our prepending a method to allow our tests to fail on a timeout.

This has been reported as a bug at https://github.com/minitest/minitest/issues/1007

.github/workflows/tests.yml
```diff
@@ -30,6 +30,7 @@ jobs:
       PUMA_TEST_DEBUG: true
       TESTOPTS: -v
       PUMA_NO_RUBOCOP: true
+      TERM: BugMinitest
```
zenspider, line 33: This just hit my radar yesterday and I didn't change anything to do with how pride works in this release... did GHA recently remove the TERM variable?? What's going on? This is new, yeah?

(funny fix btw... @flavorjones disabled pride in mechanize)

test/helper.rb
```diff
@@ -80,17 +80,48 @@ def self.call(host = '127.0.0.1')
...
   end
+else
+  module TimeoutEveryTestCase
+    # our own subclass so we never confuse different timeouts
+    class TestTookTooLong < Timeout::Error
+    end
 
-  def run
-    with_info_handler do
+    def run
```
zenspider, line 120: This doesn't make sense to me. You shouldn't be calling with_info_handler... and you prolly shouldn't be overriding run like this, let alone having to duplicate all of its code. Instead, override [Minitest::Runnable.run_one_method](https://github.com/minitest/minitest/blob/master/lib/minitest.rb#L444) and you don't need ANY of this boilerplate. Should look something like this:

```ruby
def run_one_method klass, method_name, reporter
  ::Timeout.timeout($test_case_timeout, TestTookTooLong) do
    super
  end
ensure
  clean_tmp_paths if respond_to? :clean_tmp_paths
end
```

because it is a class method, you'll prolly have to `extend` rather than `include`... I would make it a subclass of Runnable and subclass it. Happy to provide a PR if you want to see it in full.

zenspider (commented):
Some feedback and questions.

test/helper.rb
MSP-Greg: The code for this dates back many years.  Thanks for the suggestion.  I haven't checked, we may have needed `clean_tmp_paths` to run after the 'teardown' methods.  Not.sure.  Some files we may check whether Puma removed them, so `clean_tmp_paths` is just for cleanup after **all the test code* runs...

.github/workflows/tests.yml
MSP-Greg: For Runner info, I created a repo https://github.com/MSP-Greg/actions-image-testing/actions, see the 'show ENV' step.

At present, it seems none of the runners define TERM.  the older line shown above will accept `nil`, `match?` will not.  No biggy.  Was this a RuboCop change? Maybe their docs should mention the `nil`` issue...  Haven't looked.

test/helper.rb
MSP-Greg: After refreshing, we could probably use `clean_tmp_paths` in `after_teardown`...

PR #3452, opened 2024-08-16
Clean up minitest usage: use a subclass of Minitest::Test, hook run_one_method not run, etc

zenspider:
### Description

This changes the test class in every file, so it looks ugly, but it
should make things cleaner in the long run.

test_helper.rb could use some more cleanup. It is still including into
Minitest::Test when it could just PUT everything in TimeoutTestCase
now.

I can take this further... just say the word.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -30,7 +30,6 @@ jobs:
       PUMA_TEST_DEBUG: true
       TESTOPTS: -v
       PUMA_NO_RUBOCOP: true
-      TERM: BugMinitest
 
     runs-on: ${{ matrix.os }}
     if: |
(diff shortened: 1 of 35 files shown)


zenspider:
I've got 2 jobs timing out... presumably because what was there before was acting as 2x the timeout and my version is only 1x... should I multiply it?

dentarg (approved):
So `run_one_method` covers setup and teardown methods? Neat

dentarg:
If you rebase this you should find that CI has become more stable (well, turbo-rails just [did a thing](https://github.com/hotwired/turbo-rails/issues/681) but besides that ...)

MSP-Greg (approved):

nateberkopec:
Merged in #3604

PR #3466, opened 2024-08-27
server.rb - fixup for `closed_socket?` on JRuby aarch64/ARM64

### LINKED ISSUE #3465: NoMethodError on receiving requests when running on aarch64 platform with JRuby
**Describe the bug**
Hi Puma team,

I am running a Rails server with Puma in the cloud. I am currently running it on x86_64 hardware and was trying to migrate to aarch64. Specifically to AWS Graviton EC2 instances. With no other change being made besides the underlying hardware, the Puma server now throws a `NoMethodError` upon receiving any requests. I am running Puma using JRuby v9.3.9.0 on JDK 17.

```
# jruby -v
jruby 9.3.9.0 (2.6.8) 2022-10-24 537cd1f8bc OpenJDK 64-Bit Server VM 17.0.12+8-LTS on 17.0.12+8-LTS +jit [aarch64-linux]
```

The error -

```
Error reached top of thread-pool: undefined method `>=' for nil:NilClass (NoMethodError)
```

The initial error did not come with a stacktrace so I modified the Puma source code a little bit and recompiled the gem to print that as well. This is the stacktrace -

```
Error reached top of thread-pool: undefined method `>=' for nil:NilClass (NoMethodError)
jruby9.x/lib/ruby/gems/shared/gems/puma-6.4.2-java/lib/puma/server.rb:206:in `closed_socket?': undefined method `>=' for nil:NilClass (NoMethodError)
        from jruby9.x/lib/ruby/gems/shared/gems/puma-6.4.2-java/lib/puma/request.rb:161:in `prepare_response'
        from jruby9.x/lib/ruby/gems/shared/gems/puma-6.4.2-java/lib/puma/server.rb:577:in `response_to_error'
        from jruby9.x/lib/ruby/gems/shared/gems/puma-6.4.2-java/lib/puma/server.rb:549:in `client_error'
        from jruby9.x/lib/ruby/gems/shared/gems/puma-6.4.2-java/lib/puma/server.rb:506:in `process_client'
        from jruby9.x/lib/ruby/gems/shared/gems/puma-6.4.2-java/lib/puma/server.rb:251:in `block in run'
        from jruby9.x/lib/ruby/gems/shared/gems/puma-6.4.2-java/lib/puma/thread_pool.rb:155:in `block in spawn_thread'
```

The `state` variable [here](https://github.com/puma/puma/blame/master/lib/puma/server.rb#L201-L203) is `nil` for some reason. However, I was not able to debug why. This error is only showing up when I am passing `-J--add-opens=java.base/sun.nio.ch=org.jruby.dist -J--add-opens=java.base/java.io=org.jruby.dist` as command line arguments to the JRuby process. Without these everything works just fine. This also worked fine for x86_64. It only started breaking with aarch64. I think this is probably a JRuby issue but I am opening a bug over here to understand what exactly is Puma trying to do here that's failing, so that I can reach out to the JRuby folks with that information to debug further.

**Puma config:**

```ruby
# /tmp/config.rb

max_threads_count = 100
min_threads_count = 20
# Configure the number of threads to use to answer requests.
# https://msp-greg.github.io/puma/Puma/DSL.html#threads-instance_method

# # https://msp-greg.github.io/puma/Puma/DSL.html#environment-instance_method
environment 'production'

# Define the TCP port to bind to.
# https://msp-greg.github.io/puma/Puma/DSL.html#port-instance_method
port 8080

# How long to wait for threads to stop when shutting them down. Defaults to :forever. Specifying :immediately will cause
# ::Puma to kill the threads immediately. Otherwise the value is the number of seconds to wait.
# https://msp-greg.github.io/puma/Puma/DSL.html#force_shutdown_after-instance_method
force_shutdown_after 120

# Specify the backend for the ::IO selector.
# https://msp-greg.github.io/puma/Puma/DSL.html#io_selector_backend-instance_method
io_selector_backend :java

# Control how the remote address of the connection is set. This is configurable because to calculate the true socket
# peer address a kernel syscall is required which for very fast rack handlers slows down the handling significantly.
# https://msp-greg.github.io/puma/Puma/DSL.html#set_remote_address-instance_method
set_remote_address header: 'X-FORWARDED-FOR'
```

**To Reproduce**
Please add reproduction steps here.

```ruby
# hello.ru

```

```sh
# Command line to start the server

jruby9.x/bin/jruby -J--add-opens=java.base/sun.nio.ch=org.jruby.dist -J--add-opens=java.base/java.io=org.jruby.dist jruby9.x/lib/ruby/gems/shared/bin/puma -C /tmp/config.rb /tmp/hello.ru
```

```
curl -v http://localhost:8080
*   Trying 127.0.0.1:8080...
* Connected to localhost (127.0.0.1) port 8080
> GET / HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/8.3.0
> Accept: */*
>
* Empty reply from server
* Closing connection
curl: (52) Empty reply from server
```

**Expected behavior**
Get back 200 OK response with "hello world" from the server

 - OS: Amazon Linux 2
 - Puma Version: 6.4.2
 - Ruby version: JRuby 9.3.9.0
 - JDK version: 17.0.12+8-LTS
 - Platform: aarch64

MSP-Greg:
### Description

JRuby running on aarch64/ARM64 linux may have a bug when using

[`BasicSocket.getsockopt(Socket::IPPROTO_TCP, Socket::TCP_INFO)`](https://msp-greg.github.io/ruby_trunk/socket/BasicSocket.html#getsockopt-instance_method)

This works around the bug.

Closes #3465

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -199,6 +199,12 @@ module Puma
           false
         else
           state = tcp_info.unpack(UNPACK_TCP_STATE_FROM_TCP_INFO)[0]
+          # conditional is for JRuby on linux aarch64/ARM64
+          # https://github.com/puma/puma/issues/3465
+          unless state
+            @precheck_closing = false
+            return false
+          end
           # TIME_WAIT: 6, CLOSE: 7, CLOSE_WAIT: 8, LAST_ACK: 9, CLOSING: 11
           (state >= 6 && state <= 9) || state == 11
         end


lib/puma/server.rb
```diff
@@ -199,6 +199,11 @@ def closed_socket?(socket)
           false
         else
           state = tcp_info.unpack(UNPACK_TCP_STATE_FROM_TCP_INFO)[0]
+          # conditional is for JRuby on linux aarch64/ARM64
```
dentarg, line 202:
```suggestion
          # conditional is for JRuby on linux aarch64/ARM64
          # https://github.com/puma/puma/issues/3465
```

MSP-Greg: Added

dentarg:
Seeing this will be [fixed in a future JRuby release](https://github.com/puma/puma/issues/3465#issuecomment-2328941671), should we close this PR?

MSP-Greg:
I don't know.

Everyone doesn't always use the most recent release.  Also, it's a 'no cost' guard for all the various OS/platform combinations that we can't test on.

ahorek:
@dentarg actually, the fix for aarch64 was released in February 2023. Upgrading to supported JRuby 9.3.x (or 9.4) should be the preferable way to fix the issue.

also, CORK support can be easily disabled through monkey patching if anyone requires the older version for some reason.
```
if ENV_JAVA['os.name'].match?(/Linux/i) && ENV_JAVA['os.arch'].eql?('aarch64')
  Puma::Server.class_eval do
    def closed_socket?(socket)
      false
    end
  end
end
```
the patch suggested by @MSP-Greg works as "an extra safety measure", but I don’t believe it's necessary...

dentarg:
Thanks @ahorek 

I think we should close the PR in order to keep the Puma code base as simple as possible.

PR #3475, opened 2024-08-30
[CI] test_persistent.rb - use TestPuma::PumaSocket

MSP-Greg:
### Description

File uses blocking request socket.gets calls.

### DIFF
--- a/test/test_persistent.rb
+++ b/test/test_persistent.rb
@@ -1,11 +1,28 @@
+# frozen_string_literal: true
+
 require_relative "helper"
+require_relative "helpers/test_puma/puma_socket"
 
 class TestPersistent < Minitest::Test
+  parallelize_me!
+
+  include ::TestPuma::PumaSocket
 
   HOST = "127.0.0.1"
 
   def setup
+    @body = ["Hello"]
+
     @valid_request  = "GET / HTTP/1.1\r\nHost: test.com\r\nContent-Type: text/plain\r\n\r\n"
+
+    @valid_response = <<~RESP.gsub("\n", "\r\n").rstrip
+      HTTP/1.1 200 OK
+      X-Header: Works
+      Content-Length: 5
+
+      Hello
+    RESP
+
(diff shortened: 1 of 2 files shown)


test/test_persistent.rb
```diff
@@ -1,11 +1,28 @@
...
+  include ::TestPuma::PumaSocket
 
   HOST = "127.0.0.1"
 
   def setup
+    @body = ["Hello"]
+
     @valid_request  = "GET / HTTP/1.1\r\nHost: test.com\r\nContent-Type: text/plain\r\n\r\n"
+
+    @valid_response = <<~RESP.gsub("\n", "\r\n").rstrip
```
dentarg, line 18: This could be a neat helper to have somewhere, no need to take action now, just something I thought of, could be a follow-up later

dentarg (approved):

test/test_persistent.rb
MSP-Greg: What to call the method?  Maybe one of the below ('here' could also be 'heredoc')
```
to_response
str_to_response
here_to_response
here_to_resp
here2response
here2resp
```

I assume we'd add it to `String`.

I haven't looked thru a lot of the 'expected' response strings, I think I just changed the longer ones to heredocs...

dentarg: It would be cool with a DSL method

```ruby 
response do
  ...
end
```

MSP-Greg: Sorry.  I thought you were referring to something like changing:
```ruby
    @valid_response = <<~RESP.gsub("\n", "\r\n").rstrip
      HTTP/1.1 200 OK
      X-Header: Works
      Content-Length: 5

      Hello
    RESP
```

to:

```ruby
    @valid_response = <<~RESP.here2resp
      HTTP/1.1 200 OK
      X-Header: Works
      Content-Length: 5

      Hello
    RESP
```

dentarg: That looks good too and is probably much simpler to get going. I think I like `to_response` the most then.

PR #3477, opened 2024-09-01
Make preload app compatible with phased restart

joshuay03:
### Description

Phased restart is currently not performed if `preload_app` is enabled, as described [here](https://github.com/puma/puma/pull/3477/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L139). The goal here is to remove this restriction by (re-)preloading the app in the master process before workers are forked for the new phase.

This means that both the initial boot and restarts to serve the new application can benefit from [copy-on-write](https://en.wikipedia.org/wiki/Copy-on-write) (and other custom configs like heap compaction via [`Process.warmup`](https://ruby-doc.org/3.3.4/Process.html#method-c-warmup) before fork), thereby reducing overall memory consumption.

### DIFF
--- a/test/test_worker_gem_independence.rb
+++ b/test/test_worker_gem_independence.rb
@@ -19,14 +19,32 @@ class TestWorkerGemIndependence < TestIntegration
   end
 
   def test_changing_nio4r_version_during_phased_restart
-    change_gem_version_during_phased_restart old_app_dir: 'worker_gem_independence_test/old_nio4r',
+    change_gem_version_during_phased_restart server_opts: "--prune-bundler",
+                                             old_app_dir: 'worker_gem_independence_test/old_nio4r',
+                                             old_version: '2.7.1',
+                                             new_app_dir: 'worker_gem_independence_test/new_nio4r',
+                                             new_version: '2.7.2'
+  end
+
+  def test_changing_nio4r_version_during_phased_restart_with_preload_app
+    change_gem_version_during_phased_restart server_opts: "--preload",
+                                             old_app_dir: 'worker_gem_independence_test/old_nio4r',
+                                             old_version: '2.7.1',
+                                             new_app_dir: 'worker_gem_independence_test/new_nio4r',
+                                             new_version: '2.7.2'
+  end
+
+  def test_changing_nio4r_version_during_phased_restart_with_prune_bundler_and_preload_app
+    change_gem_version_during_phased_restart server_opts: "--prune-bundler --preload",
+                                             old_app_dir: 'worker_gem_independence_test/old_nio4r',
                                              old_version: '2.7.1',
                                              new_app_dir: 'worker_gem_independence_test/new_nio4r',
                                              new_version: '2.7.2'
   end
 
   def test_changing_json_version_during_phased_restart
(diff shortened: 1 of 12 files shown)


joshuay03:
> What needs to happen to accomplish this? I know the "description" here is WIP but to me it would be natural to spell out what needs to be addressed in order to solve the problem before trying to solve it :-)

The real answer is, I don't really know, there could be a bunch of things I haven't considered. I just wanted to try to get it working and put up a draft with what I had so I can get some early feedback, if any, and come back to it later.

Right now it seems to be working, at least in this basic test. I think I'll do some manual testing next weekend and try to come up with a more useful test that doesn't just check for a log line inclusion.

dentarg:
How can we know it is working if we don't know what we're trying to solve? 🤔

joshuay03:
> How can we know it is working if we don't know what we're trying to solve? 🤔

Ah right, my apologies, I assumed the title would be enough for readers with context on this incompatibility.

Phased restart is currently not performed if `preload_app` is enabled, as described [here](https://github.com/puma/puma/pull/3477/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L139). The goal here is to remove this restriction by (re-)preloading the app in the master process before workers are forked for the new phase.

This means that both the initial boot and restarts to serve the new application can benefit from [copy-on-write](https://en.wikipedia.org/wiki/Copy-on-write) (and other things like heap compaction via [`Process.warmup`](https://ruby-doc.org/3.3.4/Process.html#method-c-warmup) after load and before fork).

I'll copy this into the description for now :)

test/test_worker_gem_independence.rb
```diff
@@ -19,22 +19,32 @@ def teardown
...
   def test_changing_nio4r_version_during_phased_restart
-    change_gem_version_during_phased_restart old_app_dir: 'worker_gem_independence_test/old_nio4r',
+    change_gem_version_during_phased_restart server_opts: "--prune-bundler",
+                                             old_app_dir: 'worker_gem_independence_test/old_nio4r',
+                                             old_version: '2.7.1',
+                                             new_app_dir: 'worker_gem_independence_test/new_nio4r',
+                                             new_version: '2.7.2'
+  end
+
+  def test_changing_nio4r_version_during_phased_restart_with_preload_app
```
joshuay03, line 29: This is passing 🎉

test/test_worker_gem_independence.rb
```diff
@@ -19,22 +19,40 @@ def teardown
...
+    change_gem_version_during_phased_restart server_opts: "--preload",
+                                             old_app_dir: 'worker_gem_independence_test/old_nio4r',
+                                             old_version: '2.7.1',
+                                             new_app_dir: 'worker_gem_independence_test/new_nio4r',
+                                             new_version: '2.7.2'
+  end
+
+  def test_changing_nio4r_version_during_phased_restart_with_prune_bundler_and_preload_app
```
joshuay03, line 37: Doesn't really make sense to have a setup like this but it doesn't do any harm either  🤷🏽‍♂️

joshuay03:
I’m closing this because I’ve exceeded the amount of effort I was willing to put into it.

PR #3462, opened 2024-08-22
[CI] test/test_launcher.rb - fixups

MSP-Greg:
### Description

The recent test PR's have several job failures, and one test failed in this file.

1. One test created a thread, and the thread contained an assert.  If the assert fails, the job stops.  Fixed.
2. One test set an ENV variable, but if the test failed, it wasn't unset.  Fixed using `ensure`.
3. One test used Minitest's `capture_io`, which should be avoided.  Removed.

Additionally, some refactoring.

### Your checklist for this pull request
- [x] I have reviewed the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.
- [x] I have added (or updated) appropriate tests if this PR fixes a bug or adds a feature.
- [x] My pull request is 100 lines added/removed or less so that it can be easily reviewed.
- [ ] If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.
- [ ] If this closes any issues, I have added "Closes `#issue`" to the PR description or my commit messages.
- [ ] I have updated the documentation accordingly.
- [x] All new and existing tests passed, including Rubocop.

PR #3479, opened 2024-09-04
[DRAFT] - test only

MSP-Greg:
### Description

This branch contains commits continuing to switch tests over to PumaSocket for client connections.  I've got a 'pro' GH account, so my CI concurrency limits are different than this repo.

I want to see how long it takes running here...

I'll separate the commits here into several PR's, so this shouldn't be merged.  After the PR's are merged, we can start on shared 'server creation' code.  After that, tests will be portable, so we won't need to have test files with 125 tests in them.

Maybe we can have some new files like:

```text
test_request_valid
test_request_invalid
```

At present, tests aren't grouped logically, and it's easy to replicate tests.  Also, we may be able to have both in-process server tests (`Puma::Server.new`) and 'spawn' tests (AKA integration) in the same test file...


MSP-Greg:
Just rebased (thru PR #3462)

schneems:
Are you still working on this?

MSP-Greg:
@schneems

I'll close.  Yes, this is still something I intend to finish, but this will/should never be merged.

Issue #3443, opened 2024-08-13
Puma serves requests with significantly higher latency on HTTP/1.1 compared to HTTP/1.0


amitsaxena:
**Describe the bug**
While benchmarking our app against a similar Heroku and AWS stack during an AWS migration we encountered > 20X latencies on AWS compared to what we were getting on Heroku. Our AWS setup consists of Application load balancer + ECS + Fargate. We have ensured there's no CPU or memory contention on the Fargate instance and still Puma is responding to requests in a painfully slow manner when hit with higher concurrency. After several days of debugging we have come to the conclusion that the load balancer keep alive setting is causing the significant differences in latency. We were eventually able to reproduce the issue by hitting the public IP of the container directly with (HTTP/1.1) and without (HTTP/1.0) keep alive. Below are a couple of load tests showing the difference in latencies:
```
$ ab -n 5000 -c 100 http://xx.xx.xx.xx:9292/health/ping          
This is ApacheBench, Version 2.3 <$Revision: 1913912 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Document Path:          /health/ping
Document Length:        61 bytes

Concurrency Level:      100
Time taken for tests:   6.857 seconds
Complete requests:      5000
Failed requests:        0
Total transferred:      2720000 bytes
HTML transferred:       305000 bytes
Requests per second:    729.16 [#/sec] (mean)
Time per request:       137.144 [ms] (mean)
Time per request:       1.371 [ms] (mean, across all concurrent requests)
Transfer rate:          387.37 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:       44   58  28.0     48    1074
Processing:    46   76  20.1     69     283
Waiting:       46   76  20.1     69     280
Total:         93  134  38.8    121    1174

Percentage of the requests served within a certain time (ms)
  50%    121
  66%    133
  75%    144
  80%    154
  90%    188
  95%    206
  98%    228
  99%    245
 100%   1174 (longest request)

$ ab -k -n 5000 -c 100 http://xx.xx.xx.xx:9292/health/ping
This is ApacheBench, Version 2.3 <$Revision: 1913912 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Document Path:          /health/ping
[...]

Issue #2258, opened 2020-05-12
Puma 5 Experimental Feature Reports


nateberkopec:
Puma 5 contains 3 new experimental performance features:

* `fork_worker` option and `refork` command for reduced memory usage by forking from a worker process instead of the master process. (#2099) [Additional docs on this feature are here.](https://github.com/puma/puma/blob/master/docs/fork_worker.md) Intended result: If enabled, should reduce memory usage.
* `wait_for_less_busy_worker` config. This may reduce latency on MRI through inserting a small delay before re-listening on the socket if worker is busy (#2079). Intended result: If enabled, should reduce latency in high-load (>50% utilization) Puma clusters.
* Added `nakayoshi_fork` config option. Reduce memory usage in preloaded cluster-mode apps by GCing before fork and compacting, where available. (#2093, #2256) Intended result: If enabled, should reduce memory usage.

Note that all of these experiments are only for **cluster mode** Puma configs running on **MRI**.

Part of the reason we're calling them _experimental_ is because we're not sure if they'll actually have any benefit. People's workloads in the real world are often not what we anticipate, and synthetic benchmarks are usually not of any help in figuring out if a change will be beneficial or not.

We do not believe any of the new features will have a negative effect or impact the stability of your application. This is either a "it works" or "it does nothing" experiment.

If any of the features turn out to be particularly beneficial, we may make them defaults in future versions of Puma.

**If you are using any of the 3 new features, please post before and after results or screenshots to this issue.** Note that "it didn't do anything" is still a useful report. Posting ~24 hours of "before" and ~24 hours of "after" data is preferred.

nateberkopec:
Codetriage memory before/after adding nakayoshi_fork seems more or less unaffected
<img width="1219" alt="Screen Shot 2020-05-12 at 12 50 38 PM" src="https://user-images.githubusercontent.com/845662/81636620-3e7e1900-944f-11ea-818a-c76e91ac890a.png">

Codetriage initially had a wee latency spike (2nd to last deploy on the right here) but since calmed down, I'm going to chalk it up to random noise or possibly a bad neighbor (it runs on Heroku 1x dynos)
<img width="1230" alt="Screen Shot 2020-05-12 at 12 52 41 PM" src="https://user-images.githubusercontent.com/845662/81636749-8ef57680-944f-11ea-8f20-2ee1ebe5e175.png">

wjordan:
[...]

PR #3488, opened 2024-09-17
Return a client to the reactor if there is any backlog in the thread pool

evanphx:
### Description
Directly following from #3487, instead of having max_fast_inline only be checked if there is a client waiting in the backlog, swing back the other direction entirely and have max_fast_inline only apply when there is no one waiting.

Put another way, if there is another client ever waiting, kick the current client to the reactor to remain fair. In the event that another client isn't waiting, we only attempt max_fast_inline request before going back to the reactor.
 
### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -479,10 +479,12 @@ module Puma
             # socket for a short time before returning to the reactor.
             fast_check = @status == :run
 
-            # Always pass the client back to the reactor after a reasonable
-            # number of inline requests if there are other requests pending.
-            fast_check = false if requests >= @max_fast_inline &&
-              @thread_pool.backlog > 0
+            # If we've already attempt our max number of fast inline requests
+            # OR there is a client waiting, we skip doing a fast check of the
+            # client and return it to the reactor.
+            if requests >= @max_fast_inline || @thread_pool.backlog > 0
+              fast_check = false
+            end
 
             next_request_ready = with_force_shutdown(client) do
               client.reset(fast_check)


schneems:
That was fast. In another Slack thread, I noted that value never seemed to be false at that location in the server.rb. I thought it was a case where we had moved the logic to the response.rb file but maybe didn't clean that up. I guess not.

Short term, we should validate manually. We can use the same setup I used before in the linked issue but pointing at this `b-starving-clients` branch. I plan on doing that shortly (but anyone else is free to do it as well).

Long term. I would like to have a regression test around this behavior. I've been thinking about it, and for the terminal case where we're seeing a 2 second latency, if we could somehow call `hey` in CI and get the output and assert it didn't take longer than 1 second at max, it would catch a significant regression. The downside is possibly flappy-ness of trying to bake in perf testing into CI. Maybe there's a more robust option? If we could assert the order of requests processed instead of timing information, it would be less prone to false positives.

evanphx:
Totally agree on the testing, both immediate and regression wise. If you're willing, we could probably take your test setup and package it into a GitHub Action.

dentarg:
See https://github.com/puma/puma/issues/1981 for discussions about benchmarks/perf testing.

schneems:
Good news, bad news. 

It looks like that logic works correctly now, but we're still seeing one request with a really long tail. With 5 puma threads, and default max_fast_inline (10), when I hit it with 6 concurrent connections:

```
$ hey -c 6 -n 200 http://localhost:3000/

Summary:
  Total:	9.6162 secs
  Slowest:	2.2525 secs
  Fastest:	0.2008 secs
  Average:	0.2420 secs
  Requests/sec:	20.5902

  Total data:	2574 bytes
  Size/request:	13 bytes

Response time histogram:
  0.201 [1]	|
  0.406 [196]	|■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■
  0.611 [0]	|
  0.816 [0]	|
  1.021 [0]	|
  1.227 [0]	|
  1.432 [0]	|
  1.637 [0]	|
  1.842 [0]	|
  2.047 [0]	|
  2.253 [1]	|
```

I see a 2.2 second response time for one request. I'm not sure why.

I verified that the variable is getting toggled using `bundle open puma` and adding debugging:

```
            puts "#{Thread.current.object_id} requests: #{requests}, backlog: #{@thread_pool.backlog} fast_check: #{fast_check}"
```

I see:

```
1900 requests: 9, backlog: 0 fast_check: true
1940 requests: 9, backlog: 0 fast_check: true
1920 requests: 9, backlog: 0 fast_check: true
1920 requests: 10, backlog: 0 fast_check: false
1940 requests: 10, backlog: 0 fast_check: false
1900 requests: 10, backlog: 0 fast_check: false
1860 requests: 1, backlog: 1 fast_check: false
1940 requests: 1, backlog: 1 fast_check: false
1880 requests: 1, backlog: 0 fast_check: true
1920 requests: 1, backlog: 0 fast_check: true
1900 requests: 1, backlog: 0 fast_check: true
1860 requests: 1, backlog: 1 fast_check: false
```

So when puma sees there's a request waiting, then correctly toggles the value. I'm not sure where that two second request is coming from. I think this logic in this PR is correct and what we want, but there's some piece of the puzzle we're still missing.

evanphx:
@schneems Just to get an additional data point, what does it look like with 2 related scenarios:

1. Remove the fast_max_inline check completely from the conditional I changed.
2. disable fast_check 100% of the time

dentarg:
What about `FAST_TRACK_KA_TIMEOUT`? `10 * 0.2 = 2` https://github.com/puma/puma/issues/3443#issuecomment-2287416197

dentarg:
This diff

```diff
$ git diff
diff --git a/lib/puma/server.rb b/lib/puma/server.rb
index 6c7fac74..e53d2901 100644
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -487,7 +487,13 @@ module Puma
             end

             next_request_ready = with_force_shutdown(client) do
-              client.reset(fast_check)
+              start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+              res = client.reset(fast_check)
+              stop = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+              waited = stop - start
+
+              puts "#{Thread.current.object_id} waited #{waited}" if waited > 0.01
+              res
             end

             unless next_request_ready
```

Running Puma like this: `echo 'app { sleep(0.2); [200, {}, ["OK"]] }' | PUMA_LOG_CONFIG=1 ruby -Ilib bin/puma -w 0 -t 5:5 --config /dev/stdin`

and `hey` like this: `hey -n 200 -c 6 http://127.0.0.1:9292`

gave this output from Puma

```Puma starting in single mode...
* Puma version: 6.4.2 (ruby 3.2.4-p170) ("The Eagle of Durango")
*  Min threads: 5
*  Max threads: 5
*  Environment: development
*          PID: 42098
Use Ctrl-C to stop
1080 waited 0.20534300012513995
1100 waited 0.20073299994692206
1120 waited 0.20517600001767278
1140 waited 0.2051510000601411
1160 waited 0.20305499993264675
```

dentarg:
What are we trying to solve? Isn't it impossible to service more clients than we have resources for? Which is 6 concurrent keep-alive connections against Puma with 5 threads is?

evanphx:
@dentarg We're looking to improve fairness between clients to reduce long tail latency.

dentarg:
> 1. Remove the fast_max_inline check completely from the conditional I changed.

```diff
diff --git a/lib/puma/server.rb b/lib/puma/server.rb
index 6c7fac74..59f85c9e 100644
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -482,12 +482,18 @@ module Puma
             # If we've already attempt our max number of fast inline requests
             # OR there is a client waiting, we skip doing a fast check of the
             # client and return it to the reactor.
-            if requests >= @max_fast_inline || @thread_pool.backlog > 0
+            if @thread_pool.backlog > 0
               fast_check = false
             end
```

```
$ hey -n 200 -c 6 http://127.0.0.1:9292

Summary:
  Total:	8.8261 secs
  Slowest:	2.2528 secs
  Fastest:	0.2005 secs
  Average:	0.2409 secs
  Requests/sec:	22.4334

  Total data:	396 bytes
  Size/request:	2 bytes

Response time histogram:
  0.201 [1]	|
  0.406 [193]	|■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■
  0.611 [0]	|
  0.816 [0]	|
  1.021 [1]	|
  1.227 [0]	|
  1.432 [0]	|
  1.637 [0]	|
  1.842 [0]	|
  2.048 [1]	|
  2.253 [2]	|


Latency distribution:
  10% in 0.2014 secs
  25% in 0.2025 secs
  50% in 0.2047 secs
  75% in 0.2055 secs
  90% in 0.2124 secs
  95% in 0.2297 secs
  99% in 2.2528 secs

Details (average, fastest, slowest):
  DNS+dialup:	0.0001 secs, 0.2005 secs, 2.2528 secs
  DNS-lookup:	0.0000 secs, 0.0000 secs, 0.0000 secs
  req write:	0.0000 secs, 0.0000 secs, 0.0002 secs
  resp wait:	0.2406 secs, 0.2004 secs, 2.2498 secs
  resp read:	0.0001 secs, 0.0000 secs, 0.0008 secs

Status code distribution:
  [200]	198 responses
```

```
* Puma version: 6.4.2 (ruby 3.2.4-p170) ("The Eagle of Durango")
*  Min threads: 5
*  Max threads: 5
*  Environment: development
*          PID: 50987
Use Ctrl-C to stop
960 waited 0.20460700010880828
1000 waited 0.20186699996702373
960 waited 0.2026129998266697
1060 waited 0.20115999993868172
1120 waited 0.03513900004327297
```

dentarg:
> 2\. disable fast_check 100% of the time

```diff
diff --git a/lib/puma/server.rb b/lib/puma/server.rb
index 6c7fac74..7b9a0d00 100644
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -482,12 +482,16 @@ module Puma
             # If we've already attempt our max number of fast inline requests
             # OR there is a client waiting, we skip doing a fast check of the
             # client and return it to the reactor.
-            if requests >= @max_fast_inline || @thread_pool.backlog > 0
-              fast_check = false
-            end
+            fast_check = false
```

```
$ hey -n 200 -c 6 http://127.0.0.1:9292

Summary:
  Total:	8.5548 secs
  Slowest:	0.4142 secs
  Fastest:	0.2005 secs
  Average:	0.2408 secs
  Requests/sec:	23.1448

  Total data:	396 bytes
  Size/request:	2 bytes

Response time histogram:
  0.201 [1]	|
  0.222 [161]	|■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■
  0.243 [0]	|
  0.265 [0]	|
  0.286 [0]	|
  0.307 [0]	|
  0.329 [0]	|
  0.350 [0]	|
  0.371 [0]	|
  0.393 [3]	|■
  0.414 [33]	|■■■■■■■■


Latency distribution:
  10% in 0.2024 secs
  25% in 0.2038 secs
  50% in 0.2056 secs
  75% in 0.2109 secs
  90% in 0.3992 secs
  95% in 0.4043 secs
  99% in 0.4142 secs

Details (average, fastest, slowest):
  DNS+dialup:	0.0001 secs, 0.2005 secs, 0.4142 secs
  DNS-lookup:	0.0000 secs, 0.0000 secs, 0.0000 secs
  req write:	0.0000 secs, 0.0000 secs, 0.0005 secs
  resp wait:	0.2405 secs, 0.2004 secs, 0.4106 secs
  resp read:	0.0001 secs, 0.0000 secs, 0.0014 secs

Status code distribution:
  [200]	198 responses
```

evanphx:
@dentarg Thanks! Makes me think maybe we should just end the max_inline path entirely, esp now that the code paths for the reactor, etc have gone over years of optimization.

MSP-Greg:
Sorry for popping in late.  I think if we want to see consistent output between keep-alive connections and not when Puma is overloaded, maybe lose all the logic for reading the next request after the response has been sent?

Change the loop that is currently from lines 465 - 500 in server.rb to:
```ruby
while true
  @requests_count += 1
  case handle_request(client, requests + 1)
  when false
    break
  when :async
    close_socket = false
    break
  when true
    ThreadPool.clean_thread_locals if clean_thread_locals

    requests += 1

    break unless @queue_requests
    client.set_timeout @persistent_timeout
    if @reactor.add client
      close_socket = false
      break
    end
  end
end
```

It removes all the code to see if the keep-alive connection contains another request and sends it to Reactor.  Timing is similar for keep-alive and non.  If Puma has more connections than it can handle, things will still slow down (as @dentarg mentioned), but the timing doesn't show the long tail problems where some requests are blocked by the keep-alive connections...

I have tested the above with hey, but haven't checked it with Puma's tests...

evanphx:
@MSP-Greg Yup, exactly what i'm thinking.

MSP-Greg:
@evanphx

I'm not sure whether to add the connection to the Reactor (as above) or whether to add it to the ThreadPool, but adding to the Reactor makes the code much cleaner.  More work later.

Do you think this is a major change, as in Puma 7?

dentarg:
> running hey with keep-alive connections shows all connections completed successfully, but Puma is showing errors

I saw it too, I added some logging to `Server#client_error`

```
2024-09-18 08:50:53 +0200 Read: #<NoMethodError: undefined method `write' for nil:NilClass>
/Users/dentarg/src/puma/lib/puma/client.rb:487:in `read_body': undefined method `write' for nil:NilClass (NoMethodError)

      remain -= @body.write(chunk)
                     ^^^^^^
	from /Users/dentarg/src/puma/lib/puma/client.rb:248:in `try_to_finish'
	from /Users/dentarg/src/puma/lib/puma/server.rb:301:in `reactor_wakeup'
	from /Users/dentarg/src/puma/lib/puma/server.rb:251:in `block in run'
	from /Users/dentarg/src/puma/lib/puma/reactor.rb:119:in `wakeup!'
	from /Users/dentarg/src/puma/lib/puma/reactor.rb:76:in `block in select_loop'
	from /Users/dentarg/src/puma/lib/puma/reactor.rb:76:in `select'
	from /Users/dentarg/src/puma/lib/puma/reactor.rb:76:in `select_loop'
	from /Users/dentarg/src/puma/lib/puma/reactor.rb:39:in `block in run'
```

It is this line in `Client#read_body`

https://github.com/puma/puma/blob/796d8c6af139a22e29e57e156c05c66ed3082634/lib/puma/client.rb#L487

schneems:
I was thinking about this in the ride home after dropping my kids off at school. I think I know where that 2-second request is coming from. I like the idea of reducing/removing logic from the server thread, but I also think it's worth understanding why this PR cause the behavior it did. The better we understand the system, the better we know how to move forward.

Short: The logic in puma that prevents reading in more TCP connections than one process can handle might be interacting with this logic that expects `@thread_pool.backlog` to be non-zero if there's another request waiting.

Long: Puma has several "queues"

- TCP socket can accept up to 1024 requests by default on linux, this is a queue
- Each worker has a reactor, this has an internal pool of unbuffered requests. Multiple workers share the same TCP socket queue
- Buffered requests ready for being responded to is another queue (this is where things go after the reactor is done buffering). When a thread is done responding it pulls from here. Multiple threads share the same buffered request queue.

The queues serve as a place of accumulation. You want to make sure that the threads serving requests are always fed and never waiting on work that could have been ready. However, if a prior queue was too greedy, it might starve it's "co-worker" operating in another process. For example, if 10 requests come in, and one worker accepts all of them into it's reactor but it only has 2 threads to feed, then the second worker is starved as 8 requests sit fully buffered yet idle.

So, logic in puma is written to be aware of the queue after it. Reactors are aware of how many threads it needs to feed. It knows if they're busy or not, however it cannot see other "co-worker" behavior or know their state. It wants to prevent the situation where it takes in 10 requests but the system only has 2 threads. Therefore it uses the idle thread count to determine how many requests it can pull from the TCP socket. 

The logic we've got in this part of server.rb `@thread_pool.backlog` represents the final queue of fully buffered requests (I think). We've got logic that says "if someone is waiting and already buffered, let them in" however...we also have logic that says "if all threads are busy, assume another worker can share the load." therefore the reactor never accepts a new request from the TCP socket, therefore `@thread_pool.backlog` is 0...after that I'm a little fuzzy on what happens. Clearly at some point it accepts a new request from the extra client and then it starts interleaving them correctly.

We could test this theory by finding the logic where the reactor determines how many TCP connections it should read in and adding +1 to it. That would likely not be the best permanent fix, but it would possibly prove/disprove my theory

## Moving forward

I think it makes the most sense to have a keep-alive request treated as much like "any request" as possible. If it could be truly the same, then I think that would look like puma throwing the request back into the end of the TCP socket queue, or even better: slotting it in the correct order of arrival. However, puma doesn't own the TCP socket (from an implementation perspective, the OS owns and manages it and puma only has access to user-space APIs). So the second best place would be the reactor. But now we're in kind of a green-threading situation where we're taking on the complexity of scheduling and ordering that would otherwise be done by the OS (it manages the TCP socket).

We fully own the reactor and an ideal case for moving forward might be to make it so that individual threads never need to know anything about whether a request was keep-alive or not. That eliminates complexity in the server, but we still now have to manage the state in the reactor.

It needs to handle:

- Not allowing the system to be DDoS-d by keep-alive requests (i.e. if it's full of requests, but they're all never-ending keepalives, we need to balance that out with bringing in new TCP connections) 
- Fair or fair-ish request ordering (the problem described in my original issue)
- Not starving other workers by pulling in too many connections it cannot immediately work on
- Minimize making puma threads wait for new requests to work on

slizco:
@evanphx , I have not read through this entire thread, but I believe your change suggested by this PR does not take into account the `force_keep_alive` var in the request's `prepare_response` method: https://github.com/puma/puma/blob/796d8c6af139a22e29e57e156c05c66ed3082634/lib/puma/request.rb#L167-L169

I believe this value will be `false` in situations where we want to hand the client back to the reactor. 
In the code where we write the response headers, we do some and-ing with this value and set it as `resp_info[keep_alive]`:
https://github.com/puma/puma/blob/796d8c6af139a22e29e57e156c05c66ed3082634/lib/puma/request.rb#L610-L615

We then write the `Connection: close` header onto the response buffer before we even return back from the `handle_request` call in `process_client`. That means the connection has already been closed and `handle_request` will return `false`:
https://github.com/puma/puma/blob/796d8c6af139a22e29e57e156c05c66ed3082634/lib/puma/request.rb#L656-L665

dentarg:
Reading @schneems latest post made me think of `wait_for_less_busy_worker` (https://github.com/puma/puma/pull/2079, https://github.com/puma/puma/issues/2258). Thought it was worth mentioning here.

dentarg:
@evanphx wanna review https://github.com/puma/puma/pull/3506?

schneems:
I forgot all about this PR, and funny enough we landed something very close to it in https://github.com/puma/puma/pull/3794. That PR landed on a slightly different heuristic, it uses "fast inline" when there are waiting threads in the pool. I benchmarked looking at the todo queue instead and it was significantly slower.

We also ended up closing https://github.com/puma/puma/issues/3487, via Greg's PR. As of 7.1.0 support a "fast inline" capability with a higher limit of requests (as they no longer block other requests due to the @waiting heuristic) and the "longtail" behavior is gone too.

Issue #3490, opened 2024-09-18
`rake compile` does not work in JRuby 9.4 docker images


dentarg:
See https://github.com/jruby/docker-jruby/issues/87 for the full journey and commands to reproduce.

Using `jruby:9.4.8.0-jdk` this happens

```
root@0d5774b3f421:/app# bundle exec rake --trace compile
** Invoke compile (first_time)
** Invoke compile:java (first_time)
** Invoke compile:puma_http11:java (first_time)
** Invoke copy:puma_http11:java (first_time)
** Invoke lib/puma (first_time, not_needed)
** Invoke tmp/java/puma_http11/puma_http11.jar (first_time)
** Invoke tmp/java/puma_http11/.build (first_time)
** Invoke tmp/java/puma_http11 (first_time, not_needed)
** Invoke ext/puma_http11/PumaHttp11Service.java (first_time, not_needed)
** Invoke ext/puma_http11/org/jruby/puma/Http11.java (first_time, not_needed)
** Invoke ext/puma_http11/org/jruby/puma/Http11Parser.java (first_time, not_needed)
** Invoke ext/puma_http11/http11_parser.java.rl (first_time, not_needed)
** Invoke ext/puma_http11/org/jruby/puma/MiniSSL.java (first_time, not_needed)
** Execute tmp/java/puma_http11/.build
javac: invalid flag: --release=8
Usage: javac <options> <source files>
use -help for a list of possible options
rake aborted!
Command failed with status (2): [javac --release=8 -Xlint -d tmp/java/puma_http11 -extdirs /opt/java/openjdk/jre/lib/ext:/usr/java/packages/lib/ext -cp /opt/java/openjdk/jre/lib/resources.jar:/opt/java/openjdk/jre/lib/rt.jar:/opt/java/openjdk/jre/lib/sunrsasign.jar:/opt/java/openjdk/jre/lib/jsse.jar:/opt/java/openjdk/jre/lib/jce.jar:/opt/java/openjdk/jre/lib/charsets.jar:/opt/java/openjdk/jre/lib/jfr.jar:/opt/java/openjdk/jre/classes:/opt/jruby/lib/jruby.jar ext/puma_http11/PumaHttp11Service.java ext/puma_http11/org/jruby/puma/Http11.java ext/puma_http11/org/jruby/puma/Http11Parser.java ext/puma_http11/org/jruby/puma/MiniSSL.java]
...
```

It works in the `jruby:9.3.15.0-jdk` image.

[...]

PR #3491, opened 2024-09-18
Always use latest `rake-compiler`

dentarg:
No need to lock it down, seems to only cause issues.

Close https://github.com/puma/puma/issues/3490


schneems (approved):
LGTM. This is somewhat of a drive-by review though, I'm unsure if someone is a stronger stakeholder might have more thoughts or feelings. It's small enough though that I think they could revert it if they don't like it.

MSP-Greg (approved):

PR #3493, opened 2024-09-18
CI: remove turbo-rails ruby 2.7 job, does not work

dentarg:
turbo-rails added Ruby 3.0 syntax: https://github.com/hotwired/turbo-rails/issues/681

### DIFF
--- a/.github/workflows/turbo-rails.yml
+++ b/.github/workflows/turbo-rails.yml
@@ -25,7 +25,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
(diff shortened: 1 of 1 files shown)


dentarg:
Will we miss testing with Rails 6.1? Is it possible on Ruby 3.x? Rails 6.1 becomes unsupported Oct 1 https://rubyonrails.org/maintenance

dentarg:
It was a mistake and will be rectified with https://github.com/hotwired/turbo-rails/pull/682

dentarg:
Sounds like it will not happen, so I will merge this

PR #3496, opened 2024-09-19
Add opt to enable keep-alives

slizco:
### Description

Puma does not currently support a configuration option for enabdling/disabling keep-alive connections to the server. Because a thread can only handle a single connection at a time, the Puma server under default config will end up serving requests out of order, preferring to process requests off it's already in flight connection. See https://github.com/puma/puma/issues/3487 for more details.

To get around this issue, Puma will close connections every `max_fast_inline` requests. When `max_fast_inline=1`, the server should close the connection after each request, but this only happens _most of the time_. It does not happen all the time, due to the following conditional:
https://github.com/puma/puma/blob/48b89b5f595f4cd72b5e60a3cb49f4f2949afec3/lib/puma/request.rb#L163-L169
Thus it's still possible to end up with long tail latencies because a small percentage of requests will sit in the queue for awhile.

Sometimes, the number of busy threads is less than the max and sometimes, there are no new connections to accept on the socket. However, these decisions are made at a point in time and the state of the server is constantly changing. They are subject to race conditions since other threads are concurrently accessing these variables and taking actions that modify their values.

The absolute safest way to ensure a connection is handed back to the reactor class once the request has been handled is to outright disable keep-alives. This PR proposes exposing an `enable_keep_alives` option which defaults to `true`, as to least disrupt the current behavior. Disabling keep-alives may be done by setting `enable_keep_alives: false` in the Puma config.

### DIFF
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -1268,6 +1268,21 @@ module Puma
     def max_fast_inline(num_of_requests)
       @options[:max_fast_inline] = Float(num_of_requests)
     end
+    
+    # When enabled, disables keep-alives on in-bound connections.
+    # This option is the best way to ensure requests are handled by the server
+    # in the order in which they are received. Disabling keep-alives means Puma
+    # will close the connection after each request made, requiring that a new
+    # connection be opened. There is a TCP operations cost for both the client and server
+    # each time a new connection is opened/closed. This option will ensure fairness at the cost
+    # of potentially more time spent in TCP operations.
+    #
+    # @example
+    #   disable_keep_alives true
+    #   
+    def disable_keep_alives(enabled=false)
+      @options[:max_fast_inline] = enabled
+    end
 
     # Specify the backend for the IO selector.
     #
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -1268,7 +1268,7 @@ module Puma
     def max_fast_inline(num_of_requests)
       @options[:max_fast_inline] = Float(num_of_requests)
     end
-    
+
     # When enabled, disables keep-alives on in-bound connections.
     # This option is the best way to ensure requests are handled by the server
     # in the order in which they are received. Disabling keep-alives means Puma
@@ -1279,7 +1279,7 @@ module Puma
     #
     # @example
     #   disable_keep_alives true
-    #   
+    #
     def disable_keep_alives(enabled=false)
       @options[:disable_keep_alives] = enabled
     end
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -1269,14 +1269,25 @@ module Puma
       @options[:max_fast_inline] = Float(num_of_requests)
     end
 
-    # When true, keep-alives are maintained on in-bound connections. This is the default.
-    # When false, the connection is closed by Puma after each request, requiring that
-    # the client open a new request.
-    # Setting to false is the best way to ensure requests are handled by the server
-    # in the order in which they are received. There is a TCP operations cost for both
-    # the client and server each time a new connection is opened/closed.
-    # This option will ensure fairness at the cost of potentially more time spent in TCP operations.
+    # When `true`, keep-alive connections are maintained on inbound requests. 
+    # Enabling this setting reduces the number of TCP operations, reducing response 
+    # times for connections that can send multiple requests in a single connection.
     #
+    # When Puma receives more incoming connections than available Puma threads, 
+    # enabling the keep-alive behavior may result in processing requests out-of-order, 
+    # increasing overall response time variance. Increased response time variance 
+    # means that the overall average of response times might not change, but more 
+    # outliers will exist. Those long-tail outliers may significantly affect response 
+    # times for some processed requests.
+    # 
+    # When `false`, Puma closes the connection after each request, requiring the 
+    # client to open a new request. Disabling this setting guarantees that requests 
+    # will be processed in the order they are fully received, decreasing response 
+    # variance and eliminating long-tail outliers caused by keep-alive behavior. 
+    # The trade-off is that the number of TCP operations required will increase.
+    #
+    # The default is +true+.
+    # 
     # @example
     #   enable_keep_alives false
     #
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -1269,19 +1269,19 @@ module Puma
       @options[:max_fast_inline] = Float(num_of_requests)
     end
 
-    # When enabled, disables keep-alives on in-bound connections.
-    # This option is the best way to ensure requests are handled by the server
-    # in the order in which they are received. Disabling keep-alives means Puma
-    # will close the connection after each request made, requiring that a new
-    # connection be opened. There is a TCP operations cost for both the client and server
-    # each time a new connection is opened/closed. This option will ensure fairness at the cost
-    # of potentially more time spent in TCP operations.
+    # When true, keep-alives are maintained on in-bound connections. This is the default.
+    # When false, the connection is closed by Puma after each request, requiring that
+    # the client open a new request.
+    # Setting to false is the best way to ensure requests are handled by the server
+    # in the order in which they are received. There is a TCP operations cost for both
+    # the client and server each time a new connection is opened/closed.
+    # This option will ensure fairness at the cost of potentially more time spent in TCP operations.
     #
     # @example
-    #   disable_keep_alives true
+    #   enable_keep_alives false
     #
-    def disable_keep_alives(enabled=false)
-      @options[:disable_keep_alives] = enabled
+    def enable_keep_alives(enabled=true)
+      @options[:enable_keep_alives] = enabled
     end
 
     # Specify the backend for the IO selector.
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -1281,7 +1281,7 @@ module Puma
     #   disable_keep_alives true
     #   
     def disable_keep_alives(enabled=false)
-      @options[:max_fast_inline] = enabled
+      @options[:disable_keep_alives] = enabled
     end
 
     # Specify the backend for the IO selector.
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -1269,21 +1269,21 @@ module Puma
       @options[:max_fast_inline] = Float(num_of_requests)
     end
 
-    # When `true`, keep-alive connections are maintained on inbound requests. 
-    # Enabling this setting reduces the number of TCP operations, reducing response 
+    # When `true`, keep-alive connections are maintained on inbound requests.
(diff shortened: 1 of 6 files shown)


schneems (changes requested):
Thank you for the PR. I like the feature and think we should have some way to disable this behavior consistently. I have two requested and one optional change.

## Config set to enable instead of disable.

Short:

- Request: Please invert the logic to `enable_keep_alives`
- Request: Please add the default value to the docs above the method as well

Long: 

I'm thinking of how users will read the value when they're looking at their `config/puma.rb` without docs at hand and what they think it will do. The current logic for calling this code:

```
disable_keep_alives
```

That code would set `@disable_keep_alives=false`which seems backwards. I think that's a small issue of flipping the default argument, however...

I'm also looking at other config examples like https://github.com/puma/puma/blob/9ef6f83e914f6e1a1e8c1e91efbf9e117e3eaf2c/lib/puma/dsl.rb#L472 that are boolean. Many use the argument of `which` and are framed as enabling some behavior. Such as `drain_on_shutdown` or `clean_thread_locals`. I'm also thinking of how this might be viewed from a user and I like representing the state positively instead of negatively:

```
enable_keep_alives
enable_keep_alives false
```

To me, the first line would enable keep alives (already the default) while the second would disable it. Comparing to

```
disable_keep_alives
disable_keep_alives false
```

The first line I would expect to disable the keepalives, but the next line presents a double negative that's a bit hard for my brain to parse. It also has the side benefit of being able to use it directly in the conditional without `&&= !<variable>`.

## Code style

Short: Optionally update the code style to the `if/else` example below

Long:

I played with the style a bit and I found the `&&=` not entirely intuitive. It's used in other places in the file, but is not common. I tried this, but didn't love it:

```ruby
 force_keep_alive = @enable_keep_alives && (
   requests < @max_fast_inline || 
   @thread_pool.busy_threads < @max_threads || 
   !client.listener.to_io.wait_readable(0)
 )
```

That code was tight, but hard to parse. I think this is more verbose, but clearer

```ruby
force_keep_alive = if @enable_keep_alives
  requests < @max_fast_inline ||
    @thread_pool.busy_threads < @max_threads ||
    !client.listener.to_io.wait_readable(0)
else
  false
end
```

Another option is:

```ruby
 force_keep_alive = @enable_keep_alives
 force_keep_alive &&= 
   requests < @max_fast_inline || 
   @thread_pool.busy_threads < @max_threads || 
   !client.listener.to_io.wait_readable(0)
```

My main worry is that someone might miss the trailing `&&=` in the existing PR when reading. It's a bit of a bike-shed though. I don't love any of them, but the if/else is my least bad pick. Especially if we start adding in comments, it could increase the distance between the effects. I also want to update the comments that describe the code, but that should be in a new PR and not cloud up this one.

slizco:
@schneems I've updated this according to your suggestions. Let know your thoughts!

test/test_puma_server.rb
```diff
@@ -753,6 +753,40 @@ def test_http_11_close_without_body
...
+
+    req  = "GET / HTTP/1.1\r\n\r\n"
+    response = send_http_read_response req
+
+    # No "Connection: close" header.
+    assert_equal ["Content-Type: plain/text", "Content-Length: 6"], response.headers
+    assert_equal "hello\n", response.body
+  end
+
+
```
schneems, line 767:
```suggestion

```

test/test_puma_server.rb
```diff
@@ -753,6 +753,40 @@ def test_http_11_close_without_body
...
+    assert_equal "hello\n", response.body
+  end
+
+  def test_http_11_enable_keep_alives_false
+    server_run(enable_keep_alives: false) { [200, {"Content-Type" => "plain/text"}, ["hello\n"]] }
+
+    req  = "GET / HTTP/1.1\r\n\r\n"
+    response = send_http_read_response req
+
+    # No "Connection: close" header.
```
schneems, line 785:
```suggestion
    # Assert the "Connection: close" header is present with keep-alives disabled.
```

lib/puma/dsl.rb
```diff
@@ -1269,6 +1269,21 @@ def max_fast_inline(num_of_requests)
...
+    # When false, the connection is closed by Puma after each request, requiring that
+    # the client open a new request.
+    # Setting to false is the best way to ensure requests are handled by the server
+    # in the order in which they are received. There is a TCP operations cost for both
+    # the client and server each time a new connection is opened/closed.
+    # This option will ensure fairness at the cost of potentially more time spent in TCP operations.
+    #
+    # @example
+    #   enable_keep_alives false
+    #
```
schneems, line 1282:
```suggestion
    # When `true`, keep-alive connections are maintained on inbound requests. 
    # Enabling this setting reduces the number of TCP operations, reducing response 
    # times for connections that can send multiple requests in a single connection.
    #
    # When Puma receives more incoming connections than available Puma threads, 
    # enabling the keep-alive behavior may result in processing requests out-of-order, 
    # increasing overall response time variance. Increased response time variance 
    # means that the overall average of response times might not change, but more 
    # outliers will exist. Those long-tail outliers may significantly affect response 
    # times for some processed requests.
    # 
    # When `false`, Puma closes the connection after each request, requiring the 
    # client to open a new request. Disabling this setting guarantees that requests 
    # will be processed in the order they are fully received, decreasing response 
    # variance and eliminating long-tail outliers caused by keep-alive behavior. 
    # The trade-off is that the number of TCP operations required will increase.
    #
    # The default is +true+.
    # 
    # @example
    #   enable_keep_alives false
    #
```

schneems (approved):
Overall looks great. I made some suggestions to comments/docs I would like to apply. I'm approving now and will apply those changes and merge once the suite passes.

Thanks a ton for working on this!

dentarg:
Some lint fixes needed: https://github.com/puma/puma/actions/runs/11038443413/job/30661661652?pr=3496

lib/puma/dsl.rb
```diff
@@ -1269,6 +1269,32 @@ def max_fast_inline(num_of_requests)
...
+    # client to open a new request. Disabling this setting guarantees that requests 
+    # will be processed in the order they are fully received, decreasing response 
+    # variance and eliminating long-tail outliers caused by keep-alive behavior. 
+    # The trade-off is that the number of TCP operations required will increase.
+    #
+    # The default is +true+.
+    # 
+    # @example
+    #   enable_keep_alives false
+    #
```
schneems, line 1293:
```suggestion
    # When `true`, keep-alive connections are maintained on inbound requests. 
    # Enabling this setting reduces the number of TCP operations, reducing response 
    # times for connections that can send multiple requests in a single connection.
    #
    # When Puma receives more incoming connections than available Puma threads, 
    # enabling the keep-alive behavior may result in processing requests out-of-order, 
    # increasing overall response time variance. Increased response time variance 
    # means that the overall average of response times might not change, but more 
    # outliers will exist. Those long-tail outliers may significantly affect response 
    # times for some processed requests.
    # 
    # When `false`, Puma closes the connection after each request, requiring the 
    # client to open a new request. Disabling this setting guarantees that requests 
    # will be processed in the order they are fully received, decreasing response 
    # variance and eliminating long-tail outliers caused by keep-alive behavior. 
    # The trade-off is that the number of TCP operations required will increase.
    #
    # The default is +true+.
    # 
    # @example
    #   enable_keep_alives false
    #
```

schneems, line 1293:
```suggestion
    # When `true`, keep-alive connections are maintained on inbound requests.
    # Enabling this setting reduces the number of TCP operations, reducing response
    # times for connections that can send multiple requests in a single connection.
    #
    # When Puma receives more incoming connections than available Puma threads,
    # enabling the keep-alive behavior may result in processing requests out-of-order,
    # increasing overall response time variance. Increased response time variance
    # means that the overall average of response times might not change, but more
    # outliers will exist. Those long-tail outliers may significantly affect response
    # times for some processed requests.
    # 
    # When `false`, Puma closes the connection after each request, requiring the
    # client to open a new request. Disabling this setting guarantees that requests
    # will be processed in the order they are fully received, decreasing response
    # variance and eliminating long-tail outliers caused by keep-alive behavior.
    # The trade-off is that the number of TCP operations required will increase.
    #
    # The default is +true+.
    # 
    # @example
    #   enable_keep_alives false
    #
```

schneems:
@dentarg thanks. I fixed up the trailing whitespace, now I'm seeing failures in CI

```
Prepare all required actions
Getting action download info
Download action repository 'actions/checkout@v4' (SHA:692973e3d937129bcbf40652eb9f2f61becf3332)
Download action repository 'ruby/setup-ruby@v1' (SHA:f3[21](https://github.com/puma/puma/actions/runs/11039500154/job/30665340740#step:1:25)cf5a4d1533575411f8752cf25b86478b0442)
Warning: Failed to download action 'https://internal-api.service.iad.github.net/repos/ruby/setup-ruby/tarball/f321cf5a4d1533575411f8752cf25b86478b0442'. Error: Name or service not known (internal-api.service.iad.github.net:443)
Warning: Back off 12.968 seconds before retry.
Warning: Failed to download action 'https://internal-api.service.iad.github.net/repos/ruby/setup-ruby/tarball/f321cf5a4d1533575411f8752cf25b86478b0442'. Error: Name or service not known (internal-api.service.iad.github.net:443)
Warning: Back off 18.971 seconds before retry.
Error: Name or service not known (internal-api.service.iad.github.net:443)
```

Not seeing anything on GitHub's status site

lib/puma/dsl.rb
```diff
@@ -1269,6 +1269,32 @@ def max_fast_inline(num_of_requests)
...
+    # client to open a new request. Disabling this setting guarantees that requests
+    # will be processed in the order they are fully received, decreasing response
+    # variance and eliminating long-tail outliers caused by keep-alive behavior.
+    # The trade-off is that the number of TCP operations required will increase.
+    #
+    # The default is +true+.
+    # 
+    # @example
+    #   enable_keep_alives false
+    #
```
schneems, line 1293:
```suggestion
    # When `true`, keep-alive connections are maintained on inbound requests.
    # Enabling this setting reduces the number of TCP operations, reducing response
    # times for connections that can send multiple requests in a single connection.
    #
    # When Puma receives more incoming connections than available Puma threads,
    # enabling the keep-alive behavior may result in processing requests out-of-order,
    # increasing overall response time variance. Increased response time variance
    # means that the overall average of response times might not change, but more
    # outliers will exist. Those long-tail outliers may significantly affect response
    # times for some processed requests.
    #
    # When `false`, Puma closes the connection after each request, requiring the
    # client to open a new request. Disabling this setting guarantees that requests
    # will be processed in the order they are fully received, decreasing response
    # variance and eliminating long-tail outliers caused by keep-alive behavior.
    # The trade-off is that the number of TCP operations required will increase.
    #
    # The default is +true+.
    #
    # @example
    #   enable_keep_alives false
    #
```

schneems:
Third time was the charm. CI passed, merging.

slizco:
Thanks @schneems and @dentarg!

PR #3214, opened 2023-08-27
null_io.rb - add external_encoding, set_encoding, binmode, binmode?

MSP-Greg:
### Description

See https://github.com/rack/rack/pull/2115, which identifies a bug that occurs with `Rack::MockRequest.env_for`.

See also [Rack Spec: The Input Stream](https://msp-greg.github.io/rack/file.SPEC.html#label-The+Input+Stream).

More IO instance methods may need to be added to `Puma::NullIO`.

### Your checklist for this pull request
- [x] I have reviewed the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.
- [x] I have added (or updated) appropriate tests if this PR fixes a bug or adds a feature.
- [x] My pull request is 100 lines added/removed or less so that it can be easily reviewed.
- [ ] If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.
- [ ] If this closes any issues, I have added "Closes `#issue`" to the PR description or my commit messages.
- [ ] I have updated the documentation accordingly.
- [x] All new and existing tests passed, including Rubocop.

PR #3500, opened 2024-09-22
[CI] test/helpers/puma_socket.rb - framework updates

MSP-Greg:
### Description

This is all the commits from #3479 that include `test/helpers/test_puma/puma_socket.rb`

Note that one large change is subclassing Ruby's socket classes.  This was done to remove adding methods to every client socket instance that was created in the test suite.

One central concept of the new framework is that tests never create a client socket/request from a native Ruby socket.  They should always be created from methods provided in the framework.  This ultimately will allow one to easily change (for example) a tcp socket test to an ssl test with no more than one additional method call, switching both the server and the client.


MSP-Greg:
I just pushed an update which split the puma_socket.rb file into two files.  Also, updated some of the comments.

Re the comments in general, see https://msp-greg.github.io/puma_test/

It's an unpublished link to the branch used in this PR.

MSP-Greg:
After a rebase, problems with non-MRI Rubies.

MSP-Greg:
I rebased again after
null_io.rb - add `external_encoding`, `set_encoding`, `binmode`, `binmode?` (#3214)](https://github.com/puma/puma/commit/36c730760743), which contains a few test fixes since the last rebase.

MSP-Greg:
I'm looking at this again, with the intent to get it updated and merged.

Currently, there are several methods that can be used to create a socket, send a request, and return a `TestPuma::Response` or parts of it.

[#send_http_read_response](https://msp-greg.github.io/puma_test/TestPuma/PumaSocket.html#send_http_read_response-instance_method) - sends a request and returns the whole response

[#send_http_read_resp_body](https://msp-greg.github.io/puma_test/TestPuma/PumaSocket.html#send_http_read_resp_body-instance_method) - sends a request and returns the response body

[#send_http_read_resp_headers](https://msp-greg.github.io/puma_test/TestPuma/PumaSocket.html#send_http_read_resp_headers-instance_method) - sends a request and returns the response with the body removed as an array of lines

Personally, I don't like `send_http_read_resp_body` and `send_http_read_resp_headers`.  I prefer to change then to `send_http_read_body` and `send_http_read_headers`.  Any thoughts?

Also, to keep things simple, I'm thinking of adding all of the above to sockets.  Tests often send a request, wait for the response, then send another request on the same socket.  Using the same method names would make things easier.

MSP-Greg:
Lastly, a road map of the test changes I'd like to see.

1. Convert all tests to using the shared code for creating client sockets.
2. Add common code for creating Puma instances, whether that be in-process (`Puma::Server.new`) or a spawned server.
3. Convert all test files to use the common code for servers.
4. At this stage, tests should be portable and easily reorganized to make it easier to find similar tests, and where to place new ones.

Currently, test files use one or the other of the two server types, but not both.  Originally, this test update assumed the same.  I'm looking at whether we could have test files that used both types of servers.

nateberkopec:
#3873

Issue #3503, opened 2024-09-26
Support gRPC in puma `fork_worker` mode


chjasonwu:
**Is your feature request related to a problem? Please describe.**
For context, gRPC has recently added support for client-side forking, but it requires an explicit gRPC API call to signal backend threads to join before process forking. Furthermore, gRPC mandates that forking occur in the same process and thread where it was initialized. This is impossible with the current `fork_worker` mode, since worker-0 fork the child process in a new thread. https://github.com/puma/puma/discussions/3480. 

**Describe the solution you'd like**
I'd like to be able to utilize gRPC in `fork_worker` mode by having worker-0 to fork child processes in the same master thread. 

**Describe alternatives you've considered**
I didn't have other alternatives since gRPC implementation limits child process to be forked from the same thread as master process.

**Additional context**

dentarg:
Steps in https://github.com/puma/puma/issues/3213 not enough?

chjasonwu:
It's not enough. gRPC requires child process forked from the same thread of the parent process.

chjasonwu:
@wjordan  just for context, is there a specific reason why you decide to spawn a separate thread to fork child processes?
https://github.com/puma/puma/commit/613ac8e803474907ca61e5baccf6795ce1e4723d#diff-5e5a0776d5ae9287ce767e62486bbd520d16eee0e6dc3eb04576fb847efd2bc2R292-R309

wjordan:
No specific reason I recall (that was four years ago so I don't remember many details), I think it was just simplest to implement it that way at the time.

chjasonwu:
Hi @MSP-Greg @dentarg @wjordan,  

I've added Ruby gRPC support for refork mode, including a mutex for synchronization. Any feedback would be greatly 
appreciated! 🙏 

https://github.com/puma/puma/pull/3504

chjasonwu:
Hi Puma contributors,

I hope you’re doing well. I wanted to follow up on our previous discussion regarding `refork_mode`. I've added gRPC support for `refork_mode` and implemented a mutex for worker orchestration to prevent race conditions during forking. These changes are essential for enabling `refork_mode` in GCP, and they’re also critical steps toward optimizing the performance of my system.

If possible, could you provide a timeline or outline the process for when this might be reviewed? I’d truly appreciate any guidance you can share, as this work is key to advancing our project goals. Thank you so much for your time and consideration.

cc @MSP-Greg @dentarg @wjordan,@nateberkopec,@ytaben

[...]

PR #3504, opened 2024-09-27
Move fork_worker to parent thread

chjasonwu:
### Description
I'm shifting worker forking back to the same thread as `worker-0` for `refork` mode. This change is necessary to ensure compatibility with gRPC, which restricts forking in a different thread than its parent process.

Discussion reference: https://github.com/puma/puma/discussions/3480
Issue: https://github.com/puma/puma/issues/3503

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -81,7 +81,7 @@ module Puma
           pid = spawn_worker(idx, master)
         end
 
-        debug "Spawned worker: #{pid}"
+        log "Spawned worker: #{pid}, phase: #{@phase}"
         @workers << WorkerHandle.new(idx, pid, @phase, @options)
       end
 
--- a/lib/puma/cluster/worker.rb
+++ b/lib/puma/cluster/worker.rb
@@ -177,8 +177,10 @@ module Puma
           while (widx = new_workers.pop(non_block=true))
             log "spawn_sub_workers #{widx}\n"
             worker_pids << pid = spawn_worker(widx)
-            log "f#{pid}:#{widx}\n"
-            @worker_write << "#{Puma::Const::PipeRequest::FORK}#{pid}:#{idx}\n" rescue nil
+            # log "f#{pid}:#{widx}\n"
+            msg = "#{Puma::Const::PipeRequest::FORK}#{pid}:#{widx}\n"
+            log msg
+            @worker_write << msg rescue nil
           end
         rescue ThreadError
           log "queue is empty"
@@ -189,6 +191,7 @@ module Puma
 
 
       def spawn_worker(idx)
+        log "spawning new worker from worker-0: #{idx}"
         @config.run_hooks(:before_worker_fork, idx, @log_writer, @hook_data)
 
         pid = fork do
@@ -200,6 +203,8 @@ module Puma
                                   server: @server
           new_worker.run
         end
+        log "new worker spawned from worker-0: with pid #{pid}"
+
 
         if !pid
           log "! Complete inability to spawn new workers detected"
--- a/lib/puma/cluster/worker.rb
+++ b/lib/puma/cluster/worker.rb
@@ -56,8 +56,9 @@ module Puma
         # things in shape before booting the app.
         @config.run_hooks(:before_worker_boot, index, @log_writer, @hook_data)
 
+        puts "@server.nil?: #{@server.nil?}\n"
         begin
-        server = @server ||= start_server
+          server = @server ||= start_server
         rescue Exception => e
           log "! Unable to start worker"
           log e
@@ -69,9 +70,12 @@ module Puma
 
         fork_worker = @options[:fork_worker] && index == 0
 
+        # worker ids for validte hang process
+        new_workers = Queue.new
+        worker_pids = []
+
         if fork_worker
           restart_server.clear
-          worker_pids = []
           Signal.trap "SIGCHLD" do
             wakeup! if worker_pids.reject! do |p|
               Process.wait(p, Process::WNOHANG) rescue true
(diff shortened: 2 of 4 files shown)


chjasonwu:
@dentarg @MSP-Greg I kinda can't tell if these tests are failing due to regression or just being flaky, since retry in my github actions seems to yield different results. https://github.com/chjasonwu/puma/actions/runs/11151795195

MSP-Greg:
@chjasonwu

First of all, I haven't had time to review this PR.

But, the current Puma test suite may have intermittent failures, I'll re-run the failing tests.

dentarg:
Marked this as something for Puma v7 as it sounds like a change in behaviour – feel free to share any thoughts on that

chjasonwu:
@MSP-Greg @dentarg @wjordan 

I plan to move child process forking back to the same thread handling requests. Any guidance or feedback would be greatly appreciated.

chjasonwu:
Previous failure is related to this https://github.com/puma/puma/pull/3532. I'll have to rebase.

chjasonwu:
@MSP-Greg @dentarg @wjordan 

I hadn’t heard back, so I just wanted to follow up. Any feedback you could share would be really appreciated.

dentarg:
Maybe @toregeschliman can review this?

lib/puma/cluster/worker.rb
```diff
@@ -57,21 +57,26 @@ def run
...
-        restart_server = Queue.new << true << false
+        restart_server = Queue.new << Puma::Const::WorkerCmd::RESTART << Puma::Const::WorkerCmd::STOPPED
 
         fork_worker = @options[:fork_worker] && index == 0
 
+        # worker ids for validte hang process
+        worker_pids = []
+
+        # create a mutex to synchronize the server thread
+        mutex = Mutex.new
```
toregeschliman, line 76: I'm not clear on what this buys us; nothing about the new state in the queue seems to require synchronizing around this and it messes up the signal trap hooks pretty badly. Which race condition were you worried about here?

lib/puma/cluster/worker.rb
```diff
@@ -57,21 +57,26 @@ def run
...
+          server = @server ||= start_server
         rescue Exception => e
           log "! Unable to start worker"
           log e
           log e.backtrace.join("\n    ")
           exit 1
         end
 
-        restart_server = Queue.new << true << false
+        restart_server = Queue.new << Puma::Const::WorkerCmd::RESTART << Puma::Const::WorkerCmd::STOPPED
```
toregeschliman, line 68: We should probably rename the queue if it's no longer just booleans

lib/puma/cluster/worker.rb
```diff
@@ -84,25 +89,34 @@ def run
...
                 end
               elsif idx == 0 # restart server
-                restart_server << true << false
-              else # fork worker
-                worker_pids << pid = spawn_worker(idx)
-                @worker_write << "#{Puma::Const::PipeRequest::FORK}#{pid}:#{idx}\n" rescue nil
+                restart_server << Puma::Const::WorkerCmd::RESTART << Puma::Const::WorkerCmd::STOPPED
+              else
+                # spawn new worker
+                restart_server << "#{Puma::Const::WorkerCmd::SPAWN}#{idx}"
```
toregeschliman, line 103: We should definitely pass the idx state somewhere else, this is allocating multiple strings every loop for objects that are otherwise interned/intern-able.

toregeschliman (commented):
I like this in principle (keeping `Thread.main` as `Thread.main` solves a set of forking problems we've run into with `fork_worker`) but I'm not clear on the purpose of the new locking and I always prefer alternatives to avoid the common `Thread.new` solution for signal handlers. If we're going to go that way we might as well just stop using the Queue since it's really just a threadsafe tool that's usable in signal handler context.

lib/puma/cluster/worker.rb
toregeschliman: There's probably a narrower fix for CI, given the unlikelihood of the race condition in the real world; most setups are likely going to send a second signal if it doesn't terminate in time, and nobody restarts worker zero as frequently or quickly as tests do (ideally). As mentioned above if we think we need the mutex, we should change the restart_server queue to a simpler state machine, but I think the easier fix is to continue to avoid the mutex and add a second TERM to the specs.

toregeschliman:
This is close but is also likely to be superseded by upcoming work to convert worker-0 to a non-serving parent process (or "mold"), which I'm hoping to start work on in the next week. I won't block on that certainly since it's not yet in progress.

nateberkopec:
@toregeschliman I assume the idea there is that worker-0 will be killed once the next child worker is running, right?

I'm not sure that this PR really is v7 or a breaking change. I prefer to merge something like this rather than wait for a PR which is not yet opened.

nateberkopec (changes requested):
I'm OK with merging this if @toregeschliman's comments are addressed.

lib/puma/cluster.rb
```diff
@@ -81,7 +81,7 @@ def spawn_workers
           pid = spawn_worker(idx, master)
         end
 
-        debug "Spawned worker: #{pid}"
+        log "Spawned worker: #{pid}, phase: #{@phase}"
```
joshuay03, line 84: There's already a log line on worker boot: https://github.com/chjasonwu/puma/blob/38724d4666aac0980ebd0779fd33ddef1c3ec468/lib/puma/cluster.rb#L477

Do we need one for spawn as well? If we are adding it, should the formatting match the boot log to include the indentation and index?
```suggestion
        log "- Worker #{idx} (PID: #{pid}) spawned, phase: #{@phase}"
```

<br>

IMO this does seem like more of a debug-only log, but that's for the maintainers to decide.

test/test_integration_cluster.rb
```diff
@@ -472,12 +472,13 @@ def test_culling_strategy_oldest
   end
 
   def test_culling_strategy_oldest_fork_worker
-    cli_server "-w 2 test/rackup/hello.ru", config: <<~CONFIG
+    cli_server "-w 2 test/rackup/hello.ru", puma_debug: true, config: <<~CONFIG
```
joshuay03, line 475: Need to remove this before merge.

toregeschliman:
@nateberkopec started work at https://github.com/puma/puma/pull/3619/files on this; rough idea is just for worker zero to stop handling requests whenever it's asked to do its first refork (after initialization) and move to a fork_worker loop on the main thread listening for further signals; this reduces the risk of restarting for OOM, timeouts, or other things that can interrupt it while handling fork_worker state. I plan to extend it eventually with generational molding, which could include terminating legacy molds, but we'll still be keeping the active mold live to refork any workers >= 1 that terminate for other reasons.

When it's done I want to test it for a while internally before taking it out of draft so I still don't consider it blocking on this PR.

chjasonwu:
@nateberkopec, I'm closing this and will leveraging Tore's mold_worker instead. https://github.com/puma/puma/issues/3596.

The mold_worker implementation will not spawn a different thread for worker spawning, which will comply with gRPC. It's probably a better way to moving forward, instead of me patching this again instead.

PR #3505, opened 2024-09-28
Fix multi-delimiter split to get status app token

AnthonyClark:
### Description
Originally added back in https://github.com/puma/puma/commit/a91d64a560743d130ec0930640aa364ab187ce36 - I believe the intention is to support splitting on either `&` or `;` between query params.

If nobody has run into a problem with this maybe it can be simplified down to only support the single param, but as I stumbled across it reading through the code I thought I'd propose this change as something simple.

~I did hit a CI error due to https://github.com/hotwired/turbo-rails/issues/681 in the [ubuntu-20.04 Ruby 2.7 Rails 6.1](https://github.com/AnthonyClark/puma/actions/runs/11085382537/job/30801738684#logs) run~ Heh, of course I was just behind on the rebase and it was already taken care of.

### DIFF
--- a/lib/puma/app/status.rb
+++ b/lib/puma/app/status.rb
@@ -80,7 +80,7 @@ module Puma
 
       def authenticate(env)
         return true unless @auth_token
-        env['QUERY_STRING'].to_s.split('&;').include? "token=#{@auth_token}"
+        env['QUERY_STRING'].to_s.split(/[&;]/).include? "token=#{@auth_token}"
       end
 
       def rack_response(status, body, content_type='application/json')
--- a/test/test_app_status.rb
+++ b/test/test_app_status.rb
@@ -54,6 +54,14 @@ class TestAppStatus < Minitest::Test
     assert_equal 404, status
   end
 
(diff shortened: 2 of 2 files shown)


MSP-Greg:
Re the CI failure, maybe rebase on current master?

> If nobody has run into a problem with this

A quick look seemed to show that `Puma::App::Status` only supports one query parameter (`token`), so not sure about this?

AnthonyClark:
>only supports one query parameter (`token`), so not sure about this?

Yeah, I think there's the argument to just remove the `split` logic - only support the single param with an exact match check on the query param path to be explicit that's what is supported. Would that be the preferred version? (maybe throwing in a secure compare?)

dentarg:
That could be a breaking change if people for some reason had other things in the query string. I agree with using secure comparison but I think we can afford to properly parse the query string, and extract the token value, this isn't some hot path, right?

MSP-Greg:
I just checked, and the added test fails in master.

MSP-Greg (approved):

PR #3517, opened 2024-10-11
Add busy threads stat

jjb:
### Description

adds busy_threads to stats and also DRYs related code

the DRYing code (worker_handle.rb and worker.rb) can go into another PR or removed if preferred

### DIFF
--- a/lib/puma/cluster/worker.rb
+++ b/lib/puma/cluster/worker.rb
@@ -126,12 +126,8 @@ module Puma
 
             while true
               begin
-                b = server.backlog || 0
-                r = server.running || 0
-                t = server.pool_capacity || 0
-                m = server.max_threads || 0
-                rc = server.requests_count || 0
-                payload = %Q!#{base_payload}{ "backlog":#{b}, "running":#{r}, "pool_capacity":#{t}, "max_threads":#{m}, "requests_count":#{rc} }\n!
+                stat_string = Puma::Server::STAT_METHODS.map { |s| "\"#{s}\":#{server.public_send(s) || 0}" }.join(', ')
+                payload = %Q!#{base_payload}{ #{stat_string} }\n!
                 io << payload
               rescue IOError
                 Puma::Util.purge_interrupt_queue
--- a/lib/puma/cluster/worker.rb
+++ b/lib/puma/cluster/worker.rb
@@ -131,7 +131,8 @@ module Puma
                 t = server.pool_capacity || 0
                 m = server.max_threads || 0
                 rc = server.requests_count || 0
-                payload = %Q!#{base_payload}{ "backlog":#{b}, "running":#{r}, "pool_capacity":#{t}, "max_threads":#{m}, "requests_count":#{rc} }\n!
+                bt = server.busy_threads || 0
+                payload = %Q!#{base_payload}{ "backlog":#{b}, "running":#{r}, "pool_capacity":#{t}, "max_threads":#{m}, "requests_count":#{rc}, "busy_threads":#{bt} }\n!
                 io << payload
               rescue IOError
                 Puma::Util.purge_interrupt_queue
--- a/lib/puma/cluster/worker.rb
+++ b/lib/puma/cluster/worker.rb
@@ -126,8 +126,12 @@ module Puma
 
             while true
               begin
-                stat_string = Puma::Server::STAT_METHODS.map { |s| "\"#{s}\":#{server.public_send(s) || 0}" }.join(', ')
-                payload = %Q!#{base_payload}{ #{stat_string} }\n!
+                b = server.backlog || 0
+                r = server.running || 0
+                t = server.pool_capacity || 0
+                m = server.max_threads || 0
+                rc = server.requests_count || 0
+                payload = %Q!#{base_payload}{ "backlog":#{b}, "running":#{r}, "pool_capacity":#{t}, "max_threads":#{m}, "requests_count":#{rc} }\n!
                 io << payload
               rescue IOError
                 Puma::Util.purge_interrupt_queue
--- a/lib/puma/cluster/worker_handle.rb
+++ b/lib/puma/cluster/worker_handle.rb
@@ -51,7 +51,7 @@ module Puma
         @term
       end
 
-      STATUS_PATTERN = /{ "backlog":(?<backlog>\d*), "running":(?<running>\d*), "pool_capacity":(?<pool_capacity>\d*), "max_threads":(?<max_threads>\d*), "requests_count":(?<requests_count>\d*) }/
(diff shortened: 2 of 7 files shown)


lib/puma/cluster/worker.rb
```diff
@@ -126,12 +126,8 @@ def run
 
             while true
               begin
-                b = server.backlog || 0
-                r = server.running || 0
-                t = server.pool_capacity || 0
-                m = server.max_threads || 0
-                rc = server.requests_count || 0
-                payload = %Q!#{base_payload}{ "backlog":#{b}, "running":#{r}, "pool_capacity":#{t}, "max_threads":#{m}, "requests_count":#{rc} }\n!
+                stat_string = Puma::Server::STAT_METHODS.map { |s| "\"#{s}\":#{server.public_send(s) || 0}" }.join(', ')
```
schneems, line 129: Please revert this logic. Calling `Array#map` allocates a new array https://ruby-doc.org/3.3.5/Array.html#method-i-map. And then N String allocations instead of a single string allocation. Calling public send is slower than calling the method directly.

It's not a wild amount of overhead, but I want calling stats to be as low overhead as possible.

lib/puma/cluster/worker_handle.rb
```diff
@@ -51,7 +51,9 @@ def term?
         @term
       end
 
-      STATUS_PATTERN = /{ "backlog":(?<backlog>\d*), "running":(?<running>\d*), "pool_capacity":(?<pool_capacity>\d*), "max_threads":(?<max_threads>\d*), "requests_count":(?<requests_count>\d*) }/
+      regex_string = Puma::Server::STAT_METHODS.map { |s| "\"#{s}\":(?<#{s}>\\d*)"  }.join(', ')
```
schneems, line 54: I like this code, it's much cleaner and it's not on the hot path. 

However, I think this will also have to be reverted if you change the above method of stats collection unless we want to normalize the order of the values (such as alphabetical). I'm assuming we have tests that exercise this status pattern.

lib/puma/server.rb
```diff
@@ -233,6 +233,11 @@ def pool_capacity
       @thread_pool&.pool_capacity
     end
 
+    # @!attribute [r] busy_threads
+    def busy_threads
+      @thread_pool&.busy_threads
```
schneems, line 238: This call locks the mutex:

```
    # @!attribute [r] busy_threads
    # @version 5.0.0
    def busy_threads
      with_mutex { @spawned - @waiting + @todo.size }
    end
```

I'm wondering (aloud) if we can reduce the requirement that this be 100% accurate at the sake of reducing locking contention. As-is, there is a race condition in the other metrics we're collecting.

schneems (changes requested):
I like the idea. I didn't actually realize `running` wasn't this metric until you mentioned it in slack chat.

I've made a perf change request. I also ask that you update the `doc/stats.md` to document this new feature.

I made a comment about a locking call to the thread pool. I would like to think more on it or hear more thoughts on it before figuring out how I think we should move. Basically: What do we think our guarantees are or should be for these metrics regarding threadsafety and race conditions?

lib/puma/cluster/worker.rb
jjb: counterpoint: calling stats typically happens every 30 or 60 seconds, so the performance impact is irrelevant

🎭🪞counterpoint counterpoint: new stats are added very rarely, so the maintainability win of DRYing this is irrelevant

lib/puma/cluster/worker_handle.rb
jjb: sounds good

fwiw the particular order is already necessary because worker.rb and worker_handler.rb have a producer/consumer relationship

lib/puma/server.rb
jjb: do you mean, don't lock only the case of stats collection? or in all cases?

## if you mean don't lock only for stats collection

is that to avoid adding overhead caused by stats collection? i'm pretty sure it's not relevant, since this method is also frequently called in business logic (see bellow). but yeah, sacrificing 100% accurate is definitely acceptable for this stat AFAIK

## if you mean don't lock ever

busy_threads is used to make business logic decisions about work scheduling elsewhere in the code, so i believe in that case the mutex is necessary

jjb:
my attempt at documentation here: https://github.com/puma/puma/pull/3517/commits/b2cc4beec65d04b14ec6407a5ecb4d4f7e7cfd84

please note that I have now written most of this document, starting in https://github.com/puma/puma/pull/2525 when i created it. this is, frankly, ridiculous 😅 - someone with a better understanding of this stuff should spruce is up and fill in the `?`

lib/puma/cluster/worker.rb
jjb: reverted

lib/puma/cluster/worker_handle.rb
jjb: reverted

lib/puma/server.rb
MSP-Greg: I've also wondered about the locking code in ThreadPool.  Changing some of the variables to Queues might allow some simplification.  That work was interrupted.

At present, `busy_threads` is used in:

https://github.com/puma/puma/blob/a59afabe782ec8869bb709ca991b32c0b2cc95fd/lib/puma/request.rb#L169

With the recent focus on keep-alive long-tail issues, that may be removed.  In the meantime, maybe it's best to leave the locking place?

jjb:
i propose that improvements to stat locking should go into another PR, since status quo is already inconsistent locking for stats

dentarg:
> i propose that improvements to stat locking should go into another PR, since status quo is already inconsistent locking for stats

Not sure about that, don't we want stats we can trust?

jjb:
> Not sure about that, don't we want stats we can trust?

I’m saying, this PR doesn’t introduce new phenomenon or behavior - the stats in master already don’t use the same mutex-controlled timeslice. putting all stats in the same mutex could be done in a PR before or after this one (I could put it together if interested).

jjb:
@dentarg WDYT?

MSP-Greg:
@jjb

I'm assuming that #3527 addresses @schneems issue with multiple locking calls.  My question to you is if #3527 and #3528 are merged/accepted, are you ok with rebasing and adjusting this to work with the updates?

Or, we could flip the order and merge this first.  Not sure which is easier 'collectively'...

jjb:
definitely discuss and merge the other PRs first

lib/puma/server.rb
jjb: https://github.com/puma/puma/pull/3527

jjb:
@schneems @dentarg @MSP-Greg ready! i'll clean up the commits after discussion

MSP-Greg:
@jjb

> definitely discuss and merge the other PRs first

I've only merged the 'lib' PR, but I can update the 'test' for this.  LGTM, I'll wait a bit to see if anyone else looks at it.  Thanks.

MSP-Greg (approved):

MSP-Greg:
@schneems You ok with this as is?

PR #3527, opened 2024-10-17
Add ThreadPool#stats and adjust Server#stats to use it

MSP-Greg:
### Description

Many of the variables in ThreadPool are locked when changing their values.  Current stats code may enter that lock more than once. See https://github.com/puma/puma/pull/3517#discussion_r1799652829.

This PR creates a `ThreadPool#stats` method that returns a hash of the stats data that comes from ThreadPool, using a single lock.

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -651,8 +651,12 @@ module Puma
     # Returns a hash of stats about the running server for reporting purposes.
     # @version 5.0.0
     # @!attribute [r] stats
+    # @return [Hash] hash containing stat info from `Server` and `ThreadPool`
     def stats
-      STAT_METHODS.map {|name| [name, send(name) || 0]}.to_h
+      stats = @thread_pool&.stats || {}
+      stats[:max_threads]    = @max_threads
+      stats[:requests_count] = @requests_count
+      stats
     end
 
     # below are 'delegations' to binder
--- a/lib/puma/thread_pool.rb
+++ b/lib/puma/thread_pool.rb
@@ -85,6 +85,17 @@ module Puma
       end
     end
 
+    # generate stats hash so as not to perform multiple locks
+    # @return [Hash] hash containing stat info from ThreadPool
+    def stats
+      with_mutex do
+        { backlog: @todo.size,
+          running: @spawned,
+          pool_capacity: @waiting + (@max - @spawned)
+        }
+      end
+    end
+
     # How many objects have yet to be processed by the pool?
     #
     def backlog


MSP-Greg:
A few questions:

There are several methods that can be removed with this.  Should we mark then as deprecated as of Puma 7?

Several of current 'stats' tests check the generated 'stats' string.  These have long, hard to read match strings/regexes.  Are we concerned about the string representation, or should we be checking them as objects returned via `JSON.parse`?

In addition to #3517, I'd like to add two more items to stats, `backlog_max`, which would be then maximum size of the ThreadPool `@todo` array, and also something like `selector_size_max`, which would contain the maximum number of clients/sockets contained in the Reactor.  Another question about these is whether they should be reset after every 'stats' call.

These look at Puma behavior when it may see 'bursts' of traffic, which is the main cause of the `long-tail` issue...

lib/puma/thread_pool.rb
```diff
@@ -85,6 +85,17 @@ def self.clean_thread_locals
...
 
+    # generate stats hash so as not to perform multiple locks
+    # @return [Hash] hash containing stat info from ThreadPool
+    def stats
+      with_mutex do
+        { backlog: @todo.size,
+          running: @spawned,
+          pool_capacity: @waiting + (@max - @spawned)
+        }
+      end
```
jjb, line 96: here's the approach i was thinking, in order to avoid repeating code (naming can change of course). WDYT?

```ruby
def backlog_logic
  @todo.size
end
def backlog
  with_mutext { backlog_logic }
end
def stats
  with_mutex do
    { backlog: backlog_logic,
      ...
    }
  end
end
```

MSP-Greg: Actually, the three methods exist only so `Server#stats` can obtain the information.  There is one exception, and that may be removed with work currently being looked at to help with the 'long tail' issue.

So, thanks for the example.  As I mentioned, they may be deprecated and removed in Puma 7.

jjb: ahh gotcha, i thought others were used for business logic like busy_threads

lib/puma/server.rb
```diff
@@ -648,11 +648,15 @@ def shutting_down?
     # @version 5.0.0
     STAT_METHODS = [:backlog, :running, :pool_capacity, :max_threads, :requests_count].freeze
 
+
```
schneems, line 651:
```suggestion
```

schneems (approved):
Seems good. Asked to remove a single extra whitespace line.

lib/puma/server.rb
MSP-Greg: Note to self - always look at changes before committing...

Fixed

jjb (commented):
concern: does existing test suite really test the behavior of `stats` method fully after this PR? could we, in this PR, remove the mutex from `backlog`, invoke the method from `stats` instead of copy/pasting the logic there, and keep things like that? (possibly moving the two methods into private in the future)


if we don't do that, suggestion: add comments above the methods which are now never invoked by puma code, indicating this, maybe linking to this PR, maybe saying maybe they should be deprecated in the future (or can the deprecation decision be made now in this PR?)

MSP-Greg:
> concern: does existing test suite really test the behavior of stats method fully after this PR?

As mentioned in PR #3528, until all the stats hash values are unique, not really.  The purpose of this code was to simplify gathering the metrics from ThreadPool, so Server isn't doing them in multiple method calls, currently with more than one lock.

Re issues like `backlog`, keeping code 'dry' is important, but calling single line methods is not a good idea, especially considering the future with YJIT, etc.  When the issue comes up, I prefer comments.

Re all these stat related methods, the 'long-tail' issue may result in changes as to what metrics are used by Puma for request routing/processing.  Hence, my opinion is to leave as is, and fix up for Puma v7.

jjb:
Sorry, I should have been more clear (and should have checked first) - I assumed that pool_capacity, running, and backlog all had unit tests in addition to the (more like) integration tests in #3528, and worried that those would be deleted when those methods are retired, so we would lose the "behavior" tests - but looking now I see that there are no such tests (`test_correct_waiting_count_for_killed_threads` and `test_append_queues_on_max` use `backlog` but aren't specifically testing it.) so there isn't that risk in removing those methods eventually (can just swap in `@todo.size` for those two).

In any case - I think this PR is ready to go! (unless you want to add comments to, or even get the ball rolling on deprecating, those 3 methods in this PR.)

lib/puma/server.rb
```diff
@@ -651,8 +651,12 @@ def shutting_down?
     # Returns a hash of stats about the running server for reporting purposes.
     # @version 5.0.0
     # @!attribute [r] stats
+    # @return [Hash] hash containing stat info from `Server` and `ThreadPool`
     def stats
-      STAT_METHODS.map {|name| [name, send(name) || 0]}.to_h
```
jjb, line 655: STAT_METHODS is now only referenced from the test suite

PR #3528, opened 2024-10-19
[CI] Update stats tests

MSP-Greg:
### Description

A few of the stats tests have very long strings/regex's used with assert statements.  These make it difficult to update the tests, and also are quite a pita to work with when a test fails.

The docs state that "[Puma.stats](https://msp-greg.github.io/puma/Puma.html#stats-class_method) produces a JSON string, and the same is returned via `Puma::ControlCLI`.  JSON strings are just that, so we can change the text representation as long as they parse with the same key/value pairs.

Updates the tests to work with 'expectation' hashes, rather than 'expectation' strings/regex's.  Hopefully, others will feel that the updated tests are easier to follow.

Adds a helper method (`assert_hash`) to `test/helpers/test_puma/assertions.rb`

### DIFF
--- a/test/test_cli.rb
+++ b/test/test_cli.rb
@@ -46,6 +46,50 @@ class TestCLI < Minitest::Test
     @ready.close
   end
 
+  def check_single_stats(body, check_puma_stats = true)
+    http_hash = JSON.parse body.split("\n", 2).last
+
+    dmt = Puma::Configuration::DEFAULTS[:max_threads]
+
+    expected_single_root_keys = {
+      'started_at' => RE_8601,
+      'backlog'    => 0,
+      'running'    => 0,
+      'pool_capacity'  => dmt,
+      'busy_threads'   => 0,
+      'max_threads'    => dmt,
+      'requests_count' => 0,
+      'versions'       => Hash,
+    }
+
+    assert_hash expected_single_root_keys, http_hash
+
+    #version keys
+    expected_version_hash = {
+      'puma' => Puma::Const::VERSION,
+      'ruby' => Hash,
+    }
+    assert_hash expected_version_hash, http_hash['versions']
+
+    #version ruby keys
+    expected_version_ruby_hash = {
+      'engine'     => RUBY_ENGINE,
+      'version'    => RUBY_VERSION,
+      'patchlevel' => RUBY_PATCHLEVEL,
+    }
+
+    assert_hash expected_version_ruby_hash, http_hash['versions']['ruby']
+
+    if check_puma_stats
+      puma_stats_hash = JSON.parse(Puma.stats)
+      assert_hash expected_single_root_keys, puma_stats_hash
+      assert_hash expected_version_hash, puma_stats_hash['versions']
+      assert_hash expected_version_ruby_hash, puma_stats_hash['versions']['ruby']
+
+      assert_equal Puma.stats_hash, JSON.parse(Puma.stats, symbolize_names: true)
+    end
+  end
+
   def test_control_for_tcp
     control_port = UniquePort.call
     url = "tcp://127.0.0.1:#{control_port}/"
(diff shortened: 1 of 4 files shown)


MSP-Greg:
@jjb

Since you've been in the stats test code recently, wondering what your thoughts are.

test/test_cli.rb
```diff
@@ -46,6 +46,49 @@ def teardown
     @ready.close
   end
 
+  def check_single_stats(body, check_puma_stats = true)
+    http_hash = JSON.parse body.split("\n", 2).last
```
jjb, line 50: i might do this in two lines with variable names that indicate what the split and last achieve

test/test_integration_pumactl.rb
```diff
@@ -197,27 +197,80 @@ def test_clustered_stats
     skip_unless :fork
     skip_unless :unix
 
-    puma_version_pattern = "\\d+.\\d+.\\d+(\\.[a-z\\d]+)?"
+    min_threads = 1
+    max_threads = 2
```
jjb, line 201: is this (intentionally?) changing the behavior of the test?

maybe do that in a different standalone PR so the behavior vs. style change is clear

jjb (commented):
very nice! couple notes.

> JSON strings are just that, so we can change the text representation as long as they parse with the same key/value pairs

does this refer to the changes in this PR or to future possible changes in app code? the ordering in the app code is hardcoded in worker.rb + worker_handle.rb, and it's parsed without the json library in order to avoid making it a dependency. probably already on your radar but mentioning in case relevant.

test/test_integration_pumactl.rb
MSP-Greg: Yes, current code had the following:
```ruby
'running' => 2,
'pool_capacity'  => 2,
'max_threads'    => 2,
```

PR code has:
```ruby
'running' => 1,
'pool_capacity'  => 2,
'max_threads'    => 2,
```

So, 'running' no longer has the same value as 'max_threads'.  Changing 'pool_capacity' is a bit messier in CI...

test/test_cli.rb
MSP-Greg: Actually, that isn't needed.  My mistake.  Removed.

The tests in `test_integration_pumactl.rb` use `Puma::ControlCLI` which outputs the JSON string to `@stdout`, but prefaces it with a line "Command stats sent success".

`test_cli.rb` sends a simple request, and only the JSON string is returned.  I was thinking it also had the line that needed to be stripped...

MSP-Greg:
> does this refer to the changes in this PR or to future possible changes in app code?

No changes to lib files in this PR, only test files.  So, the JSON strings have not changed.

My point was that it's a JSON string (IIRC, JSON allows white-space characters in various places), but our tests were treating it as a 'standard' string.  One of the expectation regex strings was almost 370 characters.

By parsing the JSON string, the tests can compare hashes, and are more readable and maintainable.

I pushed another commit fixing the above.

test/test_integration_pumactl.rb
jjb: gotcha - why the change?

test/test_cli.rb
jjb: nice

is it valuable to also test the value of `first` as well?

test/test_integration_pumactl.rb
MSP-Greg: > why the change?

So fewer of the hash values have the same value.  Current code has three with the same value of 2. After the PR, only two values are identical.  Really, all hash values need to be unique for the tests to verify that they're all correct.

Off topic, but with some planned test suite updates, we should be able to have a test file something similar to `test_stats`, where all the 'full' stats tests would be located.

test/test_cli.rb
MSP-Greg: There's a few lines in `test_integration_pumactl.rb` that check it. The code in `puma/control_cli.rb` is:

https://github.com/puma/puma/blob/36c7307607432416d9a58eef317757da1eceb95a/lib/puma/control_cli.rb#L225-L226

MSP-Greg:
@jjb

> the ordering in the app code is hardcoded in worker.rb + worker_handle.rb, and it's parsed without the json library in order to avoid making it a dependency. probably already on your radar but mentioning in case relevant.

Just noticed the above.  Just to clarify, JSON is loaded in the test process, and has been for a few years.  There is a test that spawns (`IO.popen`) a Puma process and checks that JSON is not loaded.  The sub-process is needed to isolate the Puma instance from the test environment.

https://github.com/puma/puma/blob/36c7307607432416d9a58eef317757da1eceb95a/test/test_integration_cluster.rb#L386-L390

Also, the code in `Puma::JSONSerialization` only contains a public `generate` method, there is no JSON parsing in the Puma lib files.

MSP-Greg:
Just rebased

PR #3151, opened 2023-05-12
Clarify Reactor#initialize

nateberkopec:

PR #3542, opened 2024-11-02
Fix reactor initialization calling wrong JRuby Selector initialization

ning-github:
https://github.com/puma/puma/pull/3151 broke JRuby selector initialization via the following:

```
NIO::Selector.backends.delete(backend) => nil
```

What results is that `::NIO::Selector.new(nil)` is not the same as `::NIO::Selector.new`.

The former invokes
https://github.com/socketry/nio4r/blob/v2.5.2/ext/nio4r/org/nio4r/Selector.java#L47, which results in an exception being thrown.

This PR will restore the reactor initialization to call the intended https://github.com/socketry/nio4r/blob/v2.5.2/ext/nio4r/org/nio4r/Selector.java#L41

### Description

### DIFF
--- a/lib/puma/reactor.rb
+++ b/lib/puma/reactor.rb
@@ -24,8 +24,12 @@ module Puma
       unless valid_backends.include?(backend)
         raise ArgumentError.new("unsupported IO selector backend: #{backend} (available backends: #{valid_backends.join(', ')})")
       end
-
-      @selector = ::NIO::Selector.new(NIO::Selector.backends.delete(backend))
+      delete_output = NIO::Selector.backends.delete(backend)
+      if delete_output.nil?
+        @selector = ::NIO::Selector.new
+      else
+        @selector = ::NIO::Selector.new(delete_output)
+      end
       @input = Queue.new
       @timeouts = []
       @block = block
--- a/lib/puma/reactor.rb
+++ b/lib/puma/reactor.rb
@@ -24,12 +24,8 @@ module Puma
       unless valid_backends.include?(backend)
         raise ArgumentError.new("unsupported IO selector backend: #{backend} (available backends: #{valid_backends.join(', ')})")
       end
-      delete_output = NIO::Selector.backends.delete(backend)
-      if delete_output.nil?
-        @selector = ::NIO::Selector.new
-      else
-        @selector = ::NIO::Selector.new(delete_output)
-      end
+
+      @selector = ::NIO::Selector.new(NIO::Selector.backends.delete(backend))
       @input = Queue.new
       @timeouts = []
       @block = block
--- a/puma.gemspec
+++ b/puma.gemspec
@@ -14,7 +14,7 @@ Gem::Specification.new do |s|
   s.email = ["evan@phx.io"]
   s.executables = ["puma", "pumactl"]
   s.extensions = ["ext/puma_http11/extconf.rb"]
-  s.add_runtime_dependency "nio4r", "~> 2.0"
+  s.add_runtime_dependency "nio4r", "~> 2.5.5"
   if RbConfig::CONFIG['ruby_version'] >= '2.5'
     s.metadata["msys2_mingw_dependencies"] = "openssl"
   end
--- a/test/test_reactor.rb
+++ b/test/test_reactor.rb
@@ -1,7 +1,7 @@
 require 'puma/reactor'
 require_relative "helper"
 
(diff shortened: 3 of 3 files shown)


dentarg:
Can you add a test that catches the regression from #3151 ?

And solve this in a way that doesn't change the behavior for other platforms :) (see the failing test)

ning-github:
Added a test to assert on proper initialization.

However, something peculiar about this behavior is that I'm only able to reproduce this issue in certain environments.

On my puma fork, I observe:
```
jruby-9.3.15.0 :001 > require 'nio'
 => true
jruby-9.3.15.0 :002 > NIO::Selector.new(nil)
 => #<NIO::Selector:0x43756cb>
```

However, for my actual application, I see the failure:
```
irb(main):001:0> require 'nio'
=> true
irb(main):002:0> NIO::Selector.new(nil)
       10: from /nix/store/zdb3yw8slcmxa8gdw8099c339kjdn2gb-jruby-9.3.15.0/bin/irb:25:in `<main>'
        9: from org/jruby/RubyKernel.java:1053:in `load'
        8: from /nix/store/zdb3yw8slcmxa8gdw8099c339kjdn2gb-jruby-9.3.15.0/lib/ruby/gems/shared/gems/irb-1.0.0/exe/irb:11:in `<main>'
        7: from org/jruby/RubyKernel.java:1238:in `catch'
        6: from org/jruby/RubyKernel.java:1238:in `catch'
        5: from org/jruby/RubyKernel.java:1508:in `loop'
        4: from org/jruby/RubyKernel.java:1092:in `eval'
        3: from (irb):2:in `evaluate'
        2: from org/jruby/RubyClass.java:890:in `new'
        1: from org/nio4r/Selector.java:49:in `initialize'
ArgumentError (:java is the only supported backend)

```

ning-github:
FWIW, this is handled on the nio4r side as of v2.5.5 https://github.com/socketry/nio4r/blob/v2.5.5/ext/nio4r/org/nio4r/Selector.java#L46

dentarg:
> However, something peculiar about this behavior is that I'm only able to reproduce this issue in certain environments.

Did you figure out the reason for this?

ning-github:
> > However, something peculiar about this behavior is that I'm only able to reproduce this issue in certain environments.
> 
> Did you figure out the reason for this?

Yes, I figured out and it's actually ended up being quite simple :)
The only reason it was working on my local clone is because installation defaulted to the most recent version of `nio4r`, which is `2.7.4`, which means that it already included the NIO-side handling for this specific issue that was added in `2.5.5`.

When I pin the version [here](https://github.com/puma/puma/blob/v6.4.3/Gemfile#L8) in Puma's Gemfile to something below `2.5.5`, for example `2.5.2`, I see the `ArgumentError` that I'm making this change to address.

In summary, unless [this](https://github.com/puma/puma/blob/v6.4.3/Gemfile#L8) is bumped to declare that `2.5.5` is the oldest compatible version of `nio4r`, then this PR's change should be needed on the Puma side.

dentarg:
Thanks for figuring that out! 

> In summary, unless [this](https://github.com/puma/puma/blob/v6.4.3/Gemfile#L8) is bumped to declare that `2.5.5` is the oldest compatible version of `nio4r`

It is in the in the gemspec the bump should be https://github.com/puma/puma/blob/271e4d0b697aabb022de1daa4a6f425f416cfafe/puma.gemspec#L17

Seeing 2.5.5 have been available for soon 4 years (see the below snippet from https://rubygems.org/gems/nio4r/versions)

```
2.5.5 February 05, 2021 (119 KB)
2.5.5 February 05, 2021 java (132 KB)
```

...this sounds like a pretty rare edge-case – I don't think we should do anything in Puma

I assume you can update nio4r in your application?

dentarg:
I guess adjusting the gemspec doesn't hurt and make things technically more correct, so I wouldn't reject it.

ning-github:
> I assume you can update nio4r in your application?

Yup, this is what we ended up doing.

> It is in the in the gemspec the bump should be

I finally got around to doing this, but it looks like `test_changing_nio4r_version_during_phased_restart` is now failing. Regardless, the point about this being a rare edge case is reasonable, and I'm fine either way with continuing to try to get this adjustment merged or just closing it at this point.

nateberkopec:
Thanks for following up @ning-github - unless someone else wants to pipe in, I think we're good to say this is edge-casey enough to avoid making changes.

Issue #3540, opened 2024-10-31
http_content_length_limit causes Puma::HttpParserError


id-ilych:
**Describe the bug**
Hello Puma devs :wave:

I have a form with a file as one of the inputs, submitted as `multipart/form-data`.
It works just fine with various files, but when I added `http_content_length_limit` to the puma config, then for the small files it works, but if the file exceeds the given limit, then I see following in the logs
```
HTTP parse error, malformed request: #<Puma::HttpParserError: Invalid HTTP format, parsing fails. Are you trying to open an SSL connection to a non-SSL Puma?>
```
Note that I am running it on localhost (`http://localhost:3000/my_feature`)

To get more information, I've set up `lowlevel_error_handler` and figured the error is coming from this line
https://github.com/puma/puma/blob/5fc43d73b6ff193325e657a24ed76dec79133e93/lib/puma/client.rb#L268
with following values for some of the instance variables
```
{
    "parsed_bytes": 0,
    "read_header": true,
    "read_proxy": false,
    "ready": false,
    "body_read_start": 566077.315509,
    "http_content_length_limit": 5242880,
    "http_content_length_limit_exceeded": false
}
```

**Puma config:**

```
max_threads_count = ENV.fetch('RAILS_MAX_THREADS', 5)
min_threads_count = ENV.fetch('RAILS_MIN_THREADS') { max_threads_count }
threads min_threads_count, max_threads_count

worker_timeout 3600 if ENV.fetch('RAILS_ENV', 'development') == 'development'

port ENV.fetch('PORT', 3000)

environment ENV.fetch('RAILS_ENV') { 'development' }

pidfile ENV.fetch('PIDFILE') { 'server.pid' }

http_content_length_limit(2 * 1024 * 1024)

lowlevel_error_handler do
  [500, {'Content-Type' => 'text/plain'}, ['error']]
end

plugin :tmp_restart
```

**To Reproduce**
Create files of different sizes
```
dd if=/dev/zero of=one_mb.pdf bs=1M count=1
dd if=/dev/zero of=three_mb.pdf bs=1M count=3
```

Create `hello.ru` with the following content
```
run lambda { |env| [200, {"Content-Type" => "text/plain"}, ["Hello World"]] }
```

Comment out `http_content_length_limit(2 * 1024 * 1024)` in the puma config (see above).

Run the server with `bundle exec puma -C puma.rb hello.ru`

Upload files
```
[...]

PR #3548, opened 2024-11-06
Request body - increase read size to 64 kB

MSP-Greg:
### Description

While looking into #3540, I noticed that the `read_nonblock` calls used for requests were either 4k or 16k.  For large request bodies, this causes a large number of loops to read them.

Using curl, I increased the value to 128k, and checked the bytes read.  They were 64k for both `content-length` and chunked bodies.

Also, the 16k value in `Puma::Const` dates back to at least 2007, and networks are much different now.

### DIFF
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -501,7 +501,7 @@ module Puma
     def read_chunked_body
       while true
         begin
-          chunk = @io.read_nonblock(4096, @read_buffer)
+          chunk = @io.read_nonblock(CHUNK_SIZE, @read_buffer)
         rescue IO::WaitReadable
           return false
         rescue SystemCallError, IOError
--- a/lib/puma/const.rb
+++ b/lib/puma/const.rb
@@ -137,7 +137,7 @@ module Puma
     }.freeze
 
     # The basic max request size we'll try to read.
-    CHUNK_SIZE = 16 * 1024
+    CHUNK_SIZE = 64 * 1024
 
     # This is the maximum header that is allowed before a client is booted.  The parser detects
     # this, but we'd also like to do this as well.


dentarg:
I added the "perf" label here because this is more efficient, right?

How did you check with curl? If you don't mind sharing, I don't think I fully followed.

MSP-Greg:
> If you don't mind sharing

Certainly.  I was debating whether to comment out some statements, but instead removed them.

One needs to have a large file, then add debug statements that output the `bytesize` of what is read by the `@io.read_nonblock` statements in client.rb

As to curl, something like
```text
curl --form "data=@<large file path>" http://127.0.0.1:<port>/
curl --form "data=@<large file path>" --header "Transfer-Encoding: chunked" http://127.0.0.1:<port>/
```

So, one could take this PR, and up the `CHUNK_SIZE` to 128 kB, and you should see approx 64 kB as the limit for what is read.

MSP-Greg:
@dentarg

Try the branch at https://github.com/MSP-Greg/puma/tree/00-lrg-request-body-debug

I used a 4mB file generated with
```ruby
# Writes 4 mB file
txt64kb = "*" * (64 * 1_024)
File.open(<file path>, 'w+') { |f|
  64.times { f.syswrite txt64kb }
}
```

Then set
```text
$ export PUMA_DEBUG_CLIENT_READ=true
```

You can then run either of the following:
```text
test/runner -v test_puma_server_err.rb
```

or:
```text
bundle exec bin/puma -b tcp://127.0.0.1:40001 -t1:1 test/rackup/hello.ru
```
and, in another console:
```text
curl --form "data=@< 4mB file path >" http://127.0.0.1:40001/
curl --form "data=@< 4mB file path >" --header "Transfer-Encoding: chunked" http://127.0.0.1:40001/
```

The branch sets `CHUNK_SIZE` to 128 kB, mots of the bytesize logging on my system shows approx 64 kB.

MSP-Greg:
@dentarg

Finally got around to trying this on macOS Sonoma.  Using it with curl, it seems to send 128k packets, or that's what Puma reads.  As noteds above, WSL2/Ubuntu 24.04 sends 64k packets.

Hence, I think the update to 64k is okay.  Have you had a chance to look at this?

dentarg:
I have tested now, thanks for explaining! :)

dentarg:
Should we benchmark this? Can we?

dentarg:
I did this:

This PR running on port 8080

```shell
~/src/dup/puma 00-req-body-chunk-size
$ echo 'app { |env| [200, {}, ["OK"]] }' | ruby -Ilib ./bin/puma --config /dev/stdin --port 8080
* Puma version: 6.4.3 ("The Eagle of Durango")
* Ruby version: ruby 3.3.5 (2024-09-03 revision ef084cc8f4) [arm64-darwin23]
*  Min threads: 0
*  Max threads: 5
*  Environment: development
*          PID: 68485
* Listening on http://0.0.0.0:8080
Use Ctrl-C to stop
```

created a big file

```shell
$ dd if=/dev/zero of=bigfile bs=4k count=100000
...
$ ls -ahl bigfile
-rw-r--r--@ 1 dentarg  staff   391M Nov 21 22:54 bigfile
```

```shell
$ for i in {1..10} ; do curl --form "data=@bigfile" --url http://127.0.0.1:8080 --write-out " time_total: %{time_total}\n"; done
OK time_total: 0.003273
OK time_total: 0.003007
OK time_total: 0.002918
OK time_total: 0.002590
OK time_total: 0.002704
OK time_total: 0.004606
OK time_total: 0.002739
OK time_total: 0.002599
OK time_total: 0.002232
OK time_total: 0.002265
```

master (932f5d09f57ce8ec3632156148f42577e92eb4dd) running on port 8181

```shell
~/src/puma master
$ echo 'app { |env| [200, {}, ["OK"]] }' | ruby -Ilib ./bin/puma --config /dev/stdin --port 8181
* Puma version: 6.4.3 ("The Eagle of Durango")
* Ruby version: ruby 3.3.5 (2024-09-03 revision ef084cc8f4) [arm64-darwin23]
*  Min threads: 0
*  Max threads: 5
*  Environment: development
*          PID: 62559
* Listening on http://0.0.0.0:8181
Use Ctrl-C to stop
```

```shell
$ for i in {1..10} ; do curl --form "data=@bigfile" --url http://127.0.0.1:8181 --write-out " time_total: %{time_total}\n"; done
OK time_total: 0.374278
OK time_total: 0.306575
OK time_total: 0.300101
OK time_total: 0.280538
OK time_total: 0.288668
OK time_total: 0.300438
OK time_total: 0.294658
OK time_total: 0.301145
OK time_total: 0.292926
OK time_total: 0.292424
```

That's pretty good.

dentarg (approved):
Seems to make a lot of sense

MSP-Greg:
@dentarg

Thanks for testing.  I didn't check speed.  I knew it would be faster, and less 'gc pressure'.  But, two orders of magnitude is a **big jump**.

Real world networks may not have that much improvement, but they'll certainly be faster.  Also, many sites don't have large request bodies...

mcfadden:
I think this is a great change, but wanted to mention a gotcha that we ran into after this rolled out.

We run nginx in front of Puma. At some point in time we learned that responses with large headers would cause errors from nginx:

`upstream sent too big header while reading response header from upstream`

After that error, nginx would also stop using that upstream connection for a few seconds.

Ultimately, we tracked it down to responses with headers over 8k, and eventually we figured out that some of the buffer settings in nginx defaulted to 8k, but Puma would send chunks that were 16k. When the buffers didn't match, things got mad really fast. Matching the puma and nginx buffer sizes solved it!

At least, until we recently started seeing these errors again for some requests.. This week one of my coworkers noted that this value changed in this PR. We had to update our nginx settings to match the new value as well.

If you run nginx > puma you may want to take a look at your nginx config too. Specifically, the values I know to check are `proxy_buffers`, `proxy_buffer_size`, `proxy_busy_buffers_size`, and `large_client_header_buffers`.

The large requests were typically requests that were fuzzing with giant query strings, but we also saw it with some out of control preload links headers.

PR #3557, opened 2024-11-18
CI: refactor restart_does_not_drop_connections code

MSP-Greg:
### Description

There are frequent intermittent failures and retries with the shared restart code used in the clustered and single integration tests.  Also, the phased/USR1 code used a different set of code that wasn't as rigorous as the hot/USR2 code.

Updated code.

Also, refactored the assert code.  Locally, with some hacks to Puma, I forced both `Errno::ECONNRESET` and EOFError in client connections.  When tried with various browsers, they all very quickly retried the connections.

Unfortunately, even though retries are much less common with these changes, other tests started failing...

### DIFF
--- a/test/helpers/integration.rb
+++ b/test/helpers/integration.rb
@@ -506,12 +506,12 @@ class TestIntegration < Minitest::Test
           begin
             get_worker_pids phase, log: log
             # added sleep as locally 165 restarts in 7 seconds
-            sleep 0.2
+            sleep 0.15
           rescue Minitest::Assertion # Timeout
             run = false
           end
         else
-          sleep 0.25
+          sleep 0.10
         end
       end
     end
@@ -519,7 +519,7 @@ class TestIntegration < Minitest::Test
     # cycle thru threads rather than one at a time
     until client_threads.empty?
       client_threads.each_with_index do |t, i|
-        client_threads[i] = nil if t.join(2)
+        client_threads[i] = nil if t.join(1)
       end
       client_threads.compact!
     end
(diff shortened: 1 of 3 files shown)


test/helpers/integration.rb
```diff
@@ -507,30 +538,21 @@ def hot_restart_does_not_drop_connections(num_threads: 1, total_requests: 500)
     msg << "   %4d refused\n"               % replies.fetch(:refused,0)
     msg << "   %4d read timeout\n"          % replies.fetch(:read_timeout,0)
     msg << "   %4d reset\n"                 % replies.fetch(:reset,0)
+    msg << "   %4d write_errors\n"          % replies.fetch(:write_error,0)
     msg << "   %4d success\n"               % replies.fetch(:success,0)
     msg << "   %4d success after restart\n" % replies.fetch(:restart,0)
     msg << "   %4d restart count\n"         % restart_count
 
+    actual_requests = num_threads * num_requests
+    allowed_errors = (actual_requests * 0.002).round
```
dentarg, line 547: So 0.2%. 

Looks like Windows allowed 40 errors before and non-Window 1%

We test with between 300 and 3000 requests => we allow 6-60 requests with errors – this is for non-Windows. For Windows we allow any amount of reset and refused requests as long as there's no other errors

Did I get this right?

dentarg (approved):
This is probably OK :) What a beast test this `restart_does_not_drop_connections` is :-) 150+ lines. Good work 👍

test/helpers/integration.rb
MSP-Greg: > 300 and 3000 requests => we allow 6-60 requests

Are you off by a factor of ten?  I think it should be:

300 and 3000 requests => we allow 1-6 requests

Re Windows...

First of all, yes, I have used MSFT OS's forever (since DOS days).  But, I don't spend much time testing Puma on Windows. If the tests pass, I'm good.  There have been a lot of changes to test code, so the Windows limits might be able to be lowered.  Not high on my list right now...

test/test_integration_cluster.rb
```diff
@@ -21,11 +21,33 @@ def teardown
...
+  end
+
+  def test_phased_restart_does_not_drop_connections
+    restart_does_not_drop_connections num_threads: 1, total_requests: 1_000,
+      signal: :USR1
+  end
+
+  def test_phased_restart_does_not_drop_connections_threads_fork_worker
+    restart_does_not_drop_connections num_threads: 10, total_requests: 3_000,
+      signal: :USR1 #, config: 'fork_worker', log: true
```
joshuay03, line 45: Was this intentionally commented out? It seems like fork worker should be explicitly enabled here.

joshuay03: The test seems to time out when enabling `fork_worker` here, see https://github.com/puma/puma/pull/3297#issuecomment-2509608503.

I think we should bump [this sleep](https://github.com/puma/puma/pull/3557/files#diff-c117a0d21645b3453eec0b51031b67f23b45ec4acebb44e4781f473c661fb6e6R509) a lot more. When running this test locally with `fork_worker` enabled a phased restart is triggered ~10 times, which seems a bit excessive/unrealistic for 3k requests. `fork_worker` then adds to that overhead cause worker 0 doesn't accept connections until the refork cycle is complete.

dentarg: I see you created https://github.com/puma/puma/pull/3589 to address the above 👍

PR #3559, opened 2024-11-21
feat: add requests_wait_time metric

dsalahutdinov:
### Description
Introduce new metric `requests_wait_time`, describing how many time requests are waiting to be processed (wait time to get to backlog [here](https://github.com/puma/puma/blob/master/lib/puma/server.rb#L363C1-L364C98).
```ruby
                pool.wait_until_not_full
                pool.wait_for_less_busy_worker(options[:wait_for_less_busy_worker]) if @clustered
```
and wait time in backlog.

In that way backlog value is not full representable to determine the lack of capacity (f.e. to use as a scaling factor for Horizontal Pod Autoscaler) . Because backlog might be equal to zero, but the wait time could be non-zero.

Here i've created some kind of stress-tests environment in yabeda-puma-plugin gem https://github.com/yabeda-rb/yabeda-puma-plugin/pull/33 representing such case (see image below)

puma-request_wait_time

This new metric also helps to tune up `wait_for_less_busy_worker` time by measuring and comparing "request reprocessing time" before and after config change (introduced here https://github.com/puma/puma/pull/2079)

### DIFF
--- a/docs/stats.md
+++ b/docs/stats.md
@@ -61,6 +61,7 @@ When Puma runs in single mode, these stats are available at the top level. When 
 * pool_capacity: the number of requests that the server is capable of taking right now. For example, if the number is 5, then it means there are 5 threads sitting idle ready to take a request. If one request comes in, then the value would be 4 until it finishes processing. If the minimum threads allowed is zero, this number will still have a maximum value of the maximum threads allowed.
 * max_threads: the maximum number of threads Puma is configured to spool per worker
 * requests_count: the number of requests this worker has served since starting
+* requests_wait_time: the total time of all requests are waiting to be processed (milliseconds)
 
 
 ### cluster mode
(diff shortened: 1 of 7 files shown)


nateberkopec:
This can't be used as a signal for autoscaling.

Puma can only observe the time between the process picked the request up off the socket and that request being processed by a worker thread. 

It cannot observe the time the request spent waiting to be picked up by Puma on the socket, so this feature will consistently under-report the actual time spent queueing.

This is why request queue monitoring usually involves [setting a header on the request at the load balancer](https://docs.newrelic.com/docs/apm/applications-menu/features/configure-request-queue-reporting/), and then using a Rack middleware to do `Time.now - convert_to_time(env['HTTP_X_REQUEST_START'])` 

Including a feature like this which does not capture the full time that a request spends queueing, and calling the feature `requests_wait_time`, will confuse users.

Instead, users should follow the approach of stamping requests with at time, as I said above. Puma doesn't need to be involved with reading that value because it works just as well to do it in a Rack middleware.

PR #3225, opened 2023-09-15
[Fix #2786] Worker 0 timing out during phased restart

joshuay03:
### Description

Closes https://github.com/puma/puma/issues/2786.

Ensures that worker 0 is pinged on every forked worker's post-boot ping to prevent it from timing out.

~An integration test for this would be something like https://github.com/puma/puma/issues/2786#issuecomment-1704766900 but might not be practical considering the time it takes. I'll add some unit tests when I get a chance unless anyone has a better idea.~

I've added a regression test with a lowered worker timeout to simulate a high worker count + high worker timeout real world scenario.

### Your checklist for this pull request

- [x] I have reviewed the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.
- [x] I have added (or updated) appropriate tests if this PR fixes a bug or adds a feature.
- [x] My pull request is 100 lines added/removed or less so that it can be easily reviewed.
- [x] If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.
- [x] If this closes any issues, I have added "Closes `#issue`" to the PR description or my commit messages.
- [x] I have updated the documentation accordingly.
- [x] All new and existing tests passed, including Rubocop.

PR #3568, opened 2024-12-01
Only ping worker 0 during phased restart if using fork worker

joshuay03:
### Description

The ping added in https://github.com/puma/puma/pull/3225 is only needed when forking workers from worker 0 during a phased restart (i.e., fork worker is enabled); however, it is currently called during an ordinary phased restart as well.

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -491,7 +491,7 @@ module Puma
                   w.ping!(status)
                   @events.fire(:ping!, w)
 
-                  if in_phased_restart && workers_not_booted.positive? && w0 = worker_at(0)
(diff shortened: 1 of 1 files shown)


joshuay03:
> Gah I forgot a refork can be triggered with `SIGURG`, so this is not a good solution for that case. I'll open a follow-up PR soon where we don't rely on the config option for this condition.

Ignore me, I assumed that a refork can be triggered at any time with `SIGURG`, even when `fork_worker` has not been configured (that's the impression I got from [the docs](https://github.com/puma/puma/blob/master/docs/fork_worker.md)). 

So this patch is fine as is i.e. it covers both refork on hitting configured request count and refork on receiving `SIGURG`. I've got a PR up to add a test that covers the the signal case.

PR #3572, opened 2024-12-05
Fix single runner stats before the server start

barthez:
It might happen the luncher stats are called before the server even starts resulting in NoMethodError (e.g. metrics plugin). This change prevents the exception and returns default stats only.

### Description

The metrics plugin might call `stats` on the launcher before the server is instantiated resulting in `NoMethodError`. 

https://github.com/harmjanblok/puma-metrics/issues/49

It seems like the plugin is not actively maintained since [the PR](https://github.com/harmjanblok/puma-metrics/pull/68) has not been responded to in a few months.

I see that `server` is already treated like `Maybe` type, so the change only aligns this one call with the reset of the class.

### DIFF
--- a/lib/puma/single.rb
+++ b/lib/puma/single.rb
@@ -17,7 +17,7 @@ module Puma
     def stats
       {
         started_at: utc_iso8601(@started_at)
-      }.merge(@server.stats).merge(super)
+      }.merge(@server&.stats || {}).merge(super)
     end
 
     def restart
--- a/test/test_launcher.rb
+++ b/test/test_launcher.rb
@@ -85,6 +85,7 @@ class TestLauncher < PumaTest
       sleep 1.1 unless Puma.mri?
       launcher.stop
     }
+    launcher.stats # Ensure `stats` method return without errors before run
     launcher.run
     sleep 1 unless Puma.mri?
     Puma::Server::STAT_METHODS.each do |stat|


barthez:
Do you mean regression tests? I guess we could add a check that `launcher.stats` works before `launcher.run` here: https://github.com/puma/puma/blob/5ac3f408008dd3496b251ce5b252014d82668198/test/test_launcher.rb#L79-L91

In our case, we are using `puma_metrics` plugin and k8s service monitor configured to get metrics endpoint. Sometimes it triggers the issue after deployment when the plugin puma server is ready to accept requests, but the app server is not.

barthez:
Updated the test case.

barthez:
Hey @dentarg , any chance to move this forward?

nateberkopec:
I'll take a look soon 👍

dentarg:
LGTM :) I'll let Nate take a look

dentarg:
On a second thought... I guess the tests aren't really provoking the no method error that was reported here? 

Is it hard to do so?

barthez:
This is the test run, when I revert the change in `single.rb`. What do you mean that the tests aren't provoking the error? 

```
$  bundle exec rake test TEST=test/test_launcher.rb
/usr/local/bin/ginstall -c -m 0755 puma_http11.bundle ../../../../lib/puma
Run options: --seed 65500

# Running:

.....E....

Fabulous run in 6.260454s, 1.5973 runs/s, 21.7237 assertions/s.
Errors & Failures:

  1) Error:
TestLauncher#test_puma_stats:
NoMethodError: undefined method `stats' for nil:NilClass
    lib/puma/single.rb:20:in `stats'
    lib/puma/launcher.rb:117:in `stats'
    test/test_launcher.rb:87:in `test_puma_stats'
    test/helper.rb:131:in `block (3 levels) in run'
    $HOME/.rbenv/versions/3.2.1/lib/ruby/3.2.0/timeout.rb:189:in `block in timeout'
    $HOME/.rbenv/versions/3.2.1/lib/ruby/3.2.0/timeout.rb:196:in `timeout'
    test/helper.rb:129:in `block (2 levels) in run'

10 runs, 136 assertions, 0 failures, 1 errors, 0 skips
rake aborted!
```

lib/puma/single.rb
```diff
@@ -17,7 +17,7 @@ class Single < Runner
     def stats
       {
         started_at: utc_iso8601(@started_at)
-      }.merge(@server.stats).merge(super)
```
nateberkopec, line 20: I'm such an old rubyist that safe navigation still feels "new" to me. I had to look up what ruby version this was added it to make sure we support it 💀

PR #3579, opened 2024-12-14
Add test for worker 0 ping with `SIGURG`

joshuay03:
### Description

Addresses https://github.com/puma/puma/pull/3568#issuecomment-2542731536

### DIFF
--- a/test/test_integration_cluster.rb
+++ b/test/test_integration_cluster.rb
@@ -374,11 +374,11 @@ class TestIntegrationCluster < TestIntegration
     assert_equal '0', read_body(connect)
   end
 
-  def test_fork_worker_phased_restart_with_high_worker_count
+  def test_phased_restart_with_fork_worker_and_high_worker_count
     worker_count = 10
 
     cli_server "test/rackup/hello.ru", config: <<~CONFIG
-      fork_worker 0
+      fork_worker
       worker_check_interval 1
       # lower worker timeout from default (60) to avoid test timeout
       worker_timeout 2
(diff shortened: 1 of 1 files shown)


test/test_integration_cluster.rb
```diff
@@ -387,14 +387,36 @@ def test_fork_worker_phased_restart_with_high_worker_count
...
-    # below is so all of @server_log isn't output for failure
+    # below is so all of @server_log isn't simply output on failure
+    refute @server_log[/.*Terminating timed out worker.*/]
+  end
+
+  def test_refork_phased_restart_with_fork_worker_and_high_worker_count
+    worker_count = 10
+
+    cli_server "test/rackup/hello.ru", config: <<~CONFIG
+      fork_worker 1000 # not going to be reached
```
dentarg, line 404: Is it important for the test that it isn't reached?

`fork_worker 0` is used in `test_phased_restart_with_fork_worker_and_high_worker_count` but I don't see any requests made, so the value doesn't matter,  right?

test/test_integration_cluster.rb
```diff
@@ -387,14 +387,36 @@ def test_fork_worker_phased_restart_with_high_worker_count
...
+      # to simulate worker 0 timeout, total boot time for all workers
+      # needs to exceed single worker timeout
+      workers #{worker_count}
+    CONFIG
+
+    get_worker_pids 0, worker_count
+
+    Process.kill :SIGURG, @pid
+
+    get_worker_pids 1, worker_count - 1
```
dentarg, line 417: I don't understand why we do `worker_count - 1`

test/test_integration_cluster.rb
```diff
@@ -387,14 +387,36 @@ def test_fork_worker_phased_restart_with_high_worker_count
...
+
+    # below is so all of @server_log isn't simply output on failure
     refute @server_log[/.*Terminating timed out worker.*/]
```
dentarg, line 420: All in all, I think this tests is good... we send `SIGURG` and we ensure no worker timed out by looking at the logs.

Just some details I don't understand. :)

test/test_integration_cluster.rb
joshuay03: Because for `fork_worker` with two workers, worker 0 remains as is and a new worker 1 is forked from it after which the old worker 1 is terminated. Since that logic piggy backs off of phased restart the logs indicate that worker 1 is now in phase 2. We don't see any logs for worker 0 as expected, hence the `-1`.

In the previous test however we just do a plain phased restart, which will restart all workers, worker 0 included.

dentarg (approved):

test/test_integration_cluster.rb
joshuay03: Fork worker example:
Screenshot 2025-01-10 at 11 26 57 pm

joshuay03: Phased restart example:
Screenshot 2025-01-10 at 11 26 57 pm

joshuay03: We should probably update the logs to mention fork worker and not phased restart on refork...

joshuay03:
Failure seems to be unrelated/flaky.

PR #3585, opened 2024-12-27
Add Ruby 3.4 to CI

joshuay03:
Explicitly run CI against Ruby 3.4 now that it's out and [`setup-ruby` has been updated](https://github.com/ruby/setup-ruby/commit/bfefad842bb982ff05b233bcbc1571d97a87e69f).

### DIFF
--- a/.github/workflows/rack_conform.yml
+++ b/.github/workflows/rack_conform.yml
@@ -25,6 +25,7 @@ jobs:
           - { os: ubuntu-20.04 , ruby: '3.1' }
           - { os: ubuntu-20.04 , ruby: '3.2' }
           - { os: ubuntu-22.04 , ruby: '3.3' }
+          - { os: ubuntu-22.04 , ruby: '3.4' }
           - { os: ubuntu-22.04 , ruby: head  }
 
     env:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -40,7 +40,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-20.04, ubuntu-22.04, ubuntu-24.04, macos-13, macos-14, macos-15, windows-2022 ]
-        ruby: [ 2.4, 2.5, 2.6, 2.7, '3.0', 3.1, 3.2, 3.3, head ]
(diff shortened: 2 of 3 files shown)


dentarg:
Maybe make the switch to Rails 8 for hotwired/turbo-rails in a separate PR? (I don't think what's going on, something about sqlite3 gem: https://github.com/puma/puma/actions/runs/12515093445/job/34912384008#step:6:21)

dentarg:
Probably this that doesn't hold true for Rails 8

https://github.com/puma/puma/blob/339f9cd4a15f643377d8d4b393b83d2985341b9c/.github/workflows/turbo-rails.yml#L54-L55

joshuay03:
> (I don't think what's going on, something about sqlite3 gem: [puma/puma/actions/runs/12515093445/job/34912384008#step:6:21](https://github.com/puma/puma/actions/runs/12515093445/job/34912384008#step:6:21))

Ya the requirement in Rails comes from [here](https://github.com/rails/rails/blob/c31cc0b2e451b9271bdfedc68884e8ae38a7c0fb/activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb#L14), I'll need to figure out where the already installed version is coming from.

joshuay03:
> Probably this that doesn't hold true for Rails 8
> 
> https://github.com/puma/puma/blob/339f9cd4a15f643377d8d4b393b83d2985341b9c/.github/workflows/turbo-rails.yml#L54-L55

Thanks, I've removed it, lets see if it helps.

joshuay03:
Okay, the Rails issue has been resolved. However, it seems like Ruby 3.4 isn't available yet on Windows via `setup-ruby`.

dentarg:
Can you exclude the `windows-2022 3.4` job for now? We want the CI to be green whenever possible.

MSP-Greg:
Commenting it out is fine.  It should be available sometime soon, and Ruby master probably hasn't diverged much from 3.4 yet...

JFYI, almost all GHA Ruby builds are generated by repos in the Ruby organization.  Except the Windows builds...

MSP-Greg:
1. Guess one can't comment out if it's part of the matrix.  Sorry for that.
2. Decided to re-run the one failed job in CI.

Thanks.

Issue #3478, opened 2024-09-03
libev failures in CI


dentarg:
From https://github.com/puma/puma/issues/2905#issuecomment-2227468877

MRI: ubuntu-24.04 head

```
...
TestPumaServer#test_chunked_request_pause_before_closing_cr_lf = 1.00 s = .
(libev) epoll_wait: Invalid argument
/home/runner/work/_temp/c4e73e0b-6fe9-41df-b70c-71f351b3bb11.sh: line 1:  3160 Aborted                 (core dumped) test/runner --verbose
TestPumaServer#test_default_server_port_respects_x_forwarded_proto = 
Error: Process completed with exit code 134.
```

From https://github.com/puma/puma/actions/runs/10687994837/job/29626863117#step:10:609

MRI: macos-14 3.1

```
TestPumaServerSSL#test_ssl_v3_rejection = 0.00 s = S
(libev) kqueue kevent: Bad file descriptor
/Users/runner/work/_temp/65a5407d-3d8f-4479-b210-297713ee86bc.sh: line 1:  3906 Abort trap: 6           test/runner --verbose
TestPumaServerSSL#test_very_large_return = 
```

dentarg:
From https://github.com/puma/puma/issues/2905#issuecomment-2227520800

> My guess is, something is closing the file descriptor while it's still in use in the event loop.

dentarg:
We just had this fail in CI: https://github.com/puma/puma/actions/runs/10688292588/job/29627986699#step:10:970

```
/home/runner/.rubies/ruby-head/lib/ruby/3.4.0+0/socket.rb:68:in 'Socket#connect': Errno::ENOENT: No such file or directory - connect(2) for /home/runner/work/_temp/puma_socket20240903-2914-q8h9mh (Puma::SdNotify::NotifyError)
	from /home/runner/.rubies/ruby-head/lib/ruby/3.4.0+0/socket.rb:68:in 'Addrinfo#connect_internal'
	from /home/runner/.rubies/ruby-head/lib/ruby/3.4.0+0/socket.rb:141:in 'Addrinfo#connect'
	from /home/runner/work/puma/puma/lib/puma/sd_notify.rb:140:in 'Puma::SdNotify.notify'
	from /home/runner/work/puma/puma/lib/puma/sd_notify.rb:57:in 'Puma::SdNotify.reloading'
	from /home/runner/work/puma/puma/lib/puma/plugin/systemd.rb:19:in 'block in start'
[...]

PR #3589, opened 2024-12-30
Reduce `TestIntegration#restart_does_not_drop_connections` restart frequency

joshuay03:
### Description

Context:
- https://github.com/puma/puma/pull/3297#issuecomment-2509608503
- https://github.com/puma/puma/pull/3557#discussion_r1864735186

With `fork_worker` un-commented and the original `sleep` the test times out for me locally, but doesn't on this branch. We go from ~10 restarts to ~3 restarts in a non-`fork_worker` test like [this one](https://github.com/puma/puma/blob/52757c8bf96bdc07044b7243c0851a46c06c5fec/test/test_integration_cluster.rb#L23-L26), which I think is perfectly reasonable for what these tests are asserting.

Fixing this is necessary to unblock https://github.com/puma/puma/pull/3297.

cc/ @MSP-Greg 

### DIFF
--- a/test/helpers/integration.rb
+++ b/test/helpers/integration.rb
@@ -485,7 +485,8 @@ class TestIntegration < Minitest::Test
     run = true
 
     restart_thread = Thread.new do
-      sleep 0.2  # let some connections in before 1st restart
+      # Wait for some connections before first restart
+      sleep 0.2
       while run
         if Puma.windows?
           cli_pumactl 'restart'
@@ -509,13 +510,13 @@ class TestIntegration < Minitest::Test
           # If 'get_worker_pids phase' times out, error in thread shuts down CI
           begin
             get_worker_pids phase, log: log
-            # added sleep as locally 165 restarts in 7 seconds
-            sleep 0.15
+            # Wait with an exponential backoff before signaling next restart
+            sleep 0.15 * (2 ** restart_count)
           rescue Minitest::Assertion # Timeout
             run = false
           end
         else
-          sleep 0.10
+          sleep 0.1
         end
       end
     end
@@ -552,6 +553,8 @@ class TestIntegration < Minitest::Test
 
     refused = replies[:refused]
     reset   = replies[:reset]
+
+    assert_operator restart_count, :>=, 2, msg
 
     if Puma.windows?
       assert_equal actual_requests - reset - refused, replies[:success]
--- a/test/test_integration_cluster.rb
+++ b/test/test_integration_cluster.rb
@@ -42,7 +42,7 @@ class TestIntegrationCluster < TestIntegration
 
   def test_phased_restart_does_not_drop_connections_threads_fork_worker
     restart_does_not_drop_connections num_threads: 10, total_requests: 3_000,
-      signal: :USR1 #, config: 'fork_worker', log: true
+      signal: :USR1, config: 'fork_worker'
   end
 
   def test_phased_restart_does_not_drop_connections_unix


test/test_integration_cluster.rb
```diff
@@ -42,7 +42,7 @@ def test_phased_restart_does_not_drop_connections
 
   def test_phased_restart_does_not_drop_connections_threads_fork_worker
     restart_does_not_drop_connections num_threads: 10, total_requests: 3_000,
-      signal: :USR1 #, config: 'fork_worker', log: true
+      signal: :USR1, config: 'fork_worker'
   end
```
dentarg, line 46: If this is the only test that needs `sleep 1` in `restart_does_not_drop_connections`, what about passing that in from the test? Making it more clear why we need to sleep 1 second

Or does it always make sense to sleep 1 second? I'm not sure what we're waiting for, that should probably be explained better in the comment above the sleep.

joshuay03: > Or does it always make sense to sleep 1 second? I'm not sure what we're waiting for, that should probably be explained better in the comment above the sleep.

IMO it always makes sense to sleep 1 second. The sleep is just how long we wait before signalling another restart. All the tests that use `#restart_does_not_drop_connections` at most set a `total_requests` of 3000. With a sleep of 0.15 we restart ~10 times, so something like 300 requests before each restart. With a sleep of 1, ~3 times, so 1000 requests before each restart.

Seeing as the tests are to ensure connections aren't dropped:
1. I don't think the number of restarts really matters, as long as there's at least one
2. I don't think (hot/phased) restarting puma every 300 requests consecutively is a realistic scenario anyway

Essentially, this is just an arbitrary threshold that just so happens to time out one particular test, and I don't think adjusting it to the new value for all tests affects their accuracy.

dentarg: Agree, thanks for explaining!

test/helpers/integration.rb
```diff
@@ -510,7 +510,7 @@ def restart_does_not_drop_connections(
           begin
             get_worker_pids phase, log: log
             # added sleep as locally 165 restarts in 7 seconds
```
dentarg, line 512: Can we change this comment? I liked your explanation 

> The sleep is just how long we wait before signalling another restart

joshuay03: Addressed in https://github.com/puma/puma/compare/d6155e87e3bd57a39bdc2503a451bbd65c97e02c..c71fb0d6fc9f4ba179b1c4abc775767b015c6bdc.

test/test_integration_cluster.rb
joshuay03: There were still 1/2 macOS failures each run with the increased sleep. Tried increasing it further and also added an assertion for restart count to ensure there are always at least 2 restarts and the assertion started failing i.e. with that high a sleep we only restart once. I instead went for an exponential back-off which seems to both give seems to be working much better.

The remaining failure seems to be unrelated, would appreciate if someone could rerun to confirm.

MSP-Greg:
@joshuay03

Good day.  Thanks for working on this.  The test probably evolved from code tested locally.  That often can be an issue with GHA CI.  I'm sure you've had test code that ran fine locally, but failed when run in GHA.

I did a revamp of the code, where the number of restarts can be specified.  See:

https://github.com/puma/puma/compare/master...MSP-Greg:puma:00-hot-restart

I've run it in CI a few times, and it seems more reliable than the current code.  Any thoughts?  Feel free to use the code.  I haven't worked on the `fork_worker` issue...

test/test_integration_cluster.rb
dentarg: Nice, thanks.

Yes, unrelated: https://github.com/puma/puma/issues/3478#issuecomment-2582334659

dentarg (approved):

PR #3590, opened 2024-12-30
Change `HttpParserError` to be subclass of `StandardError`

### LINKED ISSUE #3552: Invalid request is swallowed and times out
**Describe the bug**

It seems like there's a bug in this optimization where after handling a request it tries to read the next request from the socket:
https://github.com/puma/puma/blob/v6.4.3/lib/puma/server.rb#L475-L486.

In this [code](https://github.com/puma/puma/blob/v6.4.3/lib/puma/client.rb#L186-L192) specifically:
```ruby
        begin
          if fast_check && @to_io.wait_readable(FAST_TRACK_KA_TIMEOUT)
            return try_to_finish
          end
        rescue IOError
          # swallow it
        end
```

If the request parsing fails, `try_to_finish` raises a `Puma::HttpParserError` which inherits `IOError` and it causes the invalid request to be swallowed and to time out after 20 seconds.

**To Reproduce**

Copy this into a file called `hello.ru`:
```
```

Run it with:
```
bundle exec puma hello.ru
```

Run this curl command:
```
curl -v --keepalive-time 30 "http://localhost:9292" --next "http://localhost:9292/$(printf 'a%.0s' {1..9000})"
```

This will send 2 requests sequentially in the same TCP connection:
- The first request succeeds with status code 200
- The second request is invalid - its path is longer than the allowed length (8192) - it hangs for 20 seconds and then the connection dies :boom: 

**Expected behavior**

The second request should return 400 immediately with the following error:
```
HTTP parse error, malformed request ("GET " - (-)): #<Puma::HttpParserError: HTTP element REQUEST_PATH is longer than the (8192) allowed length (was 9001)>
```

 - Puma Version 6.4.3

MSP-Greg:
### Description

A long time ago (02-April-2006), `HttpParserError` was [added](https://github.com/puma/puma/commit/3c804d5e15f0), and it was a subclass of `IOError`.  Today, `IOError` is more associated with errors at the `IO` level.

Puma should differentiate between socket level errors and HTTP request validity errors.  Hence, change `HttpParserError` to be subclass of `StandardError`.  This change will stop `HttpParserError` from being 'swallowed' in:

https://github.com/puma/puma/blob/f5e4b77db927628e60dfb323d06a12c96883c462/lib/puma/client.rb#L184-L190

This PR also uncomments two tests in `test/test_request_invalid.rb`, which fail on master.

Note that `@to_io.wait_readable` may throw other socket related errors.  Something to keep an eye on...

Closes #3552

### DIFF
--- a/ext/puma_http11/org/jruby/puma/Http11.java
+++ b/ext/puma_http11/org/jruby/puma/Http11.java
@@ -77,7 +77,7 @@ public class Http11 extends RubyObject {
 
     public static void createHttp11(Ruby runtime) {
         RubyModule mPuma = runtime.defineModule("Puma");
-        mPuma.defineClassUnder("HttpParserError",runtime.getClass("IOError"),runtime.getClass("IOError").getAllocator());
+        mPuma.defineClassUnder("HttpParserError",runtime.getClass("StandardError"),runtime.getClass("StandardError").getAllocator());
 
         RubyClass cHttpParser = mPuma.defineClassUnder("HttpParser",runtime.getObject(),ALLOCATOR);
         cHttpParser.defineAnnotatedMethods(Http11.class);
--- a/ext/puma_http11/puma_http11.c
+++ b/ext/puma_http11/puma_http11.c
@@ -475,7 +475,7 @@ void Init_puma_http11(void)
   DEF_GLOBAL(server_protocol, "SERVER_PROTOCOL");
   DEF_GLOBAL(request_path, "REQUEST_PATH");
 
-  eHttpParserError = rb_define_class_under(mPuma, "HttpParserError", rb_eIOError);
+  eHttpParserError = rb_define_class_under(mPuma, "HttpParserError", rb_eStandardError);
   rb_global_variable(&eHttpParserError);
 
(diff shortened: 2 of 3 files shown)


dentarg (approved):
Sounds like the right thing to do

MSP-Greg:
@dentarg

Thanks for the quick review.  I'll wait a little while to see if anyone else has comments.

As to a method to my madness (and three open PR's dealing with request error handling), this started with a realization that Puma wasn't properly closing invalid keep-alive requests, and leaving the parser with a buffer full of some or all of request 'A'.

After the error, since it was a keep-alive, the code in `Server#handle_request` then tried to see if another request was on the socket, and starts by parsing the buffer, which is part of request 'A'.  Not pretty.

Then, I noticed that code that checked and/or modified the `env` was contained in both `client.rb` and `request.rb` (which is included in `server.rb`).  That is confusing and also makes CI more involved.  The reorg for that is in PR #3582...  The code in the client tests is much simpler, I haven't yet written code to allow a 'test' request to be sent in more than one chunk with delays in between...

joshuay03 (approved):
Makes sense to me, nice one!

Scutua (approved):

PR #3594, opened 2025-01-03
CI: test_plugin_systemd.rb - fixups for intermittent retries

MSP-Greg:
### Description

The tests in this file fail and/or retry intermittently.  The changes involve code that parses the notify socket.  On GHA CI, there are status messages interlaced with other messages.  These are due to delays common on GHA, but not reproducible locally.

It seems to help...

### DIFF
--- a/lib/puma/sd_notify.rb
+++ b/lib/puma/sd_notify.rb
@@ -137,10 +137,7 @@ module Puma
       ENV.delete("NOTIFY_SOCKET") if unset_env
 
       begin
-        Addrinfo.unix(sock, :DGRAM).connect do |s|
-          s.close_on_exec = true
-          s.write(state)
-        end
+        Addrinfo.unix(sock, :DGRAM).connect { |s| s.write state }
       rescue StandardError => e
         raise NotifyError, "#{e.class}: #{e.message}", e.backtrace
       end
--- a/test/test_plugin_systemd.rb
+++ b/test/test_plugin_systemd.rb
@@ -4,7 +4,7 @@ require_relative "helper"
 require_relative "helpers/integration"
 
 class TestPluginSystemd < TestIntegration
-  parallelize_me! if ::Puma.mri?
+  parallelize_me! if ::Puma::IS_MRI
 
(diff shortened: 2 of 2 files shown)


lib/puma/sd_notify.rb
```diff
@@ -137,10 +137,7 @@ def self.notify(state, unset_env=false)
       ENV.delete("NOTIFY_SOCKET") if unset_env
 
       begin
-        Addrinfo.unix(sock, :DGRAM).connect do |s|
-          s.close_on_exec = true
```
dentarg, line 141: I assume we remove this line because (from https://docs.ruby-lang.org/en/3.3/IO.html#method-i-close_on_exec-3D)

> Ruby sets close-on-exec flags of all file descriptors by default since Ruby 2.0.0. So you don’t need to set by yourself.

test/test_plugin_systemd.rb
```diff
@@ -17,21 +17,19 @@ def setup
...
-      @socket = Socket.new(:UNIX, :DGRAM, 0)
-      socket_ai = Addrinfo.unix(sockaddr)
-      @socket.bind(socket_ai)
-      @env = {"NOTIFY_SOCKET" => sockaddr }
-    end
+    @sockaddr = tmp_path '.systemd'
+    @socket = Socket.new(:UNIX, :DGRAM, 0)
+    socket_ai = Addrinfo.unix(@sockaddr)
+    @socket.bind Addrinfo.unix(@sockaddr)
+    @env = {"NOTIFY_SOCKET" => @sockaddr }
```
dentarg, line 24:
```suggestion
    @env = { "NOTIFY_SOCKET" => @sockaddr }
```

dentarg (approved):

test/test_plugin_systemd.rb
MSP-Greg: As you suggested, and rebased.

Interesting, much of Ruby's docs have no spaces with hashes, but much of the Rails docs use spaces...

Issue #3599, opened 2025-01-10
Using fork_worker results in a large performance degradation (5x slowly)


ermolaev:
**Describe the bug**
if enabled `fork_worker` after 1000 (by default) request, app is getting very slow

1. `ab -k -n 50000 -c 100 http://127.0.0.1:3000/` => `Requests per second: 7540.16 [#/sec] (mean)`
2. `kill -SIGURG 87688`
3. `ab -k -n 50000 -c 100 http://127.0.0.1:3000/` => `Requests per second: 1456.58 [#/sec] (mean)`

**Puma config** `puma.ru`:
```
fork_worker 1_000_000 # delay reforking for benchmark reproduce

port 3000
threads 4, 32
workers 5
```

**To Reproduce**
Please add reproduction steps here.

`hello.ru`:

```
run lambda { |env|
  sleep 0.01 # emulate IO
  [200, { "Content-Type" => "text/plain" }, ["Hello World"]]
}
```

Run it with:

```
bundle exec puma -C puma.rb hello.ru
```

**Expected behavior**
should not affect performance

**Desktop (please complete the following information):**
 - OS: [Mac, Linux]
 - Puma [6.5.0]

joshuay03:
I'm able to replicate locally as well:

```
Phase 0: 7960 [#/sec]
Phase 1: 1519.51 [#/sec]
Phase 2: 1507.05 [#/sec]
Phase 3: 1589.43 [#/sec]
Phase 4: 1491.29 [#/sec]
Phase 5: 1508.87 [#/sec]
```

The fork worker implementation piggy backs off phased restart (`:USR1`), however that seems unaffected, which hopefully makes the issue easier to narrow down:

```
Phase 0: 8018.93 [#/sec]
Phase 1: 7710.63 [#/sec]
Phase 2: 7542.64 [#/sec]
Phase 3: 7804.17 [#/sec]
Phase 4: 7862.89 [#/sec]
Phase 5: 8107.95 [#/sec]
```

joshuay03:
I think there's more than one fix here, the first is https://github.com/puma/puma/compare/master...joshuay03:puma:fix-3599-1 which is a contributor but not the root cause.

I'll keep digging...

joshuay03:
Pretty sure this is it: https://github.com/puma/puma/compare/master...joshuay03:puma:fix-3599-2

Edit: PR: https://github.com/puma/puma/pull/3602

[...]

PR #3601, opened 2025-01-11
Fix incomplete worker boot state on refork

joshuay03:
### Description

Not the root cause, but a contributing factor to https://github.com/puma/puma/issues/3599

When a refork happens, worker 0 is untouched and all other workers are terminated and forked from it. Since the fork worker logic piggy backs off phased restart, `workers_not_booted` is used by the master to determine if the refork cycle has been completed. However, [the initial value for this is always set to the configured worker count](https://github.com/puma/puma/blob/edd2a1728e39d055545fd12ffdfa42d5d4ccbc58/lib/puma/cluster.rb#L454), which is incorrect for refork as worker 0 doesn't need to be booted up. This means that on every refork, [`in_phased_restart` is never set to `false`](https://github.com/puma/puma/blob/edd2a1728e39d055545fd12ffdfa42d5d4ccbc58/lib/puma/cluster.rb#L519), as `workers_not_booted` will at least be `1`.

This has two consequences:
1. The logic for pinging worker 0 during a phased restart or refork when `fork_worker` is enabled continues to be executed after all the necessary workers have been booted up, which also means worker 0's ping events (one of which is [the refork request count check](https://github.com/puma/puma/blob/edd2a1728e39d055545fd12ffdfa42d5d4ccbc58/lib/puma/cluster.rb#L309-L313)) continue to fire, which is incorrect: https://github.com/puma/puma/blob/edd2a1728e39d055545fd12ffdfa42d5d4ccbc58/lib/puma/cluster.rb#L494-L497
2. `on_booted` events are never fired and loaded extension debug logs are never logged after all the necessary workers have been booted up: https://github.com/puma/puma/blob/edd2a1728e39d055545fd12ffdfa42d5d4ccbc58/lib/puma/cluster.rb#L516-L520

### DIFF
--- a/test/test_integration_cluster.rb
+++ b/test/test_integration_cluster.rb
@@ -156,6 +156,21 @@ class TestIntegrationCluster < TestIntegration
     wait_server 15
   end
 
+  def test_on_booted_with_fork_worker_refork
+    skip_unless_signal_exist? :TERM
+
+    cli_server "-w #{workers} test/rackup/hello.ru", config: <<~CONFIG
+      fork_worker
+      on_booted  { STDOUT.syswrite "on_booted called\n"  }
+    CONFIG
+
+    assert wait_for_server_to_include("on_booted called")
+
+    Process.kill :SIGURG, @pid
+
+    assert wait_for_server_to_include("on_booted called")
+  end
+
   def test_term_worker_clean_exit
     skip_unless_signal_exist? :TERM
     cli_server "-w #{workers} test/rackup/hello.ru"
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -228,7 +228,7 @@ module Puma
     def phased_restart(refork = false)
       return false if @options[:preload_app] && !refork
 
-      @phased_restart = true
+      @phased_restart = refork ? :refork : true
       wakeup!
 
       true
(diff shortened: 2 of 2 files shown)


test/test_integration_cluster.rb
```diff
@@ -156,6 +156,21 @@ def test_on_booted_and_on_stopped
...
+    cli_server "-w #{workers} test/rackup/hello.ru", config: <<~CONFIG
+      fork_worker
+      on_booted  { STDOUT.syswrite "on_booted called\n"  }
+    CONFIG
+
+    assert wait_for_server_to_include('on_booted called')
+
+    Process.kill :SIGURG, @pid
+
+    assert wait_for_server_to_include('on_booted called')
```
joshuay03, line 171: On master:
Screenshot 2025-01-11 at 6 35 35 pm

test/test_integration_cluster.rb
```diff
@@ -156,6 +156,21 @@ def test_on_booted_and_on_stopped
...
+    cli_server "-w #{workers} test/rackup/hello.ru", config: <<~CONFIG
+      fork_worker
+      on_booted  { STDOUT.syswrite "on_booted called\n"  }
+    CONFIG
+
+    assert wait_for_server_to_include("on_booted called")
+
+    Process.kill :SIGURG, @pid
+
+    assert wait_for_server_to_include("on_booted called")
```
joshuay03, line 171: On master:
Screenshot 2025-01-11 at 6 35 35 pm

PR #3602, opened 2025-01-11
Don't share server between worker 0 and >= grandchildren on refork

joshuay03:
### Description

Fixes https://github.com/puma/puma/issues/3599

TODO: Add tests


lib/puma/cluster.rb
```diff
@@ -211,12 +211,11 @@ def worker(index, master)
         pipes[:wakeup] = @wakeup
       end
 
-      server = start_server if preload?
```
joshuay03, line 214: I'm pretty sure the only reason we start the server here and not in the new worker is to [ensure the loaded cluster app is used during server initialization](https://github.com/puma/puma/blob/edd2a1728e39d055545fd12ffdfa42d5d4ccbc58/lib/puma/runner.rb#L184). Now that we can pass that app into the worker, we can consistently start the server there instead.

lib/puma/cluster/worker.rb
```diff
@@ -165,7 +166,7 @@ def spawn_worker(idx)
                                   launcher: @launcher,
                                   pipes: { check_pipe: @check_pipe,
                                            worker_write: @worker_write },
-                                  server: @server
+                                  app: @server.app
```
joshuay03, line 169: We can't simply use `start_server` and pass the new server in here like we currently (on master) do in `Cluster#spawn_worker`, cause if worker 0 was given a server during initialization, with a preloaded app, it's `#app` will initialize a new app on the `start_server` call, which we don't want. Instead, we grab the preloaded app off its server and pass that through instead.

An alternative approach is to ensure `@app` is set in `Cluster#initialize` if a server is passed in. However, doing it this way means the worker always starts its own server, which imo is a better separation of responsibilities.

test/test_integration_cluster.rb
```diff
@@ -451,6 +451,36 @@ def test_application_is_loaded_exactly_once_if_using_preload_app
...
+    loading_app_count = @server_log.scan("Loading app").length
+    assert_equal 1, loading_app_count # loaded in master
+
+    @server_log.clear
+    Process.kill :SIGURG, @pid
+
+    get_worker_pids 1, workers - 1
+    loading_app_count = @server_log.scan("Loading app").length
+    assert_equal 0, loading_app_count
+  end
```
joshuay03, line 482: These pass on master as well, I added it to ensure I don't regress on this behaviour. I've also done some puts debugging to verify this manually.

lib/puma/cluster/worker.rb
```diff
@@ -14,7 +14,7 @@ class Cluster < Puma::Runner
     class Worker < Puma::Runner # :nodoc:
       attr_reader :index, :master
 
-      def initialize(index:, master:, launcher:, pipes:, server: nil)
+      def initialize(index:, master:, launcher:, pipes:, app: nil)
```
joshuay03, line 17: There are two options here:
1. Continue to accept `server` and simply ensure that `@app` is initialized with `@server.app` if so. Then, calling `start_server` in `#spawn_worker` will use the preloaded worker 0 app to start a new server which we can give to the new worker. This would then mirror the current `Cluster#spawn_worker` implementation.
2. Let the worker consistently be responsible for starting its own server, and instead accept the preloaded `app` where needed, which is what we really intend to preload. This makes the whole inheritance chain much easier to grok imo, and introduces a clear separation of responsibilities.

test/test_integration_cluster.rb
joshuay03, line 482: These pass on master as well, I added it to ensure I don't regress on this behaviour. I've also done some puts debugging to verify this manually.

lib/puma/cluster.rb
```diff
@@ -181,7 +181,7 @@ def check_workers
         # we need to phase any workers out (which will restart
         # in the right phase).
         #
-        w = @workers.find { |x| x.phase != @phase }
+        w = @workers.find { |x| x.phase < @phase }
```
joshuay03, line 184: This fixes a race condition bug that causes the tests I'm adding here to fail intermittently, both on master and this branch, but more frequently on the latter. `#check_workers` gets called in the `Cluster#run` loop, and when reforking, there could be a point in time where this method gets called between when worker 0's phase has been incremented, and the phase has been incremented. This results in worker 0 being returned by this find and causes it to receive a SIGTERM below just before a refork happens, which is not ideal. Seeing as that's a special case where the worker's phase could be greater than the target phase changing this ensures that we don't terminate worker 0 in that case while also ensuring workers are incremented to the right phase in an ordinary phased restart scenario.

joshuay03, line 184: This fixes a race condition bug that causes the tests I'm adding here to fail intermittently, both on master and this branch, but more frequently on the latter. `#check_workers` gets called in the `Cluster#run` loop, and when reforking, there could be a point in time where this method gets called between when worker 0's phase has been incremented, and the phase has been incremented. This results in worker 0 being returned by this find and causes it to receive a SIGTERM below just before a refork happens, which is not ideal. Seeing as that's a special case where the worker's phase could be greater than the target phase changing this ensures that we don't terminate worker 0 in that case while also ensuring workers are incremented to the right phase in an ordinary phased restart scenario.

joshuay03: Alternatively we could ensure worker 0's phase is incremented right before the cluster phase is incremented.

joshuay03: I could extract this to a separate PR, but it's difficult to test anyway. See the past CI run failures for example.

Scutua (approved):

Scutua (approved):

joshuay03:
@nateberkopec Can this be considered for v7?

dentarg:
Is this an alternative to https://github.com/puma/puma/pull/3643?

dentarg:
Or is it an alternative to https://github.com/puma/puma/pull/3629?

joshuay03:
> Is this an alternative to #3643?

Neither actually, this is purely to address https://github.com/puma/puma/issues/3599. mold_worker should also benefit from this, but I'm going to only focus my testing on fork_worker which is the topic of the issue.

I'll redo the load testing comparison in [my last comment on the issue](https://github.com/puma/puma/issues/3599#issuecomment-2609896441) now that @MSP-Greg's long-tail fix and [my other fix to address the issue](https://github.com/puma/puma/pull/3601) are on master, to confirm that this is still needed (I suspect it is).

**Edit:** This is still needed: https://github.com/puma/puma/issues/3599#issuecomment-3146891441

lib/puma/cluster.rb
dentarg: It would be good to fix this in its own PR, for better documentation of it. Sounds especially useful to me if it is hard to write tests for it.

dentarg: > See the past CI run failures for example.

From this branch/PR or have we seen them from main too? Anyway, I think such logs are gone now.

joshuay03: > It would be good to fix this in its own PR, for better documentation of it. Sounds especially useful to me if it is hard to write tests for it.

Can do 👍🏽 

> > See the past CI run failures for example.

> From this branch/PR or have we seen them from main too? Anyway, I think such logs are gone now.

From this branch, but running the test locally on main I was able to replicate the issue.

test/test_integration_cluster.rb
```diff
@@ -451,6 +451,36 @@ def test_application_is_loaded_exactly_once_if_using_preload_app
...
```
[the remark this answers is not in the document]
joshuay03: Extracted into https://github.com/puma/puma/pull/3690

lib/puma/cluster.rb
joshuay03: Extracted into https://github.com/puma/puma/pull/3690

joshuay03:
Merging because CI is green (except for the Bundler v4 failures on main) and I've left more than enough evidence in the form of benchmarks in https://github.com/puma/puma/issues/3599 that this results in a significant performance improvement to fork_worker.

PR #3606, opened 2025-01-29
CI: test/helper.rb - place timeout checks inside Minitest capture_exceptions

MSP-Greg:
### Description

Previously, the Puma test framework prepended the `run` method in `Minitest::Test`:

https://github.com/puma/puma/blob/979cc374fc525f1730736bba28ea4495544c0727/test/helper.rb#L121-L147

The above placed the `Timeout.timeout` blocks within the `capture_exception` blocks.  Recently that was changed, and the timeout errors are not working correctly, resulting in more frequent test runs that freeze and timeout via the GHA step timeout.

Code in this PR simplifies the above code by adding the `Timeout.timeout` blocks to `capture_exception`.

ping @zenspider

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -102,7 +102,7 @@ end
 
 Minitest::Test.prepend TimeoutPrepend
 
-class TimeoutTestCase < Minitest::Test # rubocop:disable Puma/TestsMustTimeout
+class PumaTest < Minitest::Test # rubocop:disable Puma/TestsMustTimeout
   def teardown
     clean_tmp_paths if respond_to? :clean_tmp_paths
   end
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -102,7 +102,7 @@ end
 
 Minitest::Test.prepend TimeoutPrepend
 
-class PumaTest < Minitest::Test # rubocop:disable Puma/TestsMustTimeout
+class PumaTest < Minitest::Test # rubocop:disable Puma/TestsMustUsePumaTest
   def teardown
     clean_tmp_paths if respond_to? :clean_tmp_paths
   end
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -34,9 +34,6 @@ Thread.abort_on_exception = true
 
 $debugging_info = []
 $debugging_hold = false   # needed for TestCLI#test_control_clustered
-$test_case_timeout = ENV.fetch("TEST_CASE_TIMEOUT") do
-  RUBY_ENGINE == "ruby" ? 45 : 60
-end.to_i
 
 require "puma"
 require "puma/detect"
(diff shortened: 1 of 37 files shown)


MSP-Greg:
Some of the GHA jobs that timed out (`Error: The action 'test' has timed out after 10 minutes.`):

https://github.com/puma/puma/actions/runs/13021148139/job/36322770926
https://github.com/puma/puma/actions/runs/13021146592/job/36322743768
https://github.com/puma/puma/actions/runs/13021133356/job/36321889398

test/helper.rb
```diff
@@ -87,17 +84,26 @@ def self.call(host = '127.0.0.1')
...
+  def capture_exceptions(&blk)
+    super do
+      ::Timeout.timeout(TEST_CASE_TIMEOUT, TestTookTooLong) { yield }
     end
   end
+end
+
+Minitest::Test.prepend TimeoutPrepend
 
+class TimeoutTestCase < Minitest::Test # rubocop:disable Puma/TestsMustTimeout
```
nateberkopec, line 106: Todo, can be separate PR: we should just rename this to PumaTest or something, since it doesn't have timeout responsibilities anymore.

nateberkopec (approved):

MSP-Greg:
@nateberkopec

I was debating whether to add something like that, but decided not tonite (long day).  Also, Ryan may have some comments...

MSP-Greg:
@nateberkopec

Renamed everything in the test system, along with the Rubocop renaming the Cop.  We may add more...

nateberkopec:
LGTM, Ryan can add comments async to this being merged if he's got an opinion 👍

zenspider:
Seems like a fine PR to me... I might suggest little tweaks like allowing a per-class timeout w/ a sensible inherited default. that really only means accessing it via `self.class::TEST_CASE_TIMEOUT` instead of vanilla const access.

Is there a reason to have a common superclass `PumaTest` but to prepend your timeout module on Minitest::Test?

MSP-Greg:
@zenspider

Good evening.  The Timeout issue is mostly due to our test suite being unstable with noisy neighbors.  Some tests also start several sub-processes, so they're taxing the 2 or 3 core GHA runners.

Of course, when running the test suite locally on a strong desktop, failures rarely happen.

> Is there a reason to have a common superclass PumaTest but to prepend your timeout module on Minitest::Test?

Not quite sure what you mean.  I've got some major revisions to the test suite, but other things keep pulling me away from working on it.  I'll revisit the class structure when I get back to it.

JFYI, over the years, many test files have their own code to create some type of Puma server, and also code to create client requests.  The revisions I'm working on consolidates that code, which makes it much easier to find 'blocking' issues, free resources, etc.

Lastly, there has been encapsulation 'leakage', so what should be simple, low resource unit tests are now forced to create a Puma server...

PR #3609, opened 2025-01-31
CI: test_launcher.rb - move `test_puma_stats_clustered` to test_integration_clustered.rb

MSP-Greg:
### Description

`Puma::Cluster` calls `Process.waitall` when shutting down.  The tests in `test_launcher.rb` are not run in a sub-process, so `Process.waitall` may affect other tests now or in the future.

Move the test to `test_integration_clustered.rb`, which runs Puma in a sub-process.

### DIFF
--- a/test/test_integration_cluster.rb
+++ b/test/test_integration_cluster.rb
@@ -221,6 +221,25 @@ class TestIntegrationCluster < TestIntegration
     worker_respawn { |phase0_worker_pids| Process.kill :USR1, @pid }
   end
 
+  def test_puma_stats_clustered
+    worker_check_interval = 2
+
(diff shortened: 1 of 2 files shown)


nateberkopec (approved):

nateberkopec:
Not blocking: we could maybe enforce such a requirement with TracePoint.

GPT generated example:

```Ruby

def detect_fork(&block)
  trace = TracePoint.new(:call) do |tp|
    if tp.defined_class == Process && [:fork, :spawn, :system, :'`'].include?(tp.method_id)
      raise "Forking or spawning is not allowed in this block: #{tp.method_id}"
    end
  end

  trace.enable
  begin
    block.call
  ensure
    trace.disable
  end
end

# Example usage:
detect_fork do
  puts "Running safe code"
  fork { puts "This should be blocked" } # Raises an error
end

```

PR #3610, opened 2025-02-01
Remove redundant signal exists guards in tests

joshuay03:
### Description

My original intent was to correct an incorrect guard I added [here](https://github.com/puma/puma/pull/3601/files#diff-355b57e321e1a1fb604586cd830ccb12f20f8198e2e4b3c191b8cfb7fa6ef902R160) (should check for URG).

However, seeing as CI was green, I realised that all except one (for INFO) of these guards is redundant in that suite as it [only runs on fork-compatible OSs](https://github.com/puma/puma/blob/801d292e174d1b32b7b95bafef498ea44c8f2a83/test/test_integration_cluster.rb#L14).

This motivated me to find similar redundant guards in other suites and remove them, which I've done here.

### DIFF
--- a/test/test_integration_cluster.rb
+++ b/test/test_integration_cluster.rb
@@ -111,18 +111,14 @@ class TestIntegrationCluster < TestIntegration
   # No more than 10 should throw Errno::ECONNRESET.
 
   def test_term_closes_listeners_tcp
-    skip_unless_signal_exist? :TERM
     term_closes_listeners unix: false
   end
 
   def test_term_closes_listeners_unix
-    skip_unless_signal_exist? :TERM
     term_closes_listeners unix: true
   end
 
   def test_term_exit_code
-    skip_unless_signal_exist? :TERM
-
     cli_server "-w #{workers} test/rackup/hello.ru"
     _, status = stop_server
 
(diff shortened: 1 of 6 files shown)


test/test_integration_cluster.rb
joshuay03, line 1: Suite is only run if fork compatible. Only leftover is a guard for INFO which Linux doesn't support.

test/test_preserve_bundler_env.rb
joshuay03, line 1: Suite is only run if fork compatible.

test/test_worker_gem_independence.rb
joshuay03, line 1: Suite is only run if fork compatible.

test/test_plugin_systemd.rb
```diff
@@ -11,8 +11,6 @@ class TestPluginSystemd < TestIntegration
 
   def setup
     skip_unless :linux
-    skip_unless :unix
-    skip_unless_signal_exist? :TERM
```
joshuay03, line 15: Unix is redundant if checking Linux (`UNIXSocket` will be defined), TERM exists on Linux.

MSP-Greg:
Well, CI passes.  I don't consider myself an expert on all the various versions of *nix, but if issues arise, the skip code can be adjusted...

Thanks, Greg

Issue #3431, opened 2024-07-17
Docs are unclear about status of --fork-worker


SpamapS:
**Describe the bug**

The docs at https://github.com/puma/puma/blob/master/docs/fork_worker.md currently start:

"Puma 5 introduces an experimental new cluster-mode configuration option, fork_worker (--fork-worker from the CLI). This mode causes Puma to fork additional workers from worker 0, instead of directly from the master process"

So, at the very least the tense is misleading as 5 is now in the past. But also, is it still experimental in 6?

Hoping we can get some clarity on its status in the docs.

**Puma config:**

N/A

**To Reproduce**

N/A

**Expected behavior**

I'd expect that it would mention the current state of the feature in the context of the most recent major release.

I'd be happy to update the docs myself, but it's not clear to me where the source of truth for the status of features is other than these docs.

Thank you for your time and effort on Puma!

**Desktop (please complete the following information):**

N/A

PR #3611, opened 2025-02-02
Remove DSL version comments, note fork_worker as experimental

joshuay03:
### Description

Closes https://github.com/puma/puma/issues/3431

It seems the only version comments are from v5 i.e. they haven't been used before v5 and haven't been kept up in v6. IMO the mention of new features in the changelog/upgrade guide is sufficient to know when they were introduced. Even without these, I would expect that users are reading the DSL at a specific tag if they are in the process of upgrading. Also, options may be added in minor versions (e.g. [this](https://github.com/puma/puma/pull/3496)), I don't think we have to tag every single one as such.

This PR removes the handful of existing version comment notes, and mentions in a comment note that fork_worker is experimental.

### DIFF
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -654,8 +654,6 @@ module Puma
     # @example
     #   state_permission 0600
     #
-    # @version 5.0.0
-    #
     def state_permission(permission)
       @options[:state_permission] = permission
     end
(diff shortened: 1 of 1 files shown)


nateberkopec:
LGTM but I'll wait for @MSP-Greg who added them.

nateberkopec (approved):

MSP-Greg:
@joshuay03

PR's with yard version comments are difficult, as one doesn't know whether the next version will be a patch or minor release.

But, `Puma::DSL` is probably the one file that relates to most user's interactions with Puma (the config file).  So, I'm okay with this PR, but sometime in the future I may add the version comments, as they are a good reference...

PR #3613, opened 2025-02-03
Update 'clustered mode' references to 'cluster mode'

joshuay03:
### Description

A majority of user facing and internal documentation refer to the two modes puma functions in as 'single mode' and 'cluster mode'. However, there are some places where the term 'clustered mode' is used. For example, the README uses 'cluster mode' in [the configuration section](https://github.com/puma/puma?tab=readme-ov-file#configuration), but 'clustered mode' in [that respective section](https://github.com/puma/puma?tab=readme-ov-file#clustered-mode).

Here I've updated all such references to consistently use 'cluster mode'. 'clustered' is still used internally to name variables and predicates (e.g. `clustered?`) which makes sense, but not to refer to the mode itself.

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -33,7 +33,7 @@ require_relative "helpers/apps"
 Thread.abort_on_exception = true
 
 $debugging_info = []
-$debugging_hold = false   # needed for TestCLI#test_control_clustered
+$debugging_hold = false
 
 require "puma"
 require "puma/detect"
@@ -224,7 +224,6 @@ class Minitest::Test
 end
 
 Minitest.after_run do
-  # needed for TestCLI#test_control_clustered
   if !$debugging_hold && ENV['PUMA_TEST_DEBUG']
     $debugging_info.sort!
     out = $debugging_info.join.strip
--- a/History.md
+++ b/History.md
@@ -163,7 +163,7 @@
 
 * Features
   * Ability to supply a custom logger ([#2770], [#2511])
-  * Warn when clustered-only hooks are defined in single mode ([#3089])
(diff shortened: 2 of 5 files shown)


test/helper.rb
```diff
@@ -33,7 +33,7 @@
 Thread.abort_on_exception = true
 
 $debugging_info = []
-$debugging_hold = false   # needed for TestCLI#test_control_clustered
```
joshuay03, line 36: Seems like the test mentioned here no longer exists.

nateberkopec: @MSP-Greg ?

nateberkopec:
Looking for @MSP-Greg's comment re the test change. The cluster rename is good with me.

test/helper.rb
dentarg: Comment was added over 5 years ago in https://github.com/puma/puma/commit/4c8d4d6921a1b9b9155ab83b09a632c377401398

dentarg: Looks like https://github.com/puma/puma/commit/9a8fcb28190673aecaa129632b7c640517430ff5 moved & renamed (slightly) the test

nateberkopec: Looks like we can remove debugging_hold and then update the comment

MSP-Greg: > Looks like we can remove `$debugging_hold` and then update the comment

Yes.  A quick check show it isn't used anymore, I.think.  Also remvoe the comment on line 227.

The mentioned test (test_control_clustered) was starting a clustered server using `Puma::CLI.new`.  It has since been been removed (or moved).

Issue #3571, opened 2024-12-04
Provide mode (single, cluster)-specfic hooks for conditionally evaluating config methods


midnight-wonderer:
**Is your feature request related to a problem? Please describe.**
I am loading JVM into Ruby's process using [RJB (Ruby-Java bridge)](https://github.com/arton/rjb).
JVM is not fork-safe, so I have to load it after fork. Hence, inside Puma's `on_worker_boot`.
I need a way to set `on_worker_boot` in `puma.rb` only if Puma is forking.

**Describe the solution you'd like**
Puma does not directly cause my problem. But it would be nice if I can
* Detect modes (cluster mode, single mode, etc.) inside `puma.rb` without relying on indirect methods (e.g., environment variables) so that I can set `on_worker_boot` in the appropriate scenario and probably `on_booted` otherwise. (Or other ways to foretell the forking from `puma.rb`.)
* To solve the issue above, I must also detect if the Rails app runs inside Puma at all (e.g., in Rails console). Is this already doable?  
This is a question somewhat unrelated to the feature request. (I have some ideas, but I also want to find some opinions or recommendations when my ideas feel like hacks.)

**Additional context**
It seems this chunk of code does something similar to what I am looking for. However, it is not part of the public API.
https://github.com/puma/puma/blob/2c9aafef0ea42058bf3bb03f53805680fffecc60/lib/puma/dsl.rb#L1433-L1434

nateberkopec:
We [warn you](https://github.com/puma/puma/blob/94461f410c444486c4516198a1379e0999bae242/lib/puma/dsl.rb#L750) if you try to set on_worker_boot in single mode, but you can just do it. Puma won't execute that hook so there's no reason not to just set both in puma.rb.

So just set both?

midnight-wonderer:
> So just set both?

Correct me if I misunderstood.
The `on_booted` hook will be fired on the master process in clustered mode, loading unused JVM instances into it.
There might be some other aspects, too, but I did not initially explore this option mainly because of that.

joshuay03:
> The `on_booted` hook will be fired on the master process in clustered mode, loading unused JVM instances into it. There might be some other aspects, too, but I did not initially explore this option mainly because of that.

[...]

PR #3438, opened 2024-08-09
Rename callback hooks

tannakartikey:
### Description

Hey Puma Maintainers,

This is my first PR for Puma, inspired by Nate's talks. Just wanted to check if I'm heading in the right direction before I invest more time into this.

Thanks a bunch!

Related PR: #3258
Closes: #3387

### Your checklist for this pull request

- [ ] I have reviewed the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.
- [ ] I have added (or updated) appropriate tests if this PR fixes a bug or adds a feature.
- [ ] My pull request is 100 lines added/removed or less so that it can be easily reviewed.
- [ ] If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.
- [ ] If this closes any issues, I have added "Closes `#issue`" to the PR description or my commit messages.
- [ ] I have updated the documentation accordingly.
- [ ] All new and existing tests passed, including Rubocop.

PR #3616, opened 2025-02-09
Ensure and enforce that configs are loaded before options are accessed

joshuay03:
### Description

Inspired by [this discussion](https://github.com/puma/puma/issues/3571#issuecomment-2634415513).

Currently, [the `Launcher` ensures that `Configuration#load` is called](https://github.com/puma/puma/blob/ca201ef69757f8830b636251b0af7a51270eb68a/lib/puma/launcher.rb#L56) before options are accessed. However, this expectation isn't maintained in many tests. To address this, this PR:
1. Adds a validation to `Configuration#options` which raises if accessed before `#load` is called
2. Updates all tests that don't `#load` configurations before their assertions
3. Updates `Binder#initialize` to accept options instead of a config

    `Binder`'s are only initialized in (two) locations where options have already been accessed/parsed, however unlike their caller classes it expects a config instead of options. In the context of this PR, that would require passing the config all the way down just to satisfy this API. However, I think it makes more sense for this API to accept options as well. Also, seeing as [the current config arg is optional](https://github.com/puma/puma/blob/ca201ef69757f8830b636251b0af7a51270eb68a/lib/puma/binder.rb#L22), we might not be using the configured options inside it as expected, like when [utilising the binder in the server](https://github.com/puma/puma/blob/ca201ef69757f8830b636251b0af7a51270eb68a/lib/puma/server.rb#L117).


dentarg:
The turbo-rails tests ([fail link](https://github.com/puma/puma/actions/runs/13226632500/job/36918427882?pr=3616#step:6:27)) are using capybara, and it seems like the changes here breaks what capybara is doing: https://github.com/teamcapybara/capybara/blob/3.40.0/lib/capybara/registrations/servers.rb#L44

Is this PR intentionally introducing breaking changes? 

Disclaimer: I have not studied this PR in detail and I don't have everything about the Puma DSL/config system in my head :-)

lib/puma/dsl.rb
```diff
@@ -1422,9 +1422,13 @@ def process_hook(options_key, key, block, meth)
...
-      # user_options (CLI) have precedence over config file
-      workers_val = @config.options.user_options[:workers] || @options[:workers] ||
-        @config.puma_default_options[:workers] || 0
+
+      # TODO: Move this to Configuration and ensure the warnings are only logged
+      #       once the entire configuration is loaded.
+      #
+      #       See https://github.com/puma/puma/issues/3571#issuecomment-2565032734
+      #       for an example of why this isn't a good place to log these warnings.
+      workers_val = @config.instance_variable_get(:@options)[:workers]
```
nateberkopec, line 1431: We should get this addressed before merging

lib/puma/launcher.rb
```diff
@@ -48,18 +48,21 @@ def initialize(conf, launcher_args={})
...
-      @config.options[:log_writer] = @log_writer
+      @config.load
+      @config.clamp
+      @options = @config.options
+
+      @options[:log_writer] = @log_writer
 
       # Advertise the Configuration
       Puma.cli_config = @config if defined?(Puma.cli_config)
+      log_config if env['PUMA_LOG_CONFIG']
```
joshuay03, line 59: I've QA'ed and confirmed that moving this here doesn't change it's position in the actual logs.

lib/puma/launcher.rb
```diff
@@ -48,18 +49,28 @@ def initialize(conf, launcher_args={})
...
+      if @options[:bind_to_activated_sockets]
+        @options[:binds] = @binder.synthesize_binds_from_activated_fs(
+          @options[:binds],
+          @options[:bind_to_activated_sockets] == 'only'
+        )
+      end
 
       # Advertise the Configuration
       Puma.cli_config = @config if defined?(Puma.cli_config)
+      log_config if env['PUMA_LOG_CONFIG']
```
joshuay03, line 62: I've QA'ed to ensure moving this here doesn't actually change the position of the configuration in the logs. It's just to group all the config advertising related code together.

lib/puma/launcher.rb
```diff
@@ -48,18 +49,28 @@ def initialize(conf, launcher_args={})
...
 
-      @config.load
-
-      @binder        = Binder.new(@log_writer, conf)
-      @binder.create_inherited_fds(ENV).each { |k| ENV.delete k }
-      @binder.create_activated_fds(ENV).each { |k| ENV.delete k }
+      @binder        = Binder.new(@log_writer, @options)
+      @binder.create_inherited_fds(env).each { |k| env.delete k }
+      @binder.create_activated_fds(env).each { |k| env.delete k }
```
joshuay03, line 71: The only time `env` is passed into the launcher is [here](https://github.com/puma/puma/blob/ca201ef69757f8830b636251b0af7a51270eb68a/lib/puma/cli.rb#L66), which happens to be `ENV` as well, which is probably why it wasn't caught that here we don't use the local var bad pass in the constant instead.

lib/puma/dsl.rb
joshuay03: Done, fortunately there's some good coverage around this.

dentarg:
> Hmm yes I just learnt about #clamp... Looks like they aren't calling #load though, which they should be doing. I think we should be validating both. Let me know what ya'll think.

I think there needs to be a released version of Capybara where this is fixed before we make Puma raise.

nateberkopec:
FWIW we're so close to a v7 release at this point (the only reason I can think of doing another 6.x minor will be for a keepalive fix) that we can just treat this as "breaking" and not have to worry about it sitting unmerged for too long.

joshuay03:
> Yep, I'll send a PR.

https://github.com/teamcapybara/capybara/pull/2794

lib/puma/configuration.rb
```diff
@@ -176,6 +180,8 @@ def initialize(user_options={}, default_options = {}, env = ENV, &block)
 
       @options     = UserFileDefaultOptions.new(user_options, default_options)
```
schneems: Request: I suggest we rename from `@options` to `@_options` to discourage someone accidentally using it instead of the `options` method in the future. It would stand out more in a PR review if someone had `@_options.method` in their code.

lib/puma/configuration.rb
```diff
@@ -125,6 +126,9 @@ def final_options
   # is done because an environment variable may have been modified while loading
   # configuration files.
   class Configuration
+    class NotLoadedError < StandardError; end
+    class NotClampedError < StandardError; end
```
schneems, line 130: Optional: The error messages for these classes are copied in a few places. We could move them to the class instead to dry them up.

```rubye
class NotLoadedError < StandardError
  def new
    super <<~EOM
        ...
    EOM
  end
end
```

lib/puma/configuration.rb
```diff
@@ -384,22 +419,33 @@ def load_rackup
...
+      return if options[:silence_fork_callback_warning]
 
-      SecureRandom.hex(16)
+      @hooks.each do |key, method|
+        options.all_of(key).each do |hook_options|
+          next unless hook_options[:cluster_only]
+
+          LogWriter.stdio.log(<<~MSG)
+            Warning: You specified code to run in a `#{method}` block,
+            but Puma is not configured to run in cluster mode (worker count > 0),
```
schneems: The words are stating the current reality "puma is not configured to run in cluster mode," but the info in the parens are reflecting the opposite of that reality (as the worker count < 0). 

I suggest we change the wording to something like this to spell out we're indicating the state the system needs to be in to avoid the warning.

```
Warning: The code in the `#{method}` block will not execute
in the current puma configuration. The `#{method}` block only 
executes in Puma's cluster mode. To fix either remove the `#{method}` 
call or increase Puma's worker count above zero.
```

schneems:
Looks like capybara is still failing. I did a rough skim and made a few suggestions. I see a lot of changes to the tests to get them to pass, but I'm not sure that I saw a test that exercises the new behavior. 

Request: Add tests that asserts the load and clamp error behaviors in configuration usage. We should also assert that different hook usage causes a warning to be emitted. To do that you'll likely need to add some dependency injection so you can pass in something you can assert against.

lib/puma/configuration.rb
joshuay03: Done

lib/puma/configuration.rb
joshuay03: The messages are slightly different in each case.

lib/puma/configuration.rb
joshuay03: Done

joshuay03:
> Looks like capybara is still failing. I did a rough skim and made a few suggestions. I see a lot of changes to the tests to get them to pass, but I'm not sure that I saw a test that exercises the new behavior.

Yes. We need https://github.com/teamcapybara/capybara/pull/2794 merged and released, and then turbo-rails to use that version. It would also be ideal if that happens before Puma releases v7, else users of both will be blocked from upgrading.

> Request: Add tests that asserts the load and clamp error behaviors in configuration usage.

Done. Thank you for the review.

> We should also assert that different hook usage causes a warning to be emitted. To do that you'll likely need to add some dependency injection so you can pass in something you can assert against.

There is already coverage for this for all hooks, see https://github.com/puma/puma/pull/3616/files#diff-f8a5111e8509bf1e3c382deae1d40eb8640626bd81cd5066577d3082d8e0a7dfR782 and where that method (`#assert_warning_for_hooks_defined_in_single_mode`) is used.

schneems:
> and where that method (#assert_warning_for_hooks_defined_in_single_mode) is used.

Great, thanks for enumerating that!

I'm aiming to get v7 out before RailsWorld, if we're blocked here on an external PR, then our options are: Come up with a non-blocking solution or move this to v8 (at which time we could STILL be blocked by an external PR). 

Looking at that PR to capybara the main thing is that it's calling load before clamp. Spitballing, what if `clamp` implicitly called `load` first? I.e. instead of making people use two operations in the correct order, allow them to use `clamp` as a "load and clamp." WDYT? Is there ever a valid time when someone would call clamp without load?

joshuay03:
> I'm aiming to get v7 out before RailsWorld

Will you be there? I will come say hi.

> if we're blocked here on an external PR, then our options are: Come up with a non-blocking solution or move this to v8 (at which time we could STILL be blocked by an external PR).
> 
> Looking at that PR to capybara the main thing is that it's calling load before clamp. Spitballing, what if `clamp` implicitly called `load` first? I.e. instead of making people use two operations in the correct order, allow them to use `clamp` as a "load and clamp." WDYT? Is there ever a valid time when someone would call clamp without load?

Hmm I like this, but I don't know the answer to your last question. I'll try this out locally and see if anything fails. If everything looks good with all the tests using just `#clamp` then I think it's fine 👍🏽

lib/puma/configuration.rb
```diff
@@ -262,7 +282,15 @@ def config_files
     # Call once all configuration (included from rackup files)
     # is loaded to flesh out any defaults
     def clamp
-      @options.finalize_values
+      raise NotLoadedError, "ensure load is called before clamping the configuration" unless @loaded
```
MSP-Greg:
```suggestion
      load unless @loaded
```

test/test_config.rb
```diff
@@ -595,18 +639,89 @@ def test_silence_fork_callback_warning_overwrite
...
 
+  def test_options_raises_not_loaded_error_when_not_loaded
+    conf = Puma::Configuration.new
+
+    error = assert_raises(Puma::Configuration::NotLoadedError) do
+      conf.options
+    end
+
+    assert_equal "ensure load is called before accessing options", error.message
+  end
```
MSP-Greg:
```suggestion
```

lib/puma/configuration.rb
joshuay03: I've just pushed this 👍🏽

test/test_config.rb
joshuay03: I've just pushed this 👍🏽

MSP-Greg:
@joshuay03

1. I noticed quite a bit of load/clamp calls in the testing.  They could be removed.  Haven't looked at the most recent changes.  Also, method comments could make clear that `load` is called in `clamp` if it hasn't been previously called.
2. Should I cancel the CI runs that are stacked up?

joshuay03:
> 1. I noticed quite a bit of load/clamp calls in the testing.  They could be removed.  Haven't looked at the most recent changes.  Also, method comments could make clear that `load` is called in `clamp` if it hasn't been previously called.

I think we want to keep them right, the idea being that if clamp isn't called, we raise an error. Currently some tests will raise if we don't clamp with these changes (cause they access `#options` for example), and some not, but the idea is to have the tests consistent with how the launcher etc. finalizes the config. Most tests are already loading, then I updated to load and clamp, now with the implicit load on clamp I've just updated to clamp.

> 2. Should I cancel the CI runs that are stacked up?

Yes please 🙏🏽 I just need the latest.

MSP-Greg:
Sorry, I meant remove the `load` call because clamp will call it.

joshuay03:
> Sorry, I meant remove the `load` call because clamp will call it.

Ah yep, I've gone ahead and done that 👍🏽

joshuay03:
Building green (including turbo-rails), so I've closed the capybara PR.

I'll wait for the build to finish and then push one last change that @MSP-Greg suggested to add a comment above `#clamp`.

**Edit:** Done.

MSP-Greg:
Thanks for closing the capybara PR and the additional changes.

We can squash commits when merging a PR.  Once a PR hit's 'last minute changes/discussion/suggested changes', if you do a commit per change, it's easier to see what changing.

I often force push before a PR is actively being discussed, but once that starts, I tend to add commits.

Then again, whatever works best for you.

And, thanks for your work on this...

schneems:
Agreed with Greg, it's easier to review new, smaller commits once a chunk has already been reviewed. It's okay to rebase against master/main and force push still.

In other news: This is now merged! Thanks for the contribution. I'm glad the "load config by clamp implicitly" seems to resolve the associated failures.

As an FYI I merged in a change to master that conflicted with your PR. I resolved those manually and merged. All tests are passing, but it wouldn't hurt to double check my merge-fu handy-work and touch up any mistakes with another PR.

joshuay03:
> As an FYI I merged in a change to master that conflicted with your PR. I resolved those manually and merged. All tests are passing, but it wouldn't hurt to double check my merge-fu handy-work and touch up any mistakes with another PR.

Ah, I think there's some code that can be removed thanks to https://github.com/puma/puma/pull/3438. I'll send a PR later.

**Edit:** https://github.com/puma/puma/pull/3710

Issue #3596, opened 2025-01-10
fork_worker and permanent worker loss


toregeschliman:
**Describe the bug**
When using `fork_worker` mode, it is possible to permanently lose other workers if worker 0 exits for any reason (including a TERM/INT/KILL signal).

**Puma config:**

Any config with `fork_worker`, though it is exacerbated by a `before_refork` hook that takes a few seconds. A minimal easy repro example would include

```
fork_worker
before_refork { sleep(10) }
workers(20)
```
though the longer the sleep and the greater the worker count, the easier the repro.

**To Reproduce**
Run any local app with the above; terminate any number of non-zero workers to trigger refork, wait a few seconds, and terminate worker zero. For repro, a simple `kill -TERM` is adequate, though in our production case this is triggered by a TERM sent from either a memory manager or from Rack::Timeout (helps prevent Thread.raise from corrupting app state). You will eventually get log messages about some workers hitting the worker_boot_timeout but they will never start again.

There is an element of chance involved here; the circumstances required are
1. There must be multiple PID's written to the `@fork_pipe` for worker zero to refork
2. The `"wrkr fork" thread on worker zero must then execute the `@fork_pipe.gets` [here](https://github.com/puma/puma/blob/master/lib/puma/cluster/worker.rb#L83) while there are multiple lines already waiting in the pipe
3. `@fork_pipe.gets` will eagerly read in from the pipe to a local buffer (on my local machine it will read up to 8192 bytes from the pipe), but only the first line will be returned; this buffer will then be drained over consecutive `gets` but the bytes are only in local state now, not the pipe anymore.
4. A termination at this point will cause all of the data read to the local buffer but _not_ processed by the app to get permanently lost.
5. Eventually the parent process will notice that the worker has not booted successfully and will complain that it has timed out, _however_ since there was never any PID returned for it to track, those workers are never cleared out of the parent process's internal state and will never get triggered to fork again ([this line](https://github.com/puma/puma/blob/master/lib/puma/cluster.rb#L566) will keep the worker around permanently despite its termed state).
6. If you repeat this often enough, eventually you will only be left with the parent process and worker zero.

**Expected behavior**
Workers that terminate should end up getting reforked eventually.

**Desktop (please complete the following information):**
 - OS: Mac & Linux
 - Puma Version: 6.5.0

**Discussion**
[...]

PR #3619, opened 2025-02-13
Mold worker zero at refork

toregeschliman:
### Description
Per discussion in https://github.com/puma/puma/issues/3596, this is an attempt to make worker-0 an idle "mold" process whenever it has to refork, to reduce the complexity of the fork_worker functionality and eliminate a class of bugs coming from the current behavior which switches Thread.main to the "puma wrkr fork" thread, effectively deleting what many Ruby gems/apps treat as global state.

This also pulled in changes to fork_pipe message handling; the current implementation pulls eagerly from the pipe into a local buffer which risks permanent data loss, so I've changed it from newline-delimited strings to a standard packed bigint payload that can be read individually. Implemented using read_/write_nonblock as I think we can improve on its performance and simplify the protocol later (if we don't have to worry about pausing a server, we can let worker zero determine when to run hooks based on when it's read everything it can) but happy to convert to a blocking read/write implementation for simplicity for now, or leave those changes out altogether.

I also made this behavior the default and only fork_worker behavior; in its most general function it will remain the same (it spawns a new worker to replace the now-idle worker 0) but there may be a _lot_ of implementations out there that have implicit or explicit expectations that the specific process will be handling requests so I may also convert it to an additional config.

### DIFF
--- a/.github/workflows/turbo-rails.yml
+++ b/.github/workflows/turbo-rails.yml
@@ -3,7 +3,7 @@ name: turbo-rails
 # Note: turbo-rails often returns an ActionDispatch::Response::RackBody for the
 # body. Also, Rack::BodyProxy or Sprockets::Asset.
 
-on: [push, pull_request, workflow_dispatch]
+on: [push, review_requested, ready_for_review, workflow_dispatch]
 
 permissions:
   contents: read # to fetch code (actions/checkout)
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -92,12 +92,28 @@ module Puma
       end
 
       if @options[:fork_worker] && all_workers_in_phase?
-        @fork_writer.restart_server
-
         if worker_at(0).phase > 0
           @fork_writer.after_refork
         end
+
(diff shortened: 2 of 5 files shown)


toregeschliman:
that's a lot of tests for a draft PR :) will have to keep that in mind

.github/workflows/turbo-rails.yml
```diff
@@ -3,7 +3,7 @@ name: turbo-rails
 # Note: turbo-rails often returns an ActionDispatch::Response::RackBody for the
 # body. Also, Rack::BodyProxy or Sprockets::Asset.
 
-on: [push, pull_request, workflow_dispatch]
+on: [push, review_requested, ready_for_review, workflow_dispatch]
```
toregeschliman, line 6: personally I think this is better in general for draft PR workflows but can remove after it's passing tests in local

toregeschliman:
Im going to go ahead and close this for now so I can keep working without triggering CI builds everytime I push to the branch; will link new PR if I can't re-open once it's ready

PR #3629, opened 2025-03-07
replace fork_pipe with more robust implementation

toregeschliman:
Trying out _just_ this as the safest change outlined in https://github.com/puma/puma/issues/3596; molding worker zero is proving hard to test safely. This will ensure that only one payload at a time is read off the fork_pipe by introducing a standard packet length, using pack and unpack1 and integers to avoid allocating any new objects during the loop, but more importantly it does _not_ eagerly read off the pipe so if anything happens to worker 0 we can only lose 1 worker at most permanently.

### DIFF
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -5,6 +5,7 @@ require_relative 'util'
 require_relative 'plugin'
 require_relative 'cluster/worker_handle'
 require_relative 'cluster/worker'
+require_relative 'cluster/fork_pipe'
 
 module Puma
   # This class is instantiated by the `Puma::Launcher` and used
@@ -73,14 +74,14 @@ module Puma
 
       master = Process.pid
       if @options[:fork_worker]
-        @fork_writer << "-1\n"
+        @fork_writer.start_refork
       end
 
       diff.times do
         idx = next_worker_index
 
         if @options[:fork_worker] && idx != 0
-          @fork_writer << "#{idx}\n"
+          @fork_writer.refork_workers(idx)
           pid = nil
         else
           pid = spawn_worker(idx, master)
@@ -91,11 +92,8 @@ module Puma
       end
 
       if @options[:fork_worker] && all_workers_in_phase?
-        @fork_writer << "0\n"
-
-        if worker_at(0).phase > 0
-          @fork_writer << "-2\n"
-        end
+        @fork_writer.after_refork if worker_at(0).phase > 0
+        @fork_writer.restart_server
       end
     end
 
@@ -419,7 +417,8 @@ module Puma
 
       # Separate pipe used by worker 0 to receive commands to
       # fork new worker processes.
-      @fork_pipe, @fork_writer = Puma::Util.pipe
+      @fork_pipe, fork_writer_pipe = Puma::Util.pipe
+      @fork_writer = ForkPipeWriter.new(fork_writer_pipe)
 
       log "Use Ctrl-C to stop"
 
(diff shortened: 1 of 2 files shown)


lib/puma/cluster/fork_pipe.rb
```diff
@@ -0,0 +1,118 @@
...
+      # avoids allocation of objects while reading
+      # only reads 1 payload at a time (minimize data loss if process exits unexpectedly)
+      # NOT thread-safe, but concurrent access not required anywhere currently
+      def initialize(pipe)
+        @pipe = pipe
+        @buffer = +""
+        @payload = +""
+      end
+
+      def read_nonblock
```
toregeschliman, line 24: the `_nonblock` methods are currently extraneous; I replaced them with simpler blocking methods below since changing to a simpler protocol (of letting the mold process handle when to call refork hooks) is a wider-sweeping change. I can delete them if preferred.

lib/puma/cluster.rb
```diff
@@ -91,11 +92,8 @@ def spawn_workers
       end
 
       if @options[:fork_worker] && all_workers_in_phase?
-        @fork_writer << "0\n"
-
-        if worker_at(0).phase > 0
```
toregeschliman, line 96: I am not sure it makes sense to run these hooks after the server has been restarted; if there is anything in there meant to reset server state it would need to go first, so I've bumped it in front of the restart command.

lib/puma/cluster/fork_pipe.rb
```diff
@@ -0,0 +1,118 @@
...
+        @buffer.clear
+        @payload.clear
+        remaining = PAYLOAD_SIZE
+        while remaining > 0
+          case @pipe.read_nonblock(remaining, @buffer, exception: false)
+          when :wait_readable
+            @pipe.wait_readable
+          when :wait_writable
+            @pipe.wait_writable
+          else
```
joshuay03, line 34: Considering that this implementation is specific to `IO.pipe`, do we need to concern ourselves with `IO::WaitWritable` here? Reading through the docs it seems it's only necessary when the io may be a `OpenSSL::SSL::SSLSocket`. Same question applies to the writer and `:wait_readable` below.

https://docs.ruby-lang.org/en/3.4/IO.html#method-i-read_nonblock

joshuay03 (commented):
I understand the problem this is meant to solve, but I'll leave it to someone more experienced than I to comment on the approach/implementation.

lib/puma/cluster/fork_pipe.rb
toregeschliman: Both _nonblock methods are extraneous for now so I'll probably just remove them; it's true that if these are limited to pipes the `:wait_writable` isn't necessary, that was just force of habit.

lib/puma/cluster.rb
toregeschliman: indeed, "after the server has temporarily stopped serving requests" is the specific timing I based this off of, which would be a very weird way to describe "and also after it's started serving them again"

joshuay03 (approved):

toregeschliman:
Tested this out on a beta service and not seeing the hoped-for results; noticing that the phased restart is extremely slow (we run 31 workers but it took over 6 minutes to get everything to phase 1) and then after a handful more worker restarts, we just stop reforking altogether until we've lost everything but worker zero. That suggests to me that we aren't getting substantial data loss in the pipe, lowering perceived priority of the fix.

schneems:

```
```

nateberkopec:
Re-requesting @joshuay03 - let's either outline next actions or merge

joshuay03:
Closing based on this discussion: https://github.com/puma/puma/pull/3643#discussion_r2115906732

@toregeschliman Feel free to reopen if that no longer holds true.

PR #3635, opened 2025-03-25
accept to_path to be nil on bodies

### LINKED ISSUE #3634: Streaming body may crash if to_path is nil
**Describe the bug**
When a class that responds nil for `to_path` is returned as a body puma errors.

**To Reproduce**
config.ru
```
run lambda { |env|
  r, w = IO.pipe
  w.write "foo"
  w.close
  [200, {"Content-Type" => "text/plain"}, r]
}
```

The `Content-Type` header is required for the bug to appear.

**Expected behavior**
The pipe should be read and served.

**Discussion**
I currently work around this issue by using
```
class RRDIO < IO
  def to_path
    ""
  end
end
```

Not ideal.

The problematic code is here:

https://github.com/puma/puma/blob/ca201ef69757f8830b636251b0af7a51270eb68a/lib/puma/request.rb#L194

I would like to submit a fix for this bug myself but I would have to know what the preferred solution is. I also think a similar bug could appear on line 202: https://github.com/puma/puma/blob/ca201ef69757f8830b636251b0af7a51270eb68a/lib/puma/request.rb#L201-L202

LevitatingBusinessMan:
### Description

The rack spec now supports bodies responding with `nil` for `to_path`.
This is the default behavior of IO objects made with `IO.pipe`.

This PR has a commit to prevent puma from crashing when `to_path` returns nil.
It also has a test for pipe response bodies which tests the presence of this bug.

https://github.com/puma/puma/issues/3634
https://github.com/rack/rack/pull/2318

Closes #3634

### DIFF
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -191,7 +191,8 @@ module Puma
           elsif res_body.is_a?(File) && res_body.respond_to?(:size)
             body = res_body
             content_length = body.size
-          elsif res_body.respond_to?(:to_path) && File.readable?(fn = res_body.to_path)
+          elsif res_body.respond_to?(:to_path) && (fn = res_body.to_path) &&
+              File.readable?(fn)
             body = File.open fn, 'rb'
             content_length = body.size
             close_body = true
@@ -199,7 +200,7 @@ module Puma
             body = res_body
           end
         elsif !res_body.is_a?(::File) && res_body.respond_to?(:to_path) &&
-            File.readable?(fn = res_body.to_path)
+            (fn = res_body.to_path) && File.readable?(fn = res_body.to_path)
           body = File.open fn, 'rb'
           content_length = body.size
           close_body = true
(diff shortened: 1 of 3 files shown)


dentarg:
(CI failures seems unrelated, I have filled https://github.com/socketry/localhost/issues/35)

MSP-Greg:
@LevitatingBusinessMan

Could you rebase this?  CI is passing now.

Also, there are some failures, most due to the issue of 'character' handling vs 'byte' handling.

I added two commits to https://github.com/MSP-Greg/puma/commits/00-pr-3535/ which contain changes to the tests and also changes to `puma_socket.rb`.  All the tests pass with the two additional commits.

Old Windows Rubies (2.4 thru 2.6) seem to be failing, so I skipped them.  All Windows Rubies from 2.7 and later are passing.

LevitatingBusinessMan:
Anything still to be done for this PR?

LevitatingBusinessMan:
Can we continue looking into this? Currently puma does not follow the rack spec and I need to use a workaround for https://github.com/puma/puma/issues/3634.

PR #3650, opened 2025-04-26
ci: Add Ubuntu and Windows arm jobs 

MSP-Greg:
### Description

GHA now supports arm64 runners on Ubuntu & Windows.

Added jobs for both.

Note that the Ubuntu jobs are affected by an unsigned vs signed char bug, probably in ragel, as the failures related to the c source code generated by it.  I tried a few things, and got Ruby 2.7 passing, but the other versions failed.  Gave up.

Disabled the 'PUMA_MAKE_WARNINGS_INTO_ERRORS=true' workflow step in the Ubuntu arm64 jobs, and the jobs passed...

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -113,7 +113,9 @@ jobs:
         run: ruby .github/workflows/github_actions_info.rb
 
       - name: set WERRORFLAG
-        if: ${{ needs.skip_duplicate_runs.outputs.should_skip != 'true' }}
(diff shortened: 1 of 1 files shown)


MSP-Greg:
When I started on this, I was pessimistic about it working.  Given that all the Ubuntu arm jobs were passing, jobs should be added in the matrix, rather than by using 'includes', as is done here.

Hence closing in favor of #3651

PR #3652, opened 2025-05-02
Update docs to show correct default value for worker_shutdown_timeout

phallstrom:
The documentation used to say 'The default is 60 seconds.' for
`worker_shutdown_timeout` but the actual default value is 30 as defined
in `lib/puma/configuration.rb` https://github.com/puma/puma/blob/bc648da28d4332276293b6ae21157c0b892ec392/lib/puma/configuration.rb#L170

### DIFF
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -1107,7 +1107,7 @@ module Puma
 
     # Set the timeout for worker shutdown.
     #
-    # The default is 60 seconds.
+    # The default is 30 seconds.
     #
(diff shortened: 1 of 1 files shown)


dentarg (approved):

dentarg:
Thanks! Looks like there was a mistake in https://github.com/puma/puma/commit/2f284bd5810b4cc6771887c28c287da6468adbc3#diff-122c03b9c96f060a9cc29b5ed3446467e1909b0054b4f5cfd95ee41cfff832e7R1082

Issue #3654, opened 2025-05-12
Introduce a way for Puma to fast fail when waiting pool is full


shayonj:
Apologies in advance if this feature already exists. I tried replicating and searching and couldn't, hence this feature request. 

**Is your feature request related to a problem? Please describe.**

When Puma's thread pool is full and `queue_requests` is set to `false`, the server currently blocks when accepting a new connection until a worker thread becomes available (`server.rb#handle_servers` calls [`@thread_pool.wait_until_not_full`](https://github.com/puma/puma/blob/979b9b0325737e7c2f2d0e7b56b7e94a6331952c/lib/puma/server.rb#L367)). This leaves the client waiting without knowing if the delay is due to application processing time or server capacity saturation.

While a client can set a `read_timeout`, this doesn't distinguish between a slow application response and waiting for an available Puma worker. Relying on a client-side `open_timeout` is also ineffective. This is because the TCP connection is typically established successfully by the operating system's kernel and queued for Puma *before* Puma's internal logic checks for worker thread availability (IIUC). The `open_timeout` applies to the kernel-level TCP handshake, not the subsequent wait for a Puma worker thread after the connection is `accept()`-ed by the server.

It would be really nice for Puma to provide immediate feedback to the client in this saturation scenario, allowing clients to implement faster fallback or retry logic instead of waiting potentially indefinitely for a worker.

**Describe the solution you'd like**

Introduce an option (e.g., `reject_when_full_no_queue`) that, when `queue_requests` is false and the thread pool is at maximum capacity, causes Puma to immediately reject newly accepted connections. Ideally, this rejection could involve sending a 503 Service Unavailable status code before closing the connection, providing clear feedback to the client.

**Describe alternatives you've considered**

For applications, particularly smaller ones or those in local network environments (e.g., with sidecars), a client or an associated component could periodically make a "pre flight like request" to a health endpoint (e.g., `/ping` or `/status`) that returns Puma's internal statistics (like `Puma.stats`). If this health check times out (e.g., within 1 second) or the stats indicate the pool is full, the client infers that the server is at capacity and doesn't send the request to the server it's meant to. This approach adds overhead, relies on an indirect inference rather than direct feedback on a specific request, and requires custom client-side logic.

Additional context

[...]

PR #3655, opened 2025-05-12
Introduce the ability for Puma to reject request when pool is full

shayonj:
### Description

This pull request introduces a mechanism for Puma to provide immediate feedback to clients when it's operating without a request queue (`queue_requests: false`) and its worker thread pool is fully saturated.

**What original problem led to this PR?**
Currently, when `queue_requests` is `false` and all worker threads are busy, Puma's main server loop blocks while waiting for a thread to become available (`@thread_pool.wait_until_not_full`). From a client's perspective, the TCP connection is established, but no response is forthcoming for an indeterminate amount of time. This makes it difficult for clients to distinguish between a temporarily saturated server and a slow application response. Standard client-side timeouts (`open_timeout`, `read_timeout`) don't provide sufficient clarity or immediate feedback in this specific scenario. The goal is to allow clients to implement faster fallback, retry, or load-shedding logic.

**Are there related issues / prior discussions?**
https://github.com/puma/puma/issues/3654

**What alternatives have been tried? Does this supersede previous attempts?**

Yes, the alternative is mentioned in https://github.com/puma/puma/issues/3654

**Solution:**

This change introduces a new boolean option, `reject_when_pool_is_full` (defaulting to `false`).

When `queue_requests` is `false` and `reject_when_pool_is_full` is `true`:
If a new connection is accepted while all worker threads (`@thread_pool.busy_threads`) are at `@max_threads`, Puma will:
1.  Accept the connection.
2.  Immediately send a minimal `HTTP/1.1 503 Service Unavailable` response with `Connection: close`.
3.  Close the connection.

This provides immediate feedback to the client that the server is temporarily unable to handle the request due to saturation, allowing for faster client-side fallback or retry logic.

If `queue_requests` is `true` or `reject_when_pool_is_full` is `false`, Puma retains its existing behavior.

**Implementation Details:**

*   Added `reject_when_pool_is_full` option to `Puma::Configuration::DEFAULTS` and `Puma::DSL`.
*   Modified `Puma::Server#initialize` to recognize the new option.
*   In `Puma::Server#handle_servers`, before the `pool.wait_until_not_full` call, a check is added for the new option and pool status (`!@queue_requests && @reject_when_pool_is_full && @thread_pool.busy_threads >= @max_threads`).
*   If the condition is met, a new private method `Puma::Server#handle_full_capacity_rejection(socket)` is called, which encapsulates the accept, 503 response, and close logic.

**Testing:**

*   Added tests in `test/test_puma_server.rb` to cover:
    *   Scenario where the option is enabled, pool is full, and 503 is correctly returned/connection is rejected.
    *   Scenario where the option is enabled, but the pool is not full, and requests are processed normally.

- [x] I have added (or updated) appropriate tests if this PR fixes a bug or adds a feature. (We've added one, and discussed others).
- [x] My pull request is 100 lines added/removed or less so that it can be easily reviewed. (Self-note: You'll need to check this. The current changes are likely within this or close.)
- [x] I have updated the documentation accordingly. (We've updated DSL comments; full README/docs would be next).
- [x] All new and existing tests passed, including Rubocop. (Assuming the tests we've added pass locally).

closes https://github.com/puma/puma/issues/3654


shayonj:
Looks like the failures are coming from the latest ruby's `HEAD` branch so likely unrelated to this PR(?)

shayonj:
Will rebase after https://github.com/puma/puma/pull/3656 is live to fix the unrelated CI failures.

schneems:
I'm not immediately saying no, but I'm a little iffy about whether or not we want this. I'm leaning cool. Talking out loud a bit: Any functionality we add adds additional constraints to future development, sometimes in non-intuitive ways. For example, the "out of band" feature impacts our ability to refactor away some locks in favor of a queue strategy: https://github.com/puma/puma/discussions/3768 (for example).

Zooming out, the goal is to expose information to distinguish between "puma is overloaded, so this was slow" and "the rack app itself was slow." I can see that information being valuable. I'm unsure of the mechanism, though. I'm also unsure if we want this to be tightly coupled to only the non-queue-requests case (which I've previously said I don't like supporting https://github.com/puma/puma/pull/3524). Or if it should be more general. For example, passing some metrics to Rack to allow an upstream-consumer (like a Rack middleware) to handle the response for us instead of Puma serving a fixed response.

Re-reading that issue of mine, this stands out:

> The problem described does not come from having a reactor that buffers request bodies, but rather from the application ACCEPT-ing a new request from the socket before it had capacity to process it. Disabling the reactor was effectively an implementation detail.

I think this changed with Puma 7 as we took out that code in the accept loop that limits the number of requests that can be stopped. As mentioned in my research, the original intent was that the request would be blocked. But your issue is from May, from before that point. So there shouldn't have ever been a case where the internal TODO queue was higher than the max threads. Or if there was, it would have been a minor blip. The majority of the queueing/waiting of the system would have been on the socket rather than internally inside of Puma.

Most systems handle this by having a proxy between the open internet and Puma and recording X-Request-Start in a trusted header (must come from a trusted source), then comparing that value against the current time to determine if queueing is happening or not. This assumes the systems' clocks are reasonably synced (NTP). That strategy can already happen in a middleware, and it's how systems like JudoScale work and how APMs like Newrelic report "queueing" metrics. 

I think that's enough to give you what you want, but I'm even more dubious that `queue_requests` is performing its original stated job of "preventing deadlocks when making requests to the same server. I'm curious why you are running with `queue_requests` turned off.

Recap:

- You can get the info you want via using a proxy and comparing the request queue time. Can you use that along with a middleware to serve a 503?
- I'm curious about your use case for turning off `queue_requests`

schneems:
I'm not immediately saying no, but I'm a little iffy about whether or not we want this. I'm leaning cool. Talking out loud a bit: Any functionality we add adds additional constraints to future development, sometimes in non-intuitive ways. For example, the "out of band" feature impacts our ability to refactor away some locks in favor of a queue strategy: https://github.com/puma/puma/discussions/3768 (for example).

Zooming out, the goal is to expose information to distinguish between "puma is overloaded, so this was slow" and "the rack app itself was slow." I can see that information being valuable. I'm unsure of the mechanism, though. I'm also unsure if we want this to be tightly coupled to only the non-queue-requests case (which I've previously said I don't like supporting https://github.com/puma/puma/pull/3524). Or if it should be more general. For example, passing some metrics to Rack to allow an upstream-consumer (like a Rack middleware) to handle the response for us instead of Puma serving a fixed response.

Re-reading that issue of mine, this stands out:

> The problem described does not come from having a reactor that buffers request bodies, but rather from the application ACCEPT-ing a new request from the socket before it had capacity to process it. Disabling the reactor was effectively an implementation detail.

I think this changed with Puma 7 as we took out that code in the accept loop that limits the number of requests that can be stopped. As mentioned in my research, the original intent was that the request would be blocked. But your issue is from May, from before that point. So there shouldn't have ever been a case where the internal TODO queue was higher than the max threads. Or if there was, it would have been a minor blip. The majority of the queueing/waiting of the system would have been on the socket rather than internally inside of Puma.

Most systems handle this by having a proxy between the open internet and Puma and recording X-Request-Start in a trusted header (must come from a trusted source), then comparing that value against the current time to determine if queueing is happening or not. This assumes the systems' clocks are reasonably synced (NTP). That strategy can already happen in a middleware, and it's how systems like JudoScale work and how APMs like Newrelic report "queueing" metrics. 

I think that's enough to give you what you want, but I'm even more dubious that `queue_requests` is performing its original stated job of "preventing deadlocks when making requests to the same server. I'm curious why you are running with `queue_requests` turned off.

Recap:

- You can get the info you want via using a proxy and comparing the request queue time. Can you use that along with a middleware to serve a 503?
- I'm curious about your use case for turning off `queue_requests`

schneems:

```
```

sagium:
@shayonj considering I'm seeing around ~30% faster response times (on high traffic only) with `queue_requests: false` on a Kubernetes setup with Nginx ingress and Puma in cluster mode (3 workers, 6 threads, 16 pods), compared to `queue_requests: true`.

I consider a mechanism of failing fast super important and useful, so thanks for bringing this to the table :+1:

schneems:
@sagium interesting, thanks for the data point. What version of puma are you using? 7.1.0 or is it older like 6.6.1? What Ruby version?

I'm wondering where the 30% is coming from. If NGINX is buffering all request bodies, then the reactor thread should mostly be empty and not running. Maybe there's an opportunity for some benchmarking and improvement, like having the reactor block on a Queue which would help hint to the OS that there's nothing for it to do (so doesn't need to be scheduled). Purely guessing

sagium:
I've checked the [main branch](https://github.com/puma/puma/blob/cf7f6bd5fd87082822f53413c0381cfdc7a40009/lib/puma/server.rb#L319-L333) and indeed the code is quite different. I'll need some time to see the changes since indeed I'm referring to the latest 6.x not 7.x of Puma.

So, this PR doesn't really apply now but the concept is still valid. Although I have yet to test v7 to production or a good enough benchmark with wrk or similar.

The problem with growing `Puma.stats` backlogs (>1) appear in production with v6 when traffic picks up. Even though I can tackle the backlog by adding threads (was mostly I/O load) or having more pods with our HPA in kubernetes. Tackling the backlog did not give me much faster responses though. 

Nevertheless, with the exact same setup and a backlog >0, when setting `queue_requests: false` I get a backlog of 0 (expected) but also 30% faster response times, meaning that the code bypassed by `@queue_requests` had to play a role in that.

Still, I would have to see the effect with the v7 since it makes little sense to reason about v6 with the code diverged.

sagium:
@schneems i see that now `queue_requests` is simply a `use_reactor?` flag.

I wonder why do we need to check it [here](https://github.com/puma/puma/blob/main/lib/puma/server.rb#L320) since the existence of the reactor depends on the flag then `reactor_wakeup` is used only when `queue_requests: true`.

I mean inside that method `queue_requests` is always `true`, no?

schneems:
> see that now queue_requests is simply a use_reactor? flag.

Correct. That’s always been the core meaning, but it’s confusing since turning it on or off affects other code, and there has been some loosely coupled behavior (like blocking new requests from accepting when all threads are busy.

> I wonder why do we need to check it [here](https://github.com/puma/puma/blob/main/lib/puma/server.rb#L320) since the existence of the reactor depends on the flag then reactor_wakeup is used only when queue_requests: true.

I’m not sure. Check the blame. I’ve not touched that class recently so the history is not fresh in my memory.

nateberkopec:
@shayonj Hey man, I really appreciate you putting this together. I'm sorry it took so long for me to get a review in on this, but I was waiting for AGI to be achieved first. Now that it's here, I have time to actually dig through our backlog.

My problem here is that most load balancers that Puma is behind do not really work well with receiving a 503. HAProxy, Nginx, and Heroku's router mesh all interpret a 503 as "this host is down, and I should stop routing traffic to it," potentially for very long periods of time. It makes the host unhealthy.

But a queuing condition, which is what you're describing here, is not unhealthy. Some level of queuing is natural in a setup, and you should be measuring that via the X-Request-Start header and using that across your whole infrastructure to figure out how saturated your hosts are.

If this were a permanent unhealthy deadlock condition, then it would make sense to return 503. We're saying, "this thing is screwed. Don't come back."

I think also that the semantics of HTTP make it very difficult for a load balancer to retry requests safely, which limits the usefulness of this behavior. On some load balancers with limited configurability, this feature is just not usable.

So I'm struggling to see the case for inclusion here. What do you think I'm missing?

After reading Richard's comments and noting the behavior change in Puma 7, with all that in mind, I think I'm going to close this. Not open to adding it in here, but I just see a substantial number of barriers, and available workarounds.

shayonj:
Sorry for the late reply. I agree yeah, it felt a bit odd place in puma. However, I really like @schneems idea of passing the `X-Request-Start` and using that from a middleware in the app to make "fast exit decisions". Thanks for the brainstorming everyone 🙏🏾

PR #3658, opened 2025-05-23
Add a new API to update the min & max thread count dynamically

yuki24:
### Description

@nateberkopec and I are working on something that automatically identifies the optional worker count on the fly and updates the thread count accordingly and dynamically. The algorithm we are actively working on will be proprietary, but we think that it makes sense to add an API to dynamically change the thread count could be part of the open source Puma, hence the PR here.

I have looked into Puma's source code, and noticed that this isn't extremely complicated, as Puma's thread pool already looks at the `min` and `max` to not allocate too many threads or trim unused threads dynamically. What this means is that in order to achieve what we want to do, we can simply set a new number to `min` or/and `max`.

This PR includes a simple change where `#update_worker_count` was added to do that. The accessors for `min` and `max` have also been added to the `ThreadPool`. I know this may seem a bit scary, but that's basically what dynamic update means so I thought it might be fine.

Please let me know what you all think.

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -680,6 +680,11 @@ module Puma
       @binder.add_unix_listener path, umask, mode, backlog
     end
 
+    def update_worker_count(min: @thread_pool.min, max: @thread_pool.max)
+      @thread_pool.min, @thread_pool.max = min, max
+      @min_threads, @max_threads = min, max
+    end
+
     # @!attribute [r] connected_ports
     def connected_ports
       @binder.connected_ports
--- a/lib/puma/thread_pool.rb
+++ b/lib/puma/thread_pool.rb
@@ -78,6 +78,7 @@ module Puma
     end
 
     attr_reader :spawned, :trim_requested, :waiting
+    attr_accessor :min, :max
 
     def self.clean_thread_locals
       Thread.current.keys.each do |key| # rubocop: disable Style/HashEachMethods
--- a/test/test_puma_server.rb
+++ b/test/test_puma_server.rb
@@ -2156,4 +2156,21 @@ class TestPumaServer < PumaTest
     server_run(min_threads: 2, max_threads: 2, auto_trim_time: 1)
     refute @pool.instance_variable_get(:@auto_trim)
   end
+
+  def test_update_worker_count
+    server_run(min_threads: 1, max_threads: 1, auto_trim_time: 1)
+
+    assert_equal 1, @server.stats[:running]
+
+    @server.update_worker_count(min: 3, max: 3)
+
+    # By making multiple requests, we can ensure that the pool is filled up to the max.
+    10.times { send_http_read_response(GET_11) }
+
+    assert_equal 3, @pool.max
+    assert_equal 3, @pool.min
+    assert_equal 3, @server.max_threads
+    assert_equal 3, @server.min_threads
+    assert_equal 3, @server.stats[:running]
+  end
 end


dentarg:
Is `update_worker_count` a good name? Could it be confused with worker _processes_?

How would one use this? Could you show and example dummy app using this?

MSP-Greg:
As @dentarg mentioned.

Maybe change `update_worker_count` to `update_thread_pool_min_max`?  I've never liked ThreadPool related methods using the term `worker`.

Re an example, it's not clear how to use this with several workers running (just a quick look)...

yuki24:
Thanks for your replies, @dentarg and @MSP-Greg! And apologies for a long delay. I have updated the method name to `update_thread_pool_min_max`. I am open to suggestions for a better name if you have any.

With regard to how we use this method, we set up a timer that periodically calculates the optimal worker count based on the metrics we collect with the `gvl_timing` gem, and then we call `update_thread_pool_min_max` to adjust the thread pool size. The code snippet below shows how we do this:

```ruby
class OurAwesomeGem
  ...

  def start_timer
    @timer ||= Thread.new do
      loop do
        avg_duration, avg_io, avg_gvl = average_metrics_from_gvl_timing

        max = calculate_optimal_worker_count(avg_duration, avg_io, avg_gvl)

        # This is a `Puma::Server` object. `min` could always be 1, so this is setting `max` only.
        @server.update_thread_pool_min_max(max: max)

        sleep interval
      end
    end
  end
  
  ...
end
```

In order to grab the `Puma::Server` object, we use a Puma plugin that hooks into the Puma lifecycle events. The code 
for the plugin is as follows:

```ruby
# our_awesome_gem/lib/puma/plugin/our_awesome_gem.rb so this could be loaded with
# `--plugin our_awesome_gem`:
Puma::Plugin.create do
  def start(launcher)
    launcher.config.configure do |user_dsl, *|
      if launcher.config.options[:workers] > 0
        # Cluster mode:
        user_dsl.on_worker_boot do
          OurAwesomeGem.start_timer
        end

        user_dsl.on_thread_start do
          OurAwesomeGem.puma_server ||= Puma::Server.current
        end
      else
        # Single mode:
        user_dsl.on_booted do
          OurAwesomeGem.puma_server = Puma::Server.current
          OurAwesomeGem.start_timer
        end
      end
    end
  end
end
```

The second commit in this PR 1a3a682c77912266fbce5af28e0ed43b6a83fe53 has been added. Without this the `Puma::Server.current` would return `nil` in the `on_thread_start` hook when running in cluster mode. Similarly, the same call in the `on_booted` hook would return `nil` in single mode without 1a3a682c77912266fbce5af28e0ed43b6a83fe53.

---------------

Another thing I noticed was that It seems like Puma's lifecycle events and hooks had been added as people needed them, and there seems to be room for improvement in the API design. I wonder if we could think about how to make these events and hooks more predictable, regardless of cluster/single mode? Issues like https://github.com/sidekiq/sidekiq/issues/6661 have already been reported, so it would be great if I could get some feedback on this.

MSP-Greg:
@yuki24

Not sure if there was a reason for not doing so, if not, please rebase?  I can have a look tomorrow.  We may need an 'integration' test for this, especially if a clustered test is done...

MSP-Greg:
@yuki24

Good day.  By using the following, each request is sent, the response is generated, and the 'client' reads it.  Then the next request is sent.  So, it's not filling up the server with a 'backlog' of requests.

```ruby
10.times { send_http_read_response(GET_11) }
```

The following does create a 'backlog' and the test passes.  It creates the requests with a small delay in the app.

```ruby
  def test_update_thread_pool_min_max
    @app = ->(env) do
      sleep 0.001
      [200, {}, [env['rack.url_scheme']]]
    end

    server_run(min_threads: 1, max_threads: 1, auto_trim_time: 1)

    assert_equal 1, @server.stats[:running]

    @server.update_thread_pool_min_max(min: 3, max: 3)

    # By making multiple requests, we can ensure that the pool is filled up to the max.
    req_ary = send_http_array 20
    resp_ary = req_ary.map { |req| req.read_response }

    assert_equal 3, @pool.max
    assert_equal 3, @pool.min
    assert_equal 3, @server.max_threads
    assert_equal 3, @server.min_threads
    assert_equal 3, @pool&.spawned
  end
```

**EDIT:**

I should probably have more coffee, but if one increases the `ThreadPool#min`, I'm not sure if it will affect a server that's 'idle'.

MSP-Greg:
@yuki24 and @nateberkopec

I pushed the test fix, CI is passing.  Question, and this pertains to the PR, and maybe Puma:

What happens if a `min` value is set higher than the `max` value?  A warning would certainly be expected, the `min` value could be ignored, or the `max` could be set to the new `min` value.  Not.sure.

**EDIT:** also, what happens if `max` is set lower than `min`?  Again, not sure if this is covered

lib/puma/server.rb
```diff
@@ -680,6 +680,11 @@ def add_unix_listener(path, umask = nil, mode = nil, backlog = 1024)
       @binder.add_unix_listener path, umask, mode, backlog
     end
 
+    def update_thread_pool_min_max(min: @thread_pool.min, max: @thread_pool.max)
+      @thread_pool.min, @thread_pool.max = min, max
+      @min_threads, @max_threads = min, max
```
MSP-Greg:
```suggestion
      if min <= max and min > 0
        @thread_pool.min, @thread_pool.max = min, max
        @min_threads, @max_threads = min, max
      end
```

lib/puma/thread_pool.rb
```diff
@@ -105,7 +106,9 @@ def backlog
 
     # @!attribute [r] pool_capacity
     def pool_capacity
-      waiting + (@max - spawned)
+      capacity = waiting + (@max - spawned)
+
+      capacity < 0 ? 0 : capacity
```
yuki24: We do need a test case for this.

yuki24:
@MSP-Greg Thank you for reviewing! Really appreciated the test fix ❤️ 

> What happens if a min value is set higher than the max value? A warning would certainly be expected, the min value could be ignored, or the max could be set to the new min value. Not.sure.
> 
> EDIT: also, what happens if max is set lower than min? Again, not sure if this is covered

I just merged your suggestion. I also think we may want to raise an exception in both the cases. At the same time, I don't think I fully understand the risk of raising an exception from within a web server. It's probably safer to just warn than an exception. What do you think?

lib/puma/thread_pool.rb
```diff
@@ -115,7 +116,9 @@ def backlog_max
...
```
schneems: Optional suggestion, I think this is a little more clear:

```suggestion
      capacity.clamp(0, Float::INFINITY)
```

lib/puma/server.rb
```diff
@@ -726,6 +726,21 @@ def add_unix_listener(path, umask = nil, mode = nil, backlog = 1024)
...
+        return
+      end
+
+      if min <= 0
+        @log_writer.log "`min' value cannot be less than 0"
+        return
+      end
+
+      @thread_pool.min, @thread_pool.max = min, max
+      @min_threads, @max_threads = min, max
```
schneems: If we want the metrics from `def stats` in the pool to be correct, I think we need to wrap setting these values in a mutex call.

There's other logic that also uses this value like `trim`

```ruby
      with_mutex do
        free = @waiting - @todo.size
        if (force or free > 0) and @spawned - @trim_requested > @min
```

And

```ruby
    def <<(work)
      with_mutex do
        if @shutdown
          raise "Unable to add work while shutting down"
        end

        @todo << work
        t = @todo.size
        @backlog_max = t if t > @backlog_max

        if @waiting < @todo.size and @spawned < @max
          spawn_thread
        end
```

We might want to introduce more, finer-grained, locks in the future. Because right now there's only one and it's "stop everything" which includes stopping enqueuing and dequing of requests.

lib/puma/cluster/worker.rb
```diff
@@ -57,14 +57,16 @@ def run
...
-        server = @server ||= start_server
+          server = @server ||= start_server
         rescue Exception => e
           log "! Unable to start worker"
           log e
           log e.backtrace.join("\n    ")
           exit 1
         end
 
+        Thread.current.puma_server = server
```
schneems: Is the puma server class already exposed as a public API somehow somewhere?

If not, instead of exposing all of the puma server, which exposes a large API and makes internal refactoring harder in the future. I propose introducing a smaller, isolated class with a limited API instead, that class can hold a reference to the server and delegate to it, but it means that only `update_thread_pool_min_max` would become stable. Something like

```ruby
# Please add documentation that explicitly states this is a public API here
#
# Add a description of what it does
class PumaServerPluginControl
  # Please note that the constructor is private and may change
  def initialize(server: )
    @server = server
    @thread_pool = server.thread_pool
   end
   
  # Please add documentation that explicitly states this is a public API here
  #
  # Add a description of what it does
  def update_thread_pool_min_max(min: @thread_pool.min, max: @thread_pool.max)
    server.update_thread_pool_min_max(min: min, max: max)
  end
end
```

lib/puma/server.rb
```diff
@@ -726,6 +726,21 @@ def add_unix_listener(path, umask = nil, mode = nil, backlog = 1024)
       @binder.add_unix_listener path, umask, mode, backlog
     end
 
+    def update_thread_pool_min_max(min: @thread_pool.min, max: @thread_pool.max)
```
schneems: Please add documentation for the new API

lib/puma/cluster/worker.rb
schneems: Actually, looking at https://github.com/puma/puma/pull/3774/files it seems were already exposing the PumaServer via thread local. So I think we can remove the duplicate setting. I'm unsure what that's used for, but it's clearly already part of the public API, so ignore that part of my request.

schneems:

```
```

lib/puma/thread_pool.rb
yuki24: Nice, TIL `#clamp`

lib/puma/server.rb
yuki24: Addressed by d2d7baa1.

nateberkopec:
Changed to waiting for changes, @yuki24 and I discussed this and I think we can do more to not require the exposure of `Puma.server`. I think we've addressed all other concerns.

yuki24:
I have updated the PR and rebased against `main`. The main change here is that there is a new `ServerPluginControl` class, which is basically what @schneems [suggested above](https://github.com/puma/puma/pull/3658#discussion_r2436818878):

```ruby
module Puma
  class ServerPluginControl
    def initialize(server)
      @server = server
      @thread_pool = server&.thread_pool
    end

    ...

    def update_thread_pool_min_max(min: @thread_pool.min, max: @thread_pool.max)
      @server.update_thread_pool_min_max(min: min, max: max)
    end
  end
end
```

The reason why this was introduced was that while the `server` object is exposed through `Thread.current`, this was never intended to be entirely public, and we may want to limit the ability for plugins to possibly e.g. mutate the `server` object too much. In the future, we may want to move the `stats` methods as we go.

The failing builds seem to be failing due to the same error, which I'm not sure how I could fix:

```
The git source https://github.com/rack/rack is not yet checked out. Please run `bundle install` before trying to start your application (Bundler::GitError)
```

lib/puma/server_plugin_control.rb
```diff
@@ -0,0 +1,33 @@
...
+    def min_threads
+      @server.min_threads
+    end
+
+    # Updates the minimum and maximum number of threads in the thread pool.
+    #
+    # @see Puma::Server#update_thread_pool_min_max
+    #
+    def update_thread_pool_min_max(min: @thread_pool.min, max: @thread_pool.max)
+      @server.update_thread_pool_min_max(min: min, max: max)
```
nateberkopec, line 29: ServerPluginControl is instantiated from ThreadPool#spawn_thread before Server#thread_pool is assigned (ThreadPool.new returns before @thread_pool is set). With @thread_pool cached as nil, default args here call @thread_pool.min/@thread_pool.max and raise NoMethodError in before_thread_start. Consider resolving the pool lazily (e.g., @server.thread_pool) when computing defaults.

lib/puma/server.rb
```diff
@@ -730,6 +731,49 @@ def add_unix_listener(path, umask = nil, mode = nil, backlog = 1024)
...
+    #   server.update_thread_pool_min_max(max: 16)
+    #
+    def update_thread_pool_min_max(min: @thread_pool.min, max: @thread_pool.max)
+      if min > max
+        @log_writer.log "`min' value cannot be greater than `max' value."
+        return
+      end
+
+      if min <= 0
+        @log_writer.log "`min' value cannot be less than 0"
```
nateberkopec, line 766: Puma's default min_threads is 0. This guard rejects min <= 0, so update_thread_pool_min_max(max: X) on a default-config server logs a warning and ignores the update. Allowing min == 0 (check min < 0) keeps it consistent with existing config semantics.

nateberkopec:
For posterity: codex 5.2 xhigh found those two issues, I watched it verify them manually, and I attest they are correct.

lib/puma/server.rb
yuki24: Addressed in the cf6dcd4a932bf45f2aa658194bf44a31015b4188.

lib/puma/server_plugin_control.rb
yuki24: Removed the reference to the `@thread_pool` and just used `@min_threads` and `@max_threads`. That's a lot safer and we already have a guard against `@thread_pool` not initialized in the `server#update_thread_pool_min_max` so this should be addressed now.

nateberkopec (approved):

lib/puma/server.rb
```diff
@@ -730,6 +730,49 @@ def add_unix_listener(path, umask = nil, mode = nil, backlog = 1024)
...
+    #
+    # @example Update only the minimum threads
+    #   server.update_thread_pool_min_max(min: 4)
+    #
+    # @example Update only the maximum threads
+    #   server.update_thread_pool_min_max(max: 16)
+    #
+    def update_thread_pool_min_max(min: @min_threads, max: @max_threads)
+      if min > max
+        @log_writer.log "`min' value cannot be greater than `max' value."
```
schneems, line 761: Consistency with the rest of the codebase

```suggestion
        @log_writer.log "`min' value cannot be greater than `max' value"
```

test/test_puma_server.rb
```diff
@@ -2261,4 +2261,66 @@ def test_auto_trim_with_fixed_pool_size
...
+    # By making multiple requests, we can ensure that the pool is filled up to the max.
+    req_ary = send_http_array 5
+    req_ary.map { |req| req.read_response }
+
+    assert_equal 3, @pool.max
+    assert_equal 3, @pool.min
+    assert_equal 3, @server.max_threads
+    assert_equal 3, @server.min_threads
+    assert_equal 3, @pool&.spawned
+  end
```
schneems, line 2287: It's better to not have to rely on sleep. This is less likely to be accidentally flappy in the future.

```suggestion
  def test_update_thread_pool_min_max
    mutex = Mutex.new
    condition_variable = ConditionVariable.new
    blocked = true

    @app = ->(env) do
      mutex.synchronize do
        condition_variable.wait(mutex) while blocked
      end
      [200, {}, [env['rack.url_scheme']]]
    end

    server_run(min_threads: 1, max_threads: 1, auto_trim_time: 1)

    assert_equal 1, @server.stats[:running]

    @server.update_thread_pool_min_max(min: 3, max: 3)

    req_ary = send_http_array 5

    start = Time.now
    Thread.pass until @pool.spawned >= 3 || Time.now - start > 5

    mutex.synchronize do
      blocked = false
      condition_variable.broadcast
    end

    req_ary.map { |req| req.read_response }

    assert_equal 3, @pool.max
    assert_equal 3, @pool.min
    assert_equal 3, @server.max_threads
    assert_equal 3, @server.min_threads
    assert_equal 3, @pool&.spawned
  end
```

schneems (approved):
No major issues, made some suggestions, please apply before merging or suggest alternatives.

yuki24:
Addressed the comments and rebased against `main` and the builds are all ✅ Let me know if there is anything else I can do here.

Issue #1889, opened 2019-08-06
Replace C-extension with pure Ruby


nateberkopec:
@ioquatix is getting good results with async-http https://github.com/socketry/async-http. 

It's 2019 and the JIT is Becoming a Thing in CRuby, which makes me think that the marginal gains of using the old Mongrel parser is probably outweighed by the simplicity of removing our native extensions.

This would also greatly improve the Windows experience.

nateberkopec:
FWIW I mean the HTTP parsing part. We may be stuck on the SSL part, but that's a separate issue.

evanphx:
No, we should not do this. The JIT'd code will never be as fast as the current code and we get nothing expect new bugs from doing it.

nateberkopec:
I don't think it will be faster, I just think it would be Fast Enough. Adding 1ms of latency to every request would be worth getting rid of C IMO.

But the fact that this is decade(?)-old battle-tested code is of course true and not to be treated lightly.

ioquatix:
Writing code in C has its own limitations but maybe I should clarify my position.

Parsing HTTP/1 is not that hard. But doing it robustly can be. I’ve done both - a Ragel parser in C and pure Ruby implementation which is done by hand and is fast enough but probably not 100% correct. My plan is to eventually either deprecate HTTP/1 or use Ragel... but the current implementation is fine for now.

I would say if you have working code don’t bother changing it... but it’s definitely a limitation if you want to support the JIT, JRuby, TruffleRuby.

As pointed out, C is definitely faster and less memory intensive, but parsing the HTTP/1 request is probably less than 1ms, and probably way less... raw response bodies with content length or connection close are almost zero overhead, chunked encoding is a bit more overhead but most of that ends up invoking native code anyway...

evanphx:
If someone can show a pure-ruby http parser that only adds 1ms to requests I'll certainly entertain the idea.

nateberkopec:
.@ioquatix's stuff could be used to a "spike" a test: https://github.com/socketry/protocol-http and https://github.com/socketry/protocol-http1.

boazsegev:
Late to the party, but I would add that C improves memory usage over time, not just speed (AFAIK).

This is related to the fact that the C parser doesn't produce as many side effects (temporary objects) as a Ruby implementation is likely to require.

[...]

PR #3660, opened 2025-06-02
[Experiment] Pure ruby parser

swebb:
### Description

As an experiment/learning exercise I decided to try implementing a HTTP parser in pure ruby. See https://github.com/puma/puma/issues/1889.  I did this purely as an exercise and not in an attempt to get it merged. I'm posting in the hope it might provide an interesting data point when comparing the C parser vs a pure ruby one.

I took inspiration from [Tenderlove's blog post](https://tenderlovemaking.com/2023/09/02/fast-tokenizers-with-stringscanner/) and used StringScanner.

# FAQ

* Why didn't you use `protocol-http`?

Somehow when I was reading through [the issue](https://github.com/puma/puma/issues/1889) initially I missed those comments. That would have been a good idea. Then again, building it from scratch was part of the fun of learning about http parsers.

* Is this a valid HTTP parser?

Probably not. I wrote it to pass the test suite and only briefly referred to the RFCs. No doubt there are bugs & security issues. Note that while it passes most of the test suite, there are still about 6 failing tests.

* Why did you change the number of connections & threads used by `wrk` in the `hello.sh` benchmark?

For reasons I don't understand if I ran `wrk` with 4 connections and 4 threads against my ruby parser it would only do 4096 requests (1024 per thread). The latency statistics would still be very low so I'm not sure what was causing the delay.

* Why do you warm up the YJIT?

In practice the pure ruby parser will be significantly slower for the first request than the C extension due to the time taken to compile the ruby into VM instructions. As Puma processes are typically long lived I felt like including the YJIT compilation step would be an unfair comparison.

* Is this a fair comparison?

I don't know ¯_(ツ)_/¯ The `hello.sh` benchmark feels interesting but also artificial to me. I don't think it's representative of a production workload, and that could skew the usefulness of these results.

I did a little performance optimisation of the ruby parser but I'm not expert in that area. It feels a little unfair to compare the optimised C parser against a poorly optimised ruby parser; but then, performance is a big part of the question around it a ruby parser so the comparison is kind of the point.

# STOP WRITING CAVEATS AND GET TO THE RESULTS!

On my x86 laptop running Linux I get:

```
====================================
C parser
====================================
Warming up YJIT
  8 threads and 16 connections
  860021 requests in 30.01s, 63.87MB read
Requests/sec:  28654.45
Transfer/sec:      2.13MB

Thread Stats         Avg       Stdev         Min         Max    +/- Stdev
     Latency:       1.33ms      1.59ms     41.00us     11.64ms   79.23%
     Req/sec:       3.60k     169.12        2.43k       3.96k    85.71%

Latency Distribution
     50%  171.00us
     75%    2.59ms
     90%    3.96ms
     99%    5.12ms

Request Stats        Avg       Stdev         Min         Max    +/- Stdev
    Req/conn:      53.75k      96.45       53.63k      53.97k    68.75%
====================================
Ruby parser
====================================
Warming up YJIT
  8 threads and 16 connections
  685083 requests in 30.01s, 50.88MB read
Requests/sec:  22826.75
Transfer/sec:      1.70MB

Thread Stats         Avg       Stdev         Min         Max    +/- Stdev
     Latency:       1.67ms      2.00ms     49.00us     12.94ms   79.05%
     Req/sec:       2.87k     147.58        1.80k       3.19k    88.42%

Latency Distribution
     50%  219.00us
     75%    3.24ms
     90%    4.98ms
     99%    6.43ms

Request Stats        Avg       Stdev         Min         Max    +/- Stdev
    Req/conn:      42.82k      60.09       42.71k      42.94k    75.00%
```

The C parser is definitely faster; but the ruby results look pretty good to me.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -39,7 +39,23 @@ file 'ext/puma_http11/org/jruby/puma/Http11Parser.java' => ['ext/puma_http11/htt
 end
 task :ragel => ['ext/puma_http11/org/jruby/puma/Http11Parser.java']
 
-if !Puma.jruby?
+if ENV["PUMA_PURE_RUBY_PARSER"]
+  task :compile => ['ext/puma_http11/puma_http11.rb'] do
+    src = 'ext/puma_http11/puma_http11.rb'
+    dest = 'lib/puma/puma_http11.rb'
+    if File.exist?(dest)
+      if File.symlink?(dest)
+        puts "Using ruby parser (already symlinked)"
+      else
+        puts "Refusing to symlink ruby parser. Did you forget to run `rake clean`?"
+      end
+    else
+      puts "Symlinking ruby parser"
+      FileUtils.symlink(src, dest, relative: true)
+    end
+  end
+  CLEAN.include "lib/puma/puma_http11.rb"
+elsif !Puma.jruby?
   # compile extensions using rake-compiler
   # C (MRI, Rubinius)
   Rake::ExtensionTask.new("puma_http11", gemspec) do |ext|
--- a/benchmarks/wrk/hello.sh
+++ b/benchmarks/wrk/hello.sh
@@ -1,8 +1,32 @@
 # You are encouraged to use @ioquatix's wrk fork, located here: https://github.com/ioquatix/wrk
 
-bundle exec bin/puma -t 4 test/rackup/hello.ru &
-PID1=$!
-sleep 5
-wrk -c 4 -d 30 --latency http://localhost:9292
+benchmark() {
+  bundle exec bin/puma -t 4 test/rackup/hello.ru &
+  PID1=$!
+  sleep 5
+  echo "Warming up YJIT"
+  wrk -c 4 -d 30 --latency http://localhost:9292 > /dev/null
+  echo "Running"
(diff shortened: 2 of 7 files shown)


schneems:
Looking at the numbers 28654.45 /  22826.75 # => 0.79x slower is not a deal breaker, but it's not ideal. The HTTP parsers don't get a ton of PRs, they're not a large maintenance burden (IMHO). I think this is an interesting undertaking, but represents a net new risk (if there are areas where a new parser diverges from the old) and is also slower. I won't discourage experimenting, and appreciate the effort. I'm leaning "not right now" on this.

swebb:
Thanks for looking at it @schneems. I think that's a sensible approach.

nateberkopec:
Definitely interesting and thank you for the effort of spiking this 🙏

tekknolagi:
@swebb any interest in adding this as a benchmark to https://github.com/ruby/ruby-bench ? We track performance of the interpreter and YJIT on these benchmarks over time. If you are interested but don't have time, you can also let us know if it's okay for us to integrate it on our own.

swebb:
@tekknolagi I'll check it out and see what I can do.

PR #3681, opened 2025-07-31
Add support for `rack.response_finished`

byroot:
It is very similar to `rack.after_reply`, but is part of the Rack spec.

It can't just be an alias because the spec state it has to invoke callbacks in reverse order.

Fix: https://github.com/puma/puma/pull/3631 (The PR seems stalled?)
Ref: https://github.com/rack/rack/issues/1777
Ref: https://github.com/rack/rack/pull/1802

### DIFF
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -52,6 +52,7 @@ module Puma
       io_buffer = client.io_buffer
       socket  = client.io   # io may be a MiniSSL::Socket
       app_body = nil
+      error = nil
 
       return false if closed_socket?(socket)
 
@@ -124,7 +125,7 @@ module Puma
         @log_writer.unknown_error error, client, "Rack app"
         @log_writer.log "Detected force shutdown of a thread"
 
-        status, headers, res_body = lowlevel_error(e, env, 503)
+        status, headers, res_body = lowlevel_error(error, env, 503)
       rescue Exception => error
         @log_writer.unknown_error error, client, "Rack app"
 


MSP-Greg:
@byroot

Thanks for these PR's.

Question, which I've wondered about, but never tested:

Given `obj`, which of the following is better:
```ruby
# assume env["rack.response_finished"] is allowed to be nil
if temp = env["rack.response_finished"]
  temp.each { # some op }
end
```
-- or --
```ruby
temp = env["rack.response_finished"] || []
temp.each { # some op }
```

Obviously, we probably shouldn't create extra arrays, or create one for every request, so the code above wouldn't be exactly the same if added to Puma...

byroot:
> Given `obj`, which of the following is better:

The first version saves an allocation, and saves a method call if the value is `nil`. That the only difference.

But since the key is pretty much guaranteed to be there (unless the application delete the key for some reason), it really doesn't make any substantial difference in reality.

MSP-Greg:
Sorry.  Posted before I consumed enough coffee.  Kind of rooted in a frustration for a method similar to `not_empty?` which would be true if an object could be enumerated with at least one loop, with the assumption that `nil` cannot.

I may do a future PR adding these keys to Binder's `@proto_env` with a `nil` value, and also add a constant `RACK_RESPONSE_FINISHED`...

Thanks.

MSP-Greg (approved):

schneems:
I merged in your other PR this one now has a conflict. I'm trying to get the keepalive fix out the door, to do that I'm trying to cut a release 6.6.1 and then have main target v7. As this is an addition, it would need to either be 6.7 (would need a new codename) or go out in 7.x.

byroot:
> also add a constant `RACK_RESPONSE_FINISHED`...

Right, I wasn't sure if all keys needed a constant or not. I can easily add it.

byroot:
Rebased.

lib/puma/request.rb
```diff
@@ -92,6 +92,7 @@ def handle_request(client, requests)
       # array, we will invoke them when the request is done.
       #
       env[RACK_AFTER_REPLY] ||= []
+      env[RACK_RESPONSE_FINISHED] ||= []
```
samuel-williams-shopify, line 96:
```suggestion
      response_finished = []
      env[RACK_RESPONSE_FINISHED] = response_finished
```

Do we actually need `||=` here?

lib/puma/request.rb
```diff
@@ -144,6 +145,16 @@ def handle_request(client, requests)
...
+
+      if response_finished = env[RACK_RESPONSE_FINISHED]
+        response_finished.reverse_each do |o|
+          begin
+            o.call
+          rescue StandardError => e
+            @log_writer.debug_error e
+          end
+        end
+      end
```
samuel-williams-shopify:
```suggestion
      response_finished.reverse_each do |o|
        begin
          o.call
        rescue StandardError => e
          @log_writer.debug_error e
        end
      end
```

MSP-Greg: @samuel-williams-shopify

Thanks.

Re the actual call, should it be something similar to:
```ruby
o.call(env, status, headers, e)
```

byroot:
I updated the PR to pass `env`, `status`, `headers` and `error`.

MSP-Greg:
@byroot

Sorry for jumping in, I noticed the CI crash & burn, assumed it was related to the `e` to `error` rename.

I'm fine with the conditional, better 'sleep factor'...

@schneems?

lib/puma/request.rb
samuel-williams-shopify: Yes, that's correct. This is now strictly clarified in the `rack-head` spec, so if you run against `Rack::Lint` in `Rack::Head` and have some tests for this, it will validate the callback.

PR #3682, opened 2025-07-31
v7.0.0.pre1

schneems:
Thanks for all the great work @MSP-Greg. New codename is an album from Return to Forever.

image

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,3 +1,8 @@
+## 7.0.0.pre1 / 2025-07-31
+
+* Changed
+  * Fix long tail response problem with keepalive connections ([#3678])
+
 ## 6.6.1 / 2025-07-30
 
 * Bugfixes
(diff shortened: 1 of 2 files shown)


joshuay03:
I noticed [this comment](https://github.com/puma/puma/pull/3678#pullrequestreview-3068713305) and was wondering if [these v7-labeled PRs](https://github.com/puma/puma/pulls?q=sort%3Aupdated-desc+is%3Apr+is%3Aopen+label%3Av7) will be considered for the next release? I've also requested that https://github.com/puma/puma/pull/3602 be considered.

MSP-Greg:
Personally, I hope PR's with and without the 'v7' label will be reviewed and hopefully merged.  Another prerelease version is also a possibility.

Testing of the long-tail fix may take some time.

schneems:
Thanks for the callout. I'm good to introduce more changing behavior as long as it wouldn't be a strong incompatibility with a lot of apps (preventing them from upgrading to the new keepalive behavior) or otherwise have possible performance implications that would complicate debugging and diagnosing issues raised with the keepalive feature. (Other reviewers don't feel blocked on a review from me now that main is targeting v7). I will also aim to review these.

If it makes sense to do a 7 and then an 8 in quick succession, I'm okay to do that too. The only real downside to more releases with Puma is needing a change in codename. In general, I'm not terribly scared of major versions (this repo is much younger but already at major version 10 https://github.com/heroku/buildpacks-ruby/blob/8fce2f32dc82d7daf90d58f4d28f6dab936b8d25/buildpacks/ruby/buildpack.toml#L5). I prefer to a closer-to-semver style of versioning than a "major is for marketing" (kinda like this satire that hits a little too close to home https://mastodon.online/@nikitonsky/113691789641950263). 

I would like to get an actual release out before RailsWorld (Sept 4th-5th) at the latest.

PR #3692, opened 2025-08-12
ci: tests.yml - set 'allow-failure' to false for non MRI jobs

MSP-Greg:
### Description

Having recently jumped back into Puma, I've noticed that non-MRI jobs (JRuby & TruffleRuby) seem to be stable, so I think it's time to remove the 'allow-failure' that happens with many non-MRI jobs.

The PR changes the setting from `true` to `false`, rather than removing the key value.

Additionally, people involved with both JRuby & TruffleRuby have contributed to Puma, so it seems appropriate to support both platforms, especially their head builds.


schneems:
Seems reasonable. Happy to try it and toggle some tests back to false if they end up being flappy.

PR #3700, opened 2025-08-19
[ci skip] Add initial Puma support document

schneems:
The goal is to encode what we're doing today into something an external observer can reference. This is NOT to be used to overrule a maintainer's future decisions or to take power out of their hands. This should eventually become a set of "sensible defaults" that make it easier for contributors and users to understand desired behavior and our general roadmap.

I'm interested in specific feedback about details listed here (please make it as an inline code suggestion if possible). I'm also interested in whether there are other external things we're relying on (rack spec is external, ruby version is external, etc.) that should be added to this document.

### DIFF
--- a/docs/puma_support.md
+++ b/docs/puma_support.md
@@ -1,0 +1,66 @@
+# Puma Support Policy
+
+## What
+
+This document's purpose is to help users and contributors of this library make future plans based on the trajectory of this project. It is loosely worded and should be expected to change over time. Maintainers and contributors are encouraged to follow the guidance here but are not strictly bound by it.
+
+## What does "not supported" mean?
+
+Puma is a community-supported and run project. Support entitles the community to open issues and file PRs. It does not guarantee or warranty any action regarding those issues and PRs. All issues and PRs can be closed at a maintainer's discretion, regardless of whether they qualify under "support." If an issue falls outside of Puma support, maintainers are not obliged to keep it open, and if a PR falls outside of Puma support, maintainers are not obliged to review it.
+
+## Puma release support policy
+
+By default, Puma supports one release version, the latest release. That means if there is a bug in 3.2.1 (made-up numbers for example only), but the latest release is 4.3.2, the bug must be reproduced in 4.3.2 to qualify for support.
+
+When a new major (and sometimes minor) version is released, Puma maintainers will try to preserve the status of the old version via branches and/or tags. Puma maintainers can choose to backport fixes to older versions and cut a release of older versions, but this behavior should not be assumed.
+
+If a bug report cannot be reproduced with the latest Puma version, it may be closed.
+
+## Ruby version support policy
+
+ "Ruby version" refers to the https://github.com/ruby/ruby implementation. Puma also supports other Ruby implementations (the origin of Puma involves an alternate Ruby implementation, rubinius) see below.
+
+Puma supports "Ruby upstream + 2 versions". Ruby upstream support is documented at this page https://www.ruby-lang.org/en/downloads/branches/. At this time of writing, it is 3 "major" versions, where a major version is X.Y with X.Y.Z being a patch release. If a Ruby version is not EOL, it is supported. Because we support "+ 2 versions", if 3.2 is the last non-EOL version, Puma will support 3.1 and 3.0 as well. The Puma team recommends you always use the latest Ruby release.
+
+When Puma supports a Ruby version, it will include that version in the CI matrix. The gemspec specifies the minimum supported Ruby versions. If a bug is reported, the report must be reproducible on a currently supported Ruby version.
+
+Puma may rely on upstream support to fix bugs in Ruby rather than working around them. That means, if a bug exists only in a "+2 version" that is no longer supported upstream, it may be closed (it's not guaranteed that we will workaround legacy Ruby issues).
+
+## Alternate implementations of Ruby (version support policy)
+
+An example of an "alternate" implementation of Ruby is JRuby. Ruby implementations typically target a specific "spec" of ruby/ruby, for example, JRuby 10.0.2.0 targets ruby/ruby version 3.4.2. The target version of a supported implementation must be within the above "Ruby version support policy." Additionally, the implementation release must be supported upstream.
+
+Puma supports the JRuby implementation. Puma supports the TruffleRuby implementation.
+
+## Rack (and SPEC) version support
+
+From the current Rack documentation https://github.com/rack/rack?tab=readme-ov-file#version-support, Rack lists support levels to be:
+
+- Bug fixes and security patches
+- Security patches only
+- End of support
+
+Puma will support any Rack version that still accepts "bug fixes" in addition to security patches.
+
+Due to the need to support "hot" reloading applications (via phased restart) which might have different Rack versions, [Puma cannot directly declare the minimum Rack version](https://github.com/puma/puma/commit/537bc21593182cd9c4c0079a3936d05b1f91fe14). Puma may use runtime or boot time checks to warn or error about Rack version support.
+
+## How to use this document
+
+You can use the guidelines here to make suggestions to contributors (anyone reporting an issue, helping with via a comment, or opening a PR) or maintainers. For example, "I could not reproduce your issue using a currently supported Ruby version. Can you please update to version X.Y and verify the issue persists?" Please link to this doc, state what action you're requesting, and why you feel that action is supported by this document. Please act in "good faith" and do not "bikeshed" (the practice of focusing on non-structural or seemingly irrelevant details).
+
+## How to update this document?
+
+Generally, you shouldn't suggest changes to policies in this document unless you have committed to the Puma project.
+
+Non-committers:
+
+- Spelling fixes are fine; do not re-grammar or "refactor" the wording.
+- Updating stale details derived from the policies is fine. For example, if a canonical URL changes or if specific upstream versions are listed.
+
+Committers:
+
+- Suggest whatever policy changes you feel are appropriate.
+- Suggest improvements to the wording or consistency or presentation of information.
+- Do not merge policy changes without oversight of other "significant" contributors.
+- When reviewing a policy change, please mark comments as either blocking or non/blocking. Please try to make blocking comments either satisfiable or falsifiable.
+


docs/puma_support.md
```diff
@@ -0,0 +1,66 @@
...
+
+When a new major (and sometimes minor) version is released, Puma maintainers will try to preserve the status of the old version via branches and/or tags. Puma maintainers can choose to backport fixes to older versions and cut a release of older versions, but this behavior should not be assumed.
+
+If a bug report cannot be reproduced with the latest Puma version, it may be closed.
+
+## Ruby version support policy
+
+ "Ruby version" refers to the https://github.com/ruby/ruby implementation. Puma also supports other Ruby implementations (the origin of Puma involves an alternate Ruby implementation, rubinius) see below.
+
+Puma supports "Ruby upstream + 2 versions". Ruby upstream support is documented at this page https://www.ruby-lang.org/en/downloads/branches/. At this time of writing, it is 3 "major" versions, where a major version is X.Y with X.Y.Z being a patch release. If a Ruby version is not EOL, it is supported. Because we support "+ 2 versions", if 3.2 is the last non-EOL version, Puma will support 3.1 and 3.0 as well. The Puma team recommends you always use the latest Ruby release.
```
MSP-Greg:
```suggestion
Puma supports "Ruby upstream + 2 versions". Ruby upstream support is documented at this page https://www.ruby-lang.org/en/downloads/branches/. At this time of writing, it is 3 "minor" versions, where a minor version is X.Y with X.Y.Z being a patch release. If a Ruby version is not EOL, it is supported. Because we support "+ 2 versions", if 3.2 is the last non-EOL version, Puma will support 3.1 and 3.0 as well. The Puma team recommends you always use the latest Ruby release.
```

docs/puma_support.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Rack (and SPEC) version support
+
+From the current Rack documentation https://github.com/rack/rack?tab=readme-ov-file#version-support, Rack lists support levels to be:
+
+- Bug fixes and security patches
+- Security patches only
+- End of support
+
+Puma will support any Rack version that still accepts "bug fixes" in addition to security patches.
```
skipkayhil, line 43: To be explicit: does this mean Puma intends to only support Rack 3.2 (the current bugfix release)?

MSP-Greg: Current Rack doc shows:

| Version  | Support                            |
|----------|------------------------------------|
|    3.2.x | Bug fixes and security patches.    |
|    3.1.x | Security patches only.             |
|    3.0.x | End of support.                    |
|    2.2.x | Security patches only.             |
| <= 2.1.x | End of support.                    |

Puma will support Rack versions 2.2.x, 3.1.x, and 3.2.x.  Current CI tests against current Rack 2 & 3 versions.

schneems: This section has my biggest unknowns. Rack has fewer releases than Ruby and the relationship with the release cadence isn't quite as understood (on my end). As written 3.2 is 100% for sure supported. In practice 3.1 is as well. Regarding 2.2 - I'm not sure how long upstream plans to maintain security patch support. I think if the 2.2 suite isn't causing problems we continue to support it.

docs/puma_support.md
```diff
@@ -0,0 +1,66 @@
+# Puma Support Policy
```
schneems, line 1: We have bundler version specific code too, should add a policy for that as well.

```
    if bundler_ver < Gem::Version.new('2.1.0')
      Bundler.with_clean_env { yield }
    else
      Bundler.with_unbundled_env { yield }
    end
```

MSP-Greg: See the various jobs in https://github.com/MSP-Greg/github-actions-ruby-info/actions.  It doesn't run as a cron job, so if they're old (or already purged) LMK.

Ruby 3.0 has bundler 2.2.33 installed.  But, some of the std items may change between patch versions.

MSP-Greg:
Should it be made clear that Puma only runs CI on the latest patch version of each Ruby minor version, and we recommend using it?

schneems:
> Should it be made clear that Puma only runs CI on the latest patch version of each Ruby minor version, and we recommend using it?

Yes. I think that's a more correct/complete description of what we're doing today. If someone files a bug with 3.3.8 that won't reproduce with 3.3.9. 

It's a bit of a question of what we do when there's a bug with the latest that is eventually fixed. I.e. if there's a bug in 3.3.9 and we add a workaround for it, that is later fixed in 3.3.10, does that workaround code stay in the codebase until 3.3 is no longer supported? What if someone files a bug with the workaround? 

Historically we've left those one-off patch problem fixes in the codebase. Though maybe we also should be adding warnings to nudge people away from those edge cases.

docs/puma_support.md
```diff
@@ -0,0 +1,66 @@
...
+
```
schneems, line 2: CPU Architecture and/or Operating system support. Right now we have 2 versions of ubuntu, 3 versions of mac OS and 1 version of windows. We also test an additional ARM architecture ubuntu image.

For windows, our support is "ci will run on at least A version of windows." For Mac there's a policy but it seems not clearly spelled out on their pages (but is observable) this is second hand https://endoflife.date/macos. For Ubuntu there's a policy https://ubuntu.com/about/release-cycle. But I also think we can support X versions even if they're not all on CI but I'm not sure what that looks like.  

I'm wondering how valuable/important some of those variants are in CI. Could we get away with only testing A version of mac (the most recent) and the most recent ubuntu on ARM and AMD architectures? That woul allow us to drop 1 ubuntu and 2 macos versions in CI. That would take us from 7 to 4 OS variants which would represent (in theory a 42% decrease in number of CI runs).

IDK how sensitive the C extensions and updates are to these OS-es.

MSP-Greg: Re CI runs, IDK.  Currently, OSS orgs have concurrency limits of 20 ubuntu/windows jobs, and 5 macOS jobs.

If a goal is to shorten CI, I think the macOS jobs are currently defining how long CI takes.  I'm not a longtime macOS user, but are there many differences between macos-14 & macos-15?  If not, we could reduce the number of jobs in one or the other.

Re Windows, we are testing on both x64 and arm, for quite a while there was a bug in MSVC that screwed Windows Ruby builds.  That's been fixed (by MSFT), so we could switch a few jobs to windows-2025.

schneems: > If a goal is to shorten CI

Right now, it's a maintainability/sustainability goal. Virtually no PR passes without having to re-run something and it takes a long time to wait for the full runs, and contributors without commit cannot trigger a re-run on their own. Fewer total tests run hopefully mean fewer flappy failures, and overall decreased PR-to-review-to-merge time. 

It's relevant to the "what we support" because how we define "supported" and how we deliver that "support" affect the costs to deliver changes to puma (negatively or positively). If we can find a policy that allows us to safely claim maximum support with minimal CI overhead that's ideal. But only if we think it's reasonable (in the same way that we will only run the latest Ruby patch version on CI).

> are there many differences between macos-14 & macos-15

As an end user, not really. But there's occasionally stuff like https://bugs.ruby-lang.org/issues/21111 which can affect compilation. I'm usually on the latest version or at most one version behind.

I kind of wish there was a better way to get some metrics like "X number of prs were blocked by suite Y" and be able to dig into which ones. I could ask on Mastodon, perhaps.

docs/puma_support.md
```diff
@@ -0,0 +1,66 @@
...
+If a bug report cannot be reproduced with the latest Puma version, it may be closed.
+
+## Ruby version support policy
+
+ "Ruby version" refers to the https://github.com/ruby/ruby implementation. Puma also supports other Ruby implementations (the origin of Puma involves an alternate Ruby implementation, rubinius) see below.
+
+Puma supports "Ruby upstream + 2 versions". Ruby upstream support is documented at this page https://www.ruby-lang.org/en/downloads/branches/. At this time of writing, it is 3 "minor" versions, where a minor version is X.Y with X.Y.Z being a patch release. If a Ruby version is not EOL, it is supported. Because we support "+ 2 versions", if 3.2 is the last non-EOL version, Puma will support 3.1 and 3.0 as well. The Puma team recommends you always use the latest Ruby release.
+
+When Puma supports a Ruby version, it will include that version in the CI matrix. The gemspec specifies the minimum supported Ruby versions. If a bug is reported, the report must be reproducible on a currently supported Ruby version.
+
```
MSP-Greg, line 26:
```suggestion
When Puma supports a Ruby version, it will include that version in the CI matrix. The gemspec specifies the minimum supported Ruby versions. If a bug is reported, the report must be reproducible using the most recent patch version of a currently supported Ruby minor version.

```

MSP-Greg: As discussed, maybe some additional text referring to 'most recent Ruby patch version'.

Not sure if the above is clear or whether an example would help.  Something like:

As an example, the current Ruby 3.4 patch version is 3.4.5.  If a Puma bug is found with Ruby 3.4.4 (or earlier) that isn’t reproducible with Ruby 3.4.5, it is up to the Puma team to decide if a Puma fix will be accepted.

docs/puma_support.md
```diff
@@ -0,0 +1,66 @@
...
+- Spelling fixes are fine; do not re-grammar or "refactor" the wording.
+- Updating stale details derived from the policies is fine. For example, if a canonical URL changes or if specific upstream versions are listed.
+
+Committers:
+
+- Suggest whatever policy changes you feel are appropriate.
+- Suggest improvements to the wording or consistency or presentation of information.
+- Do not merge policy changes without oversight of other "significant" contributors.
+- When reviewing a policy change, please mark comments as either blocking or non/blocking. Please try to make blocking comments either satisfiable or falsifiable.
+
```
schneems, line 66: Should we add big name players even if they're "downstream" from us i.e. Rails and Sinatra? Thinking of this discussion https://github.com/puma/puma/pull/3706/files/fe3ac571f834366c9e483ab27a6776777be6158b#r2301887270.

dentarg: That these projects can suggest policy changes?

schneems: The location of my comment isn't relevant, I just needed a newline to start a new discussion. Interesting thought though.

I think Aaron or Raphael should be able to ask us personally "hey can you do us a favor and support XYZ" but its ultimately our call. I don't think we suggest that those external teams control setting our policy. Though there is precedent for some overlap between teams 😅 .

We could match the Ruby version policy and say "EOL + 2 years." That policy would reinforce the decision to continue supporting Sinatra 3.2.0 as it won't be two years until December 29, 2025. Rails 7.2 is EOL August 9th 2025 https://rubyonrails.org/maintenance so supporting it until August 9th 2027 seems reasonable.

schneems:
For completeness: I just swapped the default branch name from `master` to `main`, you don't need to do anything for this PR. You'll need to update your local environment accordingly:

```
```

PR #3702, opened 2025-08-23
Fix handling of invalid Transfer-Encoding header errors

marshall-lee:
- `chunked,gzip` case was handled incorrectly. It should be reported as an unknown (unsupported) value.
- spaces were not stripped so something like `chunked, gzip, chunked` also was handled incorrectly.

### DIFF
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -414,14 +414,15 @@ module Puma
           te_ary = te_lwr.split(',').each { |te| te.gsub!(STRIP_OWS, "") }
           te_count = te_ary.count CHUNKED
           te_valid = te_ary[0..-2].all? { |e| ALLOWED_TRANSFER_ENCODING.include? e }
-          if te_ary.last == CHUNKED && te_count == 1 && te_valid
-            @env.delete TRANSFER_ENCODING2
-            return setup_chunked_body body
-          elsif te_count > 1
+          if te_count > 1
             raise HttpParserError   , "#{TE_ERR_MSG}, multiple chunked: '#{te}'"
+          elsif te_ary.last != CHUNKED
+            raise HttpParserError   , "#{TE_ERR_MSG}, last value must be chunked: '#{te}'"
           elsif !te_valid
             raise HttpParserError501, "#{TE_ERR_MSG}, unknown value: '#{te}'"
           end
+          @env.delete TRANSFER_ENCODING2
+          return setup_chunked_body body
         elsif te_lwr == CHUNKED
           @env.delete TRANSFER_ENCODING2
           return setup_chunked_body body
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -56,6 +56,11 @@ module Puma
 
     TE_ERR_MSG = 'Invalid Transfer-Encoding'
 
+    # See:
+    # https://httpwg.org/specs/rfc9110.html#rfc.section.5.6.1.1
+    # https://httpwg.org/specs/rfc9112.html#rfc.section.6.1
+    STRIP_OWS = /\A[ \t]+|[ \t]+\z/
+
     # The object used for a request with no body. All requests with
     # no body share this one object since it has no state.
     EmptyBody = NullIO.new
@@ -406,8 +411,7 @@ module Puma
       if te
         te_lwr = te.downcase
         if te.include? ','
-          te_ary = te_lwr.split ','
-          te_ary.each(&:strip!)
+          te_ary = te_lwr.split(',').each { |te| te.gsub!(STRIP_OWS, "") }
           te_count = te_ary.count CHUNKED
           te_valid = te_ary[0..-2].all? { |e| ALLOWED_TRANSFER_ENCODING.include? e }
           if te_ary.last == CHUNKED && te_count == 1 && te_valid
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -407,12 +407,13 @@ module Puma
         te_lwr = te.downcase
         if te.include? ','
           te_ary = te_lwr.split ','
+          te_ary.each(&:strip!)
           te_count = te_ary.count CHUNKED
           te_valid = te_ary[0..-2].all? { |e| ALLOWED_TRANSFER_ENCODING.include? e }
           if te_ary.last == CHUNKED && te_count == 1 && te_valid
             @env.delete TRANSFER_ENCODING2
             return setup_chunked_body body
-          elsif te_count >= 1
+          elsif te_count > 1
             raise HttpParserError   , "#{TE_ERR_MSG}, multiple chunked: '#{te}'"
           elsif !te_valid
             raise HttpParserError501, "#{TE_ERR_MSG}, unknown value: '#{te}'"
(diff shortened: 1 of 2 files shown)


MSP-Greg:
@marshall-lee

Thank you for the PR.  For reference, see [RFC 9110 Field Value Lists](https://httpwg.org/specs/rfc9110.html#rfc.section.5.6.1.1), which shows optional white space (OWS) is allowed before and after the comma.  Also, OWS is `[ \t]*` (space and htab).

Another reference is [RFC 9112 Transfer-Encoding](https://httpwg.org/specs/rfc9112.html#rfc.section.6.1)

lib/puma/client.rb
```diff
@@ -56,6 +56,12 @@ class Client # :nodoc:
 
     TE_ERR_MSG = 'Invalid Transfer-Encoding'
 
+    # See:
+    # https://httpwg.org/specs/rfc9110.html#rfc.section.5.6.1.1
+    # https://httpwg.org/specs/rfc9112.html#rfc.section.6.1
+    OWS_PREFIX = /\A[ \t]+/
+    OWS_SUFFIX = /[ \t]+\z/
```
MSP-Greg:
```suggestion
    STRIP_OWS = /\A[ \t]+|[ \t]+\z/
```

lib/puma/client.rb
```diff
@@ -406,13 +412,16 @@ def setup_body
       if te
         te_lwr = te.downcase
         if te.include? ','
-          te_ary = te_lwr.split ','
+          te_ary = te_lwr.split(',').each do |te|
+            te.gsub!(OWS_PREFIX, "")
+            te.gsub!(OWS_SUFFIX, "")
+          end
```
MSP-Greg:
```suggestion
          te_ary = te_lwr.split(',').map { |s| s.gsub(STRIP_OWS, '') }
```

MSP-Greg:
@marshall-lee

Thanks, JFYI, earlier I benched `split` with a regex vs `split` with a character, following by a map with a strip, the two step was marginally faster.  Have tried the current.

lib/puma/client.rb
MSP-Greg: Sorry, modified the above, cut and pasted incorrectly

MSP-Greg:
Question - I'm wondering about what error should be raised, especially if the value list doesn't meet the specs of [RFC 9112 6.1 - Transfer-Encoding](https://httpwg.org/specs/rfc9112.html#field.transfer-encoding)].

A 501 seems odd if 'chunked' is in the value list anywhere but as the last entry?

Or, when should anything other than 400 be raised?  If a list item is unknown?

Any thoughts?  Otherwise LGTM.  Thanks.  I'll look at #3703 soon (maybe tomorrow my time (-0500))

marshall-lee:
> A 501 seems odd if 'chunked' is in the value list anywhere but as the last entry?
> Or, when should anything other than 400 be raised? If a list item is unknown?

My guess:

I think "unknown" was an inaccurate wording initially. Here "unknown" means not an "unknown to the specs" but rather "unknown to the server". Perhaps it's better to phrase this "unsupported". When the server can't deal with some parts of the request input, it does respond with "501 Not Implemented" code. It indicates just the fact that `Transfer-Encoding` header value is technically valid but the server can't process it. For example:

 - Puma supports `Transfer-Encoding: gzip, chunked` which means that the request body was first gzipped then chunked. In this case, Puma deals with chunking layer but the gzip deflation job is performed on the Rack level by something like `Rack::Deflator`.
 - B̵u̵t̵ ̵P̵u̵m̵a̵ ̵d̵o̵e̵s̵ ̵n̵o̵t̵ ̵s̵u̵p̵p̵o̵r̵t̵ ̵`̵T̵r̵a̵n̵s̵f̵e̵r̵-̵E̵n̵c̵o̵d̵i̵n̵g̵:̵ ̵c̵h̵u̵n̵k̵e̵d̵,̵ ̵g̵z̵i̵p̵`̵ ̵(̵t̵h̵o̵u̵g̵h̵ ̵i̵t̵ ̵s̵e̵e̵m̵s̵ ̵t̵o̵ ̵b̵e̵ ̵a̵ ̵t̵e̵c̵h̵n̵i̵c̵a̵l̵l̵y̵ ̵v̵a̵l̵i̵d̵ ̵v̵a̵l̵u̵e̵)̵.̵ ̵T̵o̵ ̵s̵u̵p̵p̵o̵r̵t̵ ̵`̵c̵h̵u̵n̵k̵e̵d̵,̵ ̵g̵z̵i̵p̵`̵ ̵w̵o̵u̵l̵d̵ ̵r̵e̵q̵u̵i̵r̵e̵ ̵P̵u̵m̵a̵ ̵t̵o̵ ̵i̵m̵p̵l̵e̵m̵e̵n̵t̵ ̵t̵h̵e̵ ̵d̵e̵f̵l̵a̵t̵i̵o̵n̵ ̵o̵f̵ ̵t̵h̵e̵ ̵r̵e̵q̵u̵e̵s̵t̵ ̵b̵o̵d̵y̵.̵ ̵B̵u̵t̵ ̵n̵o̵ ̵o̵n̵e̵ ̵s̵e̵e̵m̵s̵ ̵t̵o̵ ̵r̵e̵a̵l̵l̵y̵ ̵n̵e̵e̵d̵ ̵t̵h̵i̵s̵ ̵m̵o̵d̵e̵ ̵s̵o̵ ̵t̵h̵e̵r̵e̵'̵s̵ ̵n̵o̵ ̵r̵e̵a̵s̵o̵n̵ ̵f̵o̵r̵ ̵P̵u̵m̵a̵ ̵t̵o̵ ̵b̵o̵t̵h̵e̵r̵ ̵a̵b̵o̵u̵t̵ ̵t̵h̵e̵ ̵i̵m̵p̵l̵e̵m̵e̵n̵t̵a̵t̵i̵o̵n̵.̵ ̵E̵v̵e̵r̵y̵b̵o̵d̵y̵ ̵u̵s̵e̵s̵ ̵`̵g̵z̵i̵p̵,̵ ̵c̵h̵u̵n̵k̵e̵d̵`̵ ̵n̵o̵t̵ ̵`̵c̵h̵u̵n̵k̵e̵d̵,̵ ̵g̵z̵i̵p̵`̵.̵

MSP-Greg:
> But Puma does not support Transfer-Encoding: chunked, gzip (though it seems to be a technically valid value)

Sorry, I should have explained the question more.

My point is given the below from RFC 9112, `Transfer-Encoding: chunked, gzip` is not a 'technically valid value', and hence, should be a 4** error, as the header value/list is invalid from the spec?

The spec seems to imply that there is room for 'custom' encodings (not included in `ALLOWED_TRANSFER_ENCODING`), and that would probably be a 501 error.  I.think.

The error issue is really outside of the scope of this PR, but I was wondering what your thoughts were...

From RFC 9112 -

> If any transfer coding other than chunked is applied to a request's content, the sender MUST apply chunked as the final transfer coding

marshall-lee:
Ah yes you are correct.

marshall-lee:
> My point is given the below from RFC 9112, Transfer-Encoding: chunked, gzip is not a 'technically valid value', and hence, should be a 4** error, as the header value/list is invalid from the spec?

Yes, I just added a check for that.

MSP-Greg:
@marshall-lee

Thanks.  I wasn't expecting that.  If CI passes, I'll merge.

PR #3706, opened 2025-08-26
Remove HTTP_VERSION use SERVER_PROTOCOL instead

### LINKED ISSUE #3576: Puma 6+ breaks version header
**Describe the bug**
We have a Rails API app servicing an iPhone app. iPhones send automatically a `Version` header with the current app version and we are collecting and storing that data. With puma 5.6.9 running `request.headers["Version"]` would return `"HTTP/1.1, 8.1.1"` but when we updated to 6.5.0 it would return just `"HTTP/1.1"` breaking our application.

**Puma config:**

```
threads_count = ENV.fetch("RAILS_MAX_THREADS", 3)

port ENV.fetch("PORT", 3000)

# Allow puma to be restarted by `bin/rails restart` command.
plugin :tmp_restart

# Specify the PID file. Defaults to tmp/pids/server.pid in development.
# In other environments, only set the PID file if requested.
pidfile ENV["PIDFILE"] if ENV["PIDFILE"]
```

**To Reproduce**
Create fresh rails api application

```
rails new puma-test --api
# modify puma and rack versions as needed for testing
bundle install
rails db:create
```

Add code to check versions

```
rails g controller Versions

# in versions_controller.rb
class VersionsController < ApplicationController
  def index
    render json: { versions: request.headers["Version"] }
  end
end
```

Run it with: `rails s`

Run a curl request
```
curl -H 'Content-Type: application/json' -H 'Version: 8.1.1' http://localhost:3000/versions
```

**Expected behavior**
EXPECTED (= correctly executes in puma <6)
```
{"versions":"HTTP/1.1, 8.1.1"}
```
ACTUAL (puma >= 6)
```
{"versions":"HTTP/1.1"}
```

 - OS: Mac, Linux
 - Puma Version 5 vs 6

schneems:
Going back to 2010 the rack spec (https://github.com/rack/rack/commit/859ab9f1455d42a7e5879024ea9505ac810b1726) states:

```
<tt>HTTP_</tt> Variables:: Variables corresponding to the
                           client-supplied HTTP request
                           headers (i.e., variables whose
                           names begin with <tt>HTTP_</tt>). The
                           presence or absence of these
                           variables should correspond with
                           the presence or absence of the
                           appropriate HTTP header in the
                           request.
```

Puma used HTTP_VERSION to return the server protocol information i.e. HTTP/1.0 or HTTP/1.1 to the rack app. However we can see this violated the rack spec since it does not correspond to an HTTP_ key in a header provided by a client.

Rack added SERVER_PROTOCOL as an interface in Rack 3.0 and Puma adopted this in 2022 https://github.com/puma/puma/pull/2871 but aliased HTTP_VERSION to SERVER_PROTOCOL. This removes the alias and brings Puma closer to the spec for Rack 3.1+.

Still supporting the variable as recent software (Sinatra 3.2.0) still relies on it https://github.com/sinatra/sinatra/blob/4e8fdb5172a81c1c237388f264e5684a4a15ed4f/lib/sinatra/base.rb#L303-L315 and we want them to be able to use Puma 7.

close #3576

### DIFF
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -479,9 +479,13 @@ module Puma
         env[REMOTE_ADDR] = addr
       end
 
-      # The legacy HTTP_VERSION header can be sent as a client header.
-      # Rack v4 may remove using HTTP_VERSION.  If so, remove this line.
-      env[HTTP_VERSION] = env[SERVER_PROTOCOL]
+      # This assignment of HTTP_VERSION can be removed once Rack 2 is no longer supported
+      #
+      # - It is used in recently released software (Sinatra 3.2.0)
+      #   https://github.com/sinatra/sinatra/blob/4e8fdb5172a81c1c237388f264e5684a4a15ed4f/lib/sinatra/base.rb#L303-L315
+       if Object.const_defined?(:Rack) && Rack.release < "3.1.0"
+        env[HTTP_VERSION] = env[SERVER_PROTOCOL]
+       end
     end
     private :normalize_env
 


lib/puma/request.rb
```diff
@@ -478,10 +478,6 @@ def normalize_env(env, client)
 
         env[REMOTE_ADDR] = addr
       end
-
-      # The legacy HTTP_VERSION header can be sent as a client header.
-      # Rack v4 may remove using HTTP_VERSION.  If so, remove this line.
-      env[HTTP_VERSION] = env[SERVER_PROTOCOL]
```
dentarg, line 484: Shouldn't we check the Rack version use as discussed in the issue? I see you touched on this in the issue but we do have such checks: https://github.com/puma/puma/blob/1acac470543a8892fa921c98f61118edfa24cfa0/lib/puma/rack_default.rb#L14

I think this change will make old Sinatra versions (<= 3.2.0) not work with Puma 7: https://github.com/sinatra/sinatra/blob/v3.2.0/lib/sinatra/base.rb#L303-L315

dentarg (changes requested):

lib/puma/request.rb
schneems: Where is that code loaded? I added this on line 5 and ran the test suite and it did not print or error:

```
puts caller
raise "hello"
```

Seems bad that it's not under test at all.

Looking into it: While it's not required anywhere, it's added as almost an API to be used from the Gemfile https://github.com/puma/puma/pull/194. Calling this code like that, would restrict anyone from upgrading their rack version in a phased restart (as I read/understand it).

> I think this change will make old Sinatra versions (<= 3.2.0) not work with Puma 7: https://github.com/sinatra/sinatra/blob/v3.2.0/lib/sinatra/base.rb#L303-L315

Hmm. That prompts me to add another topic of conversation to the support thread https://github.com/puma/puma/pull/3700/files#r2302288234. I see https://rubygems.org/gems/sinatra/versions/3.2.0 was released in 2023. So it's relatively recent. And we still support Rack 2.2. 

Is 3.2.x dead? Could we do a 3.2.1 sinatra release that uses `SERVER_PROTOCOL` first? Even if it's not technically in the Rack spec for 2.x, `HTTP_VERSION` violates that spec.

Trying to add code like this blows up the test suite

```diff
$ git diff
diff --git a/lib/puma/request.rb b/lib/puma/request.rb
index 4d338fa2..1a68b570 100644
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -478,9 +478,28 @@ module Puma

         env[REMOTE_ADDR] = addr
       end
+
+      # Rack 3.1 removes HTTP_VERSION (which was technically not in the spec in Rack 2, but some clients depend on it)
+      # We can remove this once Rack 2 is no longer supported
+      if rack_version < GEM_VERSION_3_1_0
+        env["HTTP_VERSION"] = env[SERVER_PROTOCOL]
+      end
     end
     private :normalize_env

+    GEM_VERSION_3_1_0 = Gem::Version.new("3.1.0")
+    private_constant :GEM_VERSION_3_1_0
+
+    private def rack_version
+      # Don't require `rack` to be defined at boot to support phased restarts (rack version can change in a phased restart).
+      @_rack_version ||= begin
+        if !Object.const_defined?(:Rack)
+          require "rack"
+        end
+        Gem::Version.new(Rack.release)
+      end
+    end
```

```
$ be rake test 
...
ES......Error reached top of thread-pool: cannot load such file -- rack (LoadError)
```

We would need to overcome those failures somehow.

dentarg: > Where is that code loaded?

Oops, I meant to link https://github.com/puma/puma/blob/1acac470543a8892fa921c98f61118edfa24cfa0/lib/rack/handler/puma.rb#L133 which is code we do have under test. I believe `lib/puma/rack_default.rb` contains the same code. It seems it is user loaded, via https://github.com/puma/puma/pull/3061#discussion_r1084237327 I found https://github.com/puma/puma/pull/194

dentarg: I don't think we need to be requiring `rack`

I tested with Sinatra 3.2.0 and this PR and this diff and it does the trick

```diff
$ git diff
diff --git a/lib/puma/const.rb b/lib/puma/const.rb
index f1167dc6..b1b1e578 100644
--- a/lib/puma/const.rb
+++ b/lib/puma/const.rb
@@ -238,6 +238,7 @@ module Puma

     HTTPS_KEY = "HTTPS"

+    HTTP_VERSION = "HTTP_VERSION"
     HTTP_CONNECTION = "HTTP_CONNECTION"
     HTTP_EXPECT = "HTTP_EXPECT"
     CONTINUE = "100-continue"
diff --git a/lib/puma/request.rb b/lib/puma/request.rb
index 4d338fa2..2f7f4e05 100644
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -478,6 +478,8 @@ module Puma

         env[REMOTE_ADDR] = addr
       end
+
+      env[HTTP_VERSION] = env[SERVER_PROTOCOL] if Object.const_defined?(:Rack) && Rack.release < "3.1.0"
     end
     private :normalize_env

```

dentarg: "Not work" above was a too strong phrase from me, I'm sorry about that. The code affected is "just" to pick the HTTP status code used for redirects, for non GET requests. So anyone using old Sinatra (<=3.2.0) with new Puma (>=7) will see non-GET requests with redirect responses go from status [303](https://en.wikipedia.org/wiki/HTTP_303) to [302](https://en.wikipedia.org/wiki/HTTP_302).

dentarg: > Is 3.2.x dead? Could we do a 3.2.1 sinatra release that uses SERVER_PROTOCOL first?

In my mind it is pretty much dead, but there has actually been some talk about a 3.2.x release: https://github.com/sinatra/sinatra/pull/2092#issuecomment-2821435886

Rails had the same check, but if I read https://github.com/rails/rails/issues/53925 correctly, there's been backports made for several old versions?

schneems: Ahh. I see, you're bypassing the requirement to have rack available.

```
 do_register = Object.const_defined?(:Rack) && ::Rack.release < '3' 
```

FWIW I benchmarked this call versus some memoization and this is what I got:

```ruby
require 'benchmark/ips'
require 'rack'

puts "Benchmarking Object.const_defined?(:Rack)"
puts "Rack constant is defined: #{Object.const_defined?(:Rack)}"
puts

Benchmark.ips do |x|
  x.report("Object.const_defined?(:Rack)") do
    Object.const_defined?(:Rack) && Rack.release < "3.1.0"
  end

  x.report("memoized call") do
    if defined?(@_memoized)
      @_memoized
    else
      @memoized = Object.const_defined?(:Rack) && Rack.release < "3.1.0"
    end
  end

  x.compare!
end
```

```
$ be ruby benchmark_const_defined.rb
Benchmarking Object.const_defined?(:Rack)
Rack constant is defined: true

ruby 3.4.5 (2025-07-16 revision 20cda200d3) +PRISM [arm64-darwin24]
Object.const_defined?(:Rack)
                       862.138k i/100ms
       memoized call   802.544k i/100ms
Object.const_defined?(:Rack)
                          8.656M (± 4.5%) i/s  (115.52 ns/i) -     43.969M in   5.093549s
       memoized call      8.065M (± 1.0%) i/s  (123.99 ns/i) -     40.930M in   5.075239s

Comparison:
Object.const_defined?(:Rack):  8656413.1 i/s
       memoized call:  8065389.6 i/s - 1.07x  slower
```

So looks like the memoization is a needless optimization. I'll make the update soon.

schneems:
Closing in favor of https://github.com/puma/puma/pull/3711

PR #3710, opened 2025-08-27
Simplify `Puma::DSL#process_hook` logic

joshuay03:
### Description

Follow-up to https://github.com/puma/puma/pull/3438 and https://github.com/puma/puma/pull/3616.

In #3438, hooks were renamed and aliases were added for their deprecated names. The `process_hook` calls were also updated to pass the new method names as the fourth positional argument. Looking back, that PR probably should have passed `__callee__`/`__method__` instead, so that cluster-only hook warnings in single mode would use the correct method name. However, I think it's acceptable that only the new method name is shown, since users will see a deprecation warning with the old name before they encounter the single mode warnings, so they won't be entirely confused about what's being referenced.

In #3616, I refactored the warning logic. This PR follows up on that work because the key used to store these hooks in the options hash now matches the new method name, making it unnecessary to pass both values. I've maintained the approach of not referencing the old name and have removed the fourth argument entirely, which allows us to simplify `process_hook` a bit. See the comments for additional details.

**Note:** All these changes have comprehensive coverage in https://github.com/puma/puma/blob/master/test/test_config.rb.

### DIFF
--- a/lib/puma/configuration.rb
+++ b/lib/puma/configuration.rb
@@ -430,15 +430,16 @@ module Puma
       return if options[:workers] > 0
       return if options[:silence_fork_callback_warning]
 
-      @hooks.each do |key, method|
-        options.all_of(key).each do |hook_options|
+      log_writer = LogWriter.stdio
+      @hooks.each_key do |hook|
+        options.all_of(hook).each do |hook_options|
           next unless hook_options[:cluster_only]
 
-          LogWriter.stdio.log(<<~MSG.tr("\n", " "))
-            Warning: The code in the `#{method}` block will not execute
-            in the current Puma configuration. The `#{method}` block only
+          log_writer.log(<<~MSG.tr("\n", " "))
+            Warning: The code in the `#{hook}` block will not execute
+            in the current Puma configuration. The `#{hook}` block only
             executes in Puma's cluster mode. To fix this, either remove the
-            `#{method}` call or increase Puma's worker count above zero.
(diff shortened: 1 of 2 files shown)


lib/puma/configuration.rb
```diff
@@ -430,15 +430,16 @@ def warn_hooks
       return if options[:workers] > 0
       return if options[:silence_fork_callback_warning]
 
-      @hooks.each do |key, method|
-        options.all_of(key).each do |hook_options|
+      log_writer = LogWriter.stdio
```
joshuay03, line 433: We were initialising a logger per-log, which is unnecessary. Ref: https://github.com/puma/puma/blob/1acac470543a8892fa921c98f61118edfa24cfa0/lib/puma/error_logger.rb#L24-L26

lib/puma/dsl.rb
```diff
@@ -1463,19 +1463,19 @@ def add_pem_values_to_options_store(opts)
       end
     end
 
-    def process_hook(options_key, key, block, method, cluster_only: false)
-      raise ArgumentError, "expected #{method} to be given a block" unless block
+    def process_hook(options_key, key, block, cluster_only: false)
+      raise ArgumentError, "expected #{options_key} to be given a block" unless block
 
-      @config.hooks[options_key] = method
+      @config.hooks[options_key] = true
```
joshuay03, line 1469: We still need `#hooks` to keep track of which hook types were registered.

MSP-Greg (approved):

schneems:
Thanks

PR #3711, opened 2025-08-28
Don't set `env['HTTP_VERSION']` for Rack > 3.1

### LINKED ISSUE #3576: Puma 6+ breaks version header

MSP-Greg:
### Description

See #3706.  This PR Is very similar, but it sets an ivar (`@env_set_http_version`) in `Server#handle_servers` to determine whether to overwrite `env['HTTP_VERSION']` with the value of `env['SERVER_PROTOCOL']`.

Adds a test to `test_rack_server.rb`

Closes #3576

### DIFF
--- a/lib/puma/commonlogger.rb
+++ b/lib/puma/commonlogger.rb
@@ -29,13 +29,13 @@ module Puma
 
     CONTENT_LENGTH       = 'Content-Length' # should be lower case from app,
                                             # Util::HeaderHash allows mixed
-    HTTP_VERSION         = Const::HTTP_VERSION
     HTTP_X_FORWARDED_FOR = Const::HTTP_X_FORWARDED_FOR
     PATH_INFO            = Const::PATH_INFO
     QUERY_STRING         = Const::QUERY_STRING
     REMOTE_ADDR          = Const::REMOTE_ADDR
     REMOTE_USER          = 'REMOTE_USER'
     REQUEST_METHOD       = Const::REQUEST_METHOD
+    SERVER_PROTOCOL      = Const::SERVER_PROTOCOL
 
     def initialize(app, logger=nil)
       @app = app
@@ -70,7 +70,7 @@ module Puma
         env[REQUEST_METHOD],
         env[PATH_INFO],
         env[QUERY_STRING].empty? ? "" : "?#{env[QUERY_STRING]}",
-        env[HTTP_VERSION],
+        env[SERVER_PROTOCOL],
         now - began_at ]
 
       write(msg)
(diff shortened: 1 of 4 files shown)


MSP-Greg:
@schneems

So I was looking at your PR and the recent CI, and trying to figure out why it failed, then I went looking for this...

Having been thru all the code affecting the env passed to the app (with the rebase and review on my old PR #3582), I thought calculating this in `Server#handle_servers` was appropriate, rather than doing so for every request.  If Rack isn't loaded by the time `Server#handle_servers` runs, it's not being used (which is probably uncommon).  Anyway, sorry for 'doubling up' on PR's.

schneems:
If you want to make that adjustment :+1: . 

Also the order that ivars are set can affect JIT caching. If we are going to set an ivar outside of initialize, we should define it as nil inside of initialize so the order of ivars is the same for every instance.

MSP-Greg:
@schneems

TIL - ivars...

See #3714

schneems:
Thanks, also linking https://www.youtube.com/watch?v=0lg9Y8gj3FI here (per my comment RE object shapes)

PR #3714, opened 2025-08-29
server.rb - initialize ivars `@reactor` and `@env_set_http_version`

MSP-Greg:
### Description

See [comment](https://github.com/puma/puma/pull/3711#issuecomment-3235392044), or

> Also the order that ivars are set can affect JIT caching. If we are going to set an ivar outside of initialize, we should define it as nil inside of initialize so the order of ivars is the same for every instance.

The above comment was directed at adding `@env_set_http_version` without initializing it.

I tried to audit whether any other ivars needed the same, and only noticed `@reactor`.

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -85,6 +85,9 @@ module Puma
 
       @thread = nil
       @thread_pool = nil
+      @reactor = nil
+
+      @env_set_http_version = nil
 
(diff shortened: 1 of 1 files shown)


schneems:
FWIW here's the "object shapes" info I was talking about RE that comment https://www.youtube.com/watch?v=0lg9Y8gj3FI

PR #3716, opened 2025-08-30
Fixes a bug where triggering hooks in the ThreadPool fails

yuki24:
### Description

https://github.com/puma/puma/pull/3616 has introduced some internal changes to the hooks object, which used to be arrays of callable objects. Now it's arrays of hashes that contain these three keys: `:id`, `:block`, and `:cluster_only`. The methods that call these hooks in the `ThreadPool` class have not been updated as part of that PR, so they started failing as it attempts to call the `#call` method on the Hash object.

cc @joshuay03

### Your checklist for this pull request

- [x] I have reviewed the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.
- [x] I have added (or updated) appropriate tests if this PR fixes a bug or adds a feature.
- [x] My pull request is 100 lines added/removed or less so that it can be easily reviewed.
- [x] If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.
- [x] If this closes any issues, I have added "Closes `#issue`" to the PR description or my commit messages.
- [x] I have updated the documentation accordingly.
- [x] All new and existing tests passed, including Rubocop.

PR #3722, opened 2025-09-03
v7.0.0

schneems:


History.md
```diff
@@ -1,3 +1,53 @@
...
+| on_worker_fork | before_worker_fork |
+
+* Features
+  * Fix long tail response problem with keepalive connections ([#3678]) (Previously released in 7.0.0.pre1, this was a high effort change)
+  * Introduce support for fiber-per-request. ([#3101])
+  * Add support for `rack.response_finished` ([#3681])
+  * Feature/support custom logger with request logs ([#3140])
+
+* Bugfixes
+  * Fixes a bug where triggering hooks in the ThreadPool fails ([#3716])
```
joshuay03, line 33: I don't think this should be in here because the fix was for something that was on master and not released yet. That is, both the original change (https://github.com/puma/puma/pull/3616) and this fix for a bug it introduced (https://github.com/puma/puma/pull/3716) were included in the same release, so this doesn't really meet the conditions of a fix for a bug in previous versions.

dentarg: Agree, a PR would be accepted :)

joshuay03: https://github.com/puma/puma/pull/3727

PR #3723, opened 2025-09-04
server.rb - improve delay conditional when workers < 2

MSP-Greg:
### Description

When workers are either 0 or 1, a delay isn't needed, which is used to send connections to the least busy worker.  Also, the current code causes slowdowns with very fast responses (e.g. 'Hello World').

Seems to help with both keep-alive requests and non-keep-alive requests.

@dentarg - have time to check this with your various tests?  I've tried several...

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -384,15 +384,10 @@ module Puma
                 # clients until the code is finished.
                 sleep 0.001 while pool.out_of_band_running
 
-                # only use delay when clustered and busy
-                if pool.busy_threads >= @max_threads
-                  if @clustered
-                    delay = 0.0001 * ((@reactor&.reactor_size || 0) + pool.busy_threads * 1.5)/max_flt
-                    sleep delay
-                  else
-                    # use small sleep for busy single worker
-                    sleep 0.0001
-                  end
+                # only use delay when clustered (more than one worker) and busy
+                if @options[:workers] > 1 && (pbr = pool.busy_threads) >= 1.5 * @max_threads
+                  delay = 0.0001 * ((@reactor&.reactor_size || 0) + pbr * 1.5)/max_flt
+                  sleep delay
                 end
 
                 io = begin


dentarg:
> have time to check this with your various tests? I've tried several...

I started something earlier, maybe I can post what I got so far..

EDIT: It will have to wait until/after the weekend

dentarg:
I think it will show that we can/should remove the warning  we added in https://github.com/puma/puma/pull/3721

lib/puma/server.rb
```diff
@@ -384,15 +384,10 @@ def handle_servers
...
-                if pool.busy_threads >= @max_threads
-                  if @clustered
-                    delay = 0.0001 * ((@reactor&.reactor_size || 0) + pool.busy_threads * 1.5)/max_flt
-                    sleep delay
-                  else
-                    # use small sleep for busy single worker
-                    sleep 0.0001
-                  end
+                # only use delay when clustered (more than one worker) and busy
+                if @options[:workers] > 1 && (pbr = pool.busy_threads) >= 1.5 * @max_threads
```
schneems, line 388: Where's the 1.5 * @max_threads coming from?

We've got two issues we need to handle, one is a sleep sort that roughly replaces https://github.com/puma/puma/pull/3678/files#diff-a95f46e7ce116caddc9b9a9aa81004246d5210d5da5f4df90a818c780630166bL251-L276. The other bit is more subtle and has to do with CPUs not evenly distributing time evenly to processes https://github.com/puma/puma/pull/2079 I commented on an alternative fix that involves using a different port option SO_REUSEPORT which helps in was that are observably documented but I've not investigated https://github.com/puma/puma/issues/2078#issuecomment-2394401709. I recently talked to @jhawthorn who mentioned it was due to an internal LIFO structure of how the OS is storing something (but I've not verified, would make sense).

If I understand the original gitlab PR, we want to always sleep a tiny bit when using 2 or more processes, then we want to sleep-sort such that the least busy worker is likely to be woken first.

schneems:
> I think it will show that we can/should remove the warning

I think we still have the problem when workers=2 for RUBY_MN_THREADS=1 so I don't think we can remove it just yet (but if benchmarks/profiles say otherwise, I'm happy to adopt new evidence).

lib/puma/server.rb
jhawthorn: I think I was mistaken in what I described. Unicorn and Pitchfork have a behaviour where one process ends up favoured on linux because of `EPOLLEXCLUSIVE` (this cloudflare article is a good write up on the topic https://blog.cloudflare.com/the-sad-state-of-linux-socket-balancing/) but it doesn't seem like puma uses `EPOLLEXCLUSIVE`, so all puma processes should be waking up and racing to be the first that calls `accept` (and then the hope is that this sleep sort selects for the least busy process).

schneems: Cool, thanks for the extra context. @MSP-Greg if you want to move forward with the proposed logic, please help fill in the blanks in terms of intent. Otherwise, I validate that changing it to check `@options[:workers] > 1` seems to be correct.

schneems:
Fixed in https://github.com/puma/puma/pull/3746

PR #3729, opened 2025-09-05
Replace` sleep 0.001 while out_of_band_running` with condition variable

bensheldon:
### Description

This refactors a bit of code mentioned in https://github.com/puma/puma/pull/3678/files#r2310028332

This replaces a sleep loop with a condition variable. Theoretically this should be better, but it doesn't seem like this behavior is tested  with the `test_out_of_band_server.rb`. 

Better because:
- The `sleep 0.001 while ...` will cause threads to be awakened unnecessarily while OOB is running

Worse, maybe, because:
- After the OOB runs, all of the Server threads will have to pass through the same mutex to continue

Also, this is written with a condition variable, but in the current implementation it will never by waited upon by the calling code because the mutex will always be locked already while OOB is running--it's actually the mutex that will hold the thread in Server. I did though want to do it _correctly_ with a condition variable in case that method is reused elsewhere, specifically within the `ThreadPool` where the condvar would be expected to be functional.

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -382,7 +382,7 @@ module Puma
               else
                 # if ThreadPool out_of_band code is running, we don't want to add
                 # clients until the code is finished.
-                sleep 0.001 while pool.out_of_band_running
+                pool.wait_while_out_of_band_running
 
                 # only use delay when clustered and busy
                 if pool.busy_threads >= @max_threads
--- a/lib/puma/thread_pool.rb
+++ b/lib/puma/thread_pool.rb
@@ -53,6 +53,7 @@ module Puma
       @block = block
       @out_of_band = options[:out_of_band]
       @out_of_band_running = false
+      @out_of_band_condvar = ConditionVariable.new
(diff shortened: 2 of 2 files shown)


schneems:
@MSP-Greg Can you take a look?

MSP-Greg:
@bensheldon

Thanks.  This fixed some code I should have noticed.  The long-tail code was something that I often picked up, then had to set down due to some long-running family issues, and this prompted me to look for some of my old 'general' notes...

We've got the below, and the comment mentions adding clients, but that operation is more about adding to the `ThreadPool`:

https://github.com/puma/puma/blob/c46840f2bfcbdc485648a8f01bbbd87bea1edde1/lib/puma/server.rb#L383-L385

So, the `Reactor` calls `Server#reactor_wakeup`, which also adds items to the `ThreadPool`.  Should that be stopped also?  I'm looking at tests, etc.

MSP-Greg:
Never mind.  Updated some tests with keep-alive connections (which are using the Reactor), and everything is fine.

MSP-Greg (approved):

PR #3731, opened 2025-09-06
Fix Puma not booting if queue_requests disabled

stanhu:
Previously if queue_requests were disabled, Puma would crash at startup:

```
puma-7.0.1/lib/puma/server.rb:688:in `stats': undefined method `reactor_max' for nil (NoMethodError)

      stats[:reactor_max] = @reactor.reactor_max
                                    ^^^^^^^^^^^^
```

Fix this by checking the presence of `@reactor`.

### Description

- [X ] All new and existing tests passed, including Rubocop.

### DIFF
--- a/lib/puma/cluster/worker_handle.rb
+++ b/lib/puma/cluster/worker_handle.rb
@@ -73,6 +73,8 @@ module Puma
         # check stat max values, we can't signal workers to reset the max values,
         # so we do so here
         WORKER_MAX_KEYS.each_with_index do |key, idx|
+          next unless hsh[key]
+
           if hsh[key] < @worker_max[idx]
             hsh[key] = @worker_max[idx]
           else
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -685,13 +685,13 @@ module Puma
       stats = @thread_pool&.stats || {}
       stats[:max_threads]    = @max_threads
       stats[:requests_count] = @requests_count
-      stats[:reactor_max] = @reactor.reactor_max
+      stats[:reactor_max] = @reactor.reactor_max if @reactor
       reset_max
       stats
     end
 
     def reset_max
-      @reactor.reactor_max = 0
+      @reactor.reactor_max = 0 if @reactor
       @thread_pool.reset_max
     end
 
--- a/test/test_integration_cluster.rb
+++ b/test/test_integration_cluster.rb
@@ -283,6 +283,14 @@ class TestIntegrationCluster < TestIntegration
     end
   end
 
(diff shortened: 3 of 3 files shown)


schneems:
Thanks! I was recently advocating getting rid of this setting (but paused on the discussion). I'm curious what prompted you to disable queuing in your app, what's your main use case? (I'm asking to help me understand how it's used in the wild)

stanhu:
@schneems I'm not actually sure. I see this comment in our config years ago, but I'm not sure if this was taken from some template config:

```
# By default, workers accept all requests and queue them to pass to handlers.
# When false, workers accept the number of simultaneous requests configured.
#
# Queueing requests generally improves performance, but can cause deadlocks if
# the app is waiting on a request to itself. See https://github.com/puma/puma/issues/612
#
# When set to false this may require a reverse proxy to handle slow clients and
# queue requests before they reach puma. This is due to disabling HTTP keepalive
```

For a while we supported a single-threaded Puma, so perhaps that's why it was there.

samsm:
> what prompted you to disable queuing in your app, what's your main use case?

Turning off queue_requests became became a preferred thing at my workplace based on observed better p95/p99 latency based on experimentation in 2021. 

I've been working backwards through this choice after seeing the above error (thanks for patching!). I suspect that updates to Puma to deal with tail latency might have already addressed this: trying to do some trial/error on that. Also, in some of our systems (especially when it's a server talking to a server) a fast error is often better than a slow success and queuing is one place we can end up waiting. 

So I think disabling queuing was motivated by some combination of those reasons, which I'm not sure how well they stack up in 2025. 😄

stanhu:
Ah, I found the discussion in https://gitlab.com/gitlab-org/gitlab-foss/-/issues/3592#note_63282991 where we indeed saw deadlocks when there was only 1 thread, so we disabled `queue_requests` out of the gate and referred to https://github.com/puma/puma/issues/612.

@schneems Could you tag a new release? We can't upgrade Puma until this fix is released. Thanks!

schneems:
Thanks for all the context!

> Turning off queue_requests became became a preferred thing at my workplace based on observed better p95/p99 latency based on experimentation in 2021.

Gotcha. Though if you're doing this, you need to be running Puma behind something that buffers requests (nginx, for example), otherwise, you're not protected from slow clients https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#slow-clients. 

> we indeed saw deadlocks when there was only 1 thread

Interesting, thanks!

I've staged 7.0.2 with the fix https://github.com/puma/puma/pull/3738.

PR #3736, opened 2025-09-08
bug: control_cli.rb - Fixup `pumactl` code to load puma.rb for `deprecate_method_change`

### LINKED ISSUE #3734: "undefined method 'deprecate_method_change' for module Puma" with pumactl
**Describe the bug**
I am able to run Puma in our staging env but when executing commands in terminal I get this error: `undefined method 'deprecate_method_change' for module Puma`.

Example:

```shell
$ bundle exec pumactl status
undefined method 'deprecate_method_change' for module Puma
$ bundle exec pumactl --version
7.0.1
```

In staging we run a custom Linux distro with custom service orchestration. But I can also reproduce this locally in MacOS where I use mise. I tried uninstalling previous Puma versions and mise reshim but the error still occurs.

 - OS: Custom Linux distro / MacOS latest
 - Puma: 7.0.1

MSP-Greg:
### Description

`pumactl` may load a config file, and config files may create 'hooks' using deprecated names.  Code for this exists in both `dsl.rb` and `events.rb', which is the reason the method was placed in the `Puma` namespace.

Fix by loading `puma.rb` when `control_cli.rb` processes a config file.

If one checks out the 2nd ci/test commit, the test will fail.

Closes #3734

### DIFF
--- a/test/test_integration_pumactl.rb
+++ b/test/test_integration_pumactl.rb
@@ -169,7 +169,6 @@ class TestIntegrationPumactl < TestIntegration
     refute_start_with out.read, 'undefined method'
   end
 
-
   # calls pumactl with both a config file and a state file,  making sure that
   # puma files are required, see https://github.com/puma/puma/issues/3186
   def test_require_dependencies


test/test_integration_pumactl.rb
```diff
@@ -163,6 +163,13 @@ def test_kill_unknown
     assert_equal(1, e.status)
   end
 
+  def test_config_deprecated
+    conf_path = "test/config/config_with_deprecated.rb"
+    out = cli_pumactl_spawn "-F #{conf_path} halt", no_bind: true
+    refute_start_with out.read, 'undefined method'
+  end
+
```
schneems, line 171:
```suggestion
```

test/test_integration_pumactl.rb
```diff
@@ -163,6 +163,12 @@ def test_kill_unknown
     assert_equal(1, e.status)
   end
 
+  def test_config_deprecated
+    conf_path = "test/config/config_with_deprecated.rb"
+    out = cli_pumactl_spawn "-F #{conf_path} halt", no_bind: true
+    refute_start_with out.read, 'undefined method'
```
schneems, line 169: Seems good for now. I'm worried about the long term utility of this test though. If the output message of the failure mode changes to something like `"Undefined method"` (with a capital) then this test would continue to pass because it's not a match. Or if there's some extra whitespace etc.

It's good enough to merge, but there's an opportunity to iterate here.

schneems:
Thanks for the PR! I left a comment (that I had to resolve to merge):

> Seems good for now. I'm worried about the long term utility of this test though. If the output message of the failure mode changes to something like "Undefined method" (with a capital) then this test would continue to pass because it's not a match. Or if there's some extra whitespace etc.
>
> It's good enough to merge, but there's an opportunity to iterate here.

MSP-Greg:
@schneems

Sorry about the double blank line.  I should have noticed that (note to self).  And yes, the test is brittle (string match).  I think the 'test framework update` used/uses `Open3.capture3` so one can get a status back...

schneems:
> Sorry about the double blank line. 

Super easy fix on my end.

> so one can get a status back...

I think so. But we need to be careful since there's some non determinism around the time between process boot and when/if  it exits.

PR #3742, opened 2025-09-10
Strip whitespace from the beginnings of header values

kenballus:
### Description

When Puma receives a header with tabs before the value, it doesn't strip them.

For example, when Puma receives the following request:
```
GET / HTTP/1.1\r\n
Host: whatever\r\n
Test: \t\t whatever\r\n
\r\n
```
...it sees the `Test` header as having value `\t\t whatever`. This is incorrect according to spec, because the OWS preceding a header value should consume the `\t` characters, leaving a value of just `whatever`.

Puma is (to my knowledge) the only web server that doesn't strip the tabs appropriately. I've tested the full list [here](https://github.com/narfindustries/http-garden?tab=readme-ov-file#targets).

This PR basically makes 2 changes:
1. Checks for leading whitespace and skips past it during header value parsing.
2. Updates the way that trailing whitespace is skipped to avoid using `isspace` (in C) and `Character.isWhitespace` (in Java) because those functions accept more characters than just space and tab, which is all we should care about according to the RFCs.

### DIFF
--- a/ext/puma_http11/org/jruby/puma/Http11.java
+++ b/ext/puma_http11/org/jruby/puma/Http11.java
@@ -109,6 +109,10 @@ public class Http11 extends RubyObject {
         return (RubyClass)runtime.getModule("Puma").getConstant("HttpParserError");
     }
 
+    private static boolean is_ows(int c) {
+        return c == ' ' || c == '\t';
+    }
+
     public static void http_field(Ruby runtime, RubyHash req, ByteList buffer, int field, int flen, int value, int vlen) {
         RubyString f;
         IRubyObject v;
@@ -127,7 +131,11 @@ public class Http11 extends RubyObject {
             }
         }
 
-        while (vlen > 0 && Character.isWhitespace(buffer.get(value + vlen - 1))) vlen--;
+        while (vlen > 0 && is_ows(buffer.get(value + vlen - 1))) vlen--;
+        while (vlen > 0 && is_ows(buffer.get(value))) {
+            vlen--;
+            value++;
+        }
 
         if (b.equals(CONTENT_LENGTH_BYTELIST) || b.equals(CONTENT_TYPE_BYTELIST)) {
           f = RubyString.newString(runtime, b);
--- a/ext/puma_http11/puma_http11.c
+++ b/ext/puma_http11/puma_http11.c
@@ -155,6 +155,10 @@ static VALUE find_common_field_value(const char *field, size_t flen)
   return Qnil;
 }
 
+static int is_ows(const char c) {
+    return c == ' ' || c == '\t';
+}
+
 void http_field(puma_parser* hp, const char *field, size_t flen,
                                  const char *value, size_t vlen)
 {
@@ -181,7 +185,11 @@ void http_field(puma_parser* hp, const char *field, size_t flen,
     f = rb_str_new(hp->buf, new_size);
   }
 
-  while (vlen > 0 && isspace(value[vlen - 1])) vlen--;
+  while (vlen > 0 && is_ows(value[vlen - 1])) vlen--;
+  while (vlen > 0 && is_ows(value[0])) {
+      vlen--;
+      value++;
+  }
 
   /* check for duplicate header */
   v = rb_hash_aref(hp->request, f);
--- a/test/test_http11.rb
+++ b/test/test_http11.rb
@@ -248,7 +248,7 @@ class Http11ParserTest < TestIntegration
   def test_trims_whitespace_from_headers
     parser = Puma::HttpParser.new
     req = {}
-    http = "GET / HTTP/1.1\r\nX-Strip-Me: Strip This       \r\n\r\n"
+    http = "GET / HTTP/1.1\r\nX-Strip-Me: \t Strip This \t      \r\n\r\n"
 
     parser.execute(req, http, 0)
 


kenballus:
The failing tests seem unrelated to my change; am I missing something?

MSP-Greg:
@kenballus

Thank you for the PR.  You are not missing something. let me see if I can fix that quickly.

MSP-Greg:
@kenballus

If you rebase, this should pass CI.

Re the test, something almost identical:
```ruby
def test_header_value_ows
  request = <<~REQ.gsub("\n", "\r\n")
    GET / HTTP/1.1
    X-Strip-Me: \t Strip This \t

  REQ

  create_client request

  if @parser.finished?
    assert_equal 'Strip This', @client.env['HTTP_X_STRIP_ME']
    assert_instance_of Puma::NullIO, @client.body
  else
    fail
  end
end
```

I was hoping to move the tests from `test/test_http11.rb` into other files  (or remove if they're duplicates).  I put the above test at line 60 of `test_request_single.rb`.

kenballus:
> If you rebase, this should pass CI.

Thanks!

> Re the test, something almost identical:
> 
> ```ruby
> def test_header_value_ows
>   request = <<~REQ.gsub("\n", "\r\n")
>     GET / HTTP/1.1
>     X-Strip-Me: \t Strip This \t
> 
>   REQ
> 
>   create_client request
> 
>   if @parser.finished?
>     assert_equal 'Strip This', @client.env['HTTP_X_STRIP_ME']
>     assert_instance_of Puma::NullIO, @client.body
>   else
>     fail
>   end
> end
> ```

Are you requesting that I change the test in this PR to match the above?

> I put the above test at line 60 of `test_request_single.rb`.

In another PR, or in a commit that's incoming?

MSP-Greg:
> Are you requesting that I change the test in this PR to match the above?

I'll catch it when `test/test_http11.rb` is changed.

MSP-Greg:
@schneems

See [RFC 9110 Sec 5.5 Field Values](https://httpwg.org/specs/rfc9110.html#fields.values), and also [RFC 9112 Sec 5 Field Syntax](https://httpwg.org/specs/rfc9112.html#header.field.syntax).

Note that the HTTP/1.1 spec (RFC 9112) shows:
```
field-line   = field-name ":" OWS field-value OWS
```
9112 Sec 5.6.3 defines OWS

kenballus:
> I'm assuming that you're referring to the HTTP spec rather than the Rack spec since some servers listed there are not using Ruby.

Yes.

> Please help me out by linking to the exact place in the HTTP spec you're looking at that describes the behavior so we can validate your interpretation? It would also be good to make sure we're handling all the cases (since this one was previously missed).

From RFC 9112:

"A field line value might be preceded and/or followed by optional whitespace (OWS); a single SP preceding the field line value is preferred for consistent readability by humans. The field line value does not include that leading or trailing whitespace: OWS occurring before the first non-whitespace octet of the field line value, or after the last non-whitespace octet of the field line value, is excluded by parsers when extracting the field line value from a field line."

> (optional) Can you give details on your test? How can we reproduce your results? You can link out to a separate repo if it's easier. This is less important than the first request, but I'm curious if it's easy to provide.

I am so glad you asked :)

[The HTTP Garden](https://github.com/narfindustries/http-garden) is a framework for comparing HTTP implementations, to find these kinds of issues.

It allows me to run tests that look like this:
```
garden> payload 'GET / HTTP/1.1\r\nHost: a\r\nTest:\twhatever\r\n\r\n' | fanout | grid
aiohttp: [
    HTTPRequest(
        method=b'GET', uri=b'/', version=b'1.1',
        headers=[
            (b'host', b'a'),
            (b'test', b'whatever'),
        ],
        body=b'',
    ),
]
apache_httpd: [
    HTTPRequest(
        method=b'GET', uri=b'/', version=b'1.1',
        headers=[
            (b'test', b'whatever'),
            (b'host', b'a'),
        ],
        body=b'',
    ),
]
apache_tomcat: [
    HTTPRequest(
        method=b'GET', uri=b'/', version=b'1.1',
        headers=[
            (b'host', b'a'),
            (b'test', b'whatever'),
        ],
        body=b'',
    ),
]
...
                            e
                            c                                     p
                    a       l e                 l             o o r
                  a p       i c                 i             p p o
                  p a     c p l                 b           n e e t
                  a c   a p s i                 m           o n n o
                  c h   w p e p   g       h     i           d b l c
                  h e   s _ _ s f o g     y   l c   l m     e s i o       u   w
                a e _   _ h g e a _ u     p   i r l i o     _ d t l   t t n u a w
                i _ t a c t r _ s s n     e   b o i g n     s _ e _   o w d v i e
                o h o p _ t i j t t i   h r   e h b h g n n t h s h   r i e i t b y
                h t m p h p z e h d c   y c k v t s t o e g d t p t p n s r c r r a
                t t c w t l z t t l o h p o t e t o t o t i l t e t u a t t o e i h
                t p a e t i l t t i r 2 e r o n p u p s t n i p e p m d e o r s c n
                p d t b p b y y p b n o r n r t d p d e y x b d d 1 a o d w n s k s
               +-------------------------------------------------------------------
aiohttp        |✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ ✓ ✓
apache_httpd   |  ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ ✓ ✓
apache_tomcat  |    ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ ✓ ✓
appweb         |      ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ ✓ ✓
aws_c_http     |        ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ ✓ ✓
cpp_httplib    |          ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ ✓ ✓
eclipse_grizzly|            ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ ✓ ✓
eclipse_jetty  |              ✓ ✓ ✓ ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ ✓ ✓
fasthttp       |                ✓ ✓ ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ ✓ ✓
go_stdlib      |                  ✓ ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ ✓ ✓
gunicorn       |                    ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ ✓ ✓
h2o            |                      ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ ✓ ✓
hyper          |                        ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ ✓ ✓
hypercorn      |                          ✓ ✓ X ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ ✓ ✓
ktor           |                            ✓ X ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ ✓ ✓
libevent       |                              ✓ X X X X X ✓ X X X X ✓ X X X X X X X
libmicrohttpd  |                                ✓ ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ ✓ ✓
libsoup        |                                  ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ ✓ ✓
lighttpd       |                                    ✓ ✓ ✓ X ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ ✓ ✓
mongoose       |                                      ✓ ✓ X ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ ✓ ✓
netty          |                                        ✓ X ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ ✓ ✓
nginx          |                                          ✓ X X X X ✓ X X X X X X X
node_stdlib    |                                            ✓ ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ ✓ ✓
openbsd_httpd  |                                              ✓ ✓ ✓ X ✓ ✓ ✓ ✓ ✓ ✓ ✓
openlitespeed  |                                                ✓ ✓ X ✓ ✓ ✓ ✓ ✓ ✓ ✓
protocol_http1 |                                                  ✓ X ✓ ✓ ✓ ✓ ✓ ✓ ✓
puma           |                                                    ✓ X X X X X X X
tornado        |                                                      ✓ ✓ ✓ ✓ ✓ ✓ ✓
twisted        |                                                        ✓ ✓ ✓ ✓ ✓ ✓
undertow       |                                                          ✓ ✓ ✓ ✓ ✓
uvicorn        |                                                            ✓ ✓ ✓ ✓
waitress       |                                                              ✓ ✓ ✓
webrick        |                                                                ✓ ✓
yahns          |                                                                  ✓
```

I stand corrected about my statement above (I had left a filter in place to remove spec deviations that I had already reported). Turns out that the HTTP parsers in Nginx and Libevent also behave weirdly when given this input. Let's see why:
```
garden> payload 'GET / HTTP/1.1\r\nHost: a\r\nTest:\twhatever\r\n\r\n' | fanout puma nginx libevent
puma: [
    HTTPRequest(
        method=b'GET', uri=b'/', version=b'1.1',
        headers=[
            (b'host', b'a'),
            (b'test', b'\twhatever'),
        ],
        body=b'',
    ),
]
nginx: [
    HTTPRequest(
        method=b'GET', uri=b'/', version=b'1.1',
        headers=[
            (b'test', b'\twhatever'),
            (b'host', b'a'),
            (b'content-length', b''),
            (b'content-type', b''),
        ],
        body=b'',
    ),
]
libevent: [
    HTTPRequest(
        method=b'GET', uri=b'/', version=b'1.1',
        headers=[
            (b'host', b'a'),
            (b'test', b'\twhatever'),
        ],
        body=b'',
    ),
]
```
...and, we can see they all have the same problem. We could also determine this automatically using the `cluster` command in the Garden, but it's less fun that way :)

ext/puma_http11/puma_http11.c
```diff
@@ -181,7 +181,11 @@ void http_field(puma_parser* hp, const char *field, size_t flen,
     f = rb_str_new(hp->buf, new_size);
   }
 
-  while (vlen > 0 && isspace(value[vlen - 1])) vlen--;
+  while (vlen > 0 && (value[vlen - 1] == ' ' || value[vlen - 1] == '\t')) vlen--;
```
schneems: (I think, C is not my daily driver) We should be using `isspace` instead of rolling our own check. The `isspace` call already handles tab:

- https://man.openbsd.org/isspace.3
- https://manpages.debian.org/testing/manpages-dev/isspace.3.en.html

The code on main already strips leading whitespace, the reason your test fails on main is because of the trailing whitespace.

ext/puma_http11/org/jruby/puma/Http11.java
```diff
@@ -127,7 +127,11 @@ public static void http_field(Ruby runtime, RubyHash req, ByteList buffer, int f
             }
         }
 
-        while (vlen > 0 && Character.isWhitespace(buffer.get(value + vlen - 1))) vlen--;
+        while (vlen > 0 && (buffer.get(value + vlen - 1) == ' ' || buffer.get(value + vlen - 1) == '\t')) vlen--;
```
schneems: Same comment as the C code, we should use a builtin instead of hand rolling manually checking space and tab.

MSP-Greg: See opening comment, paragraph right above 'Your checklist for this pull request'.  OWS space is defined as two characters, HTAB and SPACE.

schneems:
Thanks for the extra details. I took another look a the implementation this time.

It looks like the code on main strips leading values already, but your tests are failing because they include trailing values as well. The PR title and description (and commit messages) need to be updated with the new impact. We also need to assess if this is something we want to do. A very quick search lead me to https://stackoverflow.com/a/61632443/147390 which explains why there's maybe not as much support for trimming trailing whitespace, since it becomes ambiguous in the case of a streaming parser. 

Edit: I think we have a test like `Connection: a    b \r\n b    c\r\n d \r\n` in the one being modified here already `    http = "GET / HTTP/1.1\r\nX-Strip-Me: Strip This       \r\n\r\n"` since there's a space in the middle of `Strip This` already.

MSP-Greg:
I'm essentially AFK, but...

I don't think field/header values can contain an embedded space.  So `Strip This` would be invalid.  Spaces (part of OWS) are allowed as leading and trailing characters, and also in a 'field list' (comma separated list of values).

Will check tomorrow...

MSP-Greg:
First of all, regarding tests, all pass in this PR.

Current Puma will strip leading spaces and trailing OWS from a header value.

It will correctly parse the header line `Connection:  Strip This \t` with leading spaces and trailing OWS.

It **will not** correctly parse the header line `Connection: \t Strip This \t` with both leading and trailing OWS.

So, the leading OWS parsing is incorrect and is fixed by this PR.

### RFC Notes

The first reference for HTTP/1.1 header lines is [RFC 9112 HTTP/1.1 - Sec 5. Field Syntax](https://httpwg.org/specs/rfc9112.html#rfc.section.5), which shows:
```
field-line   = field-name ":" OWS field-value OWS
 ```

 It also states "Rules for parsing within field values are defined in Section 5.5 of [HTTP]".

[RFC 9110 HTTP Semantics - Sec 5.5. Field Values](https://httpwg.org/specs/rfc9110.html#rfc.section.5.5) shows:
```
field-value    = *field-content
field-content  = field-vchar
                 [ 1*( SP / HTAB / field-vchar ) field-vchar ]
field-vchar    = VCHAR / obs-text
obs-text       = %x80-FF
  ```

It also defines 'OWS' in [RFC 9110 - Section 5.6.3 Whitespace](https://httpwg.org/specs/rfc9110.html#rfc.section.5.6.3) as:
```
OWS            = *( SP / HTAB )
```

The other entities listed are defined in [RFC 5234 Augmented BNF for Syntax Specifications: ABNF - Appendix B.1. Core Rules](https://www.rfc-editor.org/rfc/rfc5234.html#appendix-B.1).

ext/puma_http11/puma_http11.c
kenballus: `isspace` strips more than *just* space and tab, though. It also strips `\x0b` and `\x0c`, for example. It's also dependent on locale. This is why I chose to compare to space and tab.

kenballus:
> Thanks for the extra details. I took another look a the implementation this time.
> 
> It looks like the code on main strips leading values already, but your tests are failing because they include trailing values as well. The PR title and description (and commit messages) need to be updated with the new impact. We also need to assess if this is something we want to do. A very quick search lead me to https://stackoverflow.com/a/61632443/147390 which explains why there's maybe not as much support for trimming trailing whitespace, since it becomes ambiguous in the case of a streaming parser.

1. My reading is that the issue they're pointing out has to do with the interaction of obs-fold and trailing OWS. Puma doesn't support obs-fold (and afaict isn't trying to be a pure-streaming parser), so this shouldn't be an issue.
2. The claim that Node and Nginx don't support headers with embedded whitespace isn't true anymore. Node correctly handles whitespace in header values. Nginx correctly handles whitespace in the middle of a header value, but doesn't permit tabs in the OWS on either side of the value. I've brought this up on nginx-devel, but the maintainers don't care to fix it because they argue that few valid clients send tabs in OWS. This is true, but somewhat beside the point imo. Nginx *does* support spaces in OWS on the right side of a header value, which shows that the lack of support is not due to the pure streaming concern referenced above.

MSP-Greg:
> Nginx correctly handles whitespace in the middle of a header value

Curious, how does it handle them?  Seems to get messy.  It might be a list, quoted text, etc.

kenballus:
> > Nginx correctly handles whitespace in the middle of a header value
> 
> Curious, how does it handle them? Seems to get messy. It might be a list, quoted text, etc.

For exactly the reasons you've mentioned, the only way to handle whitespace in the middle of a header value is to just leave it alone. This is what pretty much every web server does, because the RFCs require it.

MSP-Greg:
@kenballus

Thanks.  A bit off topic, but since you're familiar with a lot servers, I thought I'd ask.  In the past I decided that no one is complaining, so leave as is.  But, `VCHAR` does not include a `SPACE` character.

JFYI, I'm more a "Mysterious Traveller"/"Sweetnighter" type.  But "Tale Spinnin'"/"Black Market"/"Mr Gone" are also good, along with Steely Dan.

kenballus:
> Thanks. A bit off topic, but since you're familiar with a lot servers, I thought I'd ask. In the past I decided that no one is complaining, so leave as is. But, VCHAR does not include a SPACE character.

`VCHAR` doesn't, but ~~`field-vchar`~~ `field-content` does, ~~because of `obs-text`.~~


> JFYI, I'm more a "Mysterious Traveller"/"Sweetnighter" type. But "Tale Spinnin'"/"Black Market"/"Mr Gone" are also good, along with Steely Dan.

Thank you for the suggestions. Listening now :)

MSP-Greg:
> `VCHAR` doesn't, but `field-vchar` does, because of `obs-text`.

`obs-text` always struck me as odd name, as `%x80-FF` is upper ascii.  And

```
SP     =  %x20
VCHAR  =  %x21-7E
```

Another would be "Pat Metheny Group".  Years ago, saw all of them live.  I'm semi-retired, (late 60's).  Old enough to know that many "abstract smart" people don't connect with the music/artists being mentioned...

kenballus:
> obs-text always struck me as odd name, as %x80-FF is upper ascii. And

Oops! Sorry; gave the wrong explanation there. SP and HTAB are allowed in the middle of field values because of this:
>   field-content  = field-vchar [ 1*( SP / HTAB / field-vchar ) field-vchar ]

Further, RFC 9110 says this:
> Field values containing other CTL characters [than CR, LF, and NUL] are also invalid; however, recipients MAY retain such characters for the sake of robustness when they appear within a safe context (e.g., an application-specific quoted string that will not be processed by any downstream HTTP parser).

So really the only hard and fast rules that the RFC requires for field-values are that
1. A field-value can't begin or end with SP or HTAB
2. A field-value can't contain CR, LF, or NUL

> Another would be "Pat Metheny Group". Years ago, saw all of them live. I'm semi-retired, (late 60's). Old enough to know that many "abstract smart" people don't connect with the music/artists being mentioned...

Yeah, Pat Metheny is awesome! I got to see him play in Lebanon, NH a couple years ago. Also saw Steely Dan at Bethel Woods when I was a teenager, but Walter Becker was getting sick at that point and it was a little off-kilter (still great though).

MSP-Greg:
Makes sense.  Always wondered why they have a line break in the notation.

But also, "MUST either reject the message or replace each of those characters with SP before further processing or forwarding of that message." implies that SP is considered a valid character within a field value.

ext/puma_http11/puma_http11.c
byroot: Would be worth having a `is_ows(const char c)` for readability though.

Other than that the C code looks fine.

kenballus: Agreed, and now done.

schneems (approved):

PR #3696, opened 2025-08-18
Fix socket leak on monitor wakeup `NoMethodError` in `Reactor#select_loop`

joshuay03:
### Description

Closes https://github.com/puma/puma/issues/3695

This fixes the linked issue by narrowing the scope of the `NoMethodError` rescue logic in the reactor's select loop. Previously, the rescue logic targeted the entire loop iteration including wakeup processing. Now it only protects the `@selector.select` call itself, which was the original intention.

When a `lowlevel_error_handler` returns an invalid response (like a string instead of an array), the reactor thread would die during wakeup processing but leave the selector open, causing file descriptor leaks. By moving the rescue block to only wrap `@selector.select`, exceptions during wakeup processing no longer terminate the reactor thread, preventing file descriptor exhaustion under high load.

The `NoMethodError` rescue logic was originally added in commit fd259b7f1a64a505b694cd7cdef605a7de731611 to handle rare selector exceptions. It seems like it shouldn't be required, and that there's some underlying issue that should be fixed. I'd be happy to look into this in a follow-up so we can remove it altogether, but this patch should do for now.

I will come back to add a test for this when I get a chance. I thought I'd open it early for feedback.

Manual testing:

<details>
<summary>master</summary>

```
# Initial
Tue Aug 19 05:06:17 AEST 2025
Total FDs:       63

puma [master] ./slow_body_http_requests.sh valid
Sending 500 slow body HTTP requests (valid)...

# Increases
Tue Aug 19 05:06:38 AEST 2025
Total FDs:      563

# Drops
Tue Aug 19 05:06:51 AEST 2025
Total FDs:       63

puma [master] ./slow_body_http_requests.sh invalid
Sending 500 slow body HTTP requests (invalid)...

# Increases
Tue Aug 19 05:08:23 AEST 2025
Total FDs:      564

# Does not drop
Tue Aug 19 05:08:59 AEST 2025
Total FDs:      564

puma [master] ./slow_body_http_requests.sh valid
Sending 500 slow body HTTP requests (valid)...

# Increases
Tue Aug 19 05:09:40 AEST 2025
Total FDs:     1064

# Does not drop
Tue Aug 19 05:10:24 AEST 2025
Total FDs:     1064

puma [master] ./slow_body_http_requests.sh valid
Sending 500 slow body HTTP requests (valid)...

# Increases
Tue Aug 19 05:10:42 AEST 2025
Total FDs:     1564

# Does not drop
Tue Aug 19 05:11:09 AEST 2025
[...]

PR #3751, opened 2025-09-14
Rescue and validate lowlevel_error_handler

Ruto8:
# The change

In this PR, I'm adding some safety around the `lowlevel_error_handler`. Today, it's not rescued and it can be devastating for the server.

# What original problem led to this PR?

We had a small bug in our `lowlevel_error_handler`, where we used a method that doesn't exist.
```rb
lowlevel_error_handler do |err, env, status|
  Sentry.configure_scope do |scope|
    scope.set_transaction("Puma") # <- This should be set_transaction_name
    scope.set_contexts(puma: env)
  end
  Sentry.capture_exception(err)

  [status, {}, ["Puma caught this error: #{err.message} (#{err.class})\n#{err.backtrace.join("\n")}"]]
end
```

Upon receiving of some invalid requests, this proc was, rightfully so, executed. Unfortunately, there's a `NoMethodError` in there, because Sentry has `set_transaction_name` method, not `set_transaction`. What happened next? This error got propagated. The error was then logged like this:
```
Error in reactor loop escaped: undefined method `set_transaction' for an instance of Sentry::Scope (NoMethodError)
```
And if we follow that up, we see that the selector won't be closed.

https://github.com/puma/puma/blob/ef2c6aae78a5df5ceb6126045768e474fe2b51be/lib/puma/reactor.rb#L102-L118

It seems that when the selector is not closed, the reactor ends up being in an invalid state. What we experienced, is that the server was accepting but not processing requests (well, to be precise probably a worker was not processing, because some requests were handled properly, or a thread, I'm not entirely sure here about the puma architecture).
The symptoms that we've seen is that requests made to our app were timing out, but we didn't see any metrics on our server to be wrong.

# What alternatives have been tried? Does this supersede previous attempts?

To be fair, we simply used the correct method so we don't raise anymore. But since this was so hard to debug and so destructive, I figured it might be a good idea to prevent others from making the same mistake.

I've contemplated modifying the rescue in the `Reactor` to avoid rescuing `NoMethodError`, or having some kind of behavior of self-healing of the `Reactor`. But I don't have enough knowledge around the Puma internals to feel comfortable exploring these.

For context, with the help of AI, I've built this reproduction script. Commenting out `undefined_method_that_causes_nomethoderror` makes the difference in the output.

<details><summary>Reproduction script</summary>

```rb
#!/usr/bin/env ruby
# Demonstration script for Puma reactor "black hole" bug
#
# This script reproduces the exact issue you described:
# 1. A faulty lowlevel_error_handler that raises NoMethodError
# 2. This corrupts the reactor's selector (close_selector = false in reactor.rb:110)
# 3. TCP connections succeed but HTTP requests are silently lost
# 4. Clients timeout without server-side errors

require 'puma'
require 'puma/server'
require 'puma/events'
require 'socket'
require 'timeout'

class BlackHoleDemo
  def initialize
    @log_writer = Puma::LogWriter.new($stdout, $stderr)
    @events = Puma::Events.new

    # App that always raises to trigger lowlevel_error_handler
    @app = ->(env) { raise StandardError, "App error to trigger handler" }

    @handler_calls = 0

    # The faulty lowlevel_error_handler that causes the problem
    @faulty_handler = ->(err, env, status) do
      @handler_calls += 1
      puts "Handler called #{@handler_calls} times"

      # This NoMethodError will be caught in reactor.rb:102-114
      # Setting close_selector = false and corrupting the reactor
      undefined_method_that_causes_nomethoderror

      [500, {}, ["message"]]
    end

    @server = nil
    @port = nil
  end

  def start_server
    puts "🚀 Starting Puma server with faulty lowlevel_error_handler..."

    @server = Puma::Server.new(@app, @events, {
      log_writer: @log_writer,
      min_threads: 1,
      max_threads: 1,
      lowlevel_error_handler: @faulty_handler
    })

    @port = @server.add_tcp_listener("127.0.0.1", 0).addr[1]
    @server.run

    puts "Server listening on port #{@port}"
    sleep 0.2  # Let server start
  end

  def trigger_corruption
    puts "\n💥 Triggering reactor corruption..."

    socket = TCPSocket.new("127.0.0.1", @port)
    socket.write("GET /trigger-corruption HTTP/1.1\r\nHost: localhost\r\nConnection: close\r\n\r\n")

    # Give time for the corruption to occur
    sleep 0.3
    socket.close rescue nil

    puts "Faulty handler called #{@handler_calls} times (should be > 0)"
  end

  def demonstrate_blackhole
    puts "\n🕳️  Demonstrating 'black hole' behavior..."
    puts "TCP connections will succeed, but HTTP requests will be silently lost"

    successful_connections = 0
    successful_responses = 0
    timeouts = 0

    5.times do |i|
      print "Request #{i+1}: "

      begin
        # TCP connection should succeed
        socket = TCPSocket.new("127.0.0.1", @port)
        successful_connections += 1

        # Send HTTP request
        socket.write("GET /test-#{i} HTTP/1.1\r\nHost: localhost\r\nConnection: close\r\n\r\n")

        # Try to read response (this will timeout due to corruption)
        response = nil
        Timeout.timeout(2) do
          response = socket.read(1024)
        end

        p "Response: #{response}"
        if response && !response.empty?
          successful_responses += 1
          print "✅ Response received"
        else
          print "❌ Empty response"
        end

        socket.close
      rescue Timeout::Error
        timeouts += 1
        print "⏰ Timeout (no response)"
      rescue => e
        print "💥 Error: #{e.class}"
      ensure
        socket&.close rescue nil
      end

      puts
      sleep 0.1
    end

    puts "\n📊 Results:"
    puts "  - TCP connections established: #{successful_connections}/5"
    puts "  - HTTP responses received: #{successful_responses}/5"
    puts "  - Timeouts: #{timeouts}/5"
    puts "  - Handler calls: #{@handler_calls}"

    if successful_connections > 0 && successful_responses == 0
      puts "\n✅ BLACK HOLE REPRODUCED!"
      puts "  - TCP connections work (#{successful_connections}/5)"
      puts "  - But HTTP requests are silently lost (0 responses)"
      puts "  - This matches your production issue exactly"
    else
      puts "\n❓ Results unclear - may need to run again"
    end
  end

  def check_server_logs
    puts "\n📝 Server-side logging analysis:"
    puts "In production, you'd see minimal server errors despite client timeouts"
    puts "This is because the reactor is corrupted but not completely dead"
  end

  def cleanup
    puts "\n🧹 Cleaning up..."
    @server&.stop(true) rescue nil
  end

  def run
    start_server
    trigger_corruption
    demonstrate_blackhole
    check_server_logs
    cleanup

    puts "\n" + "="*60
    puts "SUMMARY: Puma Reactor 'Black Hole' Bug Reproduction"
    puts "="*60
    puts "1. ✅ Faulty lowlevel_error_handler raises NoMethodError"
    puts "2. ✅ Reactor catches error, sets close_selector=false (reactor.rb:110)"
    puts "3. ✅ TCP connections establish normally (no socket-level issues)"
    puts "4. ✅ HTTP requests are silently dropped (reactor corruption)"
    puts "5. ✅ Clients timeout without meaningful server-side errors"
    puts ""
    puts "This exactly matches your production 'black hole' behavior where:"
    puts "- Clients can establish TCP connections"
    puts "- HTTP requests disappear without a trace"
    puts "- Server appears healthy but doesn't process requests"
    puts "- Socket backlog and TCP metrics appear normal"
  end
end

if __FILE__ == $0
  demo = BlackHoleDemo.new

  begin
    demo.run
  rescue Interrupt
    puts "\n\n⏹️  Demo interrupted"
    demo.cleanup
  end
end
```

</details> 

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -584,7 +584,9 @@ module Puma
                      end
 
           # Validate that the handler returned a proper Rack response
-          if response.is_a?(Array) && response.length == 3
+          if response.is_a?(Array) && response.length == 3 &&
+            response[0].is_a?(Integer) && response[1].is_a?(Hash) &&
+            (response[2].respond_to?(:each) || response[2].respond_to?(:call))
             return response
           else
             @log_writer.unknown_error StandardError.new("lowlevel_error_handler returned invalid response: #{response.inspect}"), nil, "lowlevel_error_handler validation"
--- a/test/test_puma_server.rb
+++ b/test/test_puma_server.rb
@@ -2172,6 +2172,21 @@ class TestPumaServer < PumaTest
     assert_equal "something wrong happened", body
   end
 
+  def test_lowlevel_error_handler_response_with_proc
+    options = {
+      lowlevel_error_handler: ->(_error) do
+        [500, {}, ["something wrong happened"]]
+      end
+    }
+    broken_app = ->(_env) { [200, nil, ->(_env) { ["something wrong happened"] }] }
+
+    server_run(**options, &broken_app)
+
+    body = send_http_read_resp_body "GET / HTTP/1.1\r\n\r\n"
+
+    assert_equal "something wrong happened", body
+  end
+
   def test_lowlevel_error_handler_that_raises_exception
     handler_calls = 0
     faulty_handler = ->(_error, _env, _status) do
@@ -2230,6 +2245,72 @@ class TestPumaServer < PumaTest
       handler_calls += 1
       # Return array but not proper 3-element Rack response
       [500, "not a hash"]
+    end
+
+    options = {
+      lowlevel_error_handler: malformed_handler
+    }
+    broken_app = ->(_env) { raise StandardError, "app error" }
+    server_run(**options, &broken_app)
+
+    # Should get default error response when handler returns malformed response
+    response = send_http_read_response "GET / HTTP/1.1\r\n\r\n"
+
+    assert_equal "HTTP/1.1 500 Internal Server Error", response.status
+    assert_includes response.body, "Puma caught this error: app error"
+    assert handler_calls > 0, "Handler should have been called"
+  end
+
+  def test_lowlevel_error_handler_returns_malformed_status_response
+    handler_calls = 0
+    malformed_handler = ->(_error, _env, _status) do
+      handler_calls += 1
+      # Return array but not proper 3-element Rack response
+      ["not_integer", {}, [""]]
+    end
+
+    options = {
+      lowlevel_error_handler: malformed_handler
+    }
+    broken_app = ->(_env) { raise StandardError, "app error" }
+    server_run(**options, &broken_app)
+
+    # Should get default error response when handler returns malformed response
+    response = send_http_read_response "GET / HTTP/1.1\r\n\r\n"
+
+    assert_equal "HTTP/1.1 500 Internal Server Error", response.status
+    assert_includes response.body, "Puma caught this error: app error"
+    assert handler_calls > 0, "Handler should have been called"
+  end
+
+  def test_lowlevel_error_handler_returns_malformed_headers_response
+    handler_calls = 0
+    malformed_handler = ->(_error, _env, _status) do
+      handler_calls += 1
+      # Return array but not proper 3-element Rack response
+      [500, "not_hash_of_headers", [""]]
+    end
+
+    options = {
+      lowlevel_error_handler: malformed_handler
+    }
+    broken_app = ->(_env) { raise StandardError, "app error" }
+    server_run(**options, &broken_app)
+
+    # Should get default error response when handler returns malformed response
+    response = send_http_read_response "GET / HTTP/1.1\r\n\r\n"
+
+    assert_equal "HTTP/1.1 500 Internal Server Error", response.status
+    assert_includes response.body, "Puma caught this error: app error"
+    assert handler_calls > 0, "Handler should have been called"
+  end
+
+  def test_lowlevel_error_handler_returns_malformed_body_response
+    handler_calls = 0
+    malformed_handler = ->(_error, _env, _status) do
+      handler_calls += 1
+      # Return array but not proper 3-element Rack response
+      [500, {}, "not_an_array_or_proc"]
     end
 
     options = {


joshuay03:
Which version of Puma were you using when you ran into the issue? https://github.com/puma/puma/pull/3696 was meant to address this exact problem, but I see that you're referencing code in 7.0.3 that includes that patch.

Ruto8:
> Which version of Puma were you using when you ran into the issue? https://github.com/puma/puma/pull/3696 was meant to address this exact problem, but I see that you're referencing code in 7.0.3 that includes that patch.

We experienced the issue on 6.6.1. I've seen the patch, but I was still able to reproduce the issue on master (which currently has 7.0.3), so I figured the fix might not take care of this? I admit I haven't looked super deeply into this, but I did revert my changes to see if the new tests fail without them, and they did.

I also tested the script I shared in the original message with and without my changes on 7.0.3 and there's a difference in the output, leading me to think #3696 doesn't fully solve this issue 🤔

MSP-Greg:
I haven't really looked into this, but the new tests, when applied to master (no patch in `server.rb`), result in the following three failures.  So, if the tests are valid...
```
Errors & Failures:

  1) Error:
TestPumaServer#test_lowlevel_error_handler_returns_malformed_array:
EOFError: EOFError
    test/helpers/test_puma/puma_socket.rb:282:in 'block (2 levels) in <module:PumaSocket>'
    test/helpers/test_puma/puma_socket.rb:228:in 'Kernel#loop'
    test/helpers/test_puma/puma_socket.rb:228:in 'block in <module:PumaSocket>'
    test/helpers/test_puma/puma_socket.rb:175:in 'TestPuma::PumaSocket#send_http_read_response'
    test/test_puma_server.rb:2242:in 'TestPumaServer#test_lowlevel_error_handler_returns_malformed_array'

  2) Error:
TestPumaServer#test_lowlevel_error_handler_returns_invalid_response:
EOFError: EOFError
    test/helpers/test_puma/puma_socket.rb:282:in 'block (2 levels) in <module:PumaSocket>'
    test/helpers/test_puma/puma_socket.rb:228:in 'Kernel#loop'
    test/helpers/test_puma/puma_socket.rb:228:in 'block in <module:PumaSocket>'
    test/helpers/test_puma/puma_socket.rb:175:in 'TestPuma::PumaSocket#send_http_read_response'
    test/test_puma_server.rb:2216:in 'TestPumaServer#test_lowlevel_error_handler_returns_invalid_response'

  3) Error:
TestPumaServer#test_lowlevel_error_handler_that_raises_exception:
EOFError: EOFError
    test/helpers/test_puma/puma_socket.rb:282:in 'block (2 levels) in <module:PumaSocket>'
    test/helpers/test_puma/puma_socket.rb:228:in 'Kernel#loop'
    test/helpers/test_puma/puma_socket.rb:228:in 'block in <module:PumaSocket>'
    test/helpers/test_puma/puma_socket.rb:175:in 'TestPuma::PumaSocket#send_http_read_response'
    test/test_puma_server.rb:2190:in 'TestPumaServer#test_lowlevel_error_handler_that_raises_exception'
```

lib/puma/server.rb
```diff
@@ -574,12 +574,24 @@ def client_error(e, client, requests = 1)
...
+          response = if handler.arity == 1
+                       handler.call(e)
+                     elsif handler.arity == 2
+                       handler.call(e, env)
+                     else
+                       handler.call(e, env, status)
+                     end
+
+          # Validate that the handler returned a proper Rack response
+          if response.is_a?(Array) && response.length == 3
```
MSP-Greg:
```suggestion
          if response.is_a?(Array) && response.length == 3 &&
            response[0].is_a?(Integer) && response[1].is_a?(Hash) &&
            (response[2].respond_to?(:each) || response[2].respond_to?(:call))
```
Not sure if we need to add this, but it validates the response as far as can easily be done?

Ruto8: Added in https://github.com/puma/puma/pull/3751/commits/f533d4b97572b8d0a8475071ceb4e1d32137da73, with  more tests

schneems:

```
```

joshuay03:
> > Which version of Puma were you using when you ran into the issue? #3696 was meant to address this exact problem, but I see that you're referencing code in 7.0.3 that includes that patch.
> 
> We experienced the issue on 6.6.1. I've seen the patch, but I was still able to reproduce the issue on master (which currently has 7.0.3), so I figured the fix might not take care of this? I admit I haven't looked super deeply into this, but I did revert my changes to see if the new tests fail without them, and they did.
> 
> I also tested the script I shared in the original message with and without my changes on 7.0.3 and there's a difference in the output, leading me to think #3696 doesn't fully solve this issue 🤔

Yeah, you're right. I didn't completely understand these changes before commenting. #3696 only addressed the case where a `NoMethodError` was raised when calling the handler on reactor wakeup. This is a more generic fix for all errors raised by the handler, which I think is a good thing to do.

Having a look.

joshuay03:
Okay so I can't seem to replicate this outside of the tests:

```ruby
# config.rb

lowlevel_error_handler ->(_error, _env, _status) do
  p "in lowlevel_error_handler"
  raise NoMethodError, "undefined method in handler"
end
```

```ruby
# app.ru

run ->(_env) do
  p "in app"
  raise StandardError, "app error"
end
```

```
puma [main] be bin/puma -C config.rb app.ru
* Puma version: 7.1.0 ("Neon Witch")
* Ruby version: ruby 4.0.0dev (2025-11-10T10:12:35Z master 557eec792e) +YJIT +PRISM [arm64-darwin25]
*  Min threads: 0
*  Max threads: 5
*  Environment: development
*          PID: 27451
Use Ctrl-C to stop
"in app"
2025-11-18 16:47:17 +0530 Rack app ("GET /" - (127.0.0.1)): #<StandardError: app error>
"in lowlevel_error_handler"
"in lowlevel_error_handler"
Error reached top of thread-pool: undefined method in handler (NoMethodError)
```

```
puma [main] curl http://0.0.0.0:9292
curl: (52) Empty reply from server
```

@Ruto8 Based on:

> We experienced the issue on 6.6.1. I've seen the patch, but I was still able to reproduce the issue on master (which currently has 7.0.3), so I figured the fix might not take care of this? I admit I haven't looked super deeply into this, but I did revert my changes to see if the new tests fail without them, and they did.

I suspect this is only an issue in tests, possibly because helpers like `send_http_read_response` don't properly handle empty replies, which is the existing behaviour when an error is rescued in the thread pool. So your tests aren't actually replicating Puma being in an invalid state. However, you also mentioned:

> I also tested the script I shared in the original message with and without my changes on 7.0.3 and there's a difference in the output, leading me to think https://github.com/puma/puma/pull/3696 doesn't fully solve this issue 🤔

I get the same results on 6-6-stable and main:

<details>
<summary>6-6-1</summary>

```
puma [6-6-stable] be ruby test.rb
🚀 Starting Puma server with faulty lowlevel_error_handler...
Server listening on port 62026

💥 Triggering reactor corruption...
2025-11-18 16:56:35 +0530 Rack app ("GET /trigger-corruption" - (127.0.0.1)): #<StandardError: App error to trigger handler>
Handler called 1 times
Handler called 2 times
Error reached top of thread-pool: undefined local variable or method 'undefined_method_that_causes_nomethoderror' for an instance of BlackHoleDemo (NameError)
Faulty handler called 2 times (should be > 0)

🕳️  Demonstrating 'black hole' behavior...
TCP connections will succeed, but HTTP requests will be silently lost
Request 1: 2025-11-18 16:56:36 +0530 Rack app ("GET /test-0" - (127.0.0.1)): #<StandardError: App error to trigger handler>
Handler called 3 times
Handler called 4 times
Error reached top of thread-pool: undefined local variable or method 'undefined_method_that_causes_nomethoderror' for an instance of BlackHoleDemo (NameError)
"Response: "
❌ Empty response
Request 2: 2025-11-18 16:56:36 +0530 Rack app ("GET /test-1" - (127.0.0.1)): #<StandardError: App error to trigger handler>
Handler called 5 times
Handler called 6 times
Error reached top of thread-pool: undefined local variable or method 'undefined_method_that_causes_nomethoderror' for an instance of BlackHoleDemo (NameError)
"Response: "
❌ Empty response
Request 3: 2025-11-18 16:56:36 +0530 Rack app ("GET /test-2" - (127.0.0.1)): #<StandardError: App error to trigger handler>
Handler called 7 times
Handler called 8 times
Error reached top of thread-pool: undefined local variable or method 'undefined_method_that_causes_nomethoderror' for an instance of BlackHoleDemo (NameError)
"Response: "
❌ Empty response
Request 4: 2025-11-18 16:56:36 +0530 Rack app ("GET /test-3" - (127.0.0.1)): #<StandardError: App error to trigger handler>
Handler called 9 times
Handler called 10 times
Error reached top of thread-pool: undefined local variable or method 'undefined_method_that_causes_nomethoderror' for an instance of BlackHoleDemo (NameError)
"Response: "
❌ Empty response
Request 5: 2025-11-18 16:56:36 +0530 Rack app ("GET /test-4" - (127.0.0.1)): #<StandardError: App error to trigger handler>
Handler called 11 times
Handler called 12 times
Error reached top of thread-pool: undefined local variable or method 'undefined_method_that_causes_nomethoderror' for an instance of BlackHoleDemo (NameError)
"Response: "
❌ Empty response

📊 Results:
  - TCP connections established: 5/5
  - HTTP responses received: 0/5
  - Timeouts: 0/5
  - Handler calls: 12

✅ BLACK HOLE REPRODUCED!
  - TCP connections work (5/5)
  - But HTTP requests are silently lost (0 responses)
  - This matches your production issue exactly

📝 Server-side logging analysis:
In production, you'd see minimal server errors despite client timeouts
This is because the reactor is corrupted but not completely dead

🧹 Cleaning up...

============================================================
SUMMARY: Puma Reactor 'Black Hole' Bug Reproduction
============================================================
1. ✅ Faulty lowlevel_error_handler raises NoMethodError
2. ✅ Reactor catches error, sets close_selector=false (reactor.rb:110)
3. ✅ TCP connections establish normally (no socket-level issues)
4. ✅ HTTP requests are silently dropped (reactor corruption)
5. ✅ Clients timeout without meaningful server-side errors

This exactly matches your production 'black hole' behavior where:
- Clients can establish TCP connections
- HTTP requests disappear without a trace
- Server appears healthy but doesn't process requests
- Socket backlog and TCP metrics appear normal
```

</details>

<details>
<summary>main</summary>

```
puma [main] be ruby test.rb
🚀 Starting Puma server with faulty lowlevel_error_handler...
Server listening on port 61943

💥 Triggering reactor corruption...
2025-11-18 16:54:43 +0530 Rack app ("GET /trigger-corruption" - (127.0.0.1)): #<StandardError: App error to trigger handler>
Handler called 1 times
Handler called 2 times
Error reached top of thread-pool: undefined local variable or method 'undefined_method_that_causes_nomethoderror' for an instance of BlackHoleDemo (NameError)
Faulty handler called 2 times (should be > 0)

🕳️  Demonstrating 'black hole' behavior...
TCP connections will succeed, but HTTP requests will be silently lost
Request 1: 2025-11-18 16:54:43 +0530 Rack app ("GET /test-0" - (127.0.0.1)): #<StandardError: App error to trigger handler>
Handler called 3 times
Handler called 4 times
Error reached top of thread-pool: undefined local variable or method 'undefined_method_that_causes_nomethoderror' for an instance of BlackHoleDemo (NameError)
"Response: "
❌ Empty response
Request 2: 2025-11-18 16:54:43 +0530 Rack app ("GET /test-1" - (127.0.0.1)): #<StandardError: App error to trigger handler>
Handler called 5 times
Handler called 6 times
Error reached top of thread-pool: undefined local variable or method 'undefined_method_that_causes_nomethoderror' for an instance of BlackHoleDemo (NameError)
"Response: "
❌ Empty response
Request 3: 2025-11-18 16:54:43 +0530 Rack app ("GET /test-2" - (127.0.0.1)): #<StandardError: App error to trigger handler>
Handler called 7 times
Handler called 8 times
Error reached top of thread-pool: undefined local variable or method 'undefined_method_that_causes_nomethoderror' for an instance of BlackHoleDemo (NameError)
"Response: "
❌ Empty response
Request 4: 2025-11-18 16:54:43 +0530 Rack app ("GET /test-3" - (127.0.0.1)): #<StandardError: App error to trigger handler>
Handler called 9 times
Handler called 10 times
Error reached top of thread-pool: undefined local variable or method 'undefined_method_that_causes_nomethoderror' for an instance of BlackHoleDemo (NameError)
"Response: "
❌ Empty response
Request 5: 2025-11-18 16:54:43 +0530 Rack app ("GET /test-4" - (127.0.0.1)): #<StandardError: App error to trigger handler>
Handler called 11 times
Handler called 12 times
Error reached top of thread-pool: undefined local variable or method 'undefined_method_that_causes_nomethoderror' for an instance of BlackHoleDemo (NameError)
"Response: "
❌ Empty response

📊 Results:
  - TCP connections established: 5/5
  - HTTP responses received: 0/5
  - Timeouts: 0/5
  - Handler calls: 12

✅ BLACK HOLE REPRODUCED!
  - TCP connections work (5/5)
  - But HTTP requests are silently lost (0 responses)
  - This matches your production issue exactly

📝 Server-side logging analysis:
In production, you'd see minimal server errors despite client timeouts
This is because the reactor is corrupted but not completely dead

🧹 Cleaning up...

============================================================
SUMMARY: Puma Reactor 'Black Hole' Bug Reproduction
============================================================
1. ✅ Faulty lowlevel_error_handler raises NoMethodError
2. ✅ Reactor catches error, sets close_selector=false (reactor.rb:110)
3. ✅ TCP connections establish normally (no socket-level issues)
4. ✅ HTTP requests are silently dropped (reactor corruption)
5. ✅ Clients timeout without meaningful server-side errors

This exactly matches your production 'black hole' behavior where:
- Clients can establish TCP connections
- HTTP requests disappear without a trace
- Server appears healthy but doesn't process requests
- Socket backlog and TCP metrics appear normal
```

</details>

The script seems to simply be checking for an empty response, which like I mentioned above is the expected behaviour. You could argue that Puma should be returning a 500 or something instead, but that's a separate issue. We should also probably look into why the handler is being called twice.

So, unless you're able to modify my minimal reproduction to put Puma into a bad state, I think it's safe to say that your original issue was indeed fixed by https://github.com/puma/puma/pull/3696. I can also see that the script mentions the reactor in a comment above, which is where that fix was targeted.

joshuay03:
@Ruto8 I'm closing this off given the above, please reopen if you're able to replicate a bad state.

Ruto8:
@joshuay03 that makes sense, thank you for looking into it. I was not able to replicate 👍

PR #3754, opened 2025-09-15
puma_http11.c: Use interned UTF-8 strings for hash keys

byroot:
`rb_str_new2` creates a mutable, ASCII-8BIT (binary) strings.
    
Hence, when used with `rb_hash_aset`, ruby always has to lookup the equivalent string in the fstring table, likely fail to find it, then dup and free the given string, and use that as the hash key.
    
By keeping interned strings in these global variables, we can save Ruby quite a bit of work. It's not a whole lot in the grand scheme of things, but still welcome.

There is also a preliminary commit to fix `rb_global_variable` usage.


schneems:
Thanks! I'm curious to know how you identified this optimization. Were you looking at metrics of some kind?

byroot:
No, I noticed a `rb_global_variable` misuse when reviewing another PR, and from there saw these global and how they were used.

It can probably show up on a profiler using a micro-benchmark, but is likely a drop in the water on more real life applications.

PR #3757, opened 2025-09-18
Update wait_for_less_busy_worker docs

schneems:
Details in individual commits

### DIFF
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -1214,12 +1214,12 @@ module Puma
     # requests to pick up new requests first.
     #
     # The default is 0.005 seconds.
-
+    #
+    # To turn off this feature, set the value to 0.
+    #
     # @note Cluster mode with >= 2 workers only.
     #
     # @note Interpreters with forking support only.
-    #
-    # To turn off this feature, set the value to 0.
     #
     # @see Puma::Server#handle_servers
     # @see Puma::ThreadPool#wait_for_less_busy_worker


lib/puma/dsl.rb
```diff
@@ -1207,13 +1207,18 @@ def shutdown_debug(val=true)
     end
 
 
+    # Maximum delay of worker (process) accept loop (when workers >= 2)
```
joshuay03: Nits.

1. I think it's obvious to Puma users that worker == process. We don't clarify this in any other DSL docs, nor should we need to—this should be assumed knowledge at this point (imo).
2. We can use notes to highlight that this is a cluster-only config (and in this case also only when workers >= 2) to be consistent with other configs (see my suggestion below).

```suggestion
    # Maximum delay of worker accept loop.
```

lib/puma/dsl.rb
```diff
@@ -1207,13 +1207,18 @@ def shutdown_debug(val=true)
...
+    #
     # Attempts to route traffic to less-busy workers by causing them to delay
```
joshuay03: While we're here, it seems 'them' should refer to the current (busy) worker, but the way this reads, one might assume it's referring to the less-busy worker.

```suggestion
    # Attempts to route traffic to less-busy workers by causing a busy worker to delay
```

lib/puma/dsl.rb
```diff
@@ -1207,13 +1207,18 @@ def shutdown_debug(val=true)
...
-    # listening on the socket, allowing workers which are not processing any
+    # listening on the socket, allowing workers which are not processing as many
     # requests to pick up new requests first.
     #
     # The default is 0.005 seconds.
```
joshuay03, line 1216:
```suggestion
    # The default is 0.005 seconds.

    # @note Cluster mode with >= 2 workers only.
```

lib/puma/dsl.rb
```diff
@@ -1207,13 +1207,18 @@ def shutdown_debug(val=true)
...
     #
-    # Only works on MRI. For all other interpreters, this setting does nothing.
+    # Only takes effect when there are two or more workers. For interpreters that
+    # do not support forking, this setting does nothing.
```
joshuay03: This information is repetitive.

```suggestion
    # For interpreters that do not support forking, this setting does nothing.
```

joshuay03: You could also turn the rest into a note:

```suggestion
    # @note Interpreters with forking support only.
```

lib/puma/dsl.rb
```diff
@@ -1207,13 +1207,18 @@ def shutdown_debug(val=true)
...
+    #
+    # To turn off this feature, set the value to 0.
```
joshuay03: This might be more appropriately located under the mention of the default value above.

lib/puma/dsl.rb
MSP-Greg: Reminder, not a lot of people use it (especially for production), but Windows MRI does not support forking.

lib/puma/dsl.rb
schneems: > I think it's obvious to Puma users that worker == process

Disagree. It’s puma specific terminology that competes with plenty of other “workers” in the ecosystem (like background workers) and generally I see newcomers stumble on the distinction. Yes, someone who is very is familiar with Puma will find the information redundant. I think that’s okay.

Just as a heads up: I’ll be slow to update and respond. Feeling much worse today.

joshuay03: > It’s puma specific terminology that competes with plenty of other “workers” in the ecosystem (like background workers) and generally I see newcomers stumble on the distinction.

I agree with this perspective in general, but I think in the context of DSL documentation, it's sufficient for just the `#workers` method docs to clarify that worker == process. For users tuning parameters like `wait_for_less_busy_worker`, we can reasonably consider the worker/process equivalence to be assumed knowledge. Anyway, this isn't a hill I'm willing to die on.

PR #3761, opened 2025-09-21
OpenSSL: use native ruby OpenSSL instead of MiniSSL. Closes #2188.

### LINKED ISSUE #2188: [Question] What's the purpose of Puma having its own MiniSSL?

swebb:
### Description

This is a spike for using ruby's OpenSSL directly instead of via Minissl. Closes https://github.com/puma/puma/issues/2188.

This PR is still very much a work in progress so I've made it a draft. I've got it doing a basic hello world server using ssl and some of the context arguments, but there is still a lot more work to do. I figured as something is working it might be worth opening a PR for feedback.

Note, as part of this PR I renamed `Puma::MiniSSL::Socket#peercert` (no underscore) to `Puma::MiniSSL::Socket#peer_cert` (has an underscore) to make it consistent with [OpenSSL::SSL::SSLSocket ](https://ruby-doc.org/3.4.1/exts/openssl/OpenSSL/SSL/SSLSocket.html#method-i-peer_cert). This gives a consistent api for sockets ([example usage](https://github.com/puma/puma/blob/30e0d2254043f0ef8b586f23bc5f06a59a45c219/lib/puma/log_writer.rb#L127)).

The biggest challenge I encountered is that [`OpenSSL::SSL::SSLSocket`](https://ruby-doc.org/3.4.1/exts/openssl/OpenSSL/SSL/SSLServer.html) has an `#accept` method but not `#accept_nonblock`. Obviously we don't want slow clients preventing other requests from being processed. The current workaround I've used is to call `#accept_nonblock` on the underlying tcp socket to start work when a new client connects. The socket is passed to a worker thread which calls the blocking `#accept` method (this is the same way [Webrick handles it](https://github.com/ruby/webrick/blob/6c95f0c4b9ee559de02cd9435e33614d249456be/lib/webrick/server.rb#L262)).  I don't know if this is acceptable for Puma. Note, there is `[OpenSSL::SSL::SSLSocket#accept_nonblock](https://devdocs.io/ruby~3.4/openssl/ssl/sslsocket#method-i-accept_nonblock); however, I haven't looked far enough into it yet to see if we can use it.

At a minimum this PR is still missing:

- [ ]  Outstanding context flags
- [ ] Update `Puma::Binder#inherit_ssl_listener` to support Native SSL and MiniSSL.
- [ ] What about JRUBY? Will this just work :tm:?
- [ ] Tests
- [ ] Documentation

Any feedback is greatly appreciate.

### DIFF
--- a/lib/puma/ssl/server.rb
+++ b/lib/puma/ssl/server.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+module Puma
+  module SSL
+    class Server
+      def initialize(svr, ctx)
+        @svr = svr
+        @ctx = ctx
+      end
+
+      def to_io
+        @svr
+      end
+
+      def accept_nonblock
+        socket = @svr.accept_nonblock
+        OpenSSL::SSL::SSLSocket.new(socket, @ctx)
+        # At this point the client has connected/accepted at the socket layer but hasn't finished
+        # the SSL handshake. We can't do a OpenSSL::SSL::SSLSocket#accept here because a slow client
+        # would block other clients from connecting.
+      end
+    end
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -11,6 +11,7 @@ gem "minitest-retry"
 gem "minitest-proveit"
 gem "minitest-stub-const"
 gem "concurrent-ruby", "~> 1.3"
+gem "debug"
 
 case ENV['PUMA_CI_RACK']&.strip
 when 'rack2'
--- a/ext/puma_http11/mini_ssl.c
+++ b/ext/puma_http11/mini_ssl.c
@@ -818,7 +818,7 @@ void Init_mini_ssl(VALUE puma) {
    * Returns `nil` when `MiniSSL::Context#verify_mode` is set to `VERIFY_NONE`.
    * @return [String, nil] DER encoded cert
    */
-  rb_define_method(eng, "peercert", engine_peercert, 0);
+  rb_define_method(eng, "peer_cert", engine_peercert, 0);
 
   rb_define_method(eng, "ssl_vers_st", engine_ssl_vers_st, 0);
 }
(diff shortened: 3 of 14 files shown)


schneems:

```
```

lib/puma/ssl/server.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Puma
+  module SSL
+    class Server
+      def initialize(svr, ctx)
```
tenderlove, line 6: `svr` is the underlying socket object?

tenderlove:
> The biggest challenge I encountered is that [`OpenSSL::SSL::SSLServer`](https://ruby-doc.org/3.4.1/exts/openssl/OpenSSL/SSL/SSLServer.html) has an `#accept` method but not `#accept_nonblock`. Obviously we don't want slow clients preventing other requests from being processed. The current workaround I've used is to call `#accept_nonblock` on the underlying tcp socket to start work when a new client connects. The socket is passed to a worker thread which calls the blocking `#accept` method (this is the same way [Webrick handles it](https://github.com/ruby/webrick/blob/6c95f0c4b9ee559de02cd9435e33614d249456be/lib/webrick/server.rb#L262)). I don't know if this is acceptable for Puma.

This is just my 2 cents (since I'm not a maintainer), but I think this would work fine.

> Note, there is `[OpenSSL::SSL::SSLSocket#accept_nonblock](https://devdocs.io/ruby~3.4/openssl/ssl/sslsocket#method-i-accept_nonblock); however, I haven't looked far enough into it yet to see if we can use it.

I think we could use this.  It looks like `OpenSSL::SSL::SSLSocket#accept_nonblock` is for doing the SSL handshake in a non-blocking fashion.  I'm not exactly sure how to wire this in to Puma, but we could do a non-blocking accept for the TCP socket, then use `accept_nonblock` on the SSL socket, and only once that has succeeded, start reading the request from the socket in the normal fashion.

PR #3671, opened 2025-06-29
Add `on_force` option to `shutdown_debug`

joshuay03:
### Description

Closes https://github.com/puma/puma/issues/3666

### Your checklist for this pull request

- [x] I have reviewed the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.
- [x] I have added (or updated) appropriate tests if this PR fixes a bug or adds a feature.
- [x] My pull request is 100 lines added/removed or less so that it can be easily reviewed.
- [x] If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.
- [x] If this closes any issues, I have added "Closes `#issue`" to the PR description or my commit messages.
- [x] I have updated the documentation accordingly.
- [x] All new and existing tests passed, including Rubocop.

PR #3763, opened 2025-09-22
Fix Ctrl-C, set `force_shutdown_after` and `worker_shutdown_timeout` to small values

### LINKED ISSUE #3569: Puma hangs when delivering streaming responses, Ctrl-C cannot interrupt/shutdown.
Launching puma with the `config.ru` given in "To Reproduce" and opening the URL in a browser will prevent `Ctrl-C` from shutting down Puma (it's not as strange as it sounds).

If you close the browser, `puma` will shut down. That happens because the SSE connection is closed, which allows the `handle_client` to exit.

**To Reproduce**

```ruby
INDEX = <<~HTML
<!doctype html>
  <html>
  <head>
    <title>SSE Hang</title>
  </head>
  <body>
    <h1>SSE Hang</h1>
    <ul id="messages"></ul>
    <script>
      var source = new EventSource('/sse');
      source.onmessage = function(event) {
        console.log(event.data);
        var li = document.createElement('li');
        li.textContent = event.data;
        messages.appendChild(li);
      };
    </script>
  </body>
</html>
HTML

def index_response
  [200, {"content-type" => "text/html"}, [INDEX]]
end

def sse_response
  event_stream = proc do |stream|
    loop do
      stream.write "data: #{Time.now}\n\n"
      sleep 1
    end
  ensure
    stream.close
  end
  
  [200, {
    "content-type"  => "text/event-stream",
    "cache-control" => "no-cache",
  }, event_stream,]
end

run do |env|
  request = Rack::Request.new(env)
  
  case request.path
  when "/sse"
    sse_response
  else
    index_response
  end
end
```

Run it with:

```
> puma
* Puma version: 6.5.0 ("Sky's Version")
* Ruby version: ruby 3.3.6 (2024-11-05 revision 75015d4c1f) [x86_64-linux]
*  Min threads: 0
*  Max threads: 5
*  Environment: development
*          PID: 53783
Use Ctrl-C to stop
^C^C^C^C
<browser closed>
Error reached top of thread-pool: Socket timeout writing data (Puma::ConnectionError)
=== puma shutdown: 2024-12-03 12:06:01 +1300 ===
- Goodbye!
```

In comparison, `falcon` will exit gracefully.

cc @ayushn21

MSP-Greg:
### Description

Depending on the configuration and response types, using Ctrl-C to shutdown Puma may either not work or take a long time.

When `SIGINT` is received, set `force_shutdown_after` to 0.5 if it has not been set to a positive number.  When running clustered, set `worker_shutdown_timeout` to 2.

Note that `Puma::ControlCLI` (`pumactl`) does not use `SIGINT`, and most external control apps would use `SIGTERM` to shutdown Puma.

Closes #3569

### DIFF
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -451,6 +451,12 @@ module Puma
 
       begin
         Signal.trap "SIGINT" do
+          # setting the below to allow Ctrl-C to shutdown Puma, especially with
+          # hijacked responses.  See https://github.com/puma/puma/issues/3569
+          if @options.fetch(:force_shutdown_after, -1) < 0
+            @options[:force_shutdown_after] = 0.5
+            log "- Setting 'force_shutdown_after' to 0.5"
+          end
           stop
         end
       rescue Exception
--- a/lib/puma/cluster.rb
+++ b/lib/puma/cluster.rb
@@ -441,6 +441,13 @@ module Puma
       spawn_workers
 
       Signal.trap "SIGINT" do
+        # setting the below to allow Ctrl-C to shutdown Puma, especially with
+        # hijacked responses.  See https://github.com/puma/puma/issues/3569
+        @options[:worker_shutdown_timeout] = 2
+        if @options.fetch(:force_shutdown_after, -1) < 0
+          @options[:force_shutdown_after]    = 0.5
+          log "- Setting 'force_shutdown_after' to 0.5"
+        end
         stop
       end
 


schneems:
I mostly traffic in SIGTERM and SIGKILL. I'm unsure of the possible negative implications of this setting or what users may be expecting or find surprising.

I would expect that Ctrl+C would behave like SIGTERM (graceful exit) by letting requests finish. I've seen some programs that will print an output like "Shutting down press Ctrl-C again to force shutting down." Maybe instead of forcing a timeout here, we give that option back to the user by setting some state when SIGINT is first pressed, and if pressed again treating it as a SIGKILL?

Implementation wise I'm unsure if it's better to use internals like `force_shutdown_after` or to delegate to another signal like SIGTERM/SIGKILL. Looks like it was originally added in https://github.com/puma/puma/pull/1377 for MRI (before only for JRuby?). That implementation called `stop` rather than `graceful_stop`, but I'm unsure if that was intentional or not.

MSP-Greg:
> Maybe instead of forcing a timeout here, we give that option back to the user by setting some state when SIGINT is first pressed, and if pressed again treating it as a SIGKILL?

A long time ago, I think I got very frustrated with the `graceful_stop` vs `stop`, along with how `pumactl` and/or the control app responded to them.  Lastly, I think SIGTERM is wonky on Windows.  I'll have another look.

lib/puma/launcher.rb
```diff
@@ -451,6 +451,12 @@ def setup_signals
 
       begin
         Signal.trap "SIGINT" do
+          # setting the below to allow Ctrl-C to shutdown Puma, especially with
+          # hijacked responses.  See https://github.com/puma/puma/issues/3569
+          if @options.fetch(:force_shutdown_after, -1) < 0
```
dentarg, line 456: After https://github.com/puma/puma/pull/3671 we should always have a value here, no need for the `-1` fallback, right?

afurm:
The `cluster.rb` signal handler sets `worker_shutdown_timeout = 2` unconditionally, but `launcher.rb` (single mode) doesn't set it at all. Since clustered mode also runs the launcher path for the master process, will the master respect the `worker_shutdown_timeout` from cluster.rb, or does it need to be mirrored in launcher.rb?

PR #3765, opened 2025-09-22
Remove global side effect from test

schneems:
## Before

This test asserts that when certain configuration options are specified that a custom load error is raised if concurrent-ruby cannot be found. It does this by: 

- finding the real file on disk
- renaming the file
- running code and asserting an error with a specific message
- ensuring the file rename is reverted

Most developers would be surprised that running the test suite of one project might cause other code running on the same machine to fail, but because these modifications to disk are global, that's what happens.

## After

Instead of making a real system change, this mocks the interface where that file is loaded i.e. the "require" call:

```ruby
    def require_processor_counter
      require 'concurrent/utility/processor_counter'
    rescue LoadError
```

This is done by monkey patching the `require` method on the specific instance of Configuration under test. This means that not only does it have no side effects that are visible by other running processes on the same machine, other instances of the configuration class won't be affected.

To get this strategy to work we have to force the instance to re-parse the environment by explicitly calling the `puma_default_options` method (which in turn eventually calls the code that generates the warning). This is needed because the Puma::Configuration class eagerly evaluates the environment and calls this logic in the constructor

```ruby
module Puma
  # ...
  class Configuration
    # ...
    def initialize(user_options={}, default_options = {}, env = ENV, &block)
      default_options = self.puma_default_options(env).merge(default_options)
```

### DIFF
--- a/test/test_web_concurrency_auto.rb
+++ b/test/test_web_concurrency_auto.rb
@@ -45,32 +45,21 @@ class TestWebConcurrencyAuto < TestIntegration
     assert_equal expected, actual
   end
 
-  # Rename the processor_counter file, then restore
   def test_web_concurrency_with_concurrent_ruby_unavailable
     file_path = nil
     skip_unless :fork
 
-    ccr_gem = 'concurrent-ruby'
-    file_require = 'concurrent/utility/processor_counter'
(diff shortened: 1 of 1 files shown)


MSP-Greg:
I guess the fact that I segregate all my 'bundle' gems would affect this...

PR #3767, opened 2025-09-22
server.rb - fixup `closed_socket?` in case `state` is `nil`

### LINKED ISSUE #3766: `NoMethodError` on `NilClass` under JRuby
**Describe the bug**

When using the system JRuby in Debian Trixie (JRuby 9.4.8.0, Java 21), I get the following error in the console output on every request, with no backtrace:

```
Error reached top of thread-pool: undefined method `>=' for nil:NilClass (NoMethodError)
```

**Puma config:**

I'm running the command `puma` with the following `config.ru`:

```
run (Proc.new do
  [200, {}, ['hi']]
end)
```

**To Reproduce**

* Start with a fresh install of Debian 13
* `apt install jruby`
* `jruby -S gem install puma`
* `puma` (with the above `config.ru`)

**Expected behavior**

"hi" in the browser.

**Desktop**
 - Debian 13, Trixie (running on OrbStack on a Mac M1)
 - Puma 7.0.3, Romantic Warrior

MSP-Greg:
### Description

Currently, Puma has the following code to check if `BasicSocket#getsockoptgetsockopt` can determine whether a socket is closed.  For most platforms & OS's, this works.  #3766 shows that the check may not always work.

PR adds one line to verify the return of `BasicSocket#getsockoptgetsockopt`.

https://github.com/puma/puma/blob/474f4ac87b3a49884080f6611636eff2e6e61cee/lib/puma/server.rb#L168-L170

Closes #3766

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -217,6 +217,7 @@ module Puma
           false
         else
           state = tcp_info.unpack(UNPACK_TCP_STATE_FROM_TCP_INFO)[0]
+          return false unless state
           # TIME_WAIT: 6, CLOSE: 7, CLOSE_WAIT: 8, LAST_ACK: 9, CLOSING: 11
           (state >= 6 && state <= 9) || state == 11
         end


ahorek:
Hi @MSP-Greg 
this is a known bug affecting aarch64 targets, resolved in JRuby 9.3.10 nearly three years ago. While your fix is ok, it's a workaround for a few unsupported JRuby versions that shouldn't really be necessary.

ahorek:
The fix for JRuby 9.3.10.0 and 9.4.0.0 is here: https://github.com/jruby/jruby/pull/7590

so, 9.4.8.0 should NOT be affected, which is odd. It might be worth investigating how CRuby behaves on that platform...

schneems:
I confirm this removes the immediate error. I'm unsure if this returns the correct state or not. Does `state == nil` always mean that the socket is still open? Do we have more information about what's happening there?

MSP-Greg:
> Does state == nil always mean that the socket is still open?  Do we have more information about what's happening there?

No, I'm guessing it means that the constants are defined to appear to allow `getsockopt` to determine state, but it doesn't work, since the return should be an int.

I added this mostly due to the "Puma can't fail".

schneems:
I think we should log in this state. Otherwise this behavior might be confusing. Right now it will fail loudly with that nil error. If it’s failing silently, we should at least warn.

schneems:

```
```

PR #3772, opened 2025-09-23
Fix flappy `test/test_thread_pool.rb` tests 

### LINKED ISSUE #3771: Flappy test: TestThreadPool#test_shutdown_with_grace [test/test_thread_pool.rb:362]:
In an effort to make the CI more robust, I think we should open tracking issues about failing tests on CI with the idea that we should work to make them rock-solid or eliminate them if we cannot and determine they provide minimal value.

```
Errors & Failures:

  1) Failure:
TestThreadPool#test_shutdown_with_grace [test/test_thread_pool.rb:362]:
Expected: 2
  Actual: 1

695 runs, 1902 assertions, 1 failures, 0 errors, 62 skips
```

This occured on `ubuntu-22.04 jruby` at https://github.com/puma/puma/actions/runs/17949207702/job/51043652674?pr=3769. (But I'll be re-starting the test)

schneems:
You can validate this by running this command in a loop with jruby

```
$ chruby jruby 10.0.2.0                                                             
$ while be m test/test_thread_pool.rb; do echo "Test passed, running again..."; done
```

If it exits, we missed a flappy test.

Close #3771

### DIFF
--- a/test/test_thread_pool.rb
+++ b/test/test_thread_pool.rb
@@ -195,10 +195,10 @@ class TestThreadPool < PumaTest
 
     assert_equal 2, pool.spawned
 
-    pool.trim
+    pool.trim(wait: true)
     assert_equal 1, pool.spawned
 
-    pool.trim
+    pool.trim(wait: true)
     assert_equal 1, pool.spawned
   end
 
--- a/test/test_thread_pool.rb
+++ b/test/test_thread_pool.rb
@@ -340,7 +340,8 @@ class TestThreadPool < PumaTest
     timeout = 0.01
     grace = 0.01
 
-    rescued = []
+    # JRuby Array#<< is not thread-safe
+    rescued = Queue.new
     pool = mutex_pool(2, 2, pool_shutdown_grace_time: grace) do
       begin
         pool.with_force_shutdown do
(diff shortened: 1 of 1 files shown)


MSP-Greg:
Remove line 64 `Thread.pass`?

schneems:
> Remove line 64 Thread.pass?

Moved it after the mutex.

MSP-Greg (approved):

MSP-Greg:
@schneems

See https://github.com/puma/puma/commit/7d6da2c2f210da.  It did pass with two CI runs, but I'm running CI on something else and the macOS Jruby jobs often timeout during tests.  When they do, the last step ('Complete job') shows the following:
```
Cleaning up orphan processes
Terminate orphan process: pid (12124) (java)
```

schneems:
@MSP-Greg following up on your last comment, I don't totally follow the implication.

There's a possibility for code like this while loop to never exit, which isn't great. We should put some time or iteration bounds on it, like: Raise an error if it iterates for more than 5 seconds by default, and allow that value to be configured.

MSP-Greg:
@schneems

Sorry, I combined two topics.  Re JRuby failures:

* The tests run long, but they complete just at the time limit.  See [completed test step](https://github.com/MSP-Greg/puma/actions/runs/17949987142/job/51046432079#step:8:807), and there are no orphan processes, and the tests weren't frozen, at least according to the timestamps.

* The test hang, and the tests fail as shown [above](https://github.com/puma/puma/pull/3772#issuecomment-3325952631).

The above often can't be repro'd locally.

I was referring to the 'Thread.pass' and 'sleep' statements could probably be removed.

Also, regarding 'stuck' tests, a couple tests have the following code, which is hacky, but will wait for a limited time:
```
start = Time.now
Thread.pass until pool.spawned == 1 || Time.now - start > 1
```

Issue #3755, opened 2025-09-16
Command for `puma config` and `puma config --format=json`


schneems:
## Problem

When people report a problem with Puma, their configuration can affect behavior. There's currently no way to get puma to tell you what canonical information it's using. Instead people might report their config file contents or flags, but might miss or forget that some other behavior or setting is taking affect.

In addition to helping with bug reports, there are cases where a tool might wish to warn or error on known problematic configuration. For example the a buildpack might want to suggest a specific value for persistent_timeout might cause problems (for example). To be able to do this, we need a way to programmatically access resolved configuration values.

## Suggested solution

The goal is to produce human and machine-readable output for the current configuration values of Puma. It should resolve values as if `puma serve` had been run, that means respecting the configuration precedents and defaults, such as reading in the `config/puma.rb` file (if it exists).

For other puma flags that you see on `--help` those should be passed after a `--` like:

```
puma config -- --silent --threads 3 --preload
```

I'm unsure if OptionParser natively supports that syntax, but it's standard way to say "I'm not configuring the command being run (`puma config`) instead I'm giving information to that command (the flags that would otherwise be used with `puma serve`). 

While a pretty human output would be nice, I'm mostly interested in a machine-readable format, so I could run `puma config --format=json` and then parse the output to get a consistent interface.

The keys will change between Puma versions, and the schema should accommodate a way to handle that. I think it needs some metadata (or another word), such as:

```json
{
	"metadata": {
		"version": "7.0.0.pre1"
	},
	"config": {
		"http_content_length_limit": {
			"value": 2000000000,
			"set_by": "cli"
		},
		"worker_culling_strategy": {
			"value": "oldest",
			"set_by": "default"
		},
		"persistant_timeout": {
			"value": 95,
			"set_by": "env"
		},
		...
	}
}
```

When possible, prefer objects over lists for JSON formatting, as it makes adding things easier in the future (such as deprecation notices or additional information.

In the output, I would love to see not just the value but also where it came from. This helps with debugging if you're looking at the config and shouting "WHY ON EARTH IS `persistant_timeout` not correct?" it will hint that you need to look at `PUMA_PERSISTANT_TIMEOUT` env var. If this part is too difficult, we can break the task into two pieces by implementing the new CLI command first, refactoring the configuration precedence/selection code, and then implementing that behavior.

An example of prior art is this PR to add `bundle list --format=json` to bundler. https://github.com/rubygems/rubygems/pull/8728

dentarg:
[...]

PR #3773, opened 2025-09-24
add config command

PauloVilarinho:
### Description
closes https://github.com/puma/puma/issues/3755

This Pr adds a config command that adds a more elegant way to get the puma config for debugging and or asking for validation from other people. This is still a draft so I'm still working on tests and will add the json flag as requested in the issue if you guys thinks this is a good solution to go forward.

### DIFF
--- a/lib/puma/cli.rb
+++ b/lib/puma/cli.rb
@@ -29,6 +29,7 @@ module Puma
       @argv = argv.dup
       @log_writer = log_writer
       @events = events
+      @command = "run"
 
       @conf = nil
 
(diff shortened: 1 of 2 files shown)


schneems:
Thanks a ton for spiking this out. Sorry for the delayed review (OSS is feast and famine sometimes). I took a look and came to the conclusion that my original request needs some tweaking. I wrote some thoughts on what an API might look like https://github.com/puma/puma/issues/3755#issuecomment-3368542641. I don't think we can move forward with this directly as-is, but depending on how that conversation goes, I think re-using the existing config logging like you did here is the way to go, we just need some more ways to tweak it.

schneems:
Going to close for now, can move to a new PR for new work

PR #3774, opened 2025-09-24
Set `Thread.current.puma_server` in Thread init code, not every request

MSP-Greg:
### Description

Currently, the following is set in every call to `Server#process_request`:
```ruby
# Advertise this server into the thread
Thread.current.puma_server = self
```

PR moves code around to set it once in the `Thread.new` block.

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -259,8 +259,7 @@ module Puma
 
       @status = :run
 
-      @options[:puma_server] = self
-      @thread_pool = ThreadPool.new(thread_name, options) { |client| process_client client }
+      @thread_pool = ThreadPool.new(thread_name, options, server: self) { |client| process_client client }
 
       if @queue_requests
         @reactor = Reactor.new(@io_selector_backend) { |c|
--- a/lib/puma/thread_pool.rb
+++ b/lib/puma/thread_pool.rb
@@ -37,11 +37,8 @@ module Puma
     # The block passed is the work that will be performed in each
     # thread.
     #
-    def initialize(name, options = {}, &block)
-      # below is for CI
-      @server = options.respond_to?(:user_options) ?
-        options.user_options.delete(:puma_server) :
-        options.delete(:puma_server)
+    def initialize(name, options = {}, server: nil, &block)
+      @server = server
 
       @not_empty = ConditionVariable.new
       @not_full = ConditionVariable.new


lib/puma/server.rb
```diff
@@ -262,6 +259,7 @@ def run(background=true, thread_name: 'srv')
 
       @status = :run
 
+      @options[:puma_server] = self
```
joshuay03: Can we not pass this into the thread pool initialiser? It doesn't feel right to mutate the options at this stage.

lib/puma/thread_pool.rb
```diff
@@ -34,6 +38,11 @@ class ForceShutdown < RuntimeError
     # thread.
     #
     def initialize(name, options = {}, &block)
+      # below is for CI
+      @server = options.respond_to?(:user_options) ?
+        options.user_options.delete(:puma_server) :
+        options.delete(:puma_server)
```
joshuay03: My personal opinion is that having to do this negates the micro optimisation gained from the core of this change (I don't think any lib code should exist just to satisfy CI). Could you link to the specific CI code that needs this? Is it just a naming conflict?

joshuay03 (commented):
I like the overall idea, just sharing my two cents on the implementation.

lib/puma/thread_pool.rb
MSP-Greg: > negates the micro optimization gained from the core of this change

I'm not really concerned about code that runs once for each ThreadPool instance.  I wanted to add it to `options` before the call, and delete it, but `UserFileDefaultOptions` isn't a hash, it just implements some methods of Hash.

Also, something that is currently set on every request is possibly more than a micro-optimization, and removing it (and doing it in the init) mean's one can't be sidetracked by the question "Why is this set in every request?".   I really hate it when that happens...

lib/puma/server.rb
MSP-Greg: I think one test file ([test_thread_pool.rb](https://github.com/puma/puma/blob/master/test/test_thread_pool.rb)) creates a ThreadPool.  I suppose we could change the method signature to the below or something similar:
```
def initialize(name, options = {}, server = nil, &block)
```

I'll look at changing it.

MSP-Greg:
Just added a commit changing the ThreadPool init method signature.

lib/puma/thread_pool.rb
```diff
@@ -5,6 +5,10 @@
 require_relative 'io_buffer'
 
 module Puma
+
```
joshuay03, line 8: Nit - feel free to ignore if this is your personal preference.


```suggestion
```

joshuay03 (approved):
Nice!

lib/puma/thread_pool.rb
schneems: As an aside: I would like (someone) to remove all `options = {}` in favor of explicit KWARGS where possible into the future.

PR #3787, opened 2025-10-03
feat: make it possible to selectively enable control server commands

stanhu:
In production, we want to make the `/stats` and `/gc-stats` endpoints available, but we don't want to expose the ability to manipulate the state of the processes.

This commit adds a `enabled_commands` option. For example:

```
activate_control_app "tcp://127.0.0.1:8182", { enabled_commands: %w[gc-stats stats] }
```

### Description


stanhu:
@schneems What do you think of this pull request?

MSP-Greg:
I have thought about this previously.  I was thinking more a 'data_only' switch, so the command strings weren't needed.  Another option would be two passwords, one for 'admin/all' access, another (or nil) for 'data only'.

stanhu:
@MSP-Greg Ok, I updated this pull request accordingly. I didn't include `backtraces` because this might expose details only an admin should know.

MSP-Greg:
@stanhu

To the above message I should have added "Let's see if @schneems has any thoughts".  Otherwise, LGTM.

Nit pick:
```ruby
      data_only = @options[:control_data_only]
      app = Puma::App::Status.new @launcher, token, data_only
```
Since `data_only` is only used on the next line, maybe:
```ruby
      app = Puma::App::Status.new @launcher, token, @options[:control_data_only]

**Review by MSP-Greg — APPROVED** [review]

**Comment by stanhu** [comment]
Just pushed a slight correction to fix the commit message to `data_only`.

File: lib/puma/app/status.rb [hunk 1]
```diff
@@ -7,13 +7,16 @@ module App
...
     class Status
       OK_STATUS = '{ "status": "ok" }'.freeze
+      READ_ONLY_COMMANDS = %w[gc-stats stats].freeze
 
       # @param launcher [::Puma::Launcher]
       # @param token [String, nil] the token used for authentication
+      # @param data_only [Boolean] if true, restrict to read-only data commands
       #
-      def initialize(launcher, token = nil)
+      def initialize(launcher, token = nil, data_only = false)
```
schneems: Please use kwargs instead of positional args here for the newly added `data_only`.

@dentarg @MSP-Greg this class doesn't have any docs regarding public/private. I'm unsure if people use it directly, you think it's okay to change `token` to also be a kwarg or is that breaking?

lib/puma/app/status.rb
```diff
@@ -25,8 +28,13 @@ def call(env)
 
         # resp_type is processed by following case statement, return
         # is a number (status) or a string used as the body of a 200 response
+        command = env['PATH_INFO'][/\/([^\/]+)$/, 1]
+        if @enabled_commands && !@enabled_commands.include?(command)
+          return rack_response(404, "Command disabled", 'text/plain')
```
schneems: If someone hits this with a non-existant command, it's not technically "disabled" but it is unavailable. 

```suggestion
          return rack_response(404, "Command #{command.inspect} unavailable", 'text/plain')
```

Also I like showing people what the code is seeing which might help them realize there's a typo or something else is off like random accidental whitespace.

lib/puma/app/status.rb
MSP-Greg: > I'm unsure if people use it directly

I've got several "Don't break" bookmarks, and I don't recall any using this directly.

> you think it's okay to change token to also be a kwarg or is that breaking?

Do you really think these need to change to kwargs?  Adding an additional positional at the end, where the proceeding arg also has a default value shouldn't be a problem.  Also, I don't see the arg list growing.  Granted, kwargs are 'identifying', it not like the init code is used frequently...

lib/puma/app/status.rb
stanhu: 👍

lib/puma/app/status.rb
stanhu: I refactored this method using kwargs. I think it would be rare for projects to instantiate `Puma::App::Status.new` directly.

stanhu:
@schneems Thanks for the review! Could you take a look again?

lib/puma/app/status.rb
schneems: This comes down to personal preference. I have ADHD and kwargs help offload some memory of “what is this art supposed to be representing” to the name. Makes it easier for me to spot mistakes.

the downside is that it’s slightly slower than an all-positional method call. But for object instantiation, especially one that’s not in a hot loop, it’s more than worth it (and maybe even **in** a hot loop might not move the needle, I’ve not actually benchmarked.

> Don't break" bookmarks

If not already in existing docs, it would be a good PR to note this in the code.

schneems:
Thanks! Approved the change and restarted the failing test.

Issue #3715, opened 2025-08-29
TechEmpower benchmark is a lot slower with 7.0.0.pre1


p8:
### Describe the bug

The TechEmpower benchmark runs for `rack-puma-mri-sequel-raw`  and `rack-puma-jruby-sequel-raw` show a significant performance regression after upgrading to Puma 7.0.0.pre1.

**With Puma 6.6.0:**

<img width="1175" height="215" alt="Image" src="https://github.com/user-attachments/assets/49ba7ab8-9723-4fc8-956a-5f9008f77541" />

https://www.techempower.com/benchmarks/#section=test&runid=9d2e8e26-cb0c-48ff-8b97-dc2081b2db61&w=zik0zj-zik0zj-zik0xr-8vn08v&f=zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zf-zik0zj-zik0zj-zik0zj-zik0zj-18y67

**After upgrading to 7.0.0.pre1**

Note that `rack-sequel-postgres-puma-mri` and `rack-sequel-puma-mri` still run on Puma 6.6.0.

<img width="1178" height="215" alt="Image" src="https://github.com/user-attachments/assets/ae3bfe72-420a-45d2-81a6-383060bc1733" />

https://www.techempower.com/benchmarks/#section=test&runid=5db88eea-ccd9-4d03-bcd0-e5c8f90b921d&w=zik0zj-zik0zj-zik0xr-zik0zj&f=zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zf-zik0zj-zik0zj-zik0zj-zik0zj-27wr27

### Puma config:

The config can be [found in the repo](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Ruby/rack/config/puma.rb).
The [following command](https://github.com/TechEmpower/FrameworkBenchmarks/blob/b83fc4ad9067ec660640f464e3e31d061b0f7b4a/frameworks/Ruby/rack/rack.dockerfile#L24) is used for starting puma.

```bash
bundle exec puma -C config/puma.rb -b tcp://0.0.0.0:8080 -e production
```

### To Reproduce

Checkout the repo at: https://github.com/TechEmpower/FrameworkBenchmarks

Run the rack test with:

```
./tfb --test rack
```
I'm seeing things like:
[...]

PR #3794, opened 2025-10-14
Reintroduce keepalive "fast inline" behavior: 8x faster JRuby performance

### LINKED ISSUE #3788: Puma 7 on JRuby is 20 times slower on TechEmpower
**Describe the bug**
The TechEmpower benchmark runs for `rack-puma-jruby-sequel-raw` show a significant performance regression after upgrading to Puma 7. Fixes in 7.0.1 improved the results somewhat but there still is a big gap with 6.6.0.

**With Puma 6.6.0:**

Image

https://www.techempower.com/benchmarks/#section=test&runid=9d2e8e26-cb0c-48ff-8b97-dc2081b2db61&p=zik0zj-zik0zj-zih7un-zik0zj-zik0zj-cn3&w=zik0zj-zik0zj-zik0xr-8vn08v&f=zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zf-zik0zj-zik0zj-zik0zj-zik0zj-18y67

**With Puma 7.0.3:**

Image

https://www.techempower.com/benchmarks/#section=test&runid=d27544b6-7365-4269-a4d4-f908f0d21a3e&p=zik0zj-zik0zj-zih7un-zik0zj-zik0zj-1ekf&f=zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zj-zik0zf-zik0zj-zik0zj-zik0zj-zik0zj-hra0hr

**To Reproduce**

Checkout the repo at: https://github.com/TechEmpower/FrameworkBenchmarks

Run the rack test with:

```
./tfb --test rack-jruby --type fortune
```

schneems:
**Honesty in benchmarking disclaimer**: I'm having a bit of fun with the PR title, it's an 8x perf improvement from Puma 7 on JRuby, but that's because we accidentally made perf worse. Compared to 6.6.1,  JRuby performance on this specific benchmark is 1.45x better (which is still great IMHO, but not as fun as "8x"). It's faster on MRI as well, but not quite the same huge improvement. It is 1.37x faster than Puma 7 and 1.08x faster than Puma 6 on vanilla Ruby. (numbers later)

Short: This reintroduces a performance optimization dropped in Puma 7, and adds a guard to prevent "unfair" or "longtail" request distribution when the server is overloaded.

## Context

Puma 6 and before had a "fast inline" loop. A connection can have many requests, as an optimization, this "fast inline" feature would allow a Puma thread to respond to multiple requests on the same connection in a row without requiring that the connection be placed in the queue or the thread needing to go through mutex lock/unlock-ing to get it back. This improved performance but created a "longtail" situation where one keepalive connection can "monopolize" a thread for up to 10 requests in a row. If all threads are busy, this blocks an incoming request for 10x the average time to serve a request; this was reported in https://github.com/puma/puma/issues/3487. 

Puma 7 changed this behavior to ensure that all responses are "fair" and that a keepalive connection cannot "monopolize" a thread. This eliminated the "longtail" behavior. 

Recently, https://github.com/puma/puma/issues/3788 was opened to highlight that while MRI numbers on TechEmpower benchmarks look pretty decent, the JRuby ones fell off a cliff. Previously, we worked on tuning/improving the MRI performance on a similar report against 7.0.0.pre1 https://github.com/puma/puma/issues/3715. We fixed that by removing an env var from the TechEmpower setup and increasing the default keepalive connection. This new report with JRuby seems directly related to the overhead involved in coordinating between many threads. 

One way to think of this is with https://en.wikipedia.org/wiki/Amdahl%27s_law. The amount of speedup in a system by adding a new resource is directly limited by the amount of coordination needed between resources. When a request is placed on the Puma queue, other threads all coordinate via mutexes and condition variables in order to do bookkeeping and only pull the next request in line. MRI also has a GIL/GVL; in addition to these locks, JRuby does not have a GIL. Each thread can independently run Ruby code. It seems that decreasing pressure on the locking mechanisms is much more important for JRuby than for Cruby (MRI).

A key fact in the prior "longtail" reproduction is that it only occurs when all Puma threads are busy, that is, when Puma is overloaded. That is because if a thread is busy working on several requests in a row and a new request comes in, but can be picked up by another thread ... that request does not have to wait and can be served fairly.

## This change

This PR reintroduces the prior "fast inline" behavior to allow a thread to be monopolized, but with a key difference. This will only be allowed when Puma detects that there are idle threads waiting for a new request. If all threads are busy, it will put the next request on the connection in the queue for proper ordering. This approach preserves the Puma 7 "fair" ordering in addition to reintroducing the Puma 6 "fast inline" optimization.

Puma 6 had a setting `max_fast_inline` that would auto-close a connection after a number of requests on it were served (default was 10). That was needed to prevent a DoS as otherwise an attacker could saturate all threads with keepalive connections and then prevent any new requests from being served. This change does NOT reintroduce this concept. Connections will be closed after they've reached the maximum keepalive number of requests (default is 999). As long as the system has excess capacity (spare threads), it is not at risk of DoS via this vector. As soon as all threads are saturated, it will not close the connection (as before) but instead place it in the queue.

## Reproduction

Performance comparison using this script https://gist.github.com/schneems/bc7ee417007d467a63abcf793793b989/b51629f86f6e05e2afc52041875033c6bffa5012

```
$ git co <branchname> && ./tmp/profile_puma.sh
```

Results (jruby 10.0.2.0):

```
v6.6.1    Requests/sec: 116,403.14
v7.04     Requests/sec:  20,968.15
My branch Requests/sec: 169,514.68
```

Results (Ruby 3.4.7)

```
v6.6.1    Requests/sec:  38,989.26
v7.04     Requests/sec:  30,909.26
My branch Requests/sec:  42,379.40
```

Higher is better. Not only does this meet the performance of Puma 6, but it beats it.

close #3788

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -524,8 +524,14 @@ module Puma
             end
 
             if next_request_ready
-              @thread_pool << client
-              close_socket = false
+              # When Puma has spare threads, allow this one to be monopolized
+              # Perf optimization for https://github.com/puma/puma/issues/3788
+              if @thread_pool.waiting > 0
+                can_loop = true
+              else
+                @thread_pool << client
+                close_socket = false
+              end
             elsif @queue_requests
               client.set_timeout @persistent_timeout
               if @reactor.add client


schneems:
The current approach to looking at @waiting threads has two (theoretical) problems. I want to note them, and then I'll explore them:

- Way too many threads: If someone has configured this number to be absurdly high, then it will never be true that there aren't more threads in the system. At this point, we're basically giving each connection it's own thread, but the OS is in charge of load balancing them for fairness
- Thread race condition: A thread can be "waiting" without it being woken. In the situation where there's a single core, if the request does no IO, then another request would be delayed  100ms (quantum). 

Before we do that, here's a baseline I used the same reproduction from #3487 to check for "long tail" responses and this was the result: https://gist.github.com/schneems/c829b55c067e46fdce8d402f7621839c. This shows that the current branch still fixes the "long tail" problem.

To test the first variant (too many threads), I used the same reproduction from #3487; but set the threads on the server to 512 instead of 5. You can see the results here https://gist.github.com/schneems/8537301c5e204dc8c831a1842fa5dbda. It looks like all variants perform roughly equally when serving 200 requests and 2000 requests. As observed previously, the "longtail" behavior really only happens when all threads are busy. When lots of threads are used, the CPU seems to mitigate that specific issue. 

The second "thread race condition" still exists, but I think we're okay. Most "serious" apps will do a little IO to interrupt the quantum, and there's only so much we can do to work around the OS scheduling (or not) a thread to run on the CPU.  I explored an alternate metric that would (in theory) eliminate this, but TLDR; it's better than master, but much slower than the current solution proposed here...

## Alternative metric: Queued requests

A **more correct** metric would be to use the existence of another request in the backlog (internal queue). Basically, if there's another request already, we should assume it came in before our last request finished and process it first, whether or not there's another thread waiting for it.

I wanted to see the performance of this approach:

```diff
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -524,9 +524,10 @@ module Puma
             end

             if next_request_ready
-              # When Puma has spare threads, allow this one to be monopolized
+              # If other requests are waiting in the queue, serve them first. Otherwise,
+              # process the next request on the connection without returning it to the queue.
               # Perf optimization for https://github.com/puma/puma/issues/3788
-              if @thread_pool.waiting > 0
+              if @thread_pool.race_backlog_empty?
                 can_loop = true
               else
                 @thread_pool << client
diff --git a/lib/puma/thread_pool.rb b/lib/puma/thread_pool.rb
index a69bf021..c553a96d 100644
--- a/lib/puma/thread_pool.rb
+++ b/lib/puma/thread_pool.rb
@@ -107,6 +107,14 @@ module Puma
       with_mutex { @todo.size }
     end

+    # Private API
+    #
+    # Returns true if the queue is empty.
+    # Does not lock the mutex so this is subject to race conditions.
+    def race_backlog_empty?
+      @todo.empty?
+    end
```

Results are better than 7.0 but far worse than checking for waiting threads (with jruby):

```
@waiting > 0 Requests/sec: 142,209.16 (old solution)
@todo.empty? Requests/sec:  50,114.41 (new solution)
```

I think that this is because `Queue` is an atomic data structure in JRuby so under the hood it's locking and giving strong guarantees which reintroduce the original problem of adding coordination overhead to processing one request. In this case, I think the data race is fine at the method level. Because even if that's 100% correct, we're not calling the conditional in a mutex, so the interpreter can still data race (in MRI). Versus `@waiting` is a variable we are manually managing, and it has no associated internal mutex.

I also tried manually book-keeping the number of requests in @todo by incrementing and decrementing a count (similar to incrementing/decrementing `@waiting`). It was slow, about the same speed as `@todo.empty?`, so maybe the theory about Queue locks being slow wasn't entirely true. Another theory is that: it's faster to keep processing the same data on the thread, even if another request TECHNICALLY got in first, and if there's another thread ready to serve it ... the outcome is that the overall system performance is better and request distribution doesn't seem to meaningfully suffer.
## Outcome

I think we should move forward with this PR.

schneems:
Leaving this note here for future developers to know that this was intentional: The current `thread_pool.waiting` approach has a race condition, because it's not wrapped in a mutex. A partially correct fix would be:

```diff
diff --git a/lib/puma/server.rb b/lib/puma/server.rb
index 2c60cfb9..08a73ef2 100644
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -526,7 +526,7 @@ module Puma
             if next_request_ready
               # When Puma has spare threads, allow this one to be monopolized
               # Perf optimization for https://github.com/puma/puma/issues/3788
-              if @thread_pool.waiting > 0
+              if @thread_pool.with_mutex { @thread_pool.waiting > 0}
                 can_loop = true
               else
                 @thread_pool << client
```

A fully correct fix would be: 

```diff
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -526,12 +526,14 @@ module Puma
             if next_request_ready
               # When Puma has spare threads, allow this one to be monopolized
               # Perf optimization for https://github.com/puma/puma/issues/3788
-              if @thread_pool.waiting > 0
-                can_loop = true
-              else
-                @thread_pool << client
-                close_socket = false
-              end
+              @thread_pool.with_mutex {
+                if  @thread_pool.waiting > 0
+                  can_loop = true
+                else
+                  @thread_pool << client
+                  close_socket = false
+                end
+              }
```

But the name of the game here is "do less locking." It's not worth locking/blocking other threads from enqueue/dequeue-ing requests just to get a slightly-more-correct value here. In the case where we are one or two numbers off, the worst that will happen is each thread will not realize that the server is busy and will incorrectly serve another request from it's own connection. On the next iteration, it would see that `@waiting` is zero  and would stop looping. So, the "line jumping" behavior still exists, but it's limited to the situation where there's a data race and then would be limited to one request (versus previously there were 10 for every thread on every keepalive connection, forever).

Benchmarking the "full" fix, I get `Requests/sec: 123,064.30` (ish) compared to `Requests/sec: 169,514.68` (current PR with this data race). It's not a deal killer to do the more technically correct thing that number is still faster than the 6.6.1 benchmark, but I honestly think we don't need to (fameous last words). Leaving this comment so when people are digging up this change in 10 years they have enough info to determine if a future change is safe or not.

MSP-Greg:
LGTM.

> It's not worth locking/blocking other threads from enqueue/dequeue-ing requests just to get a slightly-more-correct value here.

Yes.  Or, concurrency issues aren't always binary...

p8:
Thanks @schneems !

jjb:
> Honesty in benchmarking disclaimer.....

relevant benchmark for me and probably many others would be: `7.1` vs. `6 with keepalive disabled`

schneems:
The script I used is posted at the top with instructions in the gist https://gist.github.com/schneems/bc7ee417007d467a63abcf793793b989/b51629f86f6e05e2afc52041875033c6bffa5012

You could make two of them, one that turns off KA and one that doesn’t and make that comparison. (Also to note that the thread count is absurdly high, it’s probably best to reduce it to something closer to what you actually run in prod).

jrochkind:
Very cool!

PR #3795, opened 2025-10-15
client.rb - Skip reading zero bytes when request body is buffered

MSP-Greg:
### Description

Current code in `Client` handles parsing a request's request line and headers (using `HttpParser`), along with reading the content/body if it exists.  Much of the body related code is in `Client#read_body`.  Some code is executed for every call, but the code can be skipped sometimes.

Code adds a conditional to skip the code.

The patch is best viewed with 'Hide whitespace', as several lines only have an indent added.

A larger refactor of this code could move the conditional 'higher', saving that for the future.  It also ties in with processing pipe-lined requests.

---
**Help Puma by sponsoring its maintainers**
* [dentarg](https://github.com/sponsors/dentarg)
* [MSP-Greg](https://github.com/sponsors/MSP-Greg)
* [schneems](https://github.com/sponsors/schneems)

### DIFF
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -502,11 +502,8 @@ module Puma
 
       # don't bother with reading zero bytes
       unless remain.zero?
-
-        want = remain > CHUNK_SIZE ? CHUNK_SIZE : remain
-
         begin
-          chunk = @io.read_nonblock(want, @read_buffer)
+          chunk = @io.read_nonblock(remain.clamp(0, CHUNK_SIZE), @read_buffer)
         rescue IO::WaitReadable
           return false
         rescue SystemCallError, IOError


lib/puma/client.rb
```diff
@@ -500,29 +500,29 @@ def read_body
...
+      unless remain.zero?
 
-      begin
-        chunk = @io.read_nonblock(want, @read_buffer)
-      rescue IO::WaitReadable
-        return false
-      rescue SystemCallError, IOError
-        raise ConnectionError, "Connection error detected during read"
-      end
+        want = remain > CHUNK_SIZE ? CHUNK_SIZE : remain
```
schneems: Optional Suggestion: We can also remove this variable assignment and move this logic inline later:

```ruby
        begin
          chunk = @io.read_nonblock(remain.clamp(0, CHUNK_SIZE), @read_buffer)
```

I feel like that's clearer and cleaner. I'm fine with the code as is, but would prefer `clamp` if you are willing to make that change.

schneems (approved):
Good catch. I like it. I made an optional suggestion, you can merge now as-is or make the change and then merge.

Another optional suggestion: I suggest changing the PR title to be more from the perspective of an end user and less from the perspective of the implementation internals. (I'm refraining from editing it inline to make sure you agree/approve.) Maybe something like

> Skip reading zero bytes when request body is buffered
or
> Optimize request body reading to avoid IO when body is fully buffered

lib/puma/client.rb
MSP-Greg: I can do that.

Issue #3675, opened 2025-07-08
Systemd unnecessarily restarts Puma after succesful pumactl restart


danielhoey:
After restarting puma using pumactl, systemd will sometimes restarts the puma master process due to the watchdog timeout. This is solvable by increasing the timeout but would be unnecessary if puma sent a WATCHDOG=1 notification immediately once ready.

I modified sd_notify.rb to log when the watchdog notifications are sent. Puma is configured in systemd with a 10s watchdog. The following log output demonstrates the issue:

**puma stdout:**
```
13:32:55 - SdNotify.watchdog
[87684] * Restarting...
[87684] * Enabling systemd notification integration
[87684] Puma starting in cluster mode...
[87684] * Puma version: 6.6.0 ("Return to Forever")
[87684] * Ruby version: ruby 3.3.7 (2025-01-15 revision be31f993d7) [aarch64-linux]
[87684] *  Min threads: 5
[87684] *  Max threads: 5
[87684] *  Environment: staging
[87684] *   Master PID: 87684
[87684] *      Workers: 3
[87684] *     Restarts: (✔) hot (✖) phased
[87684] * Preloading application
[87684] * Inherited unix:///var/www/onepoint/shared/sockets/puma.sock
[87684] Use Ctrl-C to stop
=== puma startup: 2025-07-08 13:33:01 +1000 ===
[87684] Pinging systemd watchdog every 5.0 sec
[87684] Sending status to systemd every 1.0 sec
[87684] * Starting control server on unix:///tmp/puma-status-1751945578344-87684
[87684] - Worker 0 (PID: 87853) booted in 0.02s, phase: 0
[87684] - Worker 1 (PID: 87867) booted in 0.01s, phase: 0
[87684] - Worker 2 (PID: 87881) booted in 0.0s, phase: 0
=== puma startup: 2025-07-08 13:33:09 +1000 ===
[87900] Pinging systemd watchdog every 5.0 sec
[87900] Sending status to systemd every 1.0 sec
[87900] * Starting control server on unix:///tmp/puma-status-1751945586968-87900
[87900] - Worker 0 (PID: 87924) booted in 0.02s, phase: 0
[87900] - Worker 1 (PID: 87938) booted in 0.01s, phase: 0
[87900] - Worker 2 (PID: 87952) booted in 0.0s, phase: 0
```

**journctl output:**
```
Jul 08 13:33:05 ip-172-31-29-230 systemd[1]: puma.service: Watchdog timeout (limit 10s)!
[...]

PR #3799, opened 2025-10-17
Send watchdog to systemd when puma is ready

danielhoey:
### Description
Fixes https://github.com/puma/puma/issues/3675

Previously puma when puma was ready it would not send a watchdog notification until the watchdog period was complete. Now we send an watchdog notification immediately as soon as we are ready. This avoids systemd unnecessarily restarting puma after a restart has just completed.

I needed to add a wait option to assert_message in test/test_plugin_systemd.rb to ensure the test doesn't wait and receive the later (periodic) watchdog notification causing test success when it should fail.

### DIFF
--- a/lib/puma/sd_notify.rb
+++ b/lib/puma/sd_notify.rb
@@ -51,6 +51,7 @@ module Puma
 
     def self.ready(unset_env=false)
       notify(READY, unset_env)
+      watchdog(unset_env) if watchdog?
     end
 
     def self.reloading(unset_env=false)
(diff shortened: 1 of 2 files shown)


MSP-Greg:
> This avoids systemd unnecessarily restarting puma after a restart has just completed.

I have a few other things to look at, but this makes sense.  Last time I was in this code I recall some odd things happening.

Somewhat off topic, when you're fixing a bug, two commits is helpful.  The first with the added/changed tests, the second with the 'lib/ext' code bug fix.  One can checkout the test commit and see the tests run on 'master'.  GitHub can squash commits when merging.  I need to add something about this to CONTRIBUTING.md...

danielhoey:
OK thanks, it's my first open source bug fix for a long while. I hadn't considered splitting tests and src but I can see that it makes sense.

schneems:

```
```

PR #3803, opened 2025-10-21
Refactor request.rb

schneems:
### Description

I'm normally very against refactors like this, but I've found this code fairly difficult to navigate. One of the problems (addressed in this PR) is isolation/separation of bounds isn't always clear. The code in `request.rb` calls methods on `server.rb` and vise-versa. Every method in `request.rb` can access every internal of `server.rb` and this has lead to logic in there that really belongs somewhere else. To improve clarity and separation of concerns I made these changes:

- All pure functions (that do not need instance variables) in `request.rb` are now module methods (10/12 methods)
- The ones left `prepare_response` and `handle_response` are made into a class, with only the minimum number of instance variable information passed to it at instantiation time.
- Connection handling helpers such as dealing with socket "corking" moved from `server.rb` to `request.rb`

There's a lot of commits, but they're mostly mechanical (to start with). Towards the end, I had to work out exactly how I wanted to structure the end code. And, I took performance into account. It would be nice if we made things faster, but the goal is to be able to isolate this code without a regression. You can see some benchmarks https://gist.github.com/schneems/f5471b42ac9d6e3ba38eb4d893befe29 with my script from before https://gist.github.com/schneems/bc7ee417007d467a63abcf793793b989/b51629f86f6e05e2afc52041875033c6bffa5012. It looks like it's about the same as `main`.

This also opens the door to future unit testing (an exercise left to future us).

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -82,9 +82,6 @@ module Puma
       @thread = nil
       @thread_pool = nil
       @reactor = nil
-
-      @env_set_http_version = nil
-
       @options = if options.is_a?(UserFileDefaultOptions)
         options
       else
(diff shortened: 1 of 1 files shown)


schneems:
This PR touches on some related changes, I need to review https://github.com/puma/puma/pull/3582

PR #3807, opened 2025-10-23
Update deployment and Kubernetes docs with Puma configuration tips [ci skip]

nateberkopec:
Closes/addresses https://github.com/puma/puma/discussions/3805


docs/deployment.md
```diff
@@ -50,14 +48,34 @@ a time. But since so many apps are waiting on IO from DBs, etc., they can
...
+* Unless you have a very I/O-heavy application (50%+ time spent waiting on IO), use the default thread count (5 for MRI). Using higher numbers of threads with low I/O wait (<50%) will lead to additional request queueing time (latency!) and additional memory usage.
+* More processes per pod reduces memory usage per process, because of copy-on-write memory and because the cost of the single master process is "amortized" over more child processes.
+* Don't run less than 4 processes per pod if you can. Low numbers of processes per pod will lead to high request queueing, which means you will have to run more pods.
+* If multithreaded, allocate 1 CPU core (see definition above!) per worker. If single threaded, allocate 0.75 cpus per worker. Most web applications spend about 25% of their time in I/O - but when you're running multi-threaded, your Puma process will have higher CPU usage and should be able to fully saturate a CPU core.
```
bensheldon: Could this be broken out more. I think there are two stories here:

- You're in a memory-constrained environment, in which case you want 1-1 CPU/worker and the work is tuning threads until you either max out your latency budget or max out memory. (probably in a PaaS where you have a limited number of container sizes/tiers to choose from)
- You're in a CPU-constrained environment, in which case you want to be single-threaded and the work is tuning worker count until you start hitting CPU utilization limits.

...and I think much of the advice here is only relevant to the first scenario.

I don't want to relitigate https://github.com/rails/rails/issues/50450#issuecomment-1880042127 but I think this section will be taken as some kind of natural law and not just as like "simple heuristics for common situations".

jrochkind:
> Set the number of threads to desired concurrent requests/number of workers

took me a minute to realize that / meant "divided by" and not "slash", and understand the sentence.

docs/deployment.md
jrochkind: I am not sure what "If multithreaded" means here -- does this refer to a puma setting, or to characteristic of the app, or both?

nateberkopec: Thanks for the feedback ❤️ I don't agree 100% @bensheldon but I see what you're getting at, will address.

PR #3811, opened 2025-10-29
Pass in server object before_thread_start event

yuki24:
### Description

I have the other pull request https://github.com/puma/puma/pull/3658 in order to add a way to change the # of threads at runtime. As suggested by https://github.com/puma/puma/pull/3658#discussion_r2436818878 (and after looking at https://github.com/puma/puma/pull/3774), the extra `Thread.current.puma_server = server` assignment looked unnecessary., so I removed a few lines that I thought was duplicate. This actually broke the build in one of the tests in our private repo - we basically lost access to the server object and now have to think about how we can grab it.

This PR updates the `before_thread_start` event call to take the `server` object. This way we will be able to grab the server object with:

```ruby
Puma::Plugin.create do
  def start(launcher)
    ...

    launcher.config.configure do |user_dsl, *|
      user_dsl.before_thread_start do |server|
        do_stuff_with_server(server)
      end
    end
  end
end
```

While there are still many questions such as:

1.  Why did we add a way to pass around the server object in the first place?
2. Why does it make sense that only the `before_thread_start` event hook takes an argument? (other events don't take arguments at all.)
3. Are there any other use cases than changing the thread number?

I'm not actually sure there are other ways to get access to the `server` object. I'd be open to suggestions as to what makes more sense here.

### DIFF
--- a/lib/puma/thread_pool.rb
+++ b/lib/puma/thread_pool.rb
@@ -198,7 +198,7 @@ module Puma
 
       @before_thread_start.each do |b|
         begin
-          b[:block].call
+          b[:block].call(@server)
         rescue Exception => e
           STDERR.puts "WARNING before_thread_start hook failed with exception (#{e.class}) #{e.message}"
         end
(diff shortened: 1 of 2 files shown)


nateberkopec:
Yuki and I discussed this offline, we're going another direction.

PR #3814, opened 2025-11-16
Various cleanups in the C extension

byroot:
I heard @nateberkopec say the C extension needed some care so I made a quick pass. Honestly there isn't much to do here.

I did a quick cleanup and modernization pass, I made independent commits to make it easier to review.


nateberkopec:
There Goes My Hero Foo Fighters

nateberkopec:
this is what we do when we see a byroot pr

PR #3815, opened 2025-11-16
[Experiment] Replace C parser with protocol-http1

swebb:
### Description

Addresses https://github.com/puma/puma/issues/1889.
Previous attempt: https://github.com/puma/puma/pull/3660.

A quick spike to replace the Ragel C parser with the protocol-http1 gem.

Running the hello world benchmark I get a 0.76% slowdown which feels pretty good considering I hacked this together (see below for details).

Implementation wise this is very rough. When running the test suite I get:

> 717 runs, 1958 assertions, 105 failures, 5 errors, 44 skips

Also, my understanding of protocol-http is that it makes blocking reads to the socket while parsing the HTTP headers. To avoid blocking I wait until all of the headers are available on the socket before I start using protocol-http to parse the request. This is obviously not optimal but was a quick way to get going.

### Benchmarks

On my x86 laptop running Linux I get:

```
====================================
C parser
====================================
  8 threads and 16 connections
  888170 requests in 30.10s, 64.39MB read
Requests/sec:  29506.52
Transfer/sec:      2.14MB

Thread Stats         Avg       Stdev         Min         Max    +/- Stdev
     Latency:     545.52us    149.59us    294.00us      2.88ms   92.73%
     Req/sec:       3.71k     254.35       10.00        4.11k    93.77%

Latency Distribution
     50%  521.00us
     75%  576.00us
     90%  645.00us
     99%    1.28ms

Request Stats        Avg       Stdev         Min         Max    +/- Stdev
    Req/conn:      55.51k       3.32       55.50k      55.52k    56.25%
====================================
Ruby parser
====================================
  8 threads and 16 connections
  677723 requests in 30.10s, 49.13MB read
Requests/sec:  22515.15
Transfer/sec:      1.63MB

Thread Stats         Avg       Stdev         Min         Max    +/- Stdev
     Latency:     709.26us    191.30us    391.00us      3.66ms   90.47%
     Req/sec:       2.83k     174.91       10.00        3.14k    87.19%

Latency Distribution
     50%  665.00us
     75%  736.00us
     90%    0.85ms
     99%    1.42ms

Request Stats        Avg       Stdev         Min         Max    +/- Stdev
    Req/conn:      42.36k       3.26       42.35k      42.36k    56.25%
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,6 +11,7 @@ gem "minitest-retry"
 gem "minitest-proveit"
 gem "minitest-stub-const"
 gem "concurrent-ruby", "~> 1.3"
+gem "protocol-http1"
 
 case ENV['PUMA_CI_RACK']&.strip
 when 'rack2'
--- a/Rakefile
+++ b/Rakefile
@@ -39,7 +39,23 @@ file 'ext/puma_http11/org/jruby/puma/Http11Parser.java' => ['ext/puma_http11/htt
 end
 task :ragel => ['ext/puma_http11/org/jruby/puma/Http11Parser.java']
 
-if !Puma.jruby?
+if ENV["PUMA_PURE_RUBY_PARSER"]
+  task :compile => ['ext/puma_http11/puma_http11.rb'] do
+    src = 'ext/puma_http11/puma_http11.rb'
+    dest = 'lib/puma/puma_http11.rb'
+    if File.exist?(dest)
+      if File.symlink?(dest)
+        puts "Using ruby parser (already symlinked)"
+      else
+        puts "Refusing to symlink ruby parser. Did you forget to run `rake clean`?"
+      end
+    else
+      puts "Symlinking ruby parser"
+      FileUtils.symlink(src, dest, relative: true)
+    end
+  end
+  CLEAN.include "lib/puma/puma_http11.rb"
+elsif !Puma.jruby?
   # compile extensions using rake-compiler
   # C (MRI, Rubinius)
   Rake::ExtensionTask.new("puma_http11", gemspec) do |ext|
--- a/benchmarks/wrk/hello.sh
+++ b/benchmarks/wrk/hello.sh
@@ -1,8 +1,32 @@
 # You are encouraged to use @ioquatix's wrk fork, located here: https://github.com/ioquatix/wrk
 
-bundle exec bin/puma -t 4 test/rackup/hello.ru &
-PID1=$!
-sleep 5
-wrk -c 4 -d 30 --latency http://localhost:9292
+benchmark() {
+  bundle exec bin/puma -t 4 test/rackup/hello.ru &
+  PID1=$!
+  sleep 5
+  echo "Warming up YJIT"
+  wrk -c 4 -d 30 --latency http://localhost:9292 > /dev/null
+  echo "Running"
+  wrk -c 16 -t 8 -d 30 --latency http://localhost:9292
+  kill $PID1
+}
 
-kill $PID1
+export RUBY_YJIT_ENABLE=1
+
+echo "===================================="
+echo "C parser"
+echo "===================================="
+
+unset PUMA_PURE_RUBY_PARSER
+rake clean
+rake compile
+benchmark
+
+echo "===================================="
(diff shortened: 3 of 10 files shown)


ioquatix:
I really like Ragel and we may eventually have a native implementation for `protocol-http1` and related gems. I'm not sure if parsing HTTP is a huge cost in all but the smallest micro-benchmarks, but would be interesting to see if we can close the gap using JIT. Nice work.

nateberkopec:
Beyond cool. Thanks again for trying this out.

@ioquatix do you mean a ragel implemenation of `protocol-http1`?

>my understanding of protocol-http is that it makes blocking reads to the socket while parsing the HTTP headers

That's an interesting constraint. Do we think this is a problem? What could we do instead?

> 0.76% slowdown

I think you mean 24% slower, right?

@swebb just from your own experience with two implementations now, what do you think about the path forward here?

ioquatix:
> Do you mean a ragel implementation of protocol-http1?

Yes.

nateberkopec:
That would considerably decrease the risk for Puma adopting it, I should think...

ioquatix:
I have no opinion about Puma adopting `protocol-http1`. But bear in mind since I have not released v1.x I may still make breaking changes. That being said, we are probably close to a v1 release.

swebb:
> > my understanding of protocol-http is that it makes blocking reads to the socket while parsing the HTTP headers
> 
> That's an interesting constraint. Do we think this is a problem? What could we do instead?

It's definitely an issue that requires attention. My (limited) understanding of Puma is that the http header passing is done before the request is buffered by the reactor. I'm not sure what terminology you use for the thread that manages that ("parent" thread maybe?). It's critical that the parent thread doesn't do any blocking because a slow client would prevent any other requests from being accepted.

Possible options:

1. I think my current approach of doing an early return if not all of the headers have been received is a reasonable solution for the time being (although not particularly elegant). However, I would switch to using a StringScanner to progressively search for the empty line instead of my current brute force approach of searching the entire thing for a newline.

2. Maybe we can move the header passing to after the reactor? I have no idea if that's possible. It a guess I'd assume there's something about the headers that the reactor uses and that's why they are passed first, but I don't know.

3. Maybe we could modify protocol-http or use it's api differently to avoid blocking requests.

> > 0.76% slowdown
> 
> I think you mean 24% slower, right?

Whoops. Yes.

> @swebb just from your own experience with two implementations now, what do you think about the path forward here?

Two things come to mind.

Firstly, it would be great to have an idea about how you would roll out a new parser. In a simple world I would replace the existing one and do a major version bump. In reality I suspect that's too risky and it would be better to make the parser configurable (as I have in this PR). It could be rolled out as an experimental configuration option that users could test out for us and provide feedback. The difficulty with a configurable parser approach is that the test suite will probably need considerable refactoring. Not only to support multiple parsers, but also some of the tests I believe will only apply to one parser. The nice thing about refactoring the tests is it would allow easy experimentation with lots of different parsers.

Secondly, @evanphx made [this excellent comment about a pure ruby parser on the issue](https://github.com/puma/puma/issues/1889#issuecomment-2629190924) :

> I'd say the only reason to revisit it is if there are real, tangible benefits (better performance, better platform support, something). Put another way, the only way we'd revisit this is if someone does the work and then we can evaluate it's working performance. If someone wants to go down that path, I'm all for considering it at that point.

We've now got a spiked implementation we can start to evaluate and I've love to have some idea about how much interest there is in replacing the parser. Performance wise I agree with @ioquatix that http parsing probably isn't a huge cost outside of micro benchmarks. With some polishing it *might* be possible to make this parser faster than the Ragel one (no promises) but even then I'm not sure it would justify the change.

As an outsider I feel like reducing the maintenance burden to support MS Windows, PRs using C, and keeping JRuby in-sync is probably the main benefit. I'm keen to hear how much of a burden they are and if it's worth investigating this further.

As an aside I love deleting code and making things simpler. Combining this PR with [this spike PR to replace MiniSSL with OpenSSL](https://github.com/puma/puma/pull/3761) I can see a possible future where Puma doesn't have any C or Java extensions and I like the idea of that. Whether it's practical is yet to be determined.

PR #3816, opened 2025-11-16
Implement `env["puma.mark_as_io_bound"]`

### LINKED ISSUE #3777: Feature: Add API to "remove" current thread from the pool

A common problem in applications is how to handle uneven mix of IO-heavy and CPU-heavy requests.
It's not rare for application that are mostly transactional to have a few "streaming", or "proxy" endpoints, that are massively IO-bound. One use case that is built into Rails is [`ActiveStorage::Streaming`](https://github.com/rails/rails/blob/07db688eda60ce0e57432f2299be0abf75da953f/actionpack/lib/action_controller/metal/streaming.rb)

To try to answer that need, over a decade ago, Rails introduced [`ActionController::Live`](https://api.rubyonrails.org/classes/ActionController/Live.html) but it's riddled with problems as it is essentially migrating the request processing onto a background thread, and run into all sorts of problems with thread local state etc.

It would be much simpler and cleaner if the application was able to signal back to Puma that this thread will be mostly IO going forward, and that as such it's alright to allow the pool to forget about it.

I think a callable in the `env` hash could work. e.g.:

```ruby
env["puma.release_thread"].call
```

Other possible terminologies: `defer_thread`

**Additional context**

I think this is a problem best solved at the web server level, because ultimately the AC::Live use case make sense, but its implementation is broken beyond repair.

I also think it needs to be a sever specific API, because if you are running a single threaded server like Unicorn or Pitchfork, you don't want to be doing any of this, and if you are running a server optimized for IO-bound workloads like Falcon, it's basically already not a concern.

byroot:
Fix: https://github.com/puma/puma/issues/3777

Workers can now be marked as "IO bound". This state last until the end of the current request.

At the end of the request, the pool decide whether the "IO bound" thread rejoin the normal population, or is abandonned.

Generally speaking, an "IO bound" thread no longer counts towards the `max_threads` configuration. However, there is a `max_io_threads` configuration, and any "IO bound" thread above that limit automatically counts as a "normal" thread.

For example, imagine a config of `max_threads: 5, max_io_threads: 10`:

If there is 3 normal threads and 12 io threads, then puma won't spawn any new thread, but the next deferred thread that complete will become "normal" again so that we now have 4 normal threads and 11 deferred threads.

Now if there is 5 normal threads and 2 IO threads, then when the next IO threads complete, it will exit.

### DIFF
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -595,6 +595,29 @@ module Puma
       @options[:max_threads] = max
     end
 
+    # Configure the max number of IO threads.
+    #
+    # When request handlers know the current requests will no longer use a significant amount
+    # of CPU, they can mark the current request as IO bound using <tt>env["puma.mark_as_io_bound"]</tt>.
+    #
+    # Threads marked as IO bound are allowed to go over the max thread limit.
+    #
+    # @example
+    #   threads 5
+    #   max_io_threads 5
+    #
+    # The above example allows for 5 regular threads and 5 IO threads to process requests concurrently.
+    # Any IO thread over the limit is counted as a regular thread, hence the above configuration also
+    # allows for 3 regular threads and 7 IO threads for example.
+    def max_io_threads(max)
+      max = Integer(max)
+      if max < 0
+        raise "The maximum number of IO threads (#{max}) must be a positive number"
+      end
+
+      @options[:max_io_threads] = max
+    end
+
     # Instead of using +bind+ and manually constructing a URI like:
     #
     #    bind 'ssl://127.0.0.1:9292?key=key_path&cert=cert_path'
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -31,7 +31,7 @@ module Puma
   # Each `Puma::Server` will have one reactor and one thread pool.
   class Server
     module FiberPerRequest
-      def handle_request(client, requests)
+      def handle_request(processor, client, requests)
         Fiber.new do
           super
         end.resume
@@ -259,7 +259,9 @@ module Puma
 
       @status = :run
 
-      @thread_pool = ThreadPool.new(thread_name, options, server: self) { |client| process_client client }
+      @thread_pool = ThreadPool.new(thread_name, options, server: self) do |processor, client|
+        process_client(processor, client)
+      end
 
       if @queue_requests
         @reactor = Reactor.new(@io_selector_backend) { |c|
@@ -304,7 +306,7 @@ module Puma
     # The method checks to see if it has the full header and body with
     # the `Puma::Client#try_to_finish` method. If the full request has been sent,
     # then the request is passed to the ThreadPool (`@thread_pool << client`)
-    # so that a "worker thread" can pick up the request and begin to execute application logic.
+    # so that a "processor thread" can pick up the request and begin to execute application logic.
     # The Client is then removed from the reactor (return `true`).
     #
     # If a client object times out, a 408 response is written, its connection is closed,
@@ -473,14 +475,14 @@ module Puma
     # Given a connection on +client+, handle the incoming requests,
     # or queue the connection in the Reactor if no request is available.
     #
-    # This method is called from a ThreadPool worker thread.
+    # This method is called from a ThreadPool processor thread.
     #
     # This method supports HTTP Keep-Alive so it may, depending on if the client
     # indicates that it supports keep alive, wait for another request before
     # returning.
     #
     # Return true if one or more requests were processed.
-    def process_client(client)
+    def process_client(processor, client)
       close_socket = true
 
       requests = 0
@@ -503,7 +505,7 @@ module Puma
         while can_loop
           can_loop = false
           @requests_count += 1
-          case handle_request(client, requests + 1)
+          case handle_request(processor, client, requests + 1)
           when :close
           when :async
             close_socket = false
@@ -671,7 +673,7 @@ module Puma
     end
     private :notify_safely
 
-    # Stops the acceptor thread and then causes the worker threads to finish
+    # Stops the acceptor thread and then causes the processor threads to finish
     # off the request queue before finally exiting.
 
     def stop(sync=false)
--- a/lib/puma/thread_pool.rb
+++ b/lib/puma/thread_pool.rb
@@ -25,6 +25,51 @@ module Puma
     class ForceShutdown < RuntimeError
     end
 
+    class ProcessorThread
+      attr_accessor :thread
+      attr_writer :marked_as_io_thread
+
+      def initialize(pool)
+        @pool = pool
+        @thread = nil
+        @marked_as_io_thread = false
+      end
+
+      def mark_as_io_thread!
+        unless @marked_as_io_thread
+          @marked_as_io_thread = true
+
+          # Immediately signal the pool that it can spawn a new thread
+          # if there's some work in the queue.
+          @pool.spawn_thread_if_needed
+        end
+      end
+
+      def marked_as_io_thread?
+        @marked_as_io_thread
+      end
+
+      def alive?
+        @thread&.alive?
+      end
+
+      def join(...)
+        @thread.join(...)
+      end
+
+      def kill(...)
+        @thread.kill(...)
+      end
+
+      def [](key)
+        @thread[key]
+      end
+
+      def raise(...)
+        @thread.raise(...)
+      end
+    end
+
     # How long, after raising the ForceShutdown of a thread during
     # forced shutdown mode, to wait for the thread to try and finish
     # up its work before leaving the thread to die on the vine.
@@ -53,6 +98,8 @@ module Puma
       @name = name
       @min = Integer(options[:min_threads])
       @max = Integer(options[:max_threads])
+      @max_io_threads = Integer(options[:max_io_threads] || 0)
+
       # Not an 'exposed' option, options[:pool_shutdown_grace_time] is used in CI
       # to shorten @shutdown_grace_time from SHUTDOWN_GRACE_TIME. Parallel CI
       # makes stubbing constants difficult.
@@ -71,7 +118,7 @@ module Puma
       @trim_requested = 0
       @out_of_band_pending = false
 
-      @workers = []
+      @processors = []
 
       @auto_trim = nil
       @reaper = nil
@@ -100,6 +147,7 @@ module Puma
           running: @spawned,
           pool_capacity: pool_capacity,
           busy_threads: @spawned - @waiting + @todo.size,
+          io_threads: @processors.count(&:marked_as_io_thread?),
           backlog_max: temp
         }
       end
@@ -140,7 +188,8 @@ module Puma
       @spawned += 1
 
       trigger_before_thread_start_hooks
-      th = Thread.new(@spawned) do |spawned|
+      processor = ProcessorThread.new(self)
+      processor.thread = Thread.new(processor, @spawned) do |processor, spawned|
         Puma.set_thread_name '%s tp %03i' % [@name, spawned]
         # Advertise server into the thread
         Thread.current.puma_server = @server
@@ -155,11 +204,23 @@ module Puma
           work = nil
 
           mutex.synchronize do
+            if processor.marked_as_io_thread?
+              if @processors.count { |t| !t.marked_as_io_thread? } < @max
+                # We're not at max processor threads, so the io thread can rejoin the normal population.
+                processor.marked_as_io_thread = false
+              else
+                # We're already at max threads, so we exit the extra io thread.
+                @processors.delete(processor)
+                trigger_before_thread_exit_hooks
+                Thread.exit
+              end
+            end
+
             while todo.empty?
               if @trim_requested > 0
                 @trim_requested -= 1
                 @spawned -= 1
-                @workers.delete th
+                @processors.delete(processor)
                 not_full.signal
                 trigger_before_thread_exit_hooks
                 Thread.exit
@@ -181,16 +242,16 @@ module Puma
           end
 
           begin
-            @out_of_band_pending = true if block.call(work)
+            @out_of_band_pending = true if block.call(processor, work)
           rescue Exception => e
             STDERR.puts "Error reached top of thread-pool: #{e.message} (#{e.class})"
           end
         end
       end
 
-      @workers << th
+      @processors << processor
 
-      th
+      processor
     end
 
     private :spawn_thread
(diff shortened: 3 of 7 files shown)


lib/puma/thread_pool.rb
```diff
@@ -24,6 +24,44 @@ class ThreadPool
     class ForceShutdown < RuntimeError
     end
 
+    class Worker
```
nateberkopec: Nit: we already call processes `workers` in too many contexts, I find the word overloaded. we're using it to refer to a thread execution context here. Doesn't Sidekiq call these `Processor`? Just anything other than Worker.

nateberkopec (commented):
I feel like I'm still chewing on the interface.

```ruby
env["puma.deferred_request"].call
```

Right interface or not?

I'm mostly concerned about what happens when someone screws this up and puts CPU-bound work in to the deferred pool, or if that work unexpectedly becomes CPU bound. In what ways are they worse off than they were before? How can we correct from this condition, if at all? 

This feature trades a risk of increased latency for more throughput. Is that a good tradeoff? Should we just say "f off, put these workloads on different machines and don't try to mix them?" Is the project fundamentally doomed?

byroot:
> Right interface or not?

Looks good to me. I'm just not 100% certain about the terminology?

> I'm mostly concerned about what happens when someone screws this up 

Is this really a concern? I think we're all consenting adults here. People already routinely crank up `max_thread` to unreasonable levels, I don't think this is any worse or more dangerous, if anything it's less dangerous.

lib/puma/thread_pool.rb
byroot: Good point, I think it shows I'm not that familiar with Puma 😅 . I'll try to come up with another name.

jclusso:
> People already routinely crank up `max_thread` to unreasonable levels, I don't think this is any worse or more dangerous, if anything it's less dangerous.

I can say I have done this while looking for a feature like this because we have a huge amount of requests that are waiting due to an API design that was optimized for the user at the cost of server resources.

> Right interface or not?

If I understand this correctly, in a Rails controller if you want to defer the request you'd call `env["puma.deferred_request"].call`? Not the prettiest code, but I assume there is no way to add something nicer.

byroot:
> if you want to defer the request you'd call `env["puma.deferred_request"].call`?

Well `request.env["puma.deferred_request"].call` but yes.

> Not the prettiest code

That's usually how rack interfaces work. I do plan to integrate it better with Rails though.

jclusso:
> That's usually how rack interfaces work. I do plan to integrate it better with Rails though.

I figured as much was likely the case that it would be on frameworks to make it work nicely.

lib/puma/thread_pool.rb
dentarg: Well, `@workers = []` was there before this PR, populated with `Thread`s, so I can see how it was natural with `Worker`. :) Is `WorkerThread` a acceptable name?

dentarg: but maybe that should have been `@threads = []`. Another suggestion: `DeferrableThread`

dentarg:
This looks pretty neat/simple to me.

dentarg:
> This feature trades a risk of increased latency for more throughput. Is that a good tradeoff? 

It think that's exactly what's being asked for? https://github.com/puma/puma/issues/3777#issuecomment-3337234298, https://github.com/puma/puma/issues/3777#issuecomment-3448936586

> Should we just say "f off, put these workloads on different machines and don't try to mix them?"

I think providing this could be seen as along the lines of this earlier reply of yours @nateberkopec https://github.com/puma/puma/discussions/3034#discussioncomment-4478480 
> there are several features we maintain like phased-restart which are intended to support a specific use-case of someone who wants to start a Puma server and just wants Puma to provide everything all by itself, without need for additional programs or tools

> Is the project fundamentally doomed?

I don't think so if people are getting value from `ActionController::Live` already?

I can see how this can make it harder to reason about Puma, number of threads and capacity. "give me some extra threads, but please don't count them" :). I think I see this as something frameworks can use instead of `Thread.new` more or less?

lib/puma/request.rb
```diff
@@ -97,6 +97,8 @@ def handle_request(client, requests)
       env[RACK_AFTER_REPLY] ||= []
       env[RACK_RESPONSE_FINISHED] ||= []
 
+      env["puma.deferred_request"] = -> { worker.defer! }
```
nateberkopec: Defer is a weird verb here because we're not _deferring_ work to _later_. When someone calls this, they've either just performed or are about to perform an I/O call which is about to make this thread idle for an extended period of time. The name should reflect that.

Suggestions:

```
puma.sleepy_request
puma.park
puma.offload
puma.mark_io_bound
puma.move_to_io_pool
puma.mark_idle
```

lib/puma/thread_pool.rb
nateberkopec: If I'm reading the PR right, all threads in the threadpool are getting a Worker, so `deferrablethread` doesn't make sense to me because _all_ threads are deferrable. This class also represents a kind of an abstract thread, something which lightly wraps the idea of a thread and context but isn't a Thread itself.

I think I just prefer Processor. Mike's really good at naming and there's a reason why he picked that for the ~similar concept in Sidekiq. Will accept other suggestions though.

nateberkopec (dismissed):
I'll remove my blocks on whether or not we should do this at all, but I do think the class and the env key need better names.

lib/puma/request.rb
byroot: Yes, I don't like `defer` either. It somewhat came to me because that's the term often used to move thing out of an event loop, but it's confusing because you usually do that for the opposite reason (CPU heavy work).

`puma.mark_io_bound` is probably the one I prefer in the list.

Other possibilities:

```
puma.release_thread
puma.unbounded_thread
```

dentarg: I had the same feeling about `defer`. I like something with "mark" in it, because that's almost all we do, mark the thread to be treated a little bit differently.

dentarg:
If Rails would take advantage how this, how would Rails detect that it's supported by the running web server?

byroot:
> how would Rails detect that it's supported by the running web server?

Like other optional rack features. Just check if the key is present in the `env` hash.

lib/puma/request.rb
byroot: We do just mark, but that's an implementation detail.

In my mind, what we are doing here is **freeing** a thread slot in the main pool.

So I kinda like `release_thread` better.

But either way, I think it's really going to be about documenting the behavior and expected use case.

byroot:
I started some renaming and added a couple unit tests. I'm traveling today, so not sure how much more I'll be able to do, but it should already look a bit better.

lib/puma/dsl.rb
```diff
@@ -595,6 +595,29 @@ def threads(min, max = min)
...
+    # When request handlers know the current requests will no longer use a significant amount
+    # of CPU, they can mark the current request as IO bound using <tt>env["puma.mark_as_io_bound"]</tt>.
+    #
+    # Threads marked as IO bound are allowed to go over the max thread limit.
+    #
+    # @example
+    #   threads 5
+    #   max_io_threads 5
+    #
+    # The above example allow for 5 regular threads and 5 IO threads to process requests concurrently.
```
joshuay03:
```suggestion
    # The above example allows for 5 regular threads and 5 IO threads to process requests concurrently.
```

lib/puma/request.rb
```diff
@@ -49,7 +49,7 @@ module Request # :nodoc:
     # @param client [Puma::Client]
     # @param requests [Integer]
     # @return [:close, :keep_alive, :async]
-    def handle_request(client, requests)
+    def handle_request(worker, client, requests)
```
joshuay03: 1. Based on the naming discussions that have taken place, should this be renamed to processor instead?

    ```suggestion
        def handle_request(processor, client, requests)
    ```
2. Could you add a matching `@param` doc above

lib/puma/server.rb
joshuay03, line 1: Same question for all the changes in this file (worker vs processor).

lib/puma/thread_pool.rb
```diff
@@ -153,11 +201,23 @@ def spawn_thread
           work = nil
 
           mutex.synchronize do
+            if processor.marked_as_io_thread?
+              if @workers.count { |w| !w.marked_as_io_thread? } < @max
+                # We're not at max workers, so the deferred worker can rejoin the normal population.
```
joshuay03:
```suggestion
                # We're not at max threads, so the io thread can rejoin the normal population.
```

lib/puma/thread_pool.rb
```diff
@@ -153,11 +201,23 @@ def spawn_thread
...
+                processor.marked_as_io_thread = false
+              else
+                # We're already at max workers, so we exit the extra io thread.
```
joshuay03:
```suggestion
                # We're already at max threads, so we exit the extra io thread.
```

joshuay03 (commented):
Very keen for this!

Just some feedback on naming based on previous reviews.

We really should rename `@workers` in `ThreadPool` to `@processors` in a follow-up, it's confusing even in the current state.

byroot:
> We really should rename `@workers` in `ThreadPool` to `@processors` in a follow-up, it's confusing even in the current state.

Agreed. But I wanted to minimize diff size for review purposes. If you are interested I can do a PR just for that, and then rebase this one on top.

byroot:
@joshuay03 I addressed your feedback.

joshuay03:
> Agreed. But I wanted to minimize diff size for review purposes. If you are interested I can do a PR just for that, and then rebase this one on top.

Yeah fair, up to you whether you want to PR and rebase or wait and follow up.

I'll leave the final thumbs up on this to @nateberkopec

byroot:
Any news?

adenta:
So if I'm reading this right this would basically fix action controller live? I love the API but it's basically unusable in its current state

dentarg:
Can and should the number of io threads be included in [the stats](https://github.com/puma/puma/blob/731b97d2a5c7838c9d736462e259e629655b00a1/docs/stats.md)?

https://github.com/puma/puma/blob/731b97d2a5c7838c9d736462e259e629655b00a1/lib/puma/thread_pool.rb#L91-L104

byroot:
@dentarg I didn't know about this, but yes it could easily be included. The question as always being how to present it to the user.

byroot:
@dentarg @nateberkopec anything I can do here 🙏 ?

dentarg:
@byroot update the stats?

dentarg:
I just randomly came here because I wanted to connect this with the datastar-ruby gem that I happened to be looking and, and noticed this code/comment https://github.com/starfederation/datastar-ruby/blob/226deb42fda9d0ac0d2cf99e9d5438718420b786/lib/datastar/dispatcher.rb#L327-L332

```
        # Now launch the control thread that actually writes to the socket
        # We don't want to block the main thread, so that servers like Puma
        # which have a limited thread pool can keep serving other requests
        # Other streamers will push any StandardError exceptions to the queue
        # So we handle them here
```

Sounds like that gem could potentially make use of this `mark_as_io_bound` feature too?

byroot:
> update the stats?

Right sorry, I guess I was asking if you had opinions on how to display it or if I should do what I think is best?

byroot:
> Sounds like that gem could potentially make use of this `mark_as_io_bound` feature too?

Yes, that code is very similar to what AS::Live does.

byroot:
> update the stats?

Done!

byroot:
👋

byroot:
I rebased the PR again. How can I get this unblocked?

joshuay03:
> I rebased the PR again. How can I get this unblocked?

Can you have a look at the latest failures. I'll aim to do a final review and merge this weekend.

byroot:
My bad, I didn't fix the merge conflict correctly. Expecting it to go green now.

adenta:
Can someone comment on what happens when this gets merged? Does this magically make Actioncable::Live non blocking so it can be used with Puma without performance issues?

byroot:
> Does this magically make Actioncable::Live non blocking so it can be used with Puma without performance issues?

No. I'll have to make changes on the Rails side first.

joshuay03 (approved):
I have some nits, but I'll follow up separately.

Thank you!

byroot:
😍

lib/puma/dsl.rb
```diff
@@ -595,6 +595,29 @@ def threads(min, max = min)
...
+    #
+    # @example
+    #   threads 5
+    #   max_io_threads 5
+    #
+    # The above example allows for 5 regular threads and 5 IO threads to process requests concurrently.
+    # Any IO thread over the limit is counted as a regular thread, hence the above configuration also
+    # allows for 3 regular threads and 7 IO threads for example.
+    def max_io_threads(max)
+      max = Integer(max)
```
joshuay03, line 613: Now wondering if we should allow this to be unbounded. Maybe something to come back to once it's in use out in the wild, i.e. Rails.

byroot: I don't think unbounded if ever a good idea. Even something that use very little CPU still use some, and it's always better to not accept quite as many concurrent requests as you could have (backpressure) than to accept too many.

adenta:
> > Does this magically make Actioncable::Live non blocking so it can be used with Puma without performance issues?
> 
> No. I'll have to make changes on the Rails side first.

@byroot this is AMAZING. Is there an issue open for the rails changes? Would this be a Rails 8.2 sort of thing? About to start building some chat interfaces, and most everyone in non rubyland is standardizing on SSE. Would love to not need falcon.

byroot:
> Is there an issue open for the rails changes?

Kinda, this change was driven by https://github.com/rails/rails/issues/55132#issuecomment-3590620159

I still need to find the time to figure out the Rails side of it. Ideally you'd just use `AC::Live` like today with no change, just the underlying implementation would be way more sound. But I'm not yet 100% sure it's possible, I'll have to see when I attempt it. If really it's not possible, then it will be a new API.

> Would this be a Rails 8.2 sort of thing?

Yes.

> About to start building some chat interfaces, and most everyone in non rubyland is standardizing on SSE. Would love to not need falcon.

I mean, you don't have to wait for Rails support. If you got this feature deployed you can use it directly with `request.env["puma.mark_as_io_bound"].call` and then use classic Rack steaming API.

rainerborene:
> I mean, you don't have to wait for Rails support. If you got this feature deployed you can use it directly with request.env["puma.mark_as_io_bound"].call and then use classic Rack steaming API.

```ruby
class MessagesController < ApplicationController
  def create
    request.env["puma.mark_as_io_bound"].call
    response.headers["Content-Type"] = "text/event-stream"
    response.headers["Last-Modified"] = Time.now.httpdate

    sse = ActionController::Live::SSE.new(response.stream)
    sse.write("stuff")
  ensure
    sse.close
  end
end
```

@byroot just like that?

byroot:
yes

PR #3823, opened 2025-11-17
Fix advertising of CLI config before config files are loaded

### LINKED ISSUE #3801: `Puma.cli_config` is nil when accessed from config/puma.rb
Hey there 👋 - first things first: Thank you guys for maintaining this great piece of code! 🥇 

**Describe the bug**
While migrating from Puma 6.6 to 7.0 I noticed my configuration setup is now broken due to a `NoMethodError`. My `config/puma.rb` looks like this:

```ruby
# frozen_string_literal: true

if Puma.cli_config.options[:workers] >= 2
  #     ^-- +nil+ now
  preload_app!   

  before_fork do
    ActiveRecord::Base.connection_pool.disconnect! if defined?(ActiveRecord)
  end

  before_worker_boot do
  end
end
```

The amount of workers is configured at the CLI-level via the `--workers` flag. With Puma 6 I was able to fetch the parsed `ARGV` options from `Puma.cli_config` within the file configuration. Now I have to dig around like this:

```ruby
if @config.instance_variable_get(:@_options)[:workers] >= 2
  # ...
end
```

**Expected behavior**

It would be great if Puma would expose an attribute to the parsed CLI/Env options (eg. `Puma.env_config`) to the file configuration DSL. I guess this was broken by https://github.com/puma/puma/pull/3297 or https://github.com/puma/puma/pull/3616. 

Maybe it would also help to distinguish the configuration patterns by a) CLI and b) programmatic Puma usage. This way one could have a CLI configuration stack like this: defaults, argv, env, file config - so that the file config is able to just read `workers` (without args) from the DSL, while also allowing to overwrite it via `workers(n)`. And the programmatic usage would be just defaults + configure block/DSL.

I'm aware of the configuration source mixing I'm doing, but wouldn't it be easier if each configuration layer is applied in series so that the next layer can directly access the current configuration attributes if needed?

joshuay03:
Closes #3801

#3616 introduced a breaking change where the CLI config was only advertised after `#clamp` (and therefore `#load`) [was called](https://github.com/puma/puma/pull/3616/files#diff-602f49141abce5814cc6115eb1c4e070adcd5bf7db662a97e0d6205b2ea5402dR54), meaning it wasn't available for use in the config files being loaded, which seems to have been [the original intent](https://github.com/puma/puma/commit/211aef15899a8e5b21174e74519193ead07768c9).

Here I've moved the setting of `Puma.cli_config` right before clamping to reintroduce that functionality, and added a regression test for it.

**Note:** Unlike the example in #3801, users will now have to access the options via `_options`, which imo is a good thing, as it makes it clear that the options are yet to be finalised and that they're only reading the values at that point.

### DIFF
--- a/test/config/cli_config.rb
+++ b/test/config/cli_config.rb
@@ -1,0 +1,3 @@
+if Puma.cli_config._options[:workers] == 2
+  Puma.cli_config._options[:workers] = 4
+end
--- a/test/test_cli.rb
+++ b/test/test_cli.rb
@@ -420,4 +420,12 @@ class TestCLI < PumaTest
 
     assert_equal true, config.options[:preload_app]
   end
+
+  def test_config_is_advertised_to_config_files
+    skip_unless :fork
+
+    Puma::CLI.new ['-w 2', '-C', File.expand_path('config/cli_config.rb', __dir__)]
+
+    assert_equal 4, Puma.cli_config._options[:workers]
+  end
 end
--- a/lib/puma/configuration.rb
+++ b/lib/puma/configuration.rb
@@ -197,7 +197,7 @@ module Puma
       @clamped = false
     end
 
-    attr_reader :plugins, :events, :hooks
+    attr_reader :plugins, :events, :hooks, :_options
 
     def options
       raise NotClampedError, "ensure clamp is called before accessing options" unless @clamped
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -45,6 +45,8 @@ module Puma
       ## Minimal initialization for a potential early restart (e.g. when pruning bundle)
 
       @config = conf
+      # Advertise the CLI Configuration before config files are loaded
+      Puma.cli_config = @config if defined?(Puma.cli_config)
       @config.clamp
 
       @options = @config.options
@@ -70,8 +72,6 @@ module Puma
 
       env = launcher_args.delete(:env) || ENV
 
-      # Advertise the Configuration
-      Puma.cli_config = @config if defined?(Puma.cli_config)
       log_config if env['PUMA_LOG_CONFIG']
 
       @binder = Binder.new(@log_writer, @options)


test/test_cli.rb
```diff
@@ -420,4 +420,12 @@ def test_config_preloads_app_with_workers
 
     assert_equal true, config.options[:preload_app]
   end
+
+  def test_config_is_advertised_to_config_files
+    skip_unless :fork
+
+    Puma::CLI.new ['-w 2', '-C', File.expand_path('config/cli_config.rb', __dir__)]
```
joshuay03, line 427: On main:
```
puma [main] be ruby -Ilib:test test/test_cli.rb --name test_config_is_advertised_to_config_files
Run options: --name test_config_is_advertised_to_config_files --seed 8544

# Running:

E

Fabulous run in 0.006702s, 149.2092 runs/s, 0.0000 assertions/s.
Errors & Failures:

  1) Error:
TestCLI#test_config_is_advertised_to_config_files:
Errno::ENOENT: No such file or directory @ rb_sysopen - /Users/joshuayoung/Projects/puma/test/config/cli_config.rb
    lib/puma/dsl.rb:136:in 'IO.read'
    lib/puma/dsl.rb:136:in 'Puma::DSL#_load_from'
    lib/puma/configuration.rb:259:in 'block in Puma::Configuration#load'
    lib/puma/configuration.rb:259:in 'Array#each'
    lib/puma/configuration.rb:259:in 'Puma::Configuration#load'
    lib/puma/configuration.rb:283:in 'Puma::Configuration#clamp'
    lib/puma/launcher.rb:48:in 'Puma::Launcher#initialize'
    lib/puma/cli.rb:64:in 'Puma::CLI#initialize'
    test/test_cli.rb:427:in 'TestCLI#test_config_is_advertised_to_config_files'

1 runs, 0 assertions, 0 failures, 1 errors, 0 skips
```

test/config/cli_config.rb
```diff
@@ -0,0 +1,3 @@
+if Puma.cli_config._options[:workers] == 2
+  Puma.cli_config._options[:workers] = 4
```
Jack12816, line 2: I assume `workers 4` works as well? (As the DSL is still available and used)

Jack12816:
```diff
diff --git a/test/config/cli_config.rb b/test/config/cli_config.rb
index 101436a5..f6f9e6e3 100644
--- a/test/config/cli_config.rb
+++ b/test/config/cli_config.rb
@@ -1,3 +1,3 @@
  1 if Puma.cli_config._options[:workers] == 2
2  -  Puma.cli_config._options[:workers] = 4
  2+  workers 4
  3 end
```

Results in:

```console
$ bundle exec ruby -Ilib:test test/test_cli.rb

Run options: --seed 40918

# Running:

..................F....

Fabulous run in 0.078789s, 291.9202 runs/s, 1599.2151 assertions/s.
Errors & Failures:

  1) Failure:
TestCLI#test_config_is_advertised_to_config_files [test/test_cli.rb:429]:
Expected: 4
  Actual: 2

23 runs, 126 assertions, 1 failures, 0 errors, 0 skips
```

joshuay03: > I assume workers 4 works as well? (As the DSL is still available and used)

Not really, because CLI args take precedence over config files, so it wouldn't have worked before either i.e., that test failure is expected.

Jack12816: Okay, then thank you for your efforts! :) 

I guess I'm making up a theoretical edge case in my head then. Your solution will solve/ease my issue! :)

PR #3827, opened 2025-11-20
Add workers :auto

nateberkopec:
It's probably not that great that you MUST use the ENV var in order to get this neat behavior.

I noticed that this rounds down, which means a cpu quota of 512 will potentially put you into single mode (rounds 0.5 to 0). That's a bit weird but probably not hit often enough IRL to matter (and it's kinda what I would prefer to happen anyway).

### DIFF
--- a/lib/puma/configuration.rb
+++ b/lib/puma/configuration.rb
@@ -239,7 +239,7 @@ module Puma
       max = env['PUMA_MAX_THREADS'] || env['MAX_THREADS']
       persistent_timeout = env['PUMA_PERSISTENT_TIMEOUT']
       workers_env = env['WEB_CONCURRENCY']
-      workers = workers_env && workers_env != "" ? parse_workers(workers_env) : nil
+      workers = workers_env && workers_env.strip != "" ? parse_workers(workers_env.strip) : nil
 
       {
         min_threads: min && min != "" && Integer(min),
@@ -385,10 +385,10 @@ module Puma
     def parse_workers(value)
       if value == :auto || value == 'auto'
         require_processor_counter
-        return Integer(::Concurrent.available_processor_count)
+        Integer(::Concurrent.available_processor_count)
+      else
+        Integer(value)
       end
-
-      Integer(value)
     rescue ArgumentError, TypeError
       raise ArgumentError, "workers must be an Integer or :auto"
     end


lib/puma/dsl.rb
```diff
@@ -669,21 +669,24 @@ def state_permission(permission)
...
+    # How many worker processes to run. Typically this is set to the number of
+    # available cores.
     #
     # The default is the value of the environment variable +WEB_CONCURRENCY+ if
-    # set, otherwise 0.
+    # set, otherwise 0. Passing +:auto+ will set the value to
+    # +Concurrent.available_processor_count+ (requires the concurrent-ruby gem).
+    # If available processor count is a Float (cpu quotas), we will round down.
     #
-    # @note Cluster mode only.
```
nateberkopec, line 678: This seems wrong to me when `workers 0` can be used to set single mode.

docs/deployment.md
```diff
@@ -35,10 +35,12 @@ For the purposes of Puma provisioning, "CPU cores" means:
 
 Set your config with the following process:
 
-* Use cluster mode and set the number of workers to the same number of CPU cores on the machine (minimum 2, otherwise use single mode!)
+* Use cluster mode and set `workers :auto` (requires the `concurrent-ruby` gem) to match the number of CPU cores on the machine (minimum 2, otherwise use single mode!). If you can't add the gem, set the worker count manually to the available CPU cores.
```
joshuay03, line 38: Nit: The bit about matching CPU cores is repetitive. What do you think about:

```suggestion
* Use cluster mode with `workers :auto` (requires the `concurrent-ruby` gem) or manually set the worker count to match the number of CPU cores on the machine (minimum 2, otherwise use single mode!).
```

README.md
```diff
@@ -115,9 +115,16 @@ Or with the `WEB_CONCURRENCY` environment variable:
...
 ```
 
+When using a config file, most applications can simply set `workers :auto` (requires the `concurrent-ruby` gem) to match the number of worker processes to the available processors:
+
+```ruby
+# config/puma.rb
+workers :auto
+```
+
 Note that threads are still used in cluster mode, and the `-t` thread flag setting is per worker, so `-w 2 -t 16:16` will spawn 32 threads in total, with 16 in each worker process.
```
joshuay03, line 127: The positioning of this line now seems awkward. It might read better before 'When using a config file ...'?

lib/puma/configuration.rb
```diff
@@ -380,12 +376,23 @@ def require_processor_counter
...
       raise
     end
 
+    def parse_workers(value)
+      if value == :auto || value == 'auto'
+        require_processor_counter
+        return Integer(::Concurrent.available_processor_count)
+      end
+
+      Integer(value)
```
MSP-Greg: Maybe update with the following:

```ruby
      if value == :auto || value == 'auto'
        require_processor_counter
        Integer(::Concurrent.available_processor_count)
      else
        Integer(value)
      end
```

MSP-Greg: @nateberkopec

Sorry, that didn't turn out right.  I think you know what I mean.

Have a good day...

byroot:
Hum, not sure if you are aware, but I did implement `.available_processors_count` specifically to use inside Rails' `puma.rb` config. But turns out a bunch of providers don't expose the cgroups limits, so we had to revert: https://github.com/rails/rails/issues/52522 /  https://github.com/rails/rails/commit/c68cea4f3317ac39f5174d55b95f3f028cf85653

But perhaps this is outdated info? Did you test the common providers?

dentarg:
> I noticed that this rounds down, which means a cpu quota of 512 will potentially put you into single mode (rounds 0.5 to 0). That's a bit weird but probably not hit often enough IRL to matter (and it's kinda what I would prefer to happen anyway).

Should we care about this?

...if single mode happens, and you have hooks for cluster mode, you will start to see warnings about `before_worker_boot`, so perhaps you remove them, but then later, `:auto` gets you multiple workers again, and you needed those hooks (disconnect database, etc.)

nateberkopec:
@byroot interesting, but this isn't a default ATM so I'm not worried about it not working on all platforms.

@dentarg That's almost more of an issue with how we do hooks probably than an issue with this :auto behavior I'd say? Worth calling out in docs though.

dentarg:
> but this isn't a default ATM

Should it ever be?

jjb:
some environments don't reveal the cpu share, and `::Concurrent.available_processor_count` will show the host not the running container. on two i just checked:

### cgroups v1

this is -1
```
cat /sys/fs/cgroup/cpu/cpu.cfs_quota_us
```

### cgroups v2

this is not present
```
cat /sys/fs/cgroup/cpu.max
```

puma could introspect this and conclude "my cores are being limited but i don't know what the limit is"

nateberkopec:
@dentarg Maybe, but then we'd have to do things like deal with @byroot's comment

@jjb just curious, which env, would help me if I could reproduce?

What would even be the correct behavior in that situation... deploy in single mode but warn? Bit of a dangerous situation.

jjb:
> which env, would help me if I could reproduce?

aptible and shipyard. i bet @bueller would be happy to give you a free environment to play with on shipyard

> What would even be the correct behavior in that situation... deploy in single mode but warn? Bit of a dangerous situation.

in all environments except prod, definitely raise

in systems i've worked with, raising in prod is also fine. halts the deploy, user sees what error is and fixes it. but could also fall back to 1 process and still be in cluster mode (in case there are cluster-specific behaviors the user is expecting). and log an error.

PR #3828, opened 2025-11-20
Implement dsize and dcompact functions for Puma::HttpParser

byroot:
`dsize` allows these objects to be properly reported in heap dumps.

`dcompact` allows the GC to move the other objects this one refers to.

I also wanted to implement write barrier, but this is made difficult by the current parser architecture. But I have ideas on how to change it that may or may not pan out.


nateberkopec:
0 days without (Lenny, Simpsons)

PR #3630, opened 2025-03-12
Support backtrace signals on Linux with SIGPWR

emilong:
### Description

I wanted to be able to get stack traces from a running Puma process on Linux. I noticed some Linux distributions alias SIGINFO to SIGPWR, so chose that as the Linux signal for the operation, falling back to its usage if SIGINFO is missing on the given system. `pumactl` shares the "detection" code with `puma` itself, so sending `info` from `pumactl` will allow backtraces in either case.

I tested this manually on Linux and a FreeBSD VM to validate it works on both as there weren't automated tests for signal handling already.

### Your checklist for this pull request
- [x] I have reviewed the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.
- [ ] I have added (or updated) appropriate tests if this PR fixes a bug or adds a feature.
- [x] My pull request is 100 lines added/removed or less so that it can be easily reviewed.
- [ ] If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.
- [x] I have updated the documentation accordingly.
- [ ] All new and existing tests passed, including Rubocop.

PR #3829, opened 2025-11-20
SIGPWR for backtrace on Linux/JRuby

nateberkopec:
This is a follow-on of #3630 which extends it to allow SIGPWR for JRuby on Linux.

### DIFF
--- a/lib/puma/detect.rb
+++ b/lib/puma/detect.rb
@@ -36,7 +36,7 @@ module Puma
   end
 
   BACKTRACE_SIGNAL =
-    if !jruby? && Signal.list.key?("INFO")
+    if Signal.list.key?("INFO")
       "SIGINFO"
     elsif Signal.list.key?("PWR")
       "SIGPWR"


lib/puma/detect.rb
```diff
@@ -35,6 +42,13 @@ def self.windows?
     IS_WINDOWS
   end
 
+  BACKTRACE_SIGNAL =
+    if !jruby? && Signal.list.key?("INFO")
```
ahorek: JRuby doesn't define SIGINFO on any platform https://github.com/jnr/jnr-constants/blob/master/src/main/java/jnr/constants/platform/linux/Signal.java

ahorek:
On Linux, CRuby also does not define Signal.list["INFO"] (BSD specific?)

SIGPWR doesn’t seem ideal for printing backtraces, but I don’t think there’s a better alternative, since the only two user-defined signals are already in use...

joshuay03:
> This is a follow-on of #3530 which extends it to allow SIGPWR for JRuby on Linux.

Did you mean https://github.com/puma/puma/pull/3630?

snickell:
Thank you so much for implementing this, absolutely saved our bacon this week, very appreciating. Would love to see this included in a puma release sometime soon, one of those "can't believe I lived without this" features.

PR #3830, opened 2025-11-20
bin->exe

nateberkopec:
Time to modernize.

### DIFF
--- a/bin/puma
+++ b/bin/puma
@@ -1,0 +1,6 @@
+#!/usr/bin/env ruby
+# This file is deprecated and will be removed in Puma 8.0.
+# Please use `puma` command directly from your PATH or exe/puma.
+
+exe = File.expand_path('../exe/puma', __dir__)
+exec exe, *ARGV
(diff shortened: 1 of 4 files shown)


nateberkopec:
I think this is gonna break systemd units. That's the one place I can think of where people are probably using an absolute path to the puma executable.

Probably gonna have to shim until Puma 8.

bin/puma
```diff
@@ -1,10 +1,6 @@
 #!/usr/bin/env ruby
-#
-# Copyright (c) 2011 Evan Phoenix
-#
+# This file is deprecated and will be removed in Puma 8.0.
+# Please use `puma` command directly from your PATH or exe/puma.
```
joshuay03, line 3: Shall we print these as warnings?


```suggestion
warn "bin/puma is deprecated and will be removed in Puma 8.0. Please use the `puma` command directly from your PATH or exe/puma."
```

joshuay03 (approved):

dentarg:
What makes `exe/` more modern over `bin/`?

dentarg:
I mean

Screenshot 2025-11-25 at 00 31 14

Screenshot 2025-11-25 at 00 31 28

https://github.com/rails/rails/blob/main/railties/exe/rails hasn't changed in 8 years

PR #3837, opened 2025-12-12
don't break if WEB_CONCURRENCY is set to a blank string

jjb:
### Description

- [ ] I have reviewed the [guidelines for contributing](../blob/main/CONTRIBUTING.md) to this repository.

### DIFF
--- a/test/test_config.rb
+++ b/test/test_config.rb
@@ -838,10 +838,12 @@ class TestConfigEnvVariables < PumaTest
   end
 
   def test_config_ignores_blank_workers_from_env
-    env = { "WEB_CONCURRENCY" => "" }
-    conf = Puma::Configuration.new({}, {}, env)
-    conf.clamp
-    assert_equal 0, conf.options.default_options[:workers]
+    [nil, '', ' ', " \n ", " \t "].each do |wc|
+      env = { "WEB_CONCURRENCY" => wc }
+      conf = Puma::Configuration.new({}, {}, env)
+      conf.clamp
+      assert_equal 0, conf.options.default_options[:workers]
+    end
   end
 
   def test_config_does_not_preload_app_if_not_using_workers
@@ -849,15 +851,6 @@ class TestConfigEnvVariables < PumaTest
     conf = Puma::Configuration.new({}, {}, env)
     conf.clamp
     assert_equal false, conf.options.default_options[:preload_app]
-  end
-
-  def test_blank_web_concurrency_does_not_break
-    [nil, '', ' ', " \n ", " \t "].each do |wc|
-      env = { "WEB_CONCURRENCY" => wc }
-      conf = Puma::Configuration.new({}, {}, env)
-      conf.clamp
-      assert_equal false, conf.options.default_options[:preload_app]
-    end
   end
 
   def test_config_preloads_app_if_using_workers


lib/puma/configuration.rb
```diff
@@ -249,7 +249,7 @@ def puma_options_from_env(env = ENV)
         min_threads: min && min != "" && Integer(min),
         max_threads: max && max != "" && Integer(max),
         persistent_timeout: persistent_timeout && persistent_timeout != "" && Integer(persistent_timeout),
-        workers: workers && workers != "" && Integer(workers),
+        workers: workers && workers.strip != "" && Integer(workers),
```
jjb: another option is

```ruby
workers: !workers&.empty? && Integer(workers),
```

joshuay03: I don't think that handles the `nil` case well.

test/test_config.rb
```diff
@@ -851,6 +851,15 @@ def test_config_does_not_preload_app_if_not_using_workers
     assert_equal false, conf.options.default_options[:preload_app]
   end
 
+  def test_weird_web_concurrency_does_not_break
+    [nil, '', ' ', 'hello', ' hello ', '!?', ' !? '].each do |wc|
```
jjb:
```suggestion
  def test_blank_web_concurrency_does_not_break
    [nil, '', ' ', " \n ", " \t "].each do |wc|
```

jjb:
> Can you add a test for this case please? I think we have a bunch for this env var config.

done, and made this to make sure it fails in main: https://github.com/puma/puma/pull/3839

although both seem stuck on github actions, not sure why. i did get runs working before, maybe i'm being throttled, will come back and check later

jjb:
(i could not get tests running locally, contributing.md seems to be out of date - if this is not expected let me know and i'll write it up)

jjb:
i just realized something obvious, any setting could have this problem, so .strip should be invoked on all settings. that's probably fine but should we 

1. add/modify tests for all cases
2. add/modify tests for no cases, just add .strip
3. make a helper/wrapper which loads things from env and strips them?

test/test_config.rb
```diff
@@ -838,10 +838,12 @@ def test_config_loads_workers_from_env
   end
 
   def test_config_ignores_blank_workers_from_env
-    env = { "WEB_CONCURRENCY" => "" }
-    conf = Puma::Configuration.new({}, {}, env)
-    conf.clamp
-    assert_equal 0, conf.options.default_options[:workers]
+    [nil, '', ' ', " \n ", " \t "].each do |wc|
```
dentarg, line 841: I don't think `nil` needs to be tested, `ENV` values are always strings

```suggestion
    ['', ' ', " \n ", " \t "].each do |wc|
```

dentarg:
> (i could not get tests running locally, contributing.md seems to be out of date - if this is not expected let me know and i'll write it up)

Feel free to make a issue or discussion about this

dentarg:
> i just realized something obvious, any setting could have this problem, so .strip should be invoked on all settings

Sure, but that could be another PR, I think this addition is worthwhile

test/test_config.rb
dentarg: I guess we can keep it as it simulates the case when the user did not set `WEB_CONCURRENCY`

PR #3825, opened 2025-11-17
puma_http11.c: pre-intern env keys

byroot:
As explained in https://byroot.github.io/ruby/json/2025/01/12/optimizing-ruby-json-part-6.html when inserting a string key in a hash, Ruby will try to lookup an equivalent string in the interned string table.

As such, in C extensions, it's generally advantageous to pre-intern that string, or at least to pre-freeze it.

~~The gain is minor, but measurable~~. The gain is ~17% once common fields are interned too.

```
ruby 3.4.6 (2025-09-16 revision dbd83256b1) +YJIT +PRISM [arm64-darwin24]
Warming up --------------------------------------
               after    27.498k i/100ms
Calculating -------------------------------------
               after    272.866k (± 1.5%) i/s    (3.66 μs/i) -      1.375M in   5.039945s

Comparison:
              before:   265747.8 i/s
               after:   272866.0 i/s - 1.03x  faster
```

```ruby
require "bundler/inline"

gemfile do
  source "https://rubygems.org"
  gem "benchmark-ips"
  gem "puma", path: "."
end

request = <<~HTTP.split("\n").join("\r\n").freeze
  GET /puma/puma HTTP/1.1
  Host: github.com
  User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:144.0) Gecko/20100101 Firefox/144.0
  Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
  Accept-Language: en-US,en;q=0.5
  Accept-Encoding: gzip, deflate, br, zstd
  Connection: keep-alive
[...]

PR #3838, opened 2025-12-13
JRuby HTTP parser improvements

headius:
This PR will contain improvements for the HTTP parser implementation for JRuby.

- [x] Pass bytes from the request handler through the parser loop and leaf methods, only copying when necessary.
- [x] Pre-allocate static header keys and avoid constructing them again while preparing the request hash.
- [x] Perfect hash for retrieving pre-allocated header keys from incoming bytes.
- [x] Refactor parser classes to reduce pointer dereferences
- [x] Eliminate dead code
- [ ] ~Intern unusual headers for future calls using zero-allocation "fake string" logic~

Contribution checklist:

### DIFF
--- a/.github/workflows/ragel.yml
+++ b/.github/workflows/ragel.yml
@@ -28,7 +28,7 @@ jobs:
       matrix:
         include:
           - { os: ubuntu-22.04 , ruby: head }
-          - { os: macos-13     , ruby: head }
+          - { os: macos-15     , ruby: head }
           # Dec-2023 - incorrect line directives with Windows
           # occurs with both MSYS2 and MSFT/vpkg versions of ragel
           # - { os: windows-2022 , ruby: ucrt } 
@@ -79,7 +79,7 @@ jobs:
           }
 
       - name: save ragel generated files on fail
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         if: ${{ failure() }}
         with:
           name: ${{ matrix.os }}-ragel-generated-files
--- a/.github/workflows/turbo-rails.yml
+++ b/.github/workflows/turbo-rails.yml
@@ -70,6 +70,11 @@ jobs:
           # DST="kw =\n    if RUBY_VERSION.start_with? '3'\n      {git: 'https://github.com/teamcapybara/capybara.git', ref: '43e32a8495'}\n    else\n      {}\n    end\n  gem 'capybara', **kw"
           # sed -i "s#$SRC#$DST#" Gemfile
           #
+          # use `minitest-mock`
+          SRC="gem ['\"]rexml['\"]"
+          DST="gem 'rexml'\n  gem 'minitest-mock'"
+          sed -i "s#$SRC#$DST#" Gemfile
+          #
           # use `stdio` for log_writer, always have one thread existing
           SRC="Silent: true"
           DST="Silent: false, Threads: '1:4'"
@@ -90,4 +95,4 @@ jobs:
 
       - name: turbo-rails test
         id: test
-        run: bin/test test/**/*_test.rb -vd
+        run: bin/test test/**/*_test.rb -v
--- a/Gemfile
+++ b/Gemfile
@@ -12,17 +12,16 @@ gem "minitest-proveit"
 gem "minitest-stub-const"
 gem "concurrent-ruby", "~> 1.3"
 
-case ENV['PUMA_CI_RACK']&.strip
-when 'rack2'
-  gem "rackup", '~> 1.0'
-  gem "rack"  , '~> 2.2'
+if ENV['PUMA_CI_RACK']&.strip == 'rack2'
+  gem "rack"  , "~> 2.2"
+  gem "rackup", "~> 1.0"
+## Temporarily disable using rack & rackup main branches
+#elsif RUBY_PATCHLEVEL == -1
+#  gem "rack"  , github: "rack/rack"  , branch: "main"
+#  gem "rackup", github: "rack/rackup", branch: "main"
 else
-  gem "rackup", '>= 2.0'
-  if RUBY_PATCHLEVEL == -1
-    gem "rack", git: "https://github.com/rack/rack", ref: "main"
-  else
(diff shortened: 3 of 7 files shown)


headius:
The first improvement, passing the read bytes through the parser without copying, is mostly done. I'm not quite clear on the life cycle of that buffer string, so my patch still makes defensive copies of the header values before putting them in the request hash. I have not looked at the C code to see if it is sharing the buffer contents with those strings. If anyone has insight here, I would appreciate it.

Next phase will be to pre-allocate the static header keys, similar to the C code, since the Java extension is currently recreating them every time.

I'm not sure I can keep these aggregate changes to less than 100 lines of modifications, so please let me know if that will be a problem for reviewing.

headius:
Related to preallocating the header strings: https://github.com/puma/puma/pull/3825

I wonder what other optimizations over the years have only been applied to the C code. 🤔

headius:
I don't love the linear search for HTTP envs but it matches the C ext now.

nateberkopec:
Legend! Thank you @headius

headius:
I've completed most of the items I found. Performance is only marginally better, probably because the code generated by Ragel is just bad (I've run into similar performance limitations with the json library), but the memory throughput should be drastically reduced. A benchmark like the one linked from #3825 now only allocates objects for the benchmark itself, unusual headers, and values from the request (other than objects required for JRuby internals).

Benchmark were run with a patched JRuby 10.1 that does not raise internal exceptions to wake a ConditionVariable and uses fast fixnum hashing. Numbers are peak performance seen after warmup stabilization.

Performance before this PR:
```
jruby 10.1.0.0-SNAPSHOT (3.4.5) 2025-12-15 8c2eca74b5 OpenJDK 64-Bit Server VM 25+36-LTS on 25+36-LTS +indy +jit [arm64-darwin]
               parse     7.103k i/100ms
               parse     70.925k (± 0.6%) i/s   (14.10 μs/i) -    355.150k in   5.007611s
```

After:
```
jruby 10.1.0.0-SNAPSHOT (3.4.5) 2025-12-15 8c2eca74b5 OpenJDK 64-Bit Server VM 25+36-LTS on 25+36-LTS +indy +jit [arm64-darwin]
               parse     7.401k i/100ms
               parse     73.935k (± 1.1%) i/s   (13.53 μs/i) -    370.050k in   5.005638s
```

The final item – using a "fake string" to cache and lookup additional header fstrings – is trickier to do in JRuby than CRuby, as the fstring cache does not currently provide a way to perform a lookup of fstrings using only an array of bytes, and any temporary strings used for this purpose must be isolated across threads. If the Http11 object is only ever used in a single thread, such a temporary string could be constructed there at the cost of an additional string per request. This may be worth it to avoid constantly creating new strings for the remaining less-common headers, such as those used in the #3825 benchmark.

headius:
An experiment with interning all incoming headers yielded a small gain (maybe) but I'm not sure it's worth the exposure. Given a large enough request with bogus headers, memory could be made to increase rapidly. I believe this behavior of Puma may be exposing an issue with the fstring cache in CRuby 4.0, as it produces very erratic numbers for the request parsing benchmark:

```
ruby 4.0.0dev (2025-11-18T23:50:36Z master 32b8f97b34) +PRISM [arm64-darwin24]
               parse    27.569k i/100ms
               parse    201.399k (±12.0%) i/s    (4.97 μs/i) -      1.020M in   5.142527s
ruby 4.0.0dev (2025-11-18T23:50:36Z master 32b8f97b34) +PRISM [arm64-darwin24]
               parse    15.937k i/100ms
               parse    118.456k (±12.5%) i/s    (8.44 μs/i) -    589.669k in   5.075580s
ruby 4.0.0dev (2025-11-18T23:50:36Z master 32b8f97b34) +PRISM [arm64-darwin24]
               parse    10.524k i/100ms
               parse     90.644k (± 9.4%) i/s   (11.03 μs/i) -    452.532k in   5.054908s
ruby 4.0.0dev (2025-11-18T23:50:36Z master 32b8f97b34) +PRISM [arm64-darwin24]
               parse     8.775k i/100ms
               parse     83.366k (± 2.0%) i/s   (12.00 μs/i) -    421.200k in   5.054538s
ruby 4.0.0dev (2025-11-18T23:50:36Z master 32b8f97b34) +PRISM [arm64-darwin24]
               parse     8.055k i/100ms
               parse     72.597k (± 4.0%) i/s   (13.77 μs/i) -    362.475k in   5.001894s
ruby 4.0.0dev (2025-11-18T23:50:36Z master 32b8f97b34) +PRISM [arm64-darwin24]
               parse     6.705k i/100ms
               parse    183.568k (±60.6%) i/s    (5.45 μs/i) -    549.810k in   5.004438s
```

I think I'm going to exclude that feature from the JRuby extension for now.

This is probably all I can do for the Ragel-based parser extension on JRuby. It's probably worth looking into a non-Ragel solution, especially for HTTP/1.1 which is quite easy to parse.

headius:
@nateberkopec @schneems I'm not sure why that one JRuby timed out, nor why those unrelated jobs have been consistently failing.

The CRuby jobs have been failing since the first commit, and I did not touch any code used by the CRuby version of the gem. Is it some oddity with PR jobs?

I'll try to reproduce the hang locally but everything else passes and I don't know why my changes would have caused that sort of failure.

MSP-Greg:
> The CRuby jobs have been failing since the first commit, and I did not touch any code used by the CRuby version of the gem. Is it some oddity with PR jobs?

Disregard the CRuby 'head' jobs.  I hoped to have time for a better look at the issue.  In the meantime, I'll post a fix so CI passes.

MSP-Greg:
@headius

Well, I didn't turn out quite as I hoped, but the CI tests seem to be much better now...

headius:
I've merged in recent changes from main to clean up CI. I'm satisfied with this round of changes and it generally performs much better on JRuby now than on CRuby.

headius:
CI is green except for two truffleruby-head builds that are unrelated to my changes.

nateberkopec:
Thanks again @headius

PR #3847, opened 2025-12-27
Use IPv6 (`::`) default for production

### LINKED ISSUE #3812: IPv6 addresses as default
IPv4 addresses like 127.0.0.1 and 0.0.0.0 are easy and familiar, but they're increasingly more expensive with provider such as AWS (https://aws.amazon.com/blogs/aws/new-aws-public-ipv4-address-charge-public-ip-insights/). IPv6 addresses are much more plentiful, and therefore cheaper (and represent the future).

Replace defaults of `::` instead of `0.0.0.0` (production). When Puma binds to the host `::`, it also works with the old IPv4 equivalents, but not the other way around. Here's a table showing compatability, The host the server is booted with is listed on the rows, and the url is listed on the columns. If it can connect it has a ✅ if it cannot it has a ❌. Chart:

| Host | localhost:3000 | 127.0.0.1:3000 | \[::1\]:3000 | 0.0.0.0:3000 | \[::\]:3000 | Network Open |
|:---:|:---:|:---:|:---:|:---:|:---:|:---:|
| `localhost` | ✅ | ✅ | ✅ | ✅ | ✅ | ❌ No |
| `127.0.0.1` | ✅ | ✅ | ❌ | ✅ | ❌ | ❌ No |
| `::1` | ✅ | ❌ | ✅ | ❌ | ✅ | ❌ No |
| `0.0.0.0` | ✅ | ✅ | ❌ | ✅ | ❌ | ✅ Yes |
| `::` | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ Yes |

So `::` responds to all of the above (on a mac). Unfortunately `::1` doesn't work for `127.0.0.1:3000` or `0.0.0.0:3000` so I think for development, we should stick with `localhost` (and this is a production optimization anyway). Here's the  config and script I used to generate that table https://gist.github.com/schneems/0e3e5fb61b6e4128499af01c9cf7ac1c.

In addition to the work in Puma, there's also an opportunity to change the default in downstream frameworks such as Rails, Sinatra, and Hanami.

Do nothing, someone can still manually pass in a host, but it's cumbersome:

```ruby
# config/puma.rb

# Support IPv6 by binding to host `::` in production instead of `0.0.0.0` and `::1` instead of `127.0.0.1` in development.
host = ENV.fetch("RAILS_ENV") { "development" } == "production" ? "::" : "::1"

port(ENV.fetch("PORT") { 9292 }, host)
```

schneems:
IPv4 addresses like 0.0.0.0 are easy and familiar, but they're increasingly more expensive with provider such as AWS (https://aws.amazon.com/blogs/aws/new-aws-public-ipv4-address-charge-public-ip-insights/). IPv6 addresses are much more plentiful, and therefore cheaper (and represent the future).

This update switches the default production address to IPv6 `::`. Here's the table of the URLs that supports:

| Host | localhost:3000 | 127.0.0.1:3000 | \[::1\]:3000 | 0.0.0.0:3000 | \[::\]:3000 | Network Open |
|:---:|:---:|:---:|:---:|:---:|:---:|:---:|
| `localhost` | ✅ | ✅ | ✅ | ✅ | ✅ | ❌ No |
| `127.0.0.1` | ✅ | ✅ | ❌ | ✅ | ❌ | ❌ No |
| `::1` | ✅ | ❌ | ✅ | ❌ | ✅ | ❌ No |
| `0.0.0.0` | ✅ | ✅ | ❌ | ✅ | ❌ | ✅ Yes |
| `::` | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ Yes |

Generated with https://gist.github.com/schneems/0e3e5fb61b6e4128499af01c9cf7ac1c.

Puma has no concept of a "development" default host, so we don't need to use `::1` instead of `127.0.0.1` or `localhost`. Also due to the behavior of `::1` not falling back to IPv4 the same way that `::` does, it seems a higher-cost, lower benefit update, so I'm not recomending it for other defaults such as Rails https://github.com/rails/rails/pull/56470.

Close #3812

- [] If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.

### DIFF
--- a/lib/puma/cli.rb
+++ b/lib/puma/cli.rb
@@ -148,7 +148,7 @@ module Puma
 
           o.on "-p", "--port PORT", "Define the TCP port to bind to",
             "Use -b for more advanced options" do |arg|
-            user_config.bind "tcp://#{Configuration::DEFAULTS[:tcp_host]}:#{arg}"
+            user_config.port arg, Configuration.default_tcp_host
           end
 
           o.on "--pidfile PATH", "Use PATH as a pidfile" do |arg|
--- a/lib/puma/configuration.rb
+++ b/lib/puma/configuration.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+require 'socket'
+require 'uri'
+
 require_relative 'plugin'
 require_relative 'const'
 require_relative 'dsl'
@@ -232,6 +235,8 @@ module Puma
 
     def puma_default_options(env = ENV)
       defaults = DEFAULTS.dup
+      defaults[:tcp_host] = self.class.default_tcp_host
+      defaults[:binds] = [self.class.default_tcp_bind]
       puma_options_from_env(env).each { |k,v| defaults[k] = v if v }
       defaults
     end
@@ -282,6 +287,7 @@ module Puma
       run_mode_hooks
       set_conditional_default_options
       @_options.finalize_values
+      rewrite_unavailable_ipv6_binds!
       @clamped = true
       warn_hooks
       options
@@ -360,6 +366,23 @@ module Puma
       options.final_options
     end
 
+    def self.default_tcp_host
+      ipv6_interface_available? ? Const::UNSPECIFIED_IPV6 : Const::UNSPECIFIED_IPV4
+    end
+
+    def self.default_tcp_bind(port = DEFAULTS[:tcp_port])
+      URI::Generic.build(scheme: 'tcp', host: default_tcp_host, port: Integer(port)).to_s
+    end
+
+    def self.ipv6_interface_available?
+      Socket.getifaddrs.any? do |ifaddr|
+        addr = ifaddr.addr
+        addr&.ipv6? && !addr&.ipv6_loopback?
+      end
+    rescue StandardError
+      false
+    end
+
     def self.temp_path
       require 'tmpdir'
 
@@ -374,6 +397,31 @@ module Puma
     end
 
     private
+
+    def rewrite_unavailable_ipv6_binds!
+      return if self.class.ipv6_interface_available?
+
+      tried_ipv6_bind = false
+      bind_schemes = ['tcp', 'ssl']
+
+      @_options[:binds] = Array(@_options[:binds]).map do |bind|
+        uri = URI.parse(bind)
+        next bind unless bind_schemes.include?(uri.scheme)
+
+        host = uri.host&.delete_prefix('[')&.delete_suffix(']')
+        next bind unless host&.include?(':')
+
+        tried_ipv6_bind = true
+        next bind unless host == Const::UNSPECIFIED_IPV6
+
+        uri.host = Const::UNSPECIFIED_IPV4
+        uri.to_s
+      rescue URI::InvalidURIError
+        bind
+      end
+
+      warn "WARNING: IPv6 bind requested but no non-loopback IPv6 interface was detected" if tried_ipv6_bind
+    end
 
     def require_processor_counter
       require 'concurrent/utility/processor_counter'
--- a/lib/puma/configuration.rb
+++ b/lib/puma/configuration.rb
@@ -131,7 +131,7 @@ module Puma
 
     DEFAULTS = {
       auto_trim_time: 30,
-      binds: ['tcp://0.0.0.0:9292'.freeze],
+      binds: ['tcp://[::]:9292'.freeze],
       debug: false,
       early_hints: nil,
       enable_keep_alives: true,
@@ -167,7 +167,7 @@ module Puma
       silence_fork_callback_warning: false,
       silence_single_worker_warning: false,
       tag: File.basename(Dir.getwd),
-      tcp_host: '0.0.0.0'.freeze,
+      tcp_host: '::'.freeze,
       tcp_port: 9292,
       wait_for_less_busy_worker: 0.005,
       worker_boot_timeout: 60,
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -260,7 +260,7 @@ module Puma
     # accepted protocols. Multiple urls can be bound to, calling +bind+ does
     # not overwrite previous bindings.
     #
-    # The default is "tcp://0.0.0.0:9292".
+    # The default is "tcp://[::]:9292".
     #
     # You can use query parameters within the url to specify options:
     #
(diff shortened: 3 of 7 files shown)


dentarg:
@schneems I think you should make tests fixes in its own branch/PR, so we can merge that independent of this PR!

schneems:
> I think you should make tests fixes in its own branch/PR, so we can merge that independent of this PR!

Done 💚

schneems:
I'm not actually sure how **large** this change is. I had to adjust the tests to get it to pass for windows (switching from 127.0.0.1 to localhost) so it is a breaking change for some users, for sure, but I believe it's a **small breaking change** (if there is such a thing). To be on the safe side, this would be a minor at least (7.1 -> 7.2). If we were doing strict semver (we aren't) this would need to be (7.1 -> 8.0).

nateberkopec:
>I'm not actually sure how large this change is.

Because it's a prod change, I'd prefer to mark this as a full breaking change.

nateberkopec (dismissed):
1. What do we do if binding to `::` fails? Currently this raises and dies with non-zero exit. Should we do something else?
2. The Rails issue notes there are ipv6-only hosts b/c it can be disabled at the kernel level. That definitely warrants breaking change to me, do we have any idea how often this occurs?

For me this is kind of a no as-is. Breaks hard with not a lot of user benefit.

schneems:
> he Rails issue notes there are ipv6-only hosts b/c it can be disabled at the kernel level. That definitely warrants breaking change to me, do we have any idea how often this occurs?

I've yet to find any sort of docker setup that reproduces that behavior. I believe it may exists, but given the difficulty to induce it makes me think the actual occurrence is small.

> Currently this raises and dies with non-zero exit. Should we do something else?

Without a system to test against it’s hard to say. In the scenario where IPv6 isn’t available I’m unclear if it still binds to IPv4 and behaves like 0.0.0.0 or if it fully fails. It might also be OS dependent.

We could catch and retry with 0.0.0.0 as an option. I would ideally not want to do that unless the error was clear and we aren’t retrying for all socket binding failures.

As a parallel thought: We could also allow setting tcp host via env var PUMA_TCP_HOST or just PUMA_HOST. Though it would only help people using something like the buildpack and not others who are running on a machine that doesn't have an IPv4 network.

Maybe adding env var support for config makes it less risky as someone could put the old 0.0.0.0 in their dot files if it failed for them.

schneems:
Adding a datapoint: I changed the Heroku docs for [recommended puma config](https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server) to use IPv6 November 22 of 2024. People leave feedback on articles, and I've not had any issues, granted these people are also likely using Heroku for prod, but I also know these docs show up in general search, and people sometimes copy and use them on other infrastructure. It's not an overwhelming signal that no one has complained, but it is **a signal**.

stadniklksndr:
Tested how this change behaves in environments where IPv6 is partially or fully disabled:

1) IPv6 disabled partially (`sysctl net.ipv6.conf.all.disable_ipv6` return `1`)
In this case, `TCPServer.new('::', 9292)` does not raise an error.
The socket is created with AF_INET6 family. However, it only accepts IPv4 traffic.

**The issue**: Puma will log `Listening on http://[::]:9292`, but IPv6 connections will fail with `"Can’t be reached"`. The logs will be misleading for users who expect real IPv6 connectivity.

In this case table for `::` looks like this: 

| Host | localhost:9292 | 127.0.0.1:9292 | \[::1\]:9292 | 0.0.0.0:9292 | \[::\]:9292 | Network Open |
|:---:|:---:|:---:|:---:|:---:|:---:|:---:|
| `::` | ✅ | ✅ | ❌ | ✅ | ❌ | ✅ Yes |

<br>

2. When the IPv6 kernel module is not loaded at all
IPv6 disabled (sudo vim /etc/default/grub; GRUB_CMDLINE_LINUX_DEFAULT="... ipv6.disable=1")
In this case, `TCPServer.new('::', 9292)` raises a fatal error:
`Errno::EAFNOSUPPORT (Address family not supported by protocol - socket(2) for "::" port 9292)`
This will break Puma startup for any user who relies on the default binding if their kernel has IPv6 disabled.

<br>

**Suggestion:** Perhaps checking for available IPv6 interfaces before deciding which address to bind to would help. For example:
```
has_real_ipv6 = Socket.getifaddrs.any? do |ifaddr|
  ifaddr.addr&.ipv6? && !ifaddr.addr&.ipv6_loopback?
end
```
Using this check (or a similar one) would allow Puma to:

Bind to 0.0.0.0 automatically if IPv6 is missing or disabled at the kernel level (partially or fully disabled), preventing startup crashes. Provide more accurate logging about which protocol is actually being used.

nateberkopec:
**Codex speaking below**.

Puma uses `TCPServer.new(host, port)` and **does not** set `IPV6_V6ONLY`, so behavior is inherited from the OS/kernel default for IPv6 sockets.

`bind "tcp://[::]:PORT"` can be either:

1. **dual-stack** (`::` accepts both IPv6 and IPv4-mapped traffic) or
2. **v6-only** (`::` accepts IPv6 only)

depending on platform/runtime configuration.

A few concrete data points:

- **Windows** docs for `IPV6_V6ONLY` say default is nonzero (v6-only):
  https://learn.microsoft.com/en-us/windows/win32/winsock/ipproto-ipv6-socket-options
- **Linux** `ipv6(7)` documents `net.ipv6.bindv6only` default `0` (dual-stack), but this is tunable:
  https://man7.org/linux/man-pages/man7/ipv6.7.html

I also reproduced the behavior with Ruby sockets:

- v6-only socket bound to `::` rejects `127.0.0.1` with `ECONNREFUSED`
- same socket accepts `::1`
- connecting to `localhost` succeeds (resolver picks a compatible address)

### User impact

- Users who explicitly set `bind`/`host` are unaffected.
- Users relying on defaults may see tooling drift if local scripts/healthchecks hardcode `127.0.0.1`.
  - e.g. `curl http://127.0.0.1:$PORT` may fail in v6-only environments
  - `curl http://localhost:$PORT` or `curl http://[::1]:$PORT` will still work

The new fallback in this PR (no non-loopback IPv6 interface => rewrite to `0.0.0.0`) helps with the "IPv6 absent" case, but not with the dual-stack-vs-v6-only semantic differences when IPv6 exists.

This is manageable as a breaking-change note: if someone depends on IPv4 loopback checks, they should either use `localhost`/`[::1]` or set an explicit IPv4 bind.

glaszig:
i fail to see the purpose here. are people using puma on public addresses directly instead of behind proxies?

hakusaro:
@glaszig

> are people using puma on public addresses directly instead of behind proxies?

`README.md`:

> With SSL support, zero-downtime rolling restarts and a built-in request bufferer, you can deploy Puma without any reverse proxy.

I believe it is an intended use case.

PR #3852, opened 2025-12-30
ci: add `ruby 4.0` and `rails 8.1`

moozzi:
Do we need to test `{... ruby: head , rails: '8.1' }` twice, for different ubuntu versions?

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -41,7 +41,7 @@ jobs:
         os: [ ubuntu-22.04, ubuntu-24.04, ubuntu-24.04-arm,
               macos-14, macos-15, macos-15-intel, macos-26,
               windows-2022 ]
-        ruby: [ '3.0', 3.1, 3.2, 3.3, 3.4, head ]
+        ruby: [ '3.0', 3.1, 3.2, 3.3, 3.4, '4.0', head ]
         no-ssl: ['']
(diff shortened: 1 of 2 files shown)


moozzi:
`turbo-rails / ubuntu-... Ruby head Rails 8.1` fails are not related to my changes. Same things happens on [`Ruby head Rails 8.0`](https://github.com/moozzi/puma/actions/runs/20594425460/job/59145983740)

moozzi:
I was also thinking about putting all Ruby versions in single quotes for consistency.

MSP-Greg:
@moozzi

Thanks for the PR.  With updated versions, I think we can use '~> 2.7' for the SQLite3 version?  It would probably work without, but probably a good idea to limit it...

MSP-Greg:
@moozzi

Sorry, I wasn't clear.  The conditional for SQLite3 can be removed and replaced with '~> 2.7', as the code was probably old, and wasn't accounting for recent releases of Rails 7.2 (or also Ruby < 3.0).

So, the following seems to work:
```ruby
DST="gem 'sqlite3', '~> 2.7'"
```

See https://github.com/MSP-Greg/puma/commit/db871a924759e996de08066ed755e398689d09ee

moozzi:
Oh yeah, I misunderstood you. It should be correct now.

PR #3857, opened 2026-01-12
Fix flaky test test_horrible_queries

tenderlove:
`test_horrible_queries` can be flaky, but it's very hard to detect. This commit fixes the flakiness.  To understand the flakiness, let me put the original test code below:

```ruby
 # then that large mangled field values are caught
10.times do |c|
  get = "GET /#{rand_data(10,120)} HTTP/1.1\r\nX-Test: #{rand_data(1024, 1024+(c*1024), false)}\r\n\r\n"
  assert_raises Puma::HttpParserError do
    parser.execute({}, get, 0)
    parser.reset
  end
end
```

The way I understand `rand_data(xxx, false)` is that the function should _always_ return invalid bytes (IOW cause the parser to raise an exception).

However, it is possible for that function to _sometimes_ return valid bytes.  The above test asserts that an error is raised when `parser.execute` is called, but if an exception gets raised, then `parser.reset` isn't called.  If `parser.reset` isn't called, then subsequent calls to `parser.execute` will raise an exception _regardless_ of whether the input is parse-able or not.  In other words, if the first iteration caused an exception, subsequent iterations are guaranteed to raise an exception.

This commit does two things:

1. Pulls `parser.reset` outside of the assert_raises so that the flake is more obvious / reproducible
2. Changes `rand_data` to always add an invalid byte so that if the random data happens to contain parsable data, we'll at least have one byte guaranteed to fail.

### Description

### DIFF
--- a/test/test_http11.rb
+++ b/test/test_http11.rb
@@ -136,7 +136,10 @@ class Http11ParserTest < TestIntegration
     if readable
       res << Digest(:SHA1).hexdigest(rand(count * 100).to_s) * (count / 40)
     else
-      res << Digest(:SHA1).digest(rand(count * 100).to_s) * (count / 20)
+      data = Digest(:SHA1).digest(rand(count * 100).to_s) * (count / 20)
+      # Guarantee there is an invalid byte at the end of the string
+      data.setbyte(data.bytesize - 1, 0x1)
+      res << data
     end
 
     res
@@ -208,8 +211,8 @@ class Http11ParserTest < TestIntegration
     http = "GET #{path} HTTP/1.1\r\n\r\n"
     assert_raises Puma::HttpParserError do
       parser.execute(req, http, 0)
-      parser.reset
     end
+    parser.reset
   end
 
   def test_horrible_queries
@@ -220,8 +223,8 @@ class Http11ParserTest < TestIntegration
       get = "GET /#{rand_data(10,120)} HTTP/1.1\r\nX-#{rand_data(1024, 1024+(c*1024))}: Test\r\n\r\n"
       assert_raises Puma::HttpParserError do
         parser.execute({}, get, 0)
-        parser.reset
       end
+      parser.reset
     end
 
     # then that large mangled field values are caught
@@ -229,8 +232,8 @@ class Http11ParserTest < TestIntegration
       get = "GET /#{rand_data(10,120)} HTTP/1.1\r\nX-Test: #{rand_data(1024, 1024+(c*1024), false)}\r\n\r\n"
       assert_raises Puma::HttpParserError do
         parser.execute({}, get, 0)
-        parser.reset
       end
+      parser.reset
     end
 
     # then large headers are rejected too
@@ -238,16 +241,16 @@ class Http11ParserTest < TestIntegration
     get += "X-Test: test\r\n" * (80 * 1024)
     assert_raises Puma::HttpParserError do
       parser.execute({}, get, 0)
-      parser.reset
     end
+    parser.reset
 
     # finally just that random garbage gets blocked all the time
     10.times do |c|
       get = "GET #{rand_data(1024, 1024+(c*1024), false)} #{rand_data(1024, 1024+(c*1024), false)}\r\n\r\n"
       assert_raises Puma::HttpParserError do
         parser.execute({}, get, 0)
-        parser.reset
       end
+      parser.reset
     end
   end
 


tenderlove:
btw, it looks like we're seeing timeouts in Truffle?  This _could_ be because of my change.  This loop:

```ruby
10.times do |c|
  get = "GET /#{rand_data(10,120)} HTTP/1.1\r\nX-Test: #{rand_data(1024, 1024+(c*1024), false)}\r\n\r\n"
  # ....
end
```

Is generating more and more data on each iteration (because of `rand_data(1024, 1024+(c*1024), false)`).  Before my change, the first iteration would usually fail, causing subsequent iterations to just never parse that data.  Now that the data is actually parsed, it's probably (definitely) slowing tests.

tenderlove:
One question for the maintainers: is this test valuable?  I mean, is the 10th iteration any more valuable than the 1st?  Genuine question because I can't really tell (and I don't think it's my call).  Especially since before my patch usually the first iteration made subsequent iterations do nothing (unless we're testing the behavior of _not_ calling `reset`?)

schneems:
First off. Thank you for contributing! You're the best.

> btw, it looks like we're seeing timeouts in Truffle? This could be because of my change. This loop:

Just merged a fix on main. I triggered a rebase of your PR, it's green now.

> One question for the maintainers: is this test valuable? I

I have resisted touching the HTTP parser code personally. This almost looks like a cross between fuzzing and regression protection. I'm aware of fuzzing, but haven't actually invested in trying it on a codebase **for real**. Also, as I understand it, it's a lengthy, expensive process that you probably wouldn't want to run in CI. Might be some opportunity there.

Running the blame, it was added circa 2006 https://github.com/puma/puma/commit/3c804d5e15f084cd5aec5f7184dbffc1d7350951

> I mean, is the 10th iteration any more valuable than the 1st? 

IMHO, no. If you want to clean it up, split it into several **real** and distinct tests with names instead of comments, and remove the iteration, seems good. 

Since you mention it, it also doesn't look like we have a test that asserts that calling `parser.reset` after a bad parse allows you to parse a good header later. It couldn't hurt to add a regression test.

tenderlove:
> This almost looks like a cross between fuzzing and regression protection.

Yes, that's what it looks like to me.

> Also, as I understand it, it's a lengthy, expensive process that you probably wouldn't want to run in CI.

As long as the current test isn't causing timeouts, I feel it's probably fine to keep as-is.

> Since you mention it, it also doesn't look like we have a test that asserts that calling `parser.reset` after a bad parse allows you to parse a good header later. It couldn't hurt to add a regression test.

Ok.  I'll send a regression test!

[This PR](https://github.com/puma/puma/pull/3660) inspired me to try my hand at a pure Ruby implementation (which is why I'm looking at this stuff). I want to study the "real" uses of the parser inside of Puma and try to add tests around that.  For example, I suspect that internally, Puma will either always call `reset` on the parser, or just always throw the parser away and allocate a new one (in which case I'm not sure how the `reset` method would be useful).  I want to be cautious not to "over specify" the parser interface which would couple us to an API that isn't actually useful.

schneems:
The parser is used in the `Client` class. The abstractions are a little off in the internal representation, but the client roughly represents a connection, which can carry multiple requests (wrote more https://www.heroku.com/blog/upgrade-to-puma-7-and-unlock-the-power-of-fair-scheduled-keep-alive/). 

So after a request is parsed successfully then the client is reset. This happens in a loop (but only when using keepalive connections):

```ruby
        while can_loop
        # ...
          case handle_request(client, requests + 1)
          # ... 
          when :keep_alive
            requests += 1

            client.reset
```

Which clears the state of the parser among other things

```ruby
    def reset
      @parser.reset
```

Not having to allocate is a benefit of doing it this way, but at the cost of having more book-keeping and logic. In general we've leaned towards performance at the cost of complexity/maintenance (with the caveat that correctness and reliability or other guarantees take precedence over raw perf).

tenderlove:
> Not having to allocate is a benefit of doing it this way, but at the cost of having more book-keeping and logic. In general we've leaned towards performance at the cost of complexity/maintenance (with the caveat that correctness and reliability or other guarantees take precedence over raw perf).

That makes sense.  The parser I've designed is a stateless singleton (with no side effects), so there isn't a reset.  It takes an offset in to a buffer (same as the current parser), and `execute` returns the number of bytes read, so if there are multiple incoming requests to parse, the client can track the offset and pass it to `execute`.

For example (pseudo code):

```ruby
offset = 0
loop do
  req = {}
  offset = AaronParser.execute(req, buff, offset)
  # If it's a POST, the post body can be extracted from buff via the `offset` + content-length if desired
  # do stuff with req
end
```

I added a stateful wrapper around my stateless implementation only so that I can make the tests pass.  Once I get the tests passing I may send a PR, but honestly it's just for fun and possibly a ZJIT benchmark.

PR #3846, opened 2025-12-21
Update to minitest 6 + minitest-mock

zenspider:
Looks like this isn't gonna work until the minimum ruby is 3.2... you willing to consider that for the ruby 4.0 release?

PR #3859, opened 2026-01-13
Use Minitest 6 where applicable

MSP-Greg:
### Description

Minitest 6 has been released, which works with Ruby >= 3.2.  Minitest 5.26 is used for Ruby 3.0 & 3.1.

minitest-mock is now a separate gem, so it is used with Minitest 6.

Note that with Ruby 3.0 and later, Bundler is correctly selecting gems, so no 'interesting' Gemfile code is needed.

Closes #3846

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -323,22 +323,12 @@ class PumaTest < Minitest::Test # rubocop:disable Puma/TestsMustUsePumaTest
   include TestSkips
   include TestTempFile
 
-  def teardown
-    clean_tmp_paths if respond_to? :clean_tmp_paths
-  end
+  prove_it!
 
   PROJECT_ROOT = File.dirname(__dir__)
 
-  if RUBY_VERSION < "3.2"
-    def self.run(reporter, options = {}) # :nodoc:
-      prove_it!
-      super
-    end
-  else
-    def self.run(klass, method_name, reporter) # :nodoc:
-      prove_it!
-      super
-    end
+  def teardown
+    clean_tmp_paths if respond_to? :clean_tmp_paths
   end
 
   def full_name


test/helper.rb
```diff
@@ -340,4 +316,32 @@ def assert_not_called_on_instance_of(klass, method_name, message = nil, &block)
...
+    def self.run(reporter, options = {}) # :nodoc:
+      prove_it!
+      super
+    end
+  else
+    def self.run(klass, method_name, reporter) # :nodoc:
+      prove_it!
+      super
+    end
+  end
```
zenspider:
```suggestion
  prove_it!
```

minitest-proveit also had a bump today so you shouldn't need any of this. A simple call at the top of the class to `prove_it!` should suffice to activate it for all subclasses of PumaTest.

test/helper.rb
```diff
@@ -340,4 +316,32 @@ def assert_not_called_on_instance_of(klass, method_name, message = nil, &block)
...
+
+  def full_name
+    "#{self.class.name}##{name}"
```
zenspider, line 335: I've got this in a couple places too (tho I don't recall where or why)... why do YOU have it here? I'm wondering if I shouldn't push this up.

MSP-Greg: We use it in logging, nothing more.

MSP-Greg:
> get in here at the last minute

Gee, that never happens, does it?  Yes, although not related, TruffleRuby builds are stopping.  Looking at it, more tonite...

schneems:
Ci is showing red on your commit, but green on all status checks which is...odd. I'm assuming this is affected by https://github.com/puma/puma/pull/3860?

PR #3870, opened 2026-01-27
Fix flaky integration tests

nateberkopec:
I've long struggled with intermittent flakiness in the Puma test suite locally. The particular flakes have shifted over time, so instead of chasing them one by one, I had an agent run the suite 10 times and then investigate and fix any failures. This PR is the result.

I've reviewed all three commits and verified they work in my environment. While the reasoning in the third commit is a bit weaker, I still believe the change itself is sound. I'd like to include all three changes.

### DIFF
--- a/test/helpers/integration.rb
+++ b/test/helpers/integration.rb
@@ -249,7 +249,7 @@ class TestIntegration < PumaTest
     rescue Errno::EADDRNOTAVAIL => e
       raise e if Process.clock_gettime(Process::CLOCK_MONOTONIC) >= deadline
       retries += 1
-      sleep 0.01 * [retries, 10].min
+      sleep 0.01 * retries.clamp(0, 10)
       retry
     end
   end


nateberkopec:
This is from a session with me and Codex just trying to get the test suite running 10x in a row locally with no fails. Vetting the changes next.

test/helpers/integration.rb
```diff
@@ -241,8 +241,21 @@ def server_gets(match_obj, time_timeout, log: false)
...
 
+  def open_client_socket(unix: false, timeout: 3)
+    deadline = Process.clock_gettime(Process::CLOCK_MONOTONIC) + timeout
+    retries = 0
+    begin
+      unix ? UNIXSocket.new(@bind_path) : TCPSocket.new(HOST, @tcp_port)
+    rescue Errno::EADDRNOTAVAIL => e
+      raise e if Process.clock_gettime(Process::CLOCK_MONOTONIC) >= deadline
+      retries += 1
+      sleep 0.01 * [retries, 10].min
```
MSP-Greg:
```suggestion
      sleep 0.01 * retries.clamp(0, 10)
```

PR #3412, opened 2024-06-20
Add support for EXTEND_TIMEOUT_USEC systemd

lkibbalam:
### Description
Closes https://github.com/puma/puma/issues/3234
add support for the EXTEND_TIMEOUT_USEC systemd. It is sent once at the beginning of launch. Please confirm if there's a need to continuously send EXTEND_TIMEOUT_USEC in a loop during runtime, although it doesn't seem necessary.

### Your checklist for this pull request

- [ ] I have reviewed the [guidelines for contributing](../blob/master/CONTRIBUTING.md) to this repository.
- [ ] I have added (or updated) appropriate tests if this PR fixes a bug or adds a feature.
- [ ] My pull request is 100 lines added/removed or less so that it can be easily reviewed.
- [ ] If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.
- [ ] If this closes any issues, I have added "Closes `#issue`" to the PR description or my commit messages.
- [ ] I have updated the documentation accordingly.
- [ ] All new and existing tests passed, including Rubocop.

PR #3871, opened 2026-01-27
Extend systemd timeout until ready (follow-up to #3412)

### LINKED ISSUE #3234: Add support for `EXTEND_TIMEOUT_USEC` to systemd notification support

Using puma in GitLab and running puma as a systemd service with type `notify`  I ran into startup issues because preloading the application took longer than the default service start timeout of 90s. 

It would be great if puma could write `EXTEND_TIMEOUT_USEC=` with some integer value regularly to the notification socket before signaling `READY=1` and before running into the timeout to extend its startup time.

The issue can be easily worked around on a case by case basis by giving puma more time to start in the service file, i.e. setting `TimeoutStartSec=` to a higher value

nateberkopec:
## Summary
- Extend systemd startup timeout notifications with a 5s buffer and max cap
- Add tests for extend-timeout messages and env parsing edge cases
- Document EXTEND_TIMEOUT_USEC/EXTEND_TIMEOUT_MAX_USEC usage

## Context
I like the original PR (#3412) by @lkibbalam, but I had some concerns about the behavior, so I've added a bit here. Hopefully this all works. I'm not a systemd user, but I think this is a good implementation.

Supersedes #3412.
Fixes #3234.

### DIFF
--- a/docs/systemd.md
+++ b/docs/systemd.md
@@ -62,6 +62,17 @@ Restart=always
 WantedBy=multi-user.target
 ~~~~
 
+### Startup timeout extension
+
+If your app can take longer than `TimeoutStartSec` to boot, Puma can ask systemd
+for more time.
+
+Set `EXTEND_TIMEOUT_USEC` to the extension interval in microseconds. Puma sends
+an extension every half interval until it reports `READY=1`.
+
+Set `EXTEND_TIMEOUT_MAX_USEC` to cap the total extra time. If you omit it, Puma
+uses the same value as `EXTEND_TIMEOUT_USEC`.
+
 See
 [systemd.exec](https://www.freedesktop.org/software/systemd/man/systemd.exec.html)
 for additional details.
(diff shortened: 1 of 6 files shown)


nateberkopec:
@lkibbalam would you mind taking a look when you have time?

nateberkopec:
Fixes pushed to systemd-extend-timeout:

- Stop global plugin registry pollution from `test/test_plugin_systemd.rb` by loading the systemd plugin only inside `TestPluginSystemdUnit#setup`, snapshotting/restoring `Puma::Plugins` (and background hooks), and disabling `parallelize_me!` to avoid cross-thread state bleed. This addresses `TestCLI#test_plugins`, which asserts the registry is empty before CLI plugin loading.
- Start the EXTEND_TIMEOUT_USEC loop immediately inside the systemd plugin `start` method (dedicated thread with `Puma.set_thread_name`) instead of waiting for `Plugins.fire_background`. `TestPluginSystemd#test_systemd_extend_timeout_notify` demonstrates the issue: with background threads firing after `load_and_bind`, the test sees `READY=1` before any EXTEND_TIMEOUT_USEC messages, meaning we weren't extending during Rack app boot.

nateberkopec:
* [ ] I think the env var may be the wrong idea. if anything it needs to be prefixed with `PUMA_`
* [ ] Maybe we don't need a max
* [ ] The 5 second buffer thing instead of halving is probably too big brain

Issue #3250, opened 2023-10-05
Puma forces casing of certain response headers


skipkayhil:
**Describe the bug**

While most headers have casing determined by what's in the app's response headers, Puma forces the casing of certain headers in the response it returns. When using Rack 2, the forced `Content-Length` type casing was likely not noticed because (most) apps would use this for their header casing. As Rails is close to a release that supports Rack 3, this behavior will likely be surprising to users who expect the same casing in the response returned by Puma as is returned by their app.

**Puma config:**

No config, just `rackup` with the `config.ru` below

**To Reproduce**

```ruby
# config.ru
App = ->(e) { [200, { "content-type" => "text/plain", "content-length" => "5" }, ["12345"]] }

run App
```

Run it with:

```
$ rackup
```

```
$ curl -I localhost:9292
HTTP/1.1 200 OK
content-type: text/plain
Content-Length: 5
```

**Expected behavior**

```
$ curl -I localhost:9292
HTTP/1.1 200 OK
content-type: text/plain
content-length: 5
```

**Desktop (please complete the following information):**
 - OS: any
 - Puma Version 6.4.0

MSP-Greg:
I haven't been in the various specs for several weeks.  See https://www.rfc-editor.org/rfc/rfc9110.html#name-content-length.  The spec also refers to [Transfer-Encoding](https://www.rfc-editor.org/rfc/rfc9112#field.transfer-encoding), etc.

The [Rack spec](https://msp-greg.github.io/rack/file.SPEC.html#label-The+Headers) specifies that header field names/keys should be lower case, but that isn't referring to the HTTP syntax, it's referring to what the app returns.

Also, most browsers' 'dev tools' show capitalized header field names...

skipkayhil:
Right, my understanding is either casing is valid in HTTP/1.1 spec, and the Rack spec doesn't actually say anything about how the server should respond.

However, I believe the current behavior is confusing (at least for me, and the user that opened an issue on Rails) because _only certain headers_ have their casing changed. With a Rack 2 app, both headers will be mixed case in the app and the Puma response will have all mixed case headers. With a Rack 3 app, all headers in the app will be downcased, but only _some_ of the headers in the Puma response will be downcased.

[...]

PR #3874, opened 2026-01-27
Perf: Cache downcased header key in str_headers

hadrienblanc:
### Description

**What original problem led to this PR?**

PR #3704 added `k.downcase` calls in `str_headers` for Rack 3 compliance. 
However, since we already call `k.downcase` in the case statement, we now 
call it twice per header instead of once.

**Are there related issues / prior discussions?**

Follow-up optimization to PR #3704 (merged in v7.0.0) and issue #3250.

**What alternatives have been tried?**

Considered naming alternatives (`k_lwr`, `lower_k`) but settled on `key` 
for readability.

**Why do you make the choices you did? What are the tradeoffs?**

- Use a local variable `key` to save the downcase result
- Zero behavior change
- Instead of calling `downcase` twice per header, we call it once. I did a `GC.stat` measurement and a micro-benchmark below.

---

#### Memory allocations

```ruby
# 10 headers × 1000 iterations
GC.stat[:total_allocated_objects]

BEFORE: 20,004 allocations
AFTER:  10,004 allocations
GAIN:   10,000 allocations (50%)
```

Each `String#downcase` call creates a new String object. With ~10 headers 
per response, this saves ~10 String allocations per request.

#### Micro-benchmark

```
BEFORE (2 downcase)  0.595148   0.003618   0.598766 (  0.674s)
AFTER  (1 downcase)  0.442885   0.002580   0.445465 (  0.490s)
```

In my runs, it was faster on the header loop (~27% in this instance), but I 
wouldn't draw strong conclusions in terms of exact metrics.

---

### DIFF
--- a/lib/puma/request.rb
+++ b/lib/puma/request.rb
@@ -652,7 +652,8 @@ module Puma
       headers.each do |k, vs|
         next if illegal_header_key?(k)
 
-        case k.downcase
+        key = k.downcase
+        case key
         when CONTENT_LENGTH2
           next if illegal_header_value?(vs)
           # nil.to_i is 0, nil&.to_i is nil
@@ -679,10 +680,10 @@ module Puma
         if ary
           ary.each do |v|
             next if illegal_header_value?(v)
-            io_buffer.append k.downcase, colon, v, line_ending
+            io_buffer.append key, colon, v, line_ending
           end
         else
-          io_buffer.append k.downcase, colon, line_ending
+          io_buffer.append key, colon, line_ending
         end
       end
 


nateberkopec:
Hello, this is Codex speaking.

Benchmarks (Docker):
- Built from `tools/Dockerfile` (ruby 4.0.1 aarch64, `bundle exec rake compile`)
- Installed `wrk` in-container
- Script: `benchmarks/local/response_time_wrk.sh -w2 -t5:5 -s tcp` (wrk `-t8 -c16 -d10s`)
- Note: container-local patch to avoid `TypeError` in `benchmarks/local/response_time_wrk.rb` (replace `@threads[/\d+\z/] || 5` with `@threads || 5`). No repo changes.

Avg RPS (mean across array/chunk/string/io)
```
Size(kB)   main avg   PR avg   delta
1          26,255     26,839   +2.2%
10         25,838     26,783   +3.7%
100        25,765     26,664   +3.5%
256        26,360     26,509   +0.6%
512        26,349     26,379   +0.1%
1024       26,053     26,478   +1.6%
2048       25,727     26,755   +4.0%
```

Latency notes
- 50% latency slightly lower in most rows (~0.01–0.03ms)
- One outlier: 256kB chunked 100% latency spike in PR run (31.88ms vs 5.57ms on main); other percentiles looked stable

nateberkopec:
Thanks for the optimization and the detailed context here!

nateberkopec:
https://opncd.ai/share/RfDREhZQ

PR #3876, opened 2026-01-28
Increase timeout for Intel JRuby tests to 12 minutes

nateberkopec:


nateberkopec:
Seeing this one timeout a number of times over last 2 weeks. Intel on JRuby is kind of a worst case scenario so a minor bump seems fine.

PR #3912, opened 2026-03-26
[ci skip] Minor correction to defaults documented in DSL

Be-ngt-oH:
### Description

I was looking for the default for `persistent_timeout` and, coincidentally, looked at the docs for the configuration DSL first. It had it at 20s which surprised me as I was expecting a value higher than 60s (which is the default for load balancers in some cloud providers). As it appears, the default value is actually 65s.

I used the opportunity to check for more discrepancies between the DSL docs and the actual configuration but could only find another one which is that `worker_boot_timeout` is derived while this is not the case anymore.

Not a big contribution but since it tripped me up, I thought it's useful to correct.

### DIFF
--- a/lib/puma/dsl.rb
+++ b/lib/puma/dsl.rb
@@ -350,7 +350,7 @@ module Puma
 
     # Define how long persistent connections can be idle before Puma closes them.
     #
-    # The default is 20 seconds.
+    # The default is 65 seconds.
     #
     # @example
     #   persistent_timeout 30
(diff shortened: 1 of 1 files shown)


nateberkopec (approved):
Thanks for the docs cleanup. Verified both defaults against Configuration::DEFAULTS ( = 65,  = 60). LGTM.

nateberkopec (commented):
Tiny correction to my previous note: verified against Configuration::DEFAULTS: persistent_timeout = 65, worker_boot_timeout = 60.

PR #3916, opened 2026-03-31
Fix IPv4-mapped IPv6 addresses in `REMOTE_ADDR` and request logs

dentarg:
Strip the `::ffff:` prefix from IPv4-mapped IPv6 addresses returned by dual-stack sockets so that `REMOTE_ADDR` and `CommonLogger` output show the original IPv4 address (e.g. `127.0.0.1` instead of `::ffff:127.0.0.1`).

This is a side-effect of #3847 which changed the default bind to `::` on hosts with IPv6 interfaces.

---

### DIFF
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -352,7 +352,7 @@ module Puma
         return hdr
       end
 
-      @peerip ||= @io.peeraddr.last
+      @peerip ||= unmap_ipv6(@io.peeraddr.last)
     end
 
     def peer_family
@@ -386,6 +386,21 @@ module Puma
     end
 
     private
+
+    IPV4_MAPPED_IPV6_PREFIX = "::ffff:"
+    IPV4_MAPPED_IPV6_PREFIX_LENGTH = IPV4_MAPPED_IPV6_PREFIX.length
+    private_constant :IPV4_MAPPED_IPV6_PREFIX, :IPV4_MAPPED_IPV6_PREFIX_LENGTH
+
+    # Converts IPv4-mapped IPv6 addresses (e.g. ::ffff:127.0.0.1) back to
+    # their IPv4 form. These addresses appear when IPv4 clients connect to
+    # a dual-stack IPv6 socket.
+    def unmap_ipv6(addr)
+      if addr.start_with?(IPV4_MAPPED_IPV6_PREFIX)
+        addr[IPV4_MAPPED_IPV6_PREFIX_LENGTH..]
+      else
+        addr
+      end
+    end
 
     # Checks the request `Transfer-Encoding` and/or `Content-Length` to see if
     # they are valid.  Raises errors if not, otherwise reads the body.
--- a/lib/puma/client.rb
+++ b/lib/puma/client.rb
@@ -347,12 +347,12 @@ module Puma
       return @peerip if @peerip
 
       if @remote_addr_header
-        hdr = (@env[@remote_addr_header] || @io.peeraddr.last).split(/[\s,]/).first
+        hdr = (@env[@remote_addr_header] || socket_peerip).split(/[\s,]/).first
         @peerip = hdr
         return hdr
       end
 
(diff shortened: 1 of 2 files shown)


dentarg:
[My conversation with Gemini around this topic and this PR](https://stack.pizza/s/cuwzoq/ipv4-mapped-ipv6-addresses-explained) and [the Claude session](https://gist.github.com/dentarg/7b8590d2f26df0bb22c32f382aa26f69)

lib/puma/client.rb
```diff
@@ -352,7 +352,7 @@ def peerip
         return hdr
       end
 
-      @peerip ||= @io.peeraddr.last
+      @peerip ||= unmap_ipv6(@io.peeraddr.last)
```
joshuay03: Does line 350 need addressing too? See the raw `@io.peeraddr.last` fallback. If so, should we abstract this out into a private method?

lib/puma/client.rb
```diff
@@ -387,6 +387,21 @@ def expect_proxy_proto=(val)
...
+
+    # Converts IPv4-mapped IPv6 addresses (e.g. ::ffff:127.0.0.1) back to
+    # their IPv4 form. These addresses appear when IPv4 clients connect to
+    # a dual-stack IPv6 socket.
+    def unmap_ipv6(addr)
+      if addr.start_with?(IPV4_MAPPED_IPV6_PREFIX)
+        addr[IPV4_MAPPED_IPV6_PREFIX_LENGTH..]
+      else
+        addr
+      end
```
joshuay03: Can we use `delete_prefix` here?


```suggestion
      addr.delete_prefix(IPV4_MAPPED_IPV6_PREFIX)
```

And get rid of `IPV4_MAPPED_IPV6_PREFIX_LENGTH`.

test/test_request_single.rb
```diff
@@ -726,3 +726,39 @@ def test_reset_clears_error_status_code
     assert_nil @client.error_status_code
   end
 end
+
+class TestPeerip < TestRequestBase
```
joshuay03: Can we keep the class name consistent with most of the others?


```suggestion
class TestRequestPeerip < TestRequestBase
```

test/test_request_single.rb
```diff
@@ -726,3 +726,39 @@ def test_reset_clears_error_status_code
...
+
+  def test_peerip_unmaps_ipv4_mapped_ipv6
+    peer_addr = -> () { ["AF_INET6", 80, "::ffff:127.0.0.1", "::ffff:127.0.0.1"] }
+    create_client("GET / HTTP/1.1\r\n\r\n") {
```
joshuay03: Nit. For this and the other new tests below:


```suggestion
    create_client("GET / HTTP/1.1\r\n\r\n") { |_client|
```

test/test_request_single.rb
```diff
@@ -726,3 +726,39 @@ def test_reset_clears_error_status_code
...
+      @rd.define_singleton_method(:peeraddr, peer_addr)
+    }
+
+    assert_equal "127.0.0.1", @client.peerip
+  end
+
+  def test_remote_addr_unmaps_ipv4_mapped_ipv6
```
joshuay03: This tests the same code path as `test_peerip_unmaps_ipv4_mapped_ipv6` since `REMOTE_ADDR` is set via `peerip` in `client_env.rb`. Worth collapsing into one test and using the freed slot to cover the untested `@remote_addr_header` fallback in `peerip` (when the header is configured but missing from `env`, it calls `@io.peeraddr.last` without unmapping).

test/test_request_single.rb
joshuay03: Might be worth updating `TestTransferEncodingInvalid` for consistency as well.

nateberkopec:
Resolved all of @joshuay03's comments

joshuay03 (approved):
:shipit:

PR #3917, opened 2026-04-01
systemd: send immediate watchdog ping on ready

### LINKED ISSUE #3675: Systemd unnecessarily restarts Puma after succesful pumactl restart

wwenrr:
## Summary
When watchdog mode is enabled, Puma currently sends `READY=1` at boot and only sends `WATCHDOG=1` on the next watchdog loop tick.

This can leave a small timeout window after restart/startup where systemd may kill Puma before the first periodic watchdog ping is emitted (as described in #3675).

This PR sends an immediate watchdog ping from `SdNotify.ready` when watchdog mode is active.

## Changes
- `lib/puma/sd_notify.rb`
  - Update `SdNotify.ready` to call `watchdog` if `watchdog?` is true
- `test/test_plugin_systemd.rb`
  - Add regression test `test_systemd_watchdog_sends_immediate_ping_after_ready`

### Automated
Ran systemd plugin tests focused on watchdog behavior:

```bash
bundle exec ruby test/test_plugin_systemd.rb -n "/test_systemd_watchdog|test_systemd_watchdog_sends_immediate_ping_after_ready/"
```

Result: `2 runs, 6 assertions, 0 failures`

### Manual
In a local Docker dev environment, simulated `NOTIFY_SOCKET` + `WATCHDOG_USEC` and invoked `Puma::SdNotify.ready`; confirmed messages are emitted in order:
1. `READY=1`
2. `WATCHDOG=1`

Fixes #3675

### DIFF
--- a/lib/puma/sd_notify.rb
+++ b/lib/puma/sd_notify.rb
@@ -49,8 +49,12 @@ module Puma
     WATCHDOG  = "WATCHDOG=1"
     FDSTORE   = "FDSTORE=1"
 
+    # @note If watchdog notifications are enabled, this also emits an
+    # immediate `WATCHDOG=1` ping to avoid a watchdog timeout window between
+    # startup completion and the first periodic watchdog loop tick.
     def self.ready(unset_env=false)
       notify(READY, unset_env)
+      watchdog(unset_env) if watchdog?
     end
 
     def self.reloading(unset_env=false)
--- a/lib/puma/sd_notify.rb
+++ b/lib/puma/sd_notify.rb
@@ -53,8 +53,12 @@ module Puma
     # immediate `WATCHDOG=1` ping to avoid a watchdog timeout window between
     # startup completion and the first periodic watchdog loop tick.
     def self.ready(unset_env=false)
-      notify(READY, unset_env)
-      watchdog(unset_env) if watchdog?
+      if watchdog?
+        notify(READY)
+        watchdog(unset_env)
+      else
+        notify(READY, unset_env)
+      end
     end
 
     def self.reloading(unset_env=false)
--- a/test/test_plugin_systemd.rb
+++ b/test/test_plugin_systemd.rb
@@ -54,6 +54,17 @@ class TestPluginSystemd < TestIntegration
     assert_message "STOPPING=1"
   end
 
+  def test_systemd_watchdog_sends_immediate_ping_after_ready
+    wd_env = @env.merge({"WATCHDOG_USEC" => "1_000_000"})
+    cli_server "test/rackup/hello.ru", env: wd_env
+
+    assert_message "READY=1"
+    assert_message "WATCHDOG=1"
+
+    stop_server
+    assert_message "STOPPING=1"
+  end
+
   def test_systemd_notify
     cli_server "test/rackup/hello.ru", env: @env
     assert_message "READY=1"
--- a/test/test_plugin_systemd.rb
+++ b/test/test_plugin_systemd.rb
@@ -55,7 +55,7 @@ class TestPluginSystemd < TestIntegration
   end
 
   def test_systemd_watchdog_sends_immediate_ping_after_ready
-    wd_env = @env.merge({"WATCHDOG_USEC" => "1_000_000"})
+    wd_env = @env.merge({"WATCHDOG_USEC" => "10_000_000"})
     cli_server "test/rackup/hello.ru", env: wd_env
 
     assert_message "READY=1"
--- a/test/test_sd_notify.rb
+++ b/test/test_sd_notify.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+require_relative "helper"
+require_relative "helpers/tmp_path"
+require_relative "../lib/puma/sd_notify"
+
+class TestSdNotify < PumaTest
+  include TmpPath
+
+  def setup
+    super
+
+    @sockaddr = tmp_path '.sd_notify'
+    @socket = Socket.new(:UNIX, :DGRAM, 0)
+    @socket.bind Addrinfo.unix(@sockaddr)
+    @message = +''
+  end
+
+  def teardown
+    @socket&.close
+    @socket = nil
+
+    super
+  end
+
+  def test_ready_with_watchdog_unsets_env_after_watchdog
+    with_temp_env("NOTIFY_SOCKET" => @sockaddr, "WATCHDOG_USEC" => "10_000_000") do
+      Puma::SdNotify.ready(true)
+
+      assert_message "READY=1"
+      assert_message "WATCHDOG=1"
+      refute ENV.key?("NOTIFY_SOCKET")
+    end
+  end
+
+  private
+
+  def assert_message(msg)
+    @socket.wait_readable 1
+    @message << @socket.sysread(msg.bytesize)
+    assert_equal msg, @message
+    @message = +''
+  end
+end


wwenrr:
Checked failing jobs: both failures are in unrelated integration tests (`TestIntegrationCluster#test_stuck_phased_restart` on ubuntu-22.04 3.2 and `TestPumaServer#test_update_thread_pool_min_max` on macos-15-intel 3.0), not in the new systemd codepath. All other matrix jobs pass. This looks flaky/environmental; please re-run the two failed jobs.

rud (approved):
Sounds reasonable

lib/puma/sd_notify.rb
```diff
@@ -49,8 +49,12 @@ class NotifyError < RuntimeError; end
     WATCHDOG  = "WATCHDOG=1"
     FDSTORE   = "FDSTORE=1"
 
+    # @note If watchdog notifications are enabled, this also emits an
+    # immediate `WATCHDOG=1` ping to avoid a watchdog timeout window between
+    # startup completion and the first periodic watchdog loop tick.
     def self.ready(unset_env=false)
       notify(READY, unset_env)
+      watchdog(unset_env) if watchdog?
```
joshuay03: Small edge case: if `ready` is called with `unset_env=true`, `notify(READY, unset_env)` deletes `NOTIFY_SOCKET` from the environment before `watchdog` gets a chance to use it. Since `notify` does `sock = ENV["NOTIFY_SOCKET"]` as its first step, the watchdog call then hits `return nil if !sock` and the ping is silently dropped.

The sd_notify spec explicitly supports newline-separated key=value pairs in a single datagram, so this could be fixed cleanly with something like:

```ruby
def self.ready(unset_env=false)
  msg = watchdog? ? "#{READY}\n#{WATCHDOG}" : READY
  notify(msg, unset_env)
end
```

That way `unset_env` only fires once, after both values are already written, and the send is atomic. The test helper would need to handle the combined datagram but that seems worth it.

test/test_plugin_systemd.rb
```diff
@@ -54,6 +54,17 @@ def test_systemd_watchdog
...
+  def test_systemd_watchdog_sends_immediate_ping_after_ready
+    wd_env = @env.merge({"WATCHDOG_USEC" => "1_000_000"})
+    cli_server "test/rackup/hello.ru", env: wd_env
+
+    assert_message "READY=1"
+    assert_message "WATCHDOG=1"
+
+    stop_server
+    assert_message "STOPPING=1"
+  end
```
joshuay03, line 66: This test is identical to `test_systemd_watchdog` above and would have passed even before your fix. With `WATCHDOG_USEC` set to 1 second the periodic watchdog fires at around 500ms, which is well within `assert_message`'s 1 second `wait_readable` timeout, so there is no way to tell whether the `WATCHDOG=1` you are asserting came from the immediate ping or the periodic loop.

Setting `WATCHDOG_USEC` to something large like `"60000000"` (60 seconds) would fix this. The periodic loop would not fire for ~30 seconds, so `assert_message "WATCHDOG=1"` immediately after `READY=1` would only succeed if the ping was sent right away. Without your fix it should timeout and fail.

nateberkopec (changes requested):
Concur with @joshuay03, please address his feedback 👍

gilesbowkett:
@nateberkopec I took a swing at addressing the feedback in https://github.com/puma/puma/pull/3980

wwenrr:
Thanks again for the review. I pushed the follow-up fix in `13f9bf58c3dfac464a1deb22e06ffc6eb92b27f6`.

This keeps `NOTIFY_SOCKET` available until after the immediate watchdog ping when `ready(true)` is used, and updates the watchdog test so the immediate ping is not masked by the periodic loop.

PR #3920, opened 2026-04-09
Use blocks for debug logging to avoid creating messages if debug disabled

p8:
### Description

If debugging is disabled, the following will still generate the message to log.

```ruby
Puma::LogWriter.stdio.debug("some expensive message")
```

By using a block the messages only need to be created when debug is enabled, saving some allocations:

```ruby
Puma::LogWriter.stdio.debug { "some expensive message" }
```

Passing blocks to the logger is also [implemented](https://github.com/ruby/logger/blob/4b19af64f502adb2396b597a99b53eed1fd03449/lib/logger.rb#L698) in the default Ruby Logger and is [used](https://github.com/rails/rails/blob/0b5af58d2f0b8916f8211fa620c48da71d6da657/actionpack/lib/action_controller/log_subscriber.rb#L26-L44) in frameworks like Rails.

There are still some other logger calls like `LogWriter#log` that could also benefit from blocks, but these don't seem tied to a log severity level.
I've also skipped some debug calls in rescue blocks, those don't seem to be on critical paths.

### DIFF
--- a/lib/puma/binder.rb
+++ b/lib/puma/binder.rb
@@ -90,7 +90,7 @@ module Puma
     # @version 5.0.0
     #
     def create_activated_fds(env_hash)
-      @log_writer.debug "ENV['LISTEN_FDS'] #{@env['LISTEN_FDS'].inspect}  env_hash['LISTEN_PID'] #{env_hash['LISTEN_PID'].inspect}"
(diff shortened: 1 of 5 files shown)


joshuay03 (approved):
Thanks!

p8:
Thanks @joshuay03 !

PR #3922, opened 2026-04-09
cache downcasing for lifetime of object

jjb:
if https://github.com/puma/puma/pull/3874 was beneficial... would it also be beneficial to cache them permanently?

### DIFF
--- a/lib/puma/response.rb
+++ b/lib/puma/response.rb
@@ -1,6 +1,14 @@
 # frozen_string_literal: true
 
-module Puma
+# for demonstration purposes in this PR
+$downcase_cache = {}
+def downcase_from_cache(s)
+  return $downcase_cache[s] if $downcase_cache[s]
+  $downcase_cache[s] = s.downcase
+  $downcase_cache[s]
+end
+
+  module Puma
   #———————————————————————— DO NOT USE — this class is for internal use only ———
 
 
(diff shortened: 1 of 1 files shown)


jjb:
no. 😆 

```
➔ ruby code.rb
ruby 3.3.10 (2025-10-23 revision 343ea05002) [arm64-darwin25]
       without cache   241.000 i/100ms
          with cache   373.000 i/100ms
       without cache      3.756k (± 4.6%) i/s  (266.24 μs/i) -      3.856k in   1.029055s
          with cache      3.744k (± 0.5%) i/s  (267.13 μs/i) -      4.103k in   1.096046s

Comparison:
       without cache:     3756.0 i/s
          with cache:     3743.5 i/s - same-ish: difference falls within error
```

```ruby
require 'benchmark/ips'
require 'securerandom'

$downcase_cache = {}
def downcase_from_cache(s)
  $downcase_cache[s] ||= s.downcase
end

$strings = []
300.times{ $strings << SecureRandom.alphanumeric.upcase}


Benchmark.ips do |x|
  x.config(warmup: 2, time: 1)

  x.report("without cache") do
    10.times{$strings.each{|s| s.downcase}}
  end

  x.report("with cache") do
    10.times{$strings.each{|s| downcase_from_cache(s)}}
  end

  x.compare!
end
```

Issue #3556, opened 2024-11-18
Puma's header normalization fixed not properly


dongvv-2538:
**Describe the bug**
related to the commit https://github.com/puma/puma/commit/f196b23be24712fb8fb16051cc124798cc84f70e. I understand that this will fix header like `X-Forwarded_For` will be invalidate.

This work correctly if there are more than 1 header in format `X-Forwarded_For` , `X-Forwarded-For`.
However, if there's no valid `X-Forwarded-For` header, the `HTTP_X_FORWARDED_FOR` value will still be set to value of `X-Forwarded_For`
**Puma config:**
puma v5.6.9

You can try with this test, this will be failed because `HTTP_X_FORWARDED_FOR = 2.2.2.2`

```
  def test_underscore_header_4
    hdrs = [
      "X-FORWARDED_FOR: 2.2.2.2",  # invalid, contains underscore
      "Content-Length: 5",
    ].join "\r\n"

    response = send_http_and_read "#{GET_PREFIX}#{hdrs}\r\n\r\nHello\r\n\r\n"

    assert_includes response, "HTTP_X_FORWARDED_FOR = []"
    refute_includes response, "2.2.2.2"
  end
```
**To Reproduce**
step 1: Go to web 
step 2: using some tools like BurpSuite proxy or using javascript to add invalid header like `X-Forwarded_For` or `X_Forwarded-For`
step 3: check the value of `HTTP_X_FORWARDED_FOR`

**Expected behavior**
The invalid headers should not be set as value of `HTTP_X_FORWARDED_FOR`

**Desktop (please complete the following information):**
 - OS: [e.g. Mac, Linux]
 - Puma Version [e.g. 4.1.1]

dentarg:
Yeah, we heard from at least one user about underscores, so defaulting to rejecting them is probably out of question. https://github.com/puma/puma/issues/3502#issuecomment-2402932875

dentarg:
Seeing https://stackoverflow.com/questions/22856136/why-do-http-servers-forbid-underscores-in-http-header-names and reading the comments makes me think this should not be the default, even in a new major version.

Should we even consider the feature?

Maybe time is better spent on https://github.com/puma/puma/issues/3397

dentarg:
> A better solution would be to expose the underlying headers in their raw form, and I'd be strongly in favour of normalized according to the RFC, which is lower case.

That sounds more like https://github.com/puma/puma/issues/3397

So let's close this issue.

Issue #3502, opened 2024-09-25
Puma merges request headers together by overwriting


jamiestrachan-shopify:
**Describe the bug**
Puma appears to be merging request headers with dash and underscore together, by overwriting.
This seems to pick one at random.

For example, the request headers `X_MYHEADER:1` and `X-MYHEADER:2` will result in only `HTTP_X_MYHEADER = 2` in the handler environment.

This bug has some interesting, possibly serious implications, such as being able to override headers set by a load balancer merely by including a similar header in the http request.

**Puma config:**
This config creates an app that simply echoes back the request headers that start with `HTTP_X`:
```
# frozen_string_literal: true

# Extract "X_" headers and print one per line
def headers(env)
  env.filter { |k, _| k.upcase.start_with?("HTTP_X") }.map{ |k,v| "#{k}: #{v}" }.join("\n")
end

App = ->(e) { [200, { "content-type" => "text/plain" }, [headers(e)]] }

run App
```

Command line:
```
rackup
```

**To Reproduce**
* Run rackup
* Post a request with curl that contains similar headers, such as `X-MyHeader` and `X_MyHeader`
* Observe the response only contains one header value.

Examples:
* underscore first:
```
curl localhost:9292 -H "X_MyHeader: false" -H "x-myheader: true"

HTTP_X_MYHEADER: true
```

* dash first:
```
curl localhost:9292 -H "X-MyHeader: false" -H "x_myheader: true"

HTTP_X_MYHEADER: false
```

* what is even happening here:
```
curl localhost:9292 -H "X_My-Header: false" -H "x_my_header: true"

HTTP_X_MY_HEADER: true


curl localhost:9292 -H "X_My-Header: false" -H "x_my_header: true" -H "X-my-header: filenotfound"

HTTP_X_MY_HEADER: filenotfound
```

* anti-example showing two headers with different casing
```
curl localhost:9292 -H "X_MyHeader: false" -H "x_myheader: true"

HTTP_X_MYHEADER: false, true
```

**Expected behavior**
I should probably expect the headers to be combined as if they both contained underscore, as in the anti-example.

**Desktop (please complete the following information):**
 - Puma Version 6.4.3

**Workaround**
Don't use any dashes or underscores in your header names!

MSP-Greg:
I don't think Puma 4 is supported anymore.

jamiestrachan-shopify:
Whoops.  Just checked my gemfile, it's 6.4.3. Updating...

[...]

Issue #3397, opened 2024-06-02
Any way of receiving headers as an array instead of a string concatenated with commas?


heyvito:
Hi there, folks! <3

**Is your feature request related to a problem? Please describe.**
Not sure. I'd like to receive my `HTTP_` headers as arrays instead of strings concatenated with commas.

**Describe the solution you'd like**
I'd like to have any way to instruct Puma to yield arrays of headers, instead of the default concatenated string. I'm not sure, though, if this would be a clear violation of the Rack spec.

**Describe alternatives you've considered**
I thought about implementing a middleware, or handling this on my application, but the header structure is lost in the parsing process, leaving little to no space to even infer the original value of those headers.

**Additional context**
For instance, consider a request that uses the following example headers:

```http
Foo: Bar
Foo: Baz, Fuz
Foo: A, B, C
```

The header I get in the application would be:

```ruby
{
  "HTTP_FOO": "Bar, Baz, Fuz, A, B, C",
}
```

There's no way to check which header is which, since all became strings with the same formatting.
Instead, it would be great if we could somehow receive them in the following form:

```ruby
{
  "HTTP_FOO": ["Bar", "Baz, Fuz", "A, B, C"],
}
```

If this aligns with the objectives of the Puma project, I can open a PR and handle any required spec, but I though about asking before just going on and opening a PR out of the blue. Specially since this may touch the configuration set, and again, I'm not sure if that would be beneficial in any way.

dentarg:
Do you know if this is possible with other web servers?

heyvito:
Hi Patrik! Answering your questions:

> What's the use case for this? I think that's interesting context

It is relevant in case we need to parse any custom header that already uses commas as a separator. I don't think that's common, but I just hit that limitation implementing an internal mechanism here on my end.

> Do you know if this is possible with other web servers?

I do! Go provides a `Header` type (that's just a hash associating strings to arrays of strings). You can see that in action here: https://go.dev/play/p/dN-uw1C5tKC. The output will be something like: 
```go
http.Header{
    "Foo": []string{
        "Bar", 
        "Baz", 
        "Foo, Zing"
    }
}
```

I made some changes on a fork, and sadly Rack's Linter was not happy with those changes. :/

MSP-Greg:
> Rack's Linter was not happy with those changes

I thought that might be a possibility...

[...]

PR #3937, opened 2026-05-01
Add underscore header hardening option

nateberkopec:
## Summary

Many Rack applications trust `X-Forwarded-*` headers, such as `X-Forwarded-For` and `X-Forwarded-Host`, because they believe those headers are set or sanitized by a trusted upstream proxy/CDN.

Rack env normalization makes hyphenated and underscored request header names indistinguishable:

```text
X-Forwarded-Host  -> HTTP_X_FORWARDED_HOST
X_Forwarded_Host  -> HTTP_X_FORWARDED_HOST
```

Puma already protects the collision case by preserving underscore-origin headers internally and preferring the canonical hyphenated header when both forms reach Puma. The remaining problem is the lone-underscore case. If an upstream proxy fails to write the canonical `X-Forwarded-*` header for some request path, or fails to remove a client-supplied underscore variant, Puma can normalize a client-controlled header into the trusted Rack env key. To the app, it can look like the trusted upstream set `env["HTTP_X_FORWARDED_HOST"]` to `evil.example`, when in fact the upstream did not set the canonical header at all.

This PR adds nginx-like underscore header handling so deployments can opt into the safer behavior now.

## Changes

- Add `allow_underscore_headers`, defaulting to `true` for compatibility in this major version.
- When `allow_underscore_headers false`, discard request headers whose field names contained `_` before exposing them through normalized Rack env keys.
- Add `env["puma.underscore_headers"]` when underscore-origin request headers are present, containing header names only, never values.
- Preserve current behavior when `allow_underscore_headers true`: underscore-origin headers are still normalized unless a canonical hyphenated equivalent already exists.

## Recommendation

We recommend that applications assess whether they depend on request headers with underscores and then set:

```ruby
allow_underscore_headers false
```

This is the more secure default for most proxy/CDN-backed deployments. The current default remains `true` for compatibility, but this setting is intended to default to `false` in the next Puma major release.

Applications can use `env["puma.underscore_headers"]` to audit traffic before changing the setting. Because that metadata is client-triggerable, external reporting should be sampled or rate-limited.

## Tests

- `bundle exec ruby -Itest test/test_request_single.rb`
- `bundle exec ruby -Itest test/test_config.rb`
- `bundle exec rubocop lib/puma/client_env.rb lib/puma/client.rb lib/puma/server.rb lib/puma/configuration.rb lib/puma/dsl.rb lib/puma/const.rb test/test_request_single.rb test/test_config.rb`

### DIFF
--- a/lib/puma/client_env.rb
+++ b/lib/puma/client_env.rb
@@ -115,27 +115,24 @@ module Puma
     def req_env_post_parse
       to_delete = nil
       to_add = nil
+      underscore_headers = nil
 
       @env.each do |k,v|
-        if k.start_with?("HTTP_") && k.include?(",") && !UNMASKABLE_HEADERS.key?(k)
-          if to_delete
-            to_delete << k
-          else
-            to_delete = [k]
-          end
+        next unless k.start_with?("HTTP_") && k.include?(",")
 
-          new_k = k.tr(",", "_")
-          if @env.key?(new_k)
-            next
-          end
+        (underscore_headers ||= []) << k.delete_prefix("HTTP_").tr("_,", "-_")
+        next if @allow_underscore_headers && UNMASKABLE_HEADERS.key?(k)
 
-          unless to_add
-            to_add = {}
-          end
+        (to_delete ||= []) << k
+        next unless @allow_underscore_headers
 
-          to_add[new_k] = v
-        end
+        new_k = k.tr(",", "_")
+        next if @env.key?(new_k)
+
+        (to_add ||= {})[new_k] = v
       end
+
+      @env[PUMA_UNDERSCORE_HEADERS] = underscore_headers if underscore_headers
 
       if to_delete # rubocop:disable Style/SafeNavigation
         to_delete.each { |k| env.delete(k) }
--- a/lib/puma/client_env.rb
+++ b/lib/puma/client_env.rb
@@ -105,9 +105,10 @@ module Puma
     # avoid allocation in the common case (ie there are no headers
     # with `,` in their names), that's why it has the extra conditionals.
     #
-    # @note If a normalized version of a `,` header already exists, we ignore
-    #       the `,` version. This prevents clobbering headers managed by proxies
-    #       but not by clients (Like X-Forwarded-For).
+    # @note If underscore headers are disallowed, `,` headers are discarded.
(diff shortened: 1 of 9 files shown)


nateberkopec:
- https://github.com/puma/puma/security/advisories/GHSA-9hf4-67fc-4vf4
- https://github.com/puma/puma/issues/3502
- https://github.com/puma/puma/issues/3556
- https://github.com/puma/puma/issues/3397
- https://nginx.org/en/docs/http/ngx_http_core_module.html#underscores_in_headers

nateberkopec:
Once `allow_underscore_headers` defaults to `false` in the next major, the `UNMASKABLE_HEADERS` codepath becomes dead — underscore-origin `Transfer-Encoding` / `Content-Length` will be deleted unconditionally along with every other underscore header. We should drop the constant and the gated branch in `req_env_post_parse` at that point.

MSP-Greg (approved):
LGTM.

See https://github.com/puma/puma/compare/main...MSP-Greg:puma:00-pr-3937 for some changes.  All can be another PR or two...

lib/puma/server.rb
```diff
@@ -107,6 +107,7 @@ def initialize(app, events = nil, options = {})
       @enable_keep_alives      &&= @queue_requests
       @io_selector_backend       = @options[:io_selector_backend]
       @http_content_length_limit = @options[:http_content_length_limit]
+      @allow_underscore_headers  = @options[:allow_underscore_headers]
```
joshuay03: `@options[:allow_underscore_headers]` returns `nil` when the key is absent, which happens any time `Puma::Server.new` is called without a full `Configuration` (e.g. in tests). In `req_env_post_parse`, `nil` is falsy, so both guards behave as if `false` were passed and underscore headers are silently dropped instead of allowed. Should we be defaulting to `true` here if unset?

```suggestion
      @allow_underscore_headers  = @options.fetch(:allow_underscore_headers, true)
```

lib/puma/client_env.rb
joshuay03: This only describes the collision-avoidance behavior. Might be worth updating to also mention that when `allow_underscore_headers false`, all underscore-origin headers are discarded unconditionally.

lib/puma/server.rb
joshuay03: Might be worth covering this in a test as well.

PR #3938, opened 2026-05-11
Defer on_booted/on_restart/on_stopped removal to Puma v9

dentarg:
These callbacks were slated for removal in v8 but shipped, so push the removal to v9. Clarify the warning to name Puma explicitly, and add a version-guard test that fails once PUMA_VERSION reaches 9 so we don't miss the removal again.

### DIFF
--- a/test/test_events.rb
+++ b/test/test_events.rb
@@ -5,9 +5,6 @@ require 'puma/const'
 require_relative "helper"
 
 class TestEvents < PumaTest
-  # The on_booted/on_restart/on_stopped methods were deprecated and are
-  # scheduled for removal in Puma v9. Fail loudly once the version bumps
-  # past 8 so we remember to remove them.
   def test_deprecated_event_methods_are_removed_in_v9
     refute_operator Gem::Version.new(Puma::Const::PUMA_VERSION), :>=, Gem::Version.new("9"),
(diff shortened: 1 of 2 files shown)


test/test_events.rb
```diff
@@ -1,9 +1,18 @@
...
 require_relative "helper"
 
 class TestEvents < PumaTest
+  # The on_booted/on_restart/on_stopped methods were deprecated and are
+  # scheduled for removal in Puma v9. Fail loudly once the version bumps
+  # past 8 so we remember to remove them.
+  def test_deprecated_event_methods_are_removed_in_v9
+    refute_operator Gem::Version.new(Puma::Const::PUMA_VERSION), :>=, Gem::Version.new("9"),
+      "Remove deprecated on_booted/on_restart/on_stopped from Puma::Events and their tests"
+  end
```
joshuay03: Good idea!

Nit: My personal opinion is that the comment doesn't add much value here

joshuay03 (approved):

test/test_events.rb
dentarg: Agree

joshuay03 (approved):

dentarg:
@nateberkopec as you are doing releases, this one is good to be aware of

joshuay03:
@dentarg Shall we merge this?

PR #3939, opened 2026-05-11
Fix `Configuration#clamp` being called twice in `Launcher`

### LINKED ISSUE #3935: Configuration#clamp should be idempotent or called only once

joshuay03:
Fixes #3935

`Launcher` calls `@config.clamp` in both `initialize` and `run`. The `initialize` call is necessary because `@config.options` on the next line raises `NotClampedError` without it, making the `run` call redundant. Because `clamp` calls `run_mode_hooks`, any callbacks registered via `single {}` or `cluster {}` blocks execute twice.

The fix removes the redundant call in `run`. Nothing between `initialize` and `run` invalidates the clamped config since all intermediate writes go to the same mutable options hash rather than to pre clamp DSL state. All three call sites that create a `Launcher` (`CLI`, `Rack::Handler::Puma`, and the test helper) go through `initialize`, so no path is missed.

Calling `clamp` on an already clamped config is now a no op that emits a deprecation warning. This will become a raise in Puma v9. Also capitalized an existing lowercase raise message for consistency with the rest of the codebase.

### DIFF
--- a/lib/puma/configuration.rb
+++ b/lib/puma/configuration.rb
@@ -283,6 +283,8 @@ module Puma
     #
     # This also calls load if it hasn't been called yet.
     def clamp
+      raise "Configuration already clamped" if @clamped
+
       load unless @loaded
       run_mode_hooks
       set_conditional_default_options
@@ -491,7 +493,7 @@ module Puma
       @_options.all_of(key).each(&:call)
 
       unless @_options[:workers] == workers_before
-        raise "cannot change the number of workers inside a #{key} configuration hook"
+        raise "Cannot change the number of workers inside a #{key} configuration hook"
       end
     end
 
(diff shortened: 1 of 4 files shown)


dentarg:
Lots of `RuntimeError: Configuration already clamped` in the tests, hehe

joshuay03:
Oops. The CLI tests were calling `conf.clamp` after `CLI.new`, which already clamps via `Launcher#initialize`. Should be fixed now in the latest push.

dentarg:
> The CLI tests were calling `conf.clamp` after `CLI.new`, which already clamps via `Launcher#initialize`.

That makes me think... others could very well be doing that and this would then be a breaking change for them? Do we need to `raise`? Should calling `clamp` a second/third/... time be a no-op? Should we log a warning?

joshuay03:
> That makes me think... others could very well be doing that and this would then be a breaking change for them? Do we need to `raise`? Should calling `clamp` a second/third/... time be a no-op? Should we log a warning?

Good point. The second call is now a no-op with a deprecation warning, and will become a raise in v9.

PR #3940, opened 2026-05-12
Fix worker shutdown hang in notify_safely (#3677)

### LINKED ISSUE #3677: Worker hangs on restart
I have puma running in clustered mode on a staging server. It is not processing any requests. When I restart using ```pumactl restart``` it often hangs until the ```worker_shutdown_timeout``` is triggered.

I have investigated by inserting log statements to try to figure out what is happening. What I have discovered is that when this issue occurs the worker receives the SIGTERM and sends the stop command. The 'srv' thread receives the STOP_COMMAND but hangs when closing the notify pipe. When the next SIGTERM is received the stop command cannot be sent due to IOError "stream closed in another thread". Subsequent SIGTERMs result in IOError "closed stream".

Note that the "stream closed in another thread" is only ever thrown when the shutdown hangs, other IOError are sometimes raised in successful worker shutdowns.

The srv thread never returns from the IO pipe close call and eventually the process has to be killed.

**Puma config:**
```
threads 5, 5
workers 3

environment "staging"
directory "/home/app/<appname>/current"

# Logging
log_dir = "/home/app/<appname>/log"
stdout_redirect "#{log_dir}/puma.out.log", "#{log_dir}/puma.err.log", true

# Set master PID and state locations
shared_dir = "/var/www/<appname>/shared"
pidfile "#{shared_dir}/pids/puma.pid"
state_path "#{shared_dir}/pids/puma.state"
bind "unix://#{shared_dir}/sockets/puma.sock"

activate_control_app

force_shutdown_after 2

preload_app!
on_worker_boot do
end
```

**To Reproduce**
This can be hard to reproduce. I haven't managed to reproduce on my development machine yet. Sometimes it won't happen for an hour, other times it happens multiple times in a few minutes. 

1. start puma:
```
bundle exec puma -C config/puma/staging.rb config.ru
```
 
2. run restarts in a loop:
```
while true; do pumactl restart; sleep 1; done
```

3. the restart command show an error when the previous restart has hung

I have produced this bug with the following config.ru, it is not related to Rails:

```
```

**Expected behavior**
Puma should restart very quickly since it is not processing any requests

 - OS:  Ubuntu 22.04.5 LTS (GNU/Linux 6.8.0-1021-aws x86_64)
 - Puma Version: 6.6.0

Does anyone have any suggestions as to how to investigate further? Why would closing an IO pipe hang?

thejspr:
Worker shutdown can hang indefinitely with the puma srv thread parked in sleep_forever inside `IO#close` on the `@notify` pipe. The race: notify_safely uses `@notify << message`, which goes through Ruby's buffered-IO write path and acquires the same internal write lock that `IO#close` (running in handle_servers' ensure block) waits on. When the two interleave, the writer wakes with `IOError "stream closed in another thread"`, but the closer never returns — leaving `server_thread.join` in `Cluster::Worker#run` to hang forever, which in turn leaves the cluster master spinning in `stop_workers` until SIGKILL'd by the orchestrator (e.g. ECS exit code 137).

Switch notify_safely to `IO#syswrite`, a single direct `write(2)` syscall that bypasses Ruby's buffered-IO write lock entirely. The operation is also async-signal-safe, which preserves notify_safely's existing usage from puma's signal handlers (TERM/USR2/INT/HUP via Server#stop / #begin_restart in launcher.rb) — a Mutex-based fix would raise "can't be called from trap context".

Adds two regression tests:

* `test_notify_safely_callable_from_trap_context` — installs a signal handler that calls notify_safely and asserts no ThreadError, plus that STOP_COMMAND is delivered through @notify. This is the test that catches the trap-context regression a Mutex.synchronize-based fix would introduce.
* `test_stop_does_not_hang_under_concurrent_notify_safely` — 100 cycles of run + concurrent notify_safely writers + stop, with a per-cycle timeout and a thread backtrace dump on hang. Catches the underlying race on platforms where it reproduces.

Closes #3677 

### DIFF
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -639,7 +639,16 @@ module Puma
     end
 
     def notify_safely(message)
-      @notify << message
+      # Use syswrite (a single write(2) syscall) rather than `<<` to
+      # bypass MRI's buffered-IO write lock. That lock can deadlock with
+      # IO#close running in handle_servers' ensure block — the closer
+      # parks in sleep_forever while a buffered writer is mid-write,
+      # even though the writer is woken with IOError (puma#3677).
+      # syswrite is also async-signal-safe, so notify_safely remains
+      # callable from puma's signal handlers (TERM/USR2/INT/HUP) without
+      # raising "can't be called from trap context".
+      notify = @notify
+      notify.syswrite(message) if notify && !notify.closed?
     rescue IOError, NoMethodError, Errno::EPIPE, Errno::EBADF
       # The server, in another thread, is shutting down
     rescue RuntimeError => e
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -455,7 +455,7 @@ module Puma
 
     # :nodoc:
     def handle_check
-      cmd = @check.read(1)
+      cmd = @check.sysread 1
 
       case cmd
       when STOP_COMMAND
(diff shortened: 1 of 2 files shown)


MSP-Greg:
Thanks for the PR.

Using main, I tried the two tests, and both passed.  If I added logging to the `notify_safely` rescue statement that is currently a 'no-op', and added a check in one of the tests, it failed on main.  Then I pushed the changes to the my fork, and it did not work with Rubies less that 3.3 (I use Ruby master for most OSS work).  Something is different with older Rubies...

A link to the patch with possible changes is at:

https://github.com/MSP-Greg/puma/commit/928c2ca6175f.patch

The tests are at:

https://github.com/MSP-Greg/puma/actions/runs/26480766509

As mentioned, all tests pass with the changes.

Otherwise, LGTM.

PR #3942, opened 2026-05-20
Clean up passing test output

nateberkopec:
No stuff in my green dots.

After this PR, only dots appear in the test output.


lib/puma/cluster/worker.rb
```diff
@@ -137,7 +137,7 @@ def run
                 # sub call properly adds 'closing' string
                 io << payload.sub(/,\z/, " }\n")
                 @server.reset_max
-              rescue IOError
+              rescue SystemCallError, IOError
```
nateberkopec, line 140: The background thread here, if the master side is already closed, will raise `EPIPE`. We can swallow that.

PR #3957, opened 2026-06-15
Enforce HTTP1.1 RFC for Host header presence

### LINKED ISSUE #3926: Improve RFC compliance
[Interesting test](https://www.http-probe.com/servers/puma/). I have not validated any of these results. Would be interested in fixing any valid ones.

stim371:
### Description
As a first contribution, I tried to pick off a few TODOs from https://github.com/puma/puma/issues/3926 to improve alignment with the HTTP RFC.

HTTP1.1 requires the Host header to be included in the request and the value not to be blank (See [Section 3.2 of the RFC](https://datatracker.ietf.org/doc/html/rfc9112#section-3.2)) and adding some presence checks seemed like a good place to start.

I noticed some discussion about [validating HOST headers](https://github.com/puma/puma/discussions/2977) and the decision not to move forward. This PR seems like a step short of "validation" and more about aligning with the RFC, but I'd be curious to hear everyone's take on it.

The PR goes above the 100 line limit, but only because the test suite needed to be updated with Host entries in several places. The core changes are 5 short tests and one function in `client_env.rb`.

### Questions
* Should these 400 errors explicitly close the connection? 
    * `@env[HTTP_CONNECTION] = 'close'`
* Should this be targeted against a future major version branch instead of main?

#### Extract validations
I think we could extract validations into a more centralized, visible area. As I dug through, there is header validation in a few areas:
* `Client#parser_execute` (translates parser errors into Client errors)
* `Client#setup_body` (validates Transfer Encoding & Content Length, and also sets up body)

#### Encapsulate error generation
There are a few facets to error creation, but I'm not 100% sure what's required. Maybe we could extract a method or change the `HttpParserError` class to hold a little more knowledge (code, closed).

* Invalid states seem to be marked by `raise HttpParserError`
* There's also a `HttpParserError501` that just switches the code, even though you could set the code via at least one other method (below)
* `Client@error_status_code` is set directly in a few places (but often implicitly falls back to 400 in `Server#client_error`)
    * The `assert_invalid` matcher expects `@error_status_code` to be set, so you can't rely on the implicit fallback when generating the error response (if you use the matcher).
* Some errors also flag to close the connection with `@env[HTTP_CONNECTION] = 'close'`

As a newbie, I only found these disparate flags by having Cursor walk me through the codebase. Tying all these levers together might reduce missed changes in the future.

### DIFF
--- a/lib/puma/client_env.rb
+++ b/lib/puma/client_env.rb
@@ -12,6 +12,22 @@ module Puma
 
     include Puma::Const
 
+    def validate_env
+      return unless @env[SERVER_PROTOCOL] == HTTP_11
+
+      unless @env.key? HTTP_HOST
+        @error_status_code = 400
+        @env[HTTP_CONNECTION] = CLOSE
+        raise HttpParserError, "Missing Host header"
+      end
+
+      if @env[HTTP_HOST].empty?
+        @error_status_code = 400
+        @env[HTTP_CONNECTION] = CLOSE
+        raise HttpParserError, "Invalid Host header"
+      end
+    end
+
     # Given a Hash +env+ for the request read from +client+, add
     # and fixup keys to comply with Rack's env guidelines.
     # @param env [Hash] see Puma::Client#env, from request
--- a/test/helpers/test_puma/shutdown_requests.rb
+++ b/test/helpers/test_puma/shutdown_requests.rb
@@ -37,10 +37,10 @@ module TestPuma
       s2 = send_http(
         post ?
           "POST /s2 HTTP/1.1\r\nHost: test.com\r\nContent-Type: text/plain\r\nContent-Length: 5\r\n\r\nhi!" :
-          "GET /s2 HTTP/1.1\r\n"
+          "GET /s2 HTTP/1.1\r\nHost: test.com\r\n"
       )
       mutex.synchronize do
-        s1 = send_http("GET /s1 HTTP/1.1\r\n\r\n")
+        s1 = send_http("GET /s1 HTTP/1.1\r\nHost: test.com\r\n\r\n")
         app_finished.wait(mutex)
         app_finished.signal if s1_complete
       end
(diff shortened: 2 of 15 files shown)


test/helpers/test_puma/shutdown_requests.rb
```diff
@@ -37,10 +37,10 @@ def add(client)
       s2 = send_http(
         post ?
           "POST /s2 HTTP/1.1\r\nHost: test.com\r\nContent-Type: text/plain\r\nContent-Length: 5\r\n\r\nhi!" :
-          "GET /s2 HTTP/1.1\r\n"
+          "GET /s2 HTTP/1.1\r\nHost: test.com\r\n"
       )
       mutex.synchronize do
-        s1 = send_http("GET /s1 HTTP/1.1\r\n\r\n")
+        s1 = send_http("GET /s1 HTTP/1.1\r\nHost: test.com\r\nHost: test.com\r\n\r\n")
```
joshuay03: Typo?


```suggestion
        s1 = send_http("GET /s1 HTTP/1.1\r\nHost: test.com\r\n\r\n")
```

lib/puma/client_env.rb
```diff
@@ -12,6 +12,22 @@ module ClientEnv # :nodoc:
 
     include Puma::Const
 
+    def validate_env
+      return unless @env[SERVER_PROTOCOL] == HTTP_11
+
+      if !@env.key? HTTP_HOST
```
joshuay03: Nit.

```suggestion
      unless @env.key? HTTP_HOST
```

joshuay03 (commented):
Thank you!

> Should these 400 errors explicitly close the connection? 

Yes, closing the connection on these 400s is correct, you're already doing it via `@env[HTTP_CONNECTION] = CLOSE`.

> Should this be targeted against a future major version branch instead of main?

No, we'll merge it into main and bundle it in the next release where appropriate.

joshuay03 (approved):

PR #3969, opened 2026-07-04
Restore admission gate when `queue_requests` is disabled

### LINKED ISSUE #3961: Increased tail latency caused by requests routed to a busy worker rather than an idle one
**Describe the bug**
In Puma 7, a busy worker in **clustered** mode will accept a brand-new connection into its **own** per-worker queue (`@todo`) even when a sibling worker is completely idle. The new connection is then pinned behind that worker's in-flight work, while idle capacity on the other worker goes unused.

This is a behavior change introduced deliberately by **#3678 ("Fix long tail response problem - revised", 7.0.0)**, which removed the hard accept admission gate from the accept loop. To be upfront: I understand this was an intentional trade-off for keepalive fairness, and the `ClusterAcceptLoopDelay` class comment itself acknowledges the risk ("if one worker has all threads busy ... accepting a new request might steal it from a less busy worker"). I'm filing this because for the **single-threaded clustered** configuration, the change reintroduces exactly the cross-worker stealing that the original admission gate (#1278, refined in #1563) was built to prevent, and there is currently no opt-in to get the old admission behavior back.

**Puma config:**

```ruby
workers Integer(ENV.fetch("WORKERS", "2"))
threads 1, 1
bind "tcp://127.0.0.1:9292"
queue_requests false
preload_app!
```

**To Reproduce**

`hello.ru` — `/slow` blocks for `SLOW_SLEEP` seconds, everything else is instant:

```ruby
sleep_seconds = Integer(ENV.fetch("SLOW_SLEEP", "8"))
app = lambda do |env|
  sleep sleep_seconds if env["PATH_INFO"] == "/slow"
  [200, { "content-type" => "text/plain" }, ["Hello World\n"]]
end
run app
```

Boot (2 workers, 1 thread each, queue_requests false):

```
SLOW_SLEEP=8 WORKERS=2 bundle exec puma -C puma_config.rb hello.ru
```

```
# 1) Start a FRESH server (2 workers, 1 thread, queue_requests false). One worker
#    will be tied up for 12s. Each version must run on its own freshly-started
#    server with :9292 free — kill the master AND its workers between runs, e.g.
#    `lsof -ti tcp:9292 | xargs kill -9`.
SLOW_SLEEP=12 WORKERS=2 bundle exec puma -C puma_config.rb hello.ru

# 2) Pin exactly one of the two workers for the whole measurement:
curl -s http://127.0.0.1:9292/slow &

# 3) Immediately measure the fast path:
ab -n 80 -c 4 http://127.0.0.1:9292/
```

Measured (`ab` throughput + "requests served within a certain time"):

| | RPS | p50 | p90 | p95 | max |
|---|---|---|---|---|---|
| **6.6.1** | 7793 | 0ms | 1ms | 1ms | 2ms |
| **7.2.1** | 7.08 | 0ms | 1ms | 11284ms | 11296ms |

On 6.6.1 the idle worker serves every fast request (≤2ms). On 7.2.1 ~10% of the fast requests are accepted into the busy worker's `@todo` and block behind the 12s `/slow` (p95/max ≈ 11.3s), so `ab` throughput collapses from ~7800 to ~7 req/s while it waits on the pinned requests. Zero failed requests in either version.

**Expected behavior**
In clustered mode with `queue_requests: false`, a worker at capacity should not pull a new connection ahead of an idle sibling. With 2 workers, 1 thread each, and one worker blocked, a burst of fast requests should be served by the idle worker (as in 6.6.1: low-ms tail), not head-of-line-blocked behind the busy worker's in-flight request.

## Ask

Provide an **opt-in** to restore hard admission behavior for clustered mode — e.g. either:

1. **Simplest / most targeted — gate on `queue_requests: false`.** That setting is already documented as *"each worker process accepts exactly as many requests as it is configured to simultaneously handle"* — a capacity bound that 6.x enforced via the unconditional `wait_until_not_full` and that 7.x dropped (the repro above shows a busy worker accepting a 2nd connection into `@todo` under `queue_requests: false`, `max_threads: 1`). Restoring `wait_until_not_full` **only** in the `queue_requests: false` branch needs **no new config surface**, restores the documented capacity bound (and honest `puma.worker.backlog` / `@todo` signaling, so an upstream LB again sees the backpressure), and has **zero conflict with #3678**: keepalive — the entire reason #3678 removed the gate — requires `queue_requests: true` (`server.rb:107`, `@enable_keep_alives &&= @queue_requests`), so it is already off whenever `queue_requests: false`. (The old bound was an approximate pre-accept check over the kernel backlog, not a hard invariant — the ask is to restore that *bound*, not to claim an exact guarantee.)
1. A config flag (e.g. restoring `wait_until_not_full` semantics) that makes the accept loop **block** rather than soft-delay when `busy_threads >= max_threads` and `workers >= 2`, so busy workers decline and idle siblings (or the kernel queue) take the connection; or
2. Make `ClusterAcceptLoopDelay` perform a true wait (not a scaled sleep) when `busy_threads >= max_threads` in clustered mode

Any of these keeps #3678's keepalive-fairness default while letting single-threaded-clustered deployments opt back into pre-7.0 cross-worker balancing

- **OS:** macOS (Darwin 25.5.0); mechanism verified against installed gem source for both versions.
- **Puma Version:** regression between 6.6.1 (good) and 7.2.1 (bad); gate removed in 7.0.0 (#3678).
- **Ruby Version:** ruby 3.4.9 (MRI).

joshuay03:
Closes #3961

In cluster mode with `queue_requests: false`, a busy worker's accept loop kept pulling new connections into its own `@todo` list even when a sibling worker was idle. The admission gate that used to prevent this was removed in #3678 to fix keep-alive tail latency. Keep-alive requires `queue_requests: true`, so restoring the gate under `!@queue_requests` recovers pre-7.0 cross-worker balancing without touching the path #3678 was fixing.

Reproduced the bug locally with 2 workers, 1 thread each, `queue_requests false`, one worker pinned by a 12s `/slow` request, then `ab -n 80 -c 4 /`:

|          | RPS     | p50 | p90 | p95     | max     |
|----------|---------|-----|-----|---------|---------|
| main     | 6.94    | 1ms | 9ms | 11493ms | 11496ms |
| fix-3961 | 2321.20 | 1ms | 7ms | 11ms    | 12ms    |

Also verified #3678's tail-latency fix still holds with a keep-alive benchmark matched to its original parameters: 2 workers, 5:5 threads, default `queue_requests`, 40 concurrent hey clients (4x oversubscribed), 200ms app delay, 30s duration, 5 runs each averaged:

|          | RPS  | p50     | p90     | p95     | p99     | slowest |
|----------|------|---------|---------|---------|---------|---------|
| main     | 48.9 | 815.2ms | 820.2ms | 822.1ms | 825.9ms | 831.2ms |
| fix-3961 | 49.1 | 814.3ms | 819.1ms | 820.6ms | 823.8ms | 827.1ms |

Sub-millisecond deltas across every percentile.

### DIFF
--- a/test/test_integration_cluster.rb
+++ b/test/test_integration_cluster.rb
@@ -295,6 +295,27 @@ class TestIntegrationCluster < TestIntegration
     assert true
   end
 
+  def test_queue_requests_disabled_does_not_pin_new_requests_to_busy_worker
+    cli_server "-w #{workers} -t 1:1 test/rackup/sleep_pid.ru", config: "queue_requests false"
+
+    get_worker_pids # wait for workers to boot
+
+    slow = Thread.new { read_body(fast_connect("sleep2"), 5).split.last.to_i }
+    sleep 0.5
+
+    mutex = Mutex.new
+    fast_pids = []
+    Array.new(8) do
+      Thread.new do
+        pid = read_body(fast_connect("sleep0"), 3).split.last.to_i
+        mutex.synchronize { fast_pids << pid }
+      end
+    end.each(&:join)
+
+    assert_equal 1, fast_pids.uniq.size
+    refute_equal slow.value, fast_pids.first
+  end
+
   def test_worker_index_is_with_in_options_limit
     cli_server "-C test/config/t3_conf.rb test/rackup/hello.ru"
 
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -389,13 +389,16 @@ module Puma
                 # clients until the code is finished.
                 pool.wait_while_out_of_band_running
 
-                # A well rested herd (cluster) runs faster
-                if @cluster_accept_loop_delay.on? && (busy_threads_plus_todo = pool.busy_threads) > 0
-                  delay = @cluster_accept_loop_delay.calculate(
-                    max_threads: @max_threads,
-                    busy_threads_plus_todo: busy_threads_plus_todo
-                  )
-                  sleep(delay)
+                if @queue_requests
+                  if @cluster_accept_loop_delay.on? && (busy_threads_plus_todo = pool.busy_threads) > 0
+                    delay = @cluster_accept_loop_delay.calculate(
+                      max_threads: @max_threads,
+                      busy_threads_plus_todo: busy_threads_plus_todo
+                    )
+                    sleep(delay)
+                  end
+                else
+                  pool.wait_until_not_full
                 end
 
                 io = begin
--- a/lib/puma/server.rb
+++ b/lib/puma/server.rb
@@ -389,16 +389,18 @@ module Puma
                 # clients until the code is finished.
                 pool.wait_while_out_of_band_running
 
-                if @queue_requests
-                  if @cluster_accept_loop_delay.on? && (busy_threads_plus_todo = pool.busy_threads) > 0
-                    delay = @cluster_accept_loop_delay.calculate(
-                      max_threads: @max_threads,
-                      busy_threads_plus_todo: busy_threads_plus_todo
-                    )
-                    sleep(delay)
+                unless shutting_down?
+                  if @queue_requests
+                    if @cluster_accept_loop_delay.on? && (busy_threads_plus_todo = pool.busy_threads) > 0
+                      delay = @cluster_accept_loop_delay.calculate(
+                        max_threads: @max_threads,
+                        busy_threads_plus_todo: busy_threads_plus_todo
+                      )
+                      sleep(delay)
+                    end
+                  else
+                    pool.wait_until_not_full
                   end
-                else
-                  pool.wait_until_not_full
                 end
 
                 io = begin
--- a/lib/puma/thread_pool.rb
+++ b/lib/puma/thread_pool.rb
@@ -315,6 +315,17 @@ module Puma
       end
     end
 
+    def wait_until_not_full
+      with_mutex do
+        while true
+          return if @shutdown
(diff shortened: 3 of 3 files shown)


test/test_integration_cluster.rb
```diff
@@ -295,6 +295,27 @@ def test_queue_requests_disabled
...
+    mutex = Mutex.new
+    fast_pids = []
+    8.times.map do
+      Thread.new do
+        pid = read_body(fast_connect("sleep0"), 3).split.last.to_i
+        mutex.synchronize { fast_pids << pid }
+      end
+    end.each(&:join)
+
+    assert_equal 1, fast_pids.uniq.size
```
joshuay03, line 315: On main:
```
puma [main] bundle exec ruby test/test_integration_cluster.rb -n test_queue_requests_disabled_does_not_pin_new_requests_to_busy_worker
Run options: -n test_queue_requests_disabled_does_not_pin_new_requests_to_busy_worker --seed 40306

# Running:

F

Fabulous run in 2.413025s, 0.4144 runs/s, 0.4144 assertions/s.
Errors & Failures:

  1) Failure:
TestIntegrationCluster#test_queue_requests_disabled_does_not_pin_new_requests_to_busy_worker [test/test_integration_cluster.rb:315]:
Expected: 1
  Actual: 2

1 runs, 1 assertions, 1 failures, 0 errors, 0 skips
```

MSP-Greg (approved):
LGTM

PR #3973, opened 2026-07-06
Add support for HTTP `QUERY` method

seanpdoyle:
### Description

As of June 2026, [RFC 10008][] is a proposed standard.

From the RFC Abstract:

> This specification defines the QUERY method for HTTP. A QUERY requests
> that the request target process the enclosed content in a safe and
> idempotent manner and then respond with the result of that processing.
> This is similar to POST requests, but QUERY requests can be
> automatically repeated or restarted without concern for partial state
> changes.

This commit adds `QUERY` to the `IANA_HTTP_METHODS` constant in order to support parsing the new HTTP method.

Once parsed, the value is transformed injected as a `Rack::REQUEST_METHOD`-compatible `QUERY` value proposed in [rack/rack#2474][].

[RFC 10008]: https://www.rfc-editor.org/rfc/rfc10008.html
[rack/rack#2474]: https://github.com/rack/rack/pull/2474

### DIFF
--- a/lib/puma/const.rb
+++ b/lib/puma/const.rb
@@ -147,12 +147,12 @@ module Puma
     REQUEST_METHOD = "REQUEST_METHOD"
     HEAD = "HEAD"
 
-    # based on https://www.rfc-editor.org/rfc/rfc9110.html#name-overview,
-    # with CONNECT removed, and PATCH added
-    SUPPORTED_HTTP_METHODS = %w[HEAD GET POST PUT DELETE OPTIONS TRACE PATCH].freeze
+    # based on https://www.rfc-editor.org/rfc/rfc9110.html#name-overview and https://www.rfc-editor.org/rfc/rfc10008.html,
+    # with CONNECT removed, and PATCH & QUERY added
+    SUPPORTED_HTTP_METHODS = %w[HEAD GET POST PUT DELETE OPTIONS TRACE PATCH QUERY].freeze
 
     # list from https://www.iana.org/assignments/http-methods/http-methods.xhtml
-    # as of 04-May-23
+    # as of 17-Jun-26, with QUERY added based on https://www.rfc-editor.org/rfc/rfc10008.html
     IANA_HTTP_METHODS = %w[
       ACL
       BASELINE-CONTROL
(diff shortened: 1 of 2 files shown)


lib/puma/const.rb
```diff
@@ -147,12 +147,12 @@ module Const
...
-    # based on https://www.rfc-editor.org/rfc/rfc9110.html#name-overview,
-    # with CONNECT removed, and PATCH added
-    SUPPORTED_HTTP_METHODS = %w[HEAD GET POST PUT DELETE OPTIONS TRACE PATCH].freeze
+    # based on https://www.rfc-editor.org/rfc/rfc9110.html#name-overview and https://www.rfc-editor.org/rfc/rfc10008.html,
+    # with CONNECT removed, and PATCH & QUERY added
+    SUPPORTED_HTTP_METHODS = %w[HEAD GET POST PUT DELETE OPTIONS TRACE PATCH QUERY].freeze
 
     # list from https://www.iana.org/assignments/http-methods/http-methods.xhtml
-    # as of 04-May-23
+    # as of 06-Jun-30, with QUERY added based on https://www.rfc-editor.org/rfc/rfc10008.html
```
joshuay03: `06-Jun-30` doesn't match the `DD-Mon-YY` format the prior line used (`04-May-23`), and `2026-06-17` is the Last Updated date on the IANA HTTP Methods registry itself.

```suggestion
    # as of 17-Jun-26, with QUERY added based on https://www.rfc-editor.org/rfc/rfc10008.html
```

test/test_request_single.rb
```diff
@@ -92,6 +92,12 @@ def test_method
     assert_equal 'GET', @client.env[REQUEST_METHOD]
   end
 
+  def test_query_method
+    create_client "QUERY / HTTP/1.1\r\nContent-Type: application/json\r\nContent-Length: 2\r\n\r\n{}"
```
joshuay03: RE: CI failures, you're missing the `Host` header here. See https://github.com/puma/puma/pull/3957.

joshuay03 (approved):

PR #3978, opened 2026-07-23
Log client IP on HTTP parse errors via socket peer address

softr8:
### Description
Problem
`HttpParserError/HttpParserError501` log lines drop the client IP:

```
HTTP parse error, malformed request ("GET /" - (-)): #<Puma::HttpParserError: Invalid HTTP format, parsing fails. Bad protocol HTTP/1.1>
```

`REMOTE_ADDR/X-Forwarded-For` are only populated in env once header parsing succeeds, so `ErrorLogger#request_title` logged `"(-)"` for the client IP on any request that failed to parse (e.g. `HttpParserError`, `HttpParserError501`) — the exact cases where knowing the source IP matters most for blocking abusive clients.

The peer address is still available on the raw socket regardless of parse state, via `Client#peerip`. Fall back to it when `REMOTE_ADDR` and `X-Forwarded-For` are both absent.

Fix
`Client#peerip` reads the peer address directly off the raw socket (`@io.peeraddr`), independent of parse state. `request_title` now falls back to it when `REMOTE_ADDR/X-Forwarded-For` are absent.

Compatibility
Behavior unchanged for any request that parses normally (`REMOTE_ADDR/X-Forwarded-For` still win). Only affects the previously-broken parse-error path. req in this call site is always a `Puma::Client`, but guarded with `respond_to?(:peerip)` since `ErrorLogger` is also invoked in tests/other contexts with plain request doubles.

Testing
Added cases to `test/test_error_logger.rb`: falls back to `peerip` when `REMOTE_ADDR` missing, prefers `REMOTE_ADDR` over `peerip` when both present, still renders - when neither is available.

### DIFF
--- a/lib/puma/error_logger.rb
+++ b/lib/puma/error_logger.rb
@@ -80,11 +80,19 @@ module Puma
 
       query_string = env[QUERY_STRING]
 
+      # REMOTE_ADDR/X-Forwarded-For are only present once header parsing has
+      # completed, so they're absent for requests that fail to parse (e.g.
+      # HttpParserError). Fall back to the socket's peer address, which
+      # `req` (a Puma::Client) can still provide since it reads directly
+      # off the connection rather than the parsed env.
+      remote_addr = env[HTTP_X_FORWARDED_FOR] || env[REMOTE_ADDR]
+      remote_addr ||= req.peerip if req.respond_to?(:peerip)
+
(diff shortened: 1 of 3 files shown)


dentarg:
This needs to consider IPv6 usage, see https://github.com/puma/puma/pull/3916

PR #3979, opened 2026-07-24
Honor trim requests between work items so busy pools can shrink

yuki24:
Right now, trim requests are only honored inside the `while todo.empty?` loop in the `ThreadPool` class, so a pool with queued work never executes them. This has been fine until the introduction of `Server#update_thread_pool_min_max` https://github.com/puma/puma/pull/3658, since the only situation where we want to trim threads is the process is idle with an empty queue.

Now that we have `Server#update_thread_pool_min_max` we are in the need of dynamically adjusting the # of threads, hence the newly added `#trim_thread_if_requested(processor)` call within the spawned thread.

There is one thing to keep in mind, which is we should never attempt to trim threads during shutdown - shutdown sets `@trim_requested = @spawned` and relies on workers draining `@todo` before exiting via the idle branch. Other than that, this change should be straightforward.

Let me know what you all think. Thanks!

### DIFF
--- a/lib/puma/thread_pool.rb
+++ b/lib/puma/thread_pool.rb
@@ -211,21 +211,19 @@ module Puma
                 processor.marked_as_io_thread = false
               else
                 # We're already at max threads, so we exit the extra io thread.
-                @spawned -= 1
-                @processors.delete(processor)
-                trigger_before_thread_exit_hooks
-                Thread.exit
+                trim_thread(processor)
               end
             end
 
+            trim_thread_if_requested(processor)
+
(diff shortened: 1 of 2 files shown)


dentarg:
What's the reasoning for closing this @yuki24?

yuki24:
Ah! This was 100% by accident as I had an issue in a private repo that happened to have a description of `closes https://github.com/puma/puma/pull/3979`. Re-opening now.

(was this ever the expected behaviour on GitHub? So weird...)

PR #3982, opened 2026-07-31
Feature: on_worker_boot hooks

### LINKED ISSUE #3628: Make exception rescuing configurable for on_worker_boot.

The `on_worker_boot` [rescue behavior](https://github.com/puma/puma/issues/1551) led to a subtle bug during our upgrade to Rails 7.1.  We were calling `schema_cache=` in a `on_worker_boot` which was [removed but not documented](https://github.com/rails/rails/issues/54622) in 7.1 and so it started raising an exception.  There was some load bearing code that runs after this call that was not being executed and so our web servers booted without an important feature functioning properly.

I would like to be able to opt out of this behavior.  I would have preferred to have had all workers crash looping since our deploy would have been halted due to health probes failing.  Instead we detected the issue only after reports by developers and found the issue due to the increased log volume.

gilesbowkett:
### Description

Closes #3628.

That ticket came from @ransombriggs after a Rails 7.1 upgrade cooked their `on_worker_boot` by removing `schema_cache=` unexpectedly. Puma was logging the resultant exceptions. The ticket requested that the server crash on exceptions instead. @dentarg proposed hooks as a compromise, paralleling the existing `lowlevel_error_handler` subsystem.

Caveat: I don't need these hooks myself. I wanted to contribute to Puma, saw this ticket, and thought it'd be a good way to get my feet wet.

Still, the logic of the ticket was that, if the hooks had existed, a bug which caused silent fails could have been caught quicker. That made sense to me, so this PR adds those hooks.

The PR adds a new `hook_error_handler config` option which replaces the default warn-and-continue for every hook that goes through `run_hooks`. The handler doesn't catch exceptions, so giving it a hook which `raise`s will give users the ability to crash on exceptions, but by default, it'll keep logging exceptions instead.

The PR exceeds the implied 100-line limit from the checklist by 1 line, unless you count tests and documentation, in which case it goes all the way up to 225 lines total. 😬 Sorry about that!

- [ ] ~~My pull request is 100 lines added/removed or less so that it can be easily reviewed.~~
- [ ] ~~If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.~~

### DIFF
--- a/docs/fork_worker.md
+++ b/docs/fork_worker.md
@@ -28,6 +28,7 @@ The `fork_worker` option allows your application to be initialized only once for
     - When initially forking the parent process to the worker 0 child, `before_fork` will trigger on the parent process and `before_worker_boot` will trigger on the worker 0 child as normal.
     - When forking the worker 0 child to grandchild workers, `before_refork` and `after_refork` will trigger on the worker 0 child, and `before_worker_boot` will trigger on each grandchild worker.
     - For clarity, `before_fork` does not trigger on worker 0, and `after_refork` does not trigger on the grandchild.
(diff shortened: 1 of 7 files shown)


gilesbowkett:
the last two commits show test fails, but all they change is comments, and the two commits before that pass. these tests run on a ton of different OSes, so this is probably just normal flakiness.

PR #3983, opened 2026-08-03
ci: run the test suite on FreeBSD

### LINKED ISSUE #2589: CI on FreeBSD
I think we can get CI going on FreeBSD using [Cirrus CI](https://cirrus-ci.org/), from https://cirrus-ci.org/features/:

>To support the Open Source community, Cirrus CI provides Linux, Windows, macOS and FreeBSD services free of charge with some limits but without a cap on how many minutes a month OSS projects can consume.

Even after applying https://github.com/puma/puma/pull/2588, there are many failing tests ([gist with test output](https://gist.github.com/dentarg/fec2da4a8c0808522eacaeb98b91e9a4)), it seems to be all that use `wait_for_server_to_boot`: https://github.com/puma/puma/blob/5e6e0737fcc92e48ec94ae52595f1c948f82a916/test/helpers/integration.rb#L102-L114

Anyone has any idea?

neilpang:
### Description

Closes #2589.

You opened this in 2021 and asked "Anyone has any idea?" about the failing
tests, then dropped the vmactions link in 2024. This is that job, plus the two
things it turned out to need.

The suite runs **862 tests, 0 failures, 0 errors** on FreeBSD 15.1 with Ruby
3.4.9 -- the same 862 the Ubuntu job runs.

https://github.com/neilpang/puma/actions/runs/30757622697/job/91522430505

```
862 runs, 2526 assertions, 0 failures, 0 errors, 16 skips
Finished in 304.446288s
```

**`Puma::IS_LINUX` is true on FreeBSD.** `detect.rb` defines it as
`!(IS_OSX || IS_WINDOWS)`, which held while CI was Linux/macOS/Windows. So
`skip_unless :linux` never skipped, and all seven `TestPluginSystemd` tests ran
here -- six passed, and `test_systemd_notify_usr1_phased_restart_cluster` hung
for 45s on `IO#wait_readable`. Defining it as `RUBY_DESCRIPTION.include? 'linux'`
(matching how `IS_OSX` is written, and not `RUBY_PLATFORM`, which is `"java"` on
JRuby) lets those classes skip as intended. `IS_LINUX` is read only by the test
suite, so no runtime behaviour changes.

**`test_http_11_req_oversize_chunked` assumes a Linux write error.** After the
413 the server closes, and the test asserts that the next write raises. I
isolated this from puma entirely -- a plain TCP server that replies and closes,
then a client writing repeatedly:

```
read response: "HTTP/1.1 413 Payload Too Large"
write 1: NO ERROR
write 2: Errno::EPIPE
```

BSD buffers the first write after the peer's FIN and reports the error on a
later one. Back-to-back writes both land in the buffer, so the second write
needs a brief pause. Linux still raises on the first write, inside the same
block.

One note on the workflow itself: `RUNNER_TOOL_CACHE` has to be passed into the
VM. `test/helper.rb:137` does `result_str.gsub! ENV['RUNNER_TOOL_CACHE'], ''`
unconditionally, and `nil` raises `TypeError` inside a parallel worker, which
hangs the whole run rather than failing it.

### DIFF
--- a/.github/workflows/test_bsd.yml
+++ b/.github/workflows/test_bsd.yml
@@ -1,0 +1,59 @@
+name: FreeBSD
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  freebsd:
+    name: freebsd ruby
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - name: repo checkout
+        uses: actions/checkout@v7
+
+      - name: test
+        uses: vmactions/freebsd-vm@v1
+        with:
+          usesh: true
+          cache-after-prepare: true
+          envs: 'CI PUMA_TEST_DEBUG TESTOPTS MT_CPU RUNNER_TOOL_CACHE GITHUB_ACTIONS'
+          prepare: |
+            pkg install -y git devel/ragel devel/ruby-gems gmake
+          run: |
+            set -e
+            # Puma's suite opens a lot of sockets; CONTRIBUTING.md recommends
+            # a file limit of 4000 or more.
+            ulimit -n 4000 || true
+            ulimit -n
+            ruby -v
+            ragel --version | head -1
+            gem install --no-document bundler
+            bundle install --jobs 4
+            bundle exec rake compile
+            rc=0
+            test/runner --verbose || rc=$?
+            # Integration servers are spawned with IO.popen and, with the
+            # default merge_err: false, inherit this session's stderr (see
+            # test/helpers/integration.rb). One that outlives the suite holds
+            # that descriptor open, the ssh command never returns, and the job
+            # then idles until it times out with every test already passed.
+            pkill -f 'bin/puma' 2>/dev/null || true
+            exit $rc
+        env:
+          CI: true
+          GITHUB_ACTIONS: true
+          PUMA_TEST_DEBUG: true
+          TESTOPTS: -v
+          MT_CPU: 2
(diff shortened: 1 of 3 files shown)


dentarg:
Thanks for the work. Is it too slow?

neilpang:
ubuntu and macos tests use about 3 minutes,
Freebsd uses about 6 minutes.
It doesn't seem too slow for me.

dentarg:
The job timed out but I re-ran it.

neilpang:
Thanks for re-running it. That timeout is worth a look, because the suite
had already finished when it happened.

From the cancelled attempt:

  03:25:03  862 runs, 2539 assertions, 0 failures, 0 errors, 16 skips
            (nothing at all for the next 23m27s)
  03:48:30  ##[error]The operation was canceled.

It never reached the copyback step, so what stalled was the command
inside the VM, not the artifact sync.

The likely mechanism: integration servers are spawned in
test/helpers/integration.rb with

  if merge_err
    @server = IO.popen(env, cmd, :err=>[:child, :out])
  else
    @server = IO.popen(env, cmd)
  end

and merge_err defaults to false, so a server inherits the caller's stderr
-- here the ssh session's. teardown reaps @pid, but a cluster worker that
outlives its master is not covered by that, and while it lives it holds
the descriptor open, so ssh never sees EOF and the job idles with every
test already green. That matches the log exactly: no output, no copyback,
no failure.

What I could not do is reproduce it. Five further runs, four of them in
parallel, all finished in 4 to 6 minutes with nothing left behind:

  --- ruby processes:
     none
  --- puma processes:
     none

So this is a guard, not a claimed fix -- one line after the suite, exit
status preserved:

  rc=0
  test/runner --verbose || rc=$?
  pkill -f 'bin/puma' 2>/dev/null || true
  exit $rc

https://github.com/neilpang/puma/actions/runs/30969208749
862 runs, 0 failures, 6m22s, copyback reached normally.

If you would rather not carry that in CI, the structural alternative is
merge_err: true for the tests that spawn cluster servers, which keeps the
descriptor off the session in the first place -- but that is a change to
your test helpers, so it seemed like your call.

Issue #3535, opened 2024-10-24
CI: hot_restart_does_not_drop_connections - what is an failure?


MSP-Greg:
CI often has failures (many pass on retry) in the `hot_restart_does_not_drop_connections` tests, which exist in both clustered and single tests.  Examples below:

```text
TestIntegrationSingle#test_hot_restart_does_not_drop_connections_threads
Failure: [test/helpers/integration.rb:521]:
Expected: 1000
Actual: 999


TestIntegrationCluster#test_hot_restart_does_not_drop_connections_threads
Failure: [test/helpers/integration.rb:521]:
Expected: 3000
Actual: 2999


TestIntegrationSingle#test_hot_restart_does_not_drop_connections_threads
Failure: [/Users/runner/work/puma/puma/test/helpers/integration.rb:511]:
0 unexpected_response
0 refused
0 read timeout
1 reset
999 success
489 success after restart
2 restart count
Expected no reset errors.
Expected: 0
Actual: 1
```

A long time ago, Puma CI classified socket errors into 'acceptable' and 'unacceptable'.  I haven't done a lot of research into it, but it's my understanding that all current browsers will retry a request on almost any socket error.

So, should we remove the error classification and set the tests to fail if too many errors occur?  The above logs show 0.1% or less errors.

@schneems any thoughts?

Issue #3114, opened 2023-03-30
Flaky test_hot_restart_does_not_drop_connections_threads test


Apteryks:
**Describe the bug**
The `test_hot_restart_does_not_drop_connections_threads` sometimes fail, like:

```
Finished in 46.952074s, 12.1400 runs/s, 35.8238 assertions/s.
Errors & Failures:

  1) Failure:
TestIntegrationCluster#test_hot_restart_does_not_drop_connections_threads [/tmp/guix-build-ruby-puma-6.2.0.drv-0/source/test/helpers/integration.rb:457]:
Expected: 3000
  Actual: 2999

570 runs, 1682 assertions, 1 failures, 0 errors, 12 skips
```

**To Reproduce**
Run the test suite multiple times until it fails.

**Expected behavior**
Test should always pass.

**Desktop (please complete the following information):**
 - OS: Guix System (GNU/Linux)
 - Puma Version 6.2.0.

Apteryks:
Another one:
```
S...........................................................................................................................SSSSS................................................Parsing options from the first comment line is deprecated!
............S...SS........................................S.................S..........................................................S........................................................................................................F..............................S.............................................................

Finished in 47.177169s, 12.0821 runs/s, 35.4621 assertions/s.
Errors & Failures:

  1) Failure:
TestIntegrationCluster#test_hot_restart_does_not_drop_connections [/tmp/guix-build-ruby-puma-6.2.0.drv-0/source/test/helpers/integration.rb:447]:
      0 unexpected_response
      0 refused
      0 read timeout
      1 reset
    999 success
    752 success after restart
      2 restart count
Expected no reset errors.
Expected: 0
  Actual: 1

570 runs, 1673 assertions, 1 failures, 0 errors, 13 skips
```

MSP-Greg:
@Apteryks I haven't had the time, but we/I will have PR's for this, probably this weekend.

joshuay03:
Unsure if related but I'm getting this intermittently on `master` using the MacOS Sonoma 14.0 Beta:

```ruby
#<Thread:0x000000010768ad00 /Users/joshuayoung/Projects/puma/test/helpers/integration.rb:388 run> terminated with exception (report_on_exception is true):
/Users/joshuayoung/Projects/puma/test/helpers/integration.rb:392:in `initialize': Can't assign requested address - connect(2) for "127.0.0.1" port 54325 (Errno::EADDRNOTAVAIL)
[...]

PR #3990, opened 2026-08-07
Make a family of flaky tests deterministic

### LINKED ISSUE #3987: test_hot_restart_does_not_drop_connections_threads operates on a race condition
**Describe the bug**

I ran into this with a PR I made, namely https://github.com/puma/puma/pull/3982.

`test_hot_restart_does_not_drop_connections_threads` starts a single Puma, then does a couple things in parallel:

- Five threads fire off a set number of TTL requests (1000 on `main`).
- One thread runs a loop which repeatedly sends `USR2`, waits for the server to boot, and increments a restarts counter.

The test passes if the restarts counter is 2 or more.

What the test literally asserts is that Puma restarts at least twice within a "timeframe" of 1K requests, and that 998 of those requests (i.e., 99.8%) get an acceptable response. You could take that for granted if it weren't for the restarts, so this is only a meaningful test if the restarts happen.

On older Linuxes, the test looks for one restart instead of two:

```ruby
    # started intermittently failing on ubuntu22, May-2026    
    assert_operator_equal = UBUNTU_VERSION && UBUNTU_VERSION <= 22 ? 1 : 2
```

(That's line 638 of `test/helpers/integration.rb`.)

I only found that while I was trying to write up this issue. The test failed for me on one of the Ubuntu GitHub Actions, the one for Ubuntu 24.04 and Ruby 4.0. I "fixed" it by increasing the set number of TTL requests. But this would have worked too. The problem is that Puma runs on faster hardware than it did when the test was first written, so a 1K "timeframe" isn't enough for two restarts. The above code fixes it by checking for just one restart instead.

**To Reproduce**

This issue will automatically recur on other PRs if my understanding is correct, and indeed the above code shows that people have run into it before.

Also it looks as if many reproductions have already happened:
- https://github.com/puma/puma/pull/3972
- https://github.com/puma/puma/pull/3960
- https://github.com/puma/puma/pull/3589
- https://github.com/puma/puma/issues/3535
- https://github.com/puma/puma/issues/3114

OK, actually, now I think anybody who cares enough to read all this probably already knew all about this issue. My bad!

I'm going to keep going because I'm hoping to make a PR for this. I think the real problem is that `test_hot_restart_does_not_drop_connections_threads` (and what appears to be a family of related tests) uses a count of requests to define a kind of quasi-timeframe, and asserts that at least X many restarts will happen with that "timeframe" of Y many requests.

I believe this means that the core engine of the test is a race condition, and therefore the test (and probably some or all of its family of tests) should use a different mechanism.

gilesbowkett:
### Description

Closes https://github.com/puma/puma/issues/3987, which is my own issue, just explaining the problem when I first found it.

I have to apologize, this PR is a lot of text (and diagrams). The actual changes are 76 lines deleted, 75 lines added. They make some complicated, flaky code more simple and reliable over the long term, so I think it'll be worth it.

-------------

## Background

I wrote a [PR](https://github.com/puma/puma/pull/3982), and an unrelated test failed. So I figured out why it was failing, and discovered that it's part of a family of flaky tests. Maintainers probably already know about this, because the flakiness has shown up already in these issues and PRs:

- https://github.com/puma/puma/pull/3972
- https://github.com/puma/puma/pull/3960
- https://github.com/puma/puma/pull/3589
- https://github.com/puma/puma/issues/3535
- https://github.com/puma/puma/issues/3114

These PRs and issues include a history of maintainers tweaking these tests, because the helper makes them flaky. There's also a couple workarounds: the tests scale back the number of requests for Windows, and the number of restarts for Ubuntu.

What's going on in this family of tests is that the test helper `restart_does_not_drop_connections` in `test/helpers/integration.rb` serves as a foundation for eight tests across `test_integration_cluster.rb` and `test_integration_single.rb`. As I explained in the related issue when I first understood it, the helper starts a single Puma, then does a couple things in parallel:

- Five threads fire off a set number of TTL requests (1000 on `main`).
- One thread runs a loop which repeatedly sends `USR2`, waits for the server to boot, and increments a restarts counter.

The test passes if the restarts counter is 2 or more.

(These specifics come from `test_hot_restart_does_not_drop_connections_threads`, the test which broke my [otherwise unrelated PR](https://github.com/puma/puma/pull/3982), but the same basic shape holds for several tests.)

What the test literally asserts is that Puma restarts at least twice within a "timeframe" of 1K requests, and that 998 of those requests (i.e., 99.8%) get an acceptable response. But that's not the goal of the tests. The goal is to show that you get good responses even after a restart. So there's this effort to construct a window, and have both responses and restarts happening within that window.

`total_requests` is functioning like it's a stopwatch, or the finish line that the test has to cross. Basically the mechanism that the test runs on is a race condition. So it's kind of doomed to always be flaky unless we restructure it, so this PR restructures it.

## The sequence of events

This is sort of what the test helper looks like on `main`. The client threads and the server thread start at the same time, and the goal is for the server to restart twice before the clients run out of requests.

```mermaid                                                            
sequenceDiagram    
    participant C as 5 client threads<br/>(budget: 1000 requests)    
    participant R as restart thread                                 
    Note over C,R: both start at t=0
    R->>R: sleep 0.2s                                                                                                            
    C-->>C: 1000 requests done at t=1.4s                                                                                    
    R->>R: restart 1 finishes booting at t=1.1s    
    R->>R: restart 2 starts at t=1.25s...    
    Note over C,R: clients exit, only one restart. FAIL    
```

Basically, the test says "send 1000 requests, and while that's happening, restart the server over and over until the requests run out." But that didn't happen on [my PR](https://github.com/puma/puma/pull/3982), for a particular Linux flavor, until I randomly bumped the number of requests from 1K to 3K. [Similar prior changes have lowered the number of restarts](https://github.com/puma/puma/pull/3589), instead of raising the number of requests.

Both of those fixes are kind of dealing with the fact that the relationship between "how many restarts" vs "how many requests" hinges on hardware performance. That means the tests have a dependency on hardware performance. Since hardware performance is generally always getting better, these tests are always getting worse.

Also, we don't care that the server restarts X times while serving Y requests. All these tests are actually trying to show is that if a client request is in flight when the server restarts, the request still gets a correct response. The restarts vs requests thing is just a windowing mechanism to make the comparison in.

We can't actually _guarantee_ that we're restarting while a request is in flight. What we can do is provide continuous load. That's what the race condition mechanism does, but there's an easier way to do it.

So in this PR, the test now says "do exactly 3 restarts." The requests keep going until the test tells them to stop. Between each restart, the test pauses and waits until 100 more requests have succeeded before it restarts again. So we're always sending enough requests that the restarts are meaningful.

```mermaid
sequenceDiagram
    participant C as client threads<br/>(run until told to stop)
    participant R as test body
    C->>R: 100 successes
    R->>C: signal, wait for boot
    C->>R: 100 more successes
    R->>C: signal, wait for boot
    C->>R: 100 more successes
    Note over C,R: restarts = 3, exactly
```

This magic 100 number has a really important difference from the magic 1000 number (1000 requests). It doesn't get worse as the hardware gets better.

The client threads keep sending requests the whole time, including during a restart. The 100 just tells the test when to fire the next signal. Each client thread also sleeps 2ms per request, so a faster machine still takes roughly the same time to reach 100 successes.

### Checklist

- ~~If this PR doesn't need tests (docs change), I added `[ci skip]` to the title of the PR.~~
- [x] I have updated the ~~documentation~~ test helper comments accordingly.

### DIFF
--- a/test/helpers/integration.rb
+++ b/test/helpers/integration.rb
@@ -483,9 +483,22 @@ class TestIntegration < PumaTest
     JSON.parse read_pipe.read.split("\n", 2).last
   end
 
+  # Restarts the server `restarts` times while `num_threads` client threads keep
+  # sending requests, then asserts that (almost) none of those requests were
+  # dropped.
+  #
+  # The number of restarts is a loop bound, not a count of however many restarts
+  # happen to fit inside a fixed number of requests.  Client threads run until
+  # they are told to stop, and each pass waits for `replies_per_gate` further
+  # successful responses before signaling again, so traffic is known to be
+  # flowing across every restart.  A fast server or a slow machine changes how
+  # long the test takes, not whether it passes.
   def restart_does_not_drop_connections(
+      restarts: 3,
       num_threads: 1,
-      total_requests: 500,
+      replies_per_gate: 100,
+      gate_timeout: 20,
+      pause: 0.002,
       config: nil,
       unix: nil,
       signal: nil,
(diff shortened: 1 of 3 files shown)


MSP-Greg:
@gilesbowkett

Good idea.

Sorry for delay.

I've got some changes at https://github.com/puma/puma/commit/3e7dcb65bfa6f0666e53ba1f2e498c51a7f574e8, please add if they're okay.  I wanted to increase the socket count on Ubuntu, so that's part of the changes...

MSP-Greg:
@gilesbowkett

Thanks.

Re 'I believe these are now unnecessary', can you remove that?  All sockets are closed in `teardown` with checks that they should be.  Or, the normal code would close them, but the code in the tests here closes them earlier.

Re 'take out one last magic number', the change is too large.

The `TestIntegrationCluster#test_hot_restart_does_not_drop_connections_threads` had 868 attempts, or 868 * 0.002, which rounds to 2 allowed_errors.  Note that all these tests passed., so a large jump isn't required.

The new code would have 30 allowed_errors.  Maybe use `1.5 * restarts`?

MSP-Greg:
@gilesbowkett

Thanks again...

gilesbowkett:
hi @MSP-Greg, sorry about the slow reply. I was digging into alternate ways to handle `allowed_errors` and had to put it off til the weekend. wrote a couple comments attempting to cover the tradeoffs but never got anywhere coherent enough to post them. anyway, thx for merging! LMK if you want me to see if I can untangle those last two commits

MSP-Greg:
@gilesbowkett

No problem.  I'm sorry for the comment about the 'I believe these are now unnecessary' commit, I was mistaken, as the `ios_to_close` is in the `open_client_socket` code.  Again, thanks for the PR.

gilesbowkett:
@MSP-Greg all good!

PR #4001, opened 2026-08-28
Run after_stopped hooks after request draining

dentarg:
The after_stopped hook currently runs before stop_blocked, despite its name. Cleanup callbacks can therefore release resources while requests are still in flight, losing final work such as exported metrics.

Fire the hook after graceful request draining completes. This also aligns single-mode shutdown with cluster mode, which fires the hook after its workers have stopped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,8 +225,8 @@ before_restart do
 end
 
 after_stopped do
-  # Add code to run in the Puma master process when it receives
-  # a stop command but before it shuts down.
+  # Add code to run in the Puma master process after outstanding
+  # requests have finished and Puma has stopped.
 end
 ```
 
--- a/lib/puma/launcher.rb
+++ b/lib/puma/launcher.rb
@@ -280,8 +280,8 @@ module Puma
     end
 
     def do_graceful_stop
-      @events.fire_after_stopped!
       @runner.stop_blocked
+      @events.fire_after_stopped!
     end
 
     def do_restart(previous_env)
--- a/test/test_launcher.rb
+++ b/test/test_launcher.rb
@@ -137,6 +137,19 @@ class TestLauncher < PumaTest
     assert is_stopped, "after_stopped not called"
   end
 
+  def test_after_stopped_fires_after_requests_drain
+    events = []
+    runner = Object.new
+    runner.define_singleton_method(:stop_blocked) { events << :requests_drained }
+    launcher = create_launcher
+    launcher.instance_variable_set(:@runner, runner)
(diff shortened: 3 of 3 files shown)


dentarg:
<details><summary>Details</summary>
<p>



`just to double check, can you show me how this aligns the behaviour with cluster mode`

> Yes. The implementations differ, but the observable ordering now matches.
> 
> Single mode after the patch:
> 
> ```text
> stop requested
> → Single#stop_blocked
> → Server#stop(true)
> → outstanding requests drain
> → after_stopped
> ```
> 
> [Launcher#do_graceful_stop](/repos/puma/lib/puma/launcher.rb:282) now calls [Single#stop_blocked](/repos/puma/lib/puma/single.rb:35) before firing the hook.
> 
> Cluster mode on `SIGTERM` already does:
> 
> ```text
> stop requested
> → stop_workers
> → workers stop their servers and join server threads
> → master waits until every worker exits
> → after_stopped
> ```
> 
> The ordering is visible in [Cluster’s SIGTERM handler](/repos/puma/lib/puma/cluster.rb:344). `stop_workers` waits until all worker processes have exited, and each [worker joins its server thread](/repos/puma/lib/puma/cluster/worker.rb:146) before exiting.
> 
> For other cluster shutdown paths, `Cluster#run` calls `stop_workers` before returning to the launcher. The launcher then performs `stop_blocked` and fires `after_stopped`.
> 
> So the common lifecycle contract becomes:
> 
> ```text
> serving processes finish graceful shutdown
> → after_stopped callbacks run
> ```
> 
> Previously, only single mode violated that ordering by firing the callback immediately before its blocking drain.

`so "master waits until every worker exits" is stop_blocked?`

> Not in the cluster `SIGTERM` path.
> 
> There, “master waits until every worker exits” happens inside `Cluster#stop_workers`:
> 
> ```ruby
> loop do
>   wait_workers
>   break if @workers.reject { |worker| worker.pid.nil? }.empty?
>   sleep 0.2
> end
> ```
> 
> Then `after_stopped` fires directly from the cluster signal handler.
> 
> `Cluster#stop_blocked` is used for shutdowns that return through `Launcher#do_graceful_stop`. It calls `Process.waitall`, but `Cluster#run` has normally already called `stop_workers` before returning to the launcher.
> 
> So the alignment is semantic, not that both modes use `stop_blocked`:
> 
> - Single mode drains through `stop_blocked`, then fires the hook.
> - Cluster mode waits through `stop_workers`, then fires the hook.

</p>
</details>

MSP-Greg (approved):
LGTM

PR #4012, opened 2026-09-04
Bound large file responses to Content-Length

nateberkopec:
## Summary

Bound optimized large-file responses to their declared `Content-Length`.

The small-file path already reads at most `content_length`, but the `IO.copy_stream` path copied through EOF. If a file was larger than its declared length, Puma wrote surplus bytes after the response boundary, which could desynchronize a persistent HTTP connection.

This PR intentionally changes exactly one line.

## Validation

- Standalone regression reproduction: before `announced=65537 serialized=131074`; after `PASS: serialized 65537 bytes`
- `bundle exec ruby -Itest test/test_response_header.rb` — 14 runs, 32 assertions
- `bundle exec ruby -Itest test/test_puma_server.rb` — 143 runs, 497 assertions
- `bundle exec rubocop lib/puma/response.rb`
- `git diff --check`

### DIFF
--- a/lib/puma/response.rb
+++ b/lib/puma/response.rb
@@ -297,7 +297,7 @@ module Puma
             fast_write_str socket, io_buffer.read_and_reset
           else
             fast_write_str socket, io_buffer.read_and_reset
-            IO.copy_stream body, socket
+            IO.copy_stream body, socket, content_length
           end
         end
       elsif body.is_a?(::Array) && body.length == 1


nateberkopec:
This change protects the HTTP response boundary.

Puma gets `Content-Length` from the Rack response headers. If the application does not supply the header, Puma can calculate the value from `File#size`. Puma then writes this value in the response header.

For small files, Puma reads only `content_length` bytes. For files larger than `IO_BODY_MAX`, Puma currently calls `IO.copy_stream` without a length. This call copies data until EOF. Thus, the two paths do not have the same behavior.

I reproduced the problem with these steps:

1. Create a file with 131,074 bytes.
2. Set the response `Content-Length` to 65,537 bytes. This value is one byte above `IO_BODY_MAX`, so Puma uses the large-file path.
3. Call `fast_write_response` with a temporary file as the output socket.
4. Read the output and compare the declared length with the serialized body length.

Before this change, the result was:

```text
announced=65537 serialized=131074
```

After this change, the result is:

```text
PASS: serialized 65537 bytes
```

A mismatch can occur if an application or middleware supplies old or incorrect metadata. It can also occur when a file grows after Puma gets its size and before the copy is complete. Therefore, Puma can produce this mismatch even when it calculates the header itself.

On a persistent connection, the client reads the declared number of bytes as the first response body. It can then read the extra bytes as part of the next response. This can cause response desynchronization. It can also send file data that is outside the intended response range.

The third argument to `IO.copy_stream` sets the maximum byte count. The change keeps the optimized copy path and makes its behavior agree with the existing small-file path.

PR #4020, opened 2026-09-17
ci: Fix Turbo CI dependencies and bound integration test waits

tedaford:
## Description

Fix incompatible JSON resolution in the Turbo integration bundle, make integration-test waits and cleanup bounded, and keep retry summaries within GitHub Actions’ upload limit.

The Turbo workflow resolves JSON 3.0.2 with ActiveSupport 8.1.3.1, whose positional `JSON.parse(json, options)` call is incompatible with JSON 3. Constrain JSON to `~> 2.18` in the Turbo checkout’s Gemfile, matching Puma’s own development constraint. The unchanged-main control reports 123 errors; the constrained bundle passes all 235 tests.

The shared integration helper discards custom log timeouts, restarts the deadline on each read attempt, and rescues the test watchdog as an ordinary I/O error. A silent server, incomplete line, or closed output pipe can therefore consume the entire ten-minute Actions step instead of producing a useful failure.

This change:

- Uses one monotonic deadline for each expected log event and honors explicit timeout values.
- Reads incrementally without blocking on a partial line, preserves buffered lines and final lines without a newline, and fails promptly at EOF. Normalizes complete CRLF lines, including CR/LF split across reads, to preserve Windows PID and readiness matching previously provided by `IO#gets`.
- Preserves watchdog exceptions and includes the last 4,096 characters of captured server output on log-wait failures.
- Bounds graceful child shutdown at 35 seconds, allowing Puma’s existing 30-second worker shutdown period; drains output during the wait, then force-terminates/reaps an unresponsive test-owned process or isolated POSIX process group and reports a failure.
- Keeps teardown cleanup running after a stop failure and bounds closing the process pipe.
- Gives shutdown-hook and hot-restart shutdown waits the explicit graceful-shutdown allowance.
- Caps accumulated retry summaries at 1,000,000 bytes, below GitHub’s 1 MiB per-step limit. Keeps complete entries, appends one truncation notice, and leaves full failure details in the job log; handles concurrent callbacks, existing content, multibyte text, and Windows newline conversion.

The workflow step limit and retry count are unchanged. This makes stalls bounded and diagnosable; it does not claim to resolve every underlying intermittent worker-start or refork failure. There are no Puma runtime or CVE changes in this branch.

## Validation

- [Unchanged-main Turbo control](https://github.com/tedaford/puma/actions/runs/35257469304): 235 runs, 281 assertions, 0 failures, 123 errors.
- [Turbo with JSON constraint](https://github.com/tedaford/puma/actions/runs/35258109102): 235 runs, 654 assertions, 0 failures, 0 errors, 0 skips (JSON 2.21.2).
- The 12 new helper regressions produce 7 failures and 1 error against the original helper, and pass with the fix: 12 runs, 40 assertions, no failures/errors.
- Final focused check on Ruby 3.3.6/macOS, including culling, shutdown hooks, hot restart, and helper regressions: **15 runs, 47 assertions, no failures/errors**.
- Full suite before the CRLF/summary follow-up, on Ruby 3.3.6/macOS (`MT_CPU=2`, seed 24962): **877 runs, 2,574 assertions, 0 failures, 0 errors, 16 platform skips**.
- The initial cross-platform matrix exposed CRLF matching failures on Windows. Both new CRLF regressions fail before the correction and pass afterward.
- Current focused validation, including helper, summary, and single-process integration tests: **40 runs, 115 assertions, no failures/errors**.
- The actual retry-summary callback was exercised with 400 simulated failures: **999,942 bytes**, one truncation notice, no overflow.
- The updated GitHub Actions matrix is pending. A prior Ubuntu 24.04/Ruby 3.1 run still reached the step timeout after a connection-restart test retry; that separate intermittent hang is not claimed fixed by this follow-up.
- RuboCop: 221 files, no offenses. Workflow YAML, shell syntax, and `git diff --check` pass.

The original 12 helper regression tests are committed separately before their implementation so the failures can be reproduced. Cross-platform validation is left to the updated GitHub Actions matrix.


tedaford:
@MSP-Greg This should fix CI so that everything gets a "green check" consistently.

